PR #9, opened 2016-06-25
Start using fern for logging

### LINKED ISSUE #7: Allow logging to file
The log / env_logger crates are great and easy to use, but I'd like to have the ability to log to a file so that logging from the server process is useful. Someone pointed me at this crate which might be the way to go:
https://github.com/daboross/fern-rs

F3real:
Fixes #7

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,7 +16,9 @@ extern crate clap;
 extern crate env_logger;
 extern crate filetime;
 extern crate kernel32;
-#[macro_use] extern crate log;
+#[macro_use] 
+extern crate log;
+extern crate fern;
 extern crate libc;
 extern crate mio;
 extern crate number_prefix;
@@ -25,6 +27,7 @@ extern crate retry;
 extern crate rusoto;
 extern crate sha1;
 extern crate tempdir;
+extern crate time;
 extern crate winapi;
 extern crate zip;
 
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -43,6 +46,18 @@ mod protocol;
...
 fn main() {
-    env_logger::init().unwrap();
+    let logger_config = fern::DispatchConfig {
+        format: Box::new(|msg: &str, level: &log::LogLevel, _location: &log::LogLocation| {
+            format!("[{}][{}] {}", time::now().strftime("%Y-%m-%d][%H:%M:%S").unwrap(), level, msg)
+        }),
+        output: vec![fern::OutputConfig::stdout(), fern::OutputConfig::file("sccache2.log")],
+        level: log::LogLevelFilter::Trace,
+    };
+    if let Err(e) = fern::init_global_logger(logger_config, log::LogLevelFilter::Trace)
```
luser: This is a good start, but I don't think we want to log at `Trace` by default, that's very verbose. I guess to do anything useful we'll have to provide some way to set the log levels, whether via environment variable (like `env_logger` does) or commandline arguments. Maybe you should set these to `Warn` as the default right now, and in a follow-up change we can figure out how to make it easy to set the log levels?

F3real:
@luser  I've changed log level.  I'll add option to set log level through  environment variable in next patch.

PR #10, opened 2016-06-27
Set log level using env variable

F3real:


src/main.rs
```diff
@@ -53,7 +55,19 @@ fn main() {
         output: vec![fern::OutputConfig::stdout(), fern::OutputConfig::file("sccache2.log")],
         level: log::LogLevelFilter::Trace,
     };
-    if let Err(e) = fern::init_global_logger(logger_config, log::LogLevelFilter::Warn)
+
+    // Set current log level depending on environment variable
+    let log_level = match &*env::var("SCCACHE2_LOG_LEVEL").unwrap_or("warn".to_string()).to_lowercase() {
```
luser: Could you move this all into a separate function, like `fn init_logging()`, and just call that from main? I like having main with a very simple body.

F3real:
@luser I've moved code to new function

luser:
Thanks! I have an idea for a follow-up to make logging even more useful. I will file a new issue.

armenzg:
\o/

PR #14, opened 2016-07-05
Started using which crate

### LINKED ISSUE #11: Switch to using `which` crate
There's a `which` crate on crates.io: https://crates.io/crates/which

I originally looked at using it but it didn't support all the things I needed, so I wrote my own implementation in commands.rs. I submitted my changes upstream and they've been merged and made it to the latest release, so we can remove the implementation in sccache2 and just use the external crate now.

F3real:
Fixes #11

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,6 +17,7 @@ dependencies = [
  "sha1 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempdir 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.35 (registry+https://github.com/rust-lang/crates.io-index)",
+ "which 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


luser:
So this is failing some tests on Windows. Did you ever get a Windows build environment working?

F3real:
Not yet, do I have to install Visual Studio for it to work ?

luser:
I haven't tried building with the mingw rust toolchain, I've been using the msvc one. To use that one you will need to install Visual C++, but Visual C++ 2015 Community Edition is free and works fine.

F3real:
@luser I've finally gotten scache2 working on windows :)

But it seems bug is upstream in which library. I am calling which_in function which is same as one from sccache2 with same arguments as function from sccache2 and I am getting following errors
**Bad absolute path**
**Cannot find binary path**
**Bad relative path**

Tests are passing on Linux with same code.

F3real:
@luser All tests are green  :) I think you can merge it now.

luser:
Thanks for the patch!

PR #17, opened 2016-07-10
Use MutByteBuf

### LINKED ISSUE #13: Use `bytes` crate for buffering data
I left a pair of FIXME comments in the server source:
https://github.com/luser/sccache2/blob/1fb11fc53ec7ddae4663ea6dea6ad3afa29b01a9/src/server.rs#L695
https://github.com/luser/sccache2/blob/1fb11fc53ec7ddae4663ea6dea6ad3afa29b01a9/src/server.rs#L742

I didn't get around to sorting those out, but it'd be nice to use the `bytes` crate for buffering data:
https://github.com/carllerche/bytes

We probably want to use `MutByteBuf` in this case:
http://carllerche.github.io/bytes/bytes/buf/struct.MutByteBuf.html

F3real:
Fixes #13

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use bytes::{
+    Buf,
+    ByteBuf,
+    MutByteBuf,
+};
 use cache::{
     Storage,
     storage_from_environment,
@@ -216,10 +221,10 @@ impl<C : CommandCreatorSync + 'static> SccacheServer<C> {
             }
         };
 
-        if let Some(token) = self.conns.insert_with(|token| 
+        if let Some(token) = self.conns.insert_with(|token|
         {
             ClientConnection::new(sock, token)
-        }) 
+        })
         {
             match self.conns[token].register(event_loop) {
                 Ok(_) => {},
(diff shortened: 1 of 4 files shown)


src/server.rs
```diff
@@ -687,19 +688,17 @@ impl<C : CommandCreatorSync + 'static> ClientConnection<C> {
...
+            match self.sock.try_read_buf(&mut buf) {
                 Ok(None) => {
                     // Read all available data.
                     trace!("try_read returned Ok(None)");
                     break;
                 },
                 Ok(Some(n)) => {
                     trace!("try_read read {} bytes", n);
-                    self.recv_buf.extend_from_slice(&buf[..n])
+                    self.recv_buf.extend_from_slice(buf.bytes())
```
luser: I think what I was hoping to do here was to make `self.recv_buf` and `self.send_queue` into `MutByteBuf`, to make the logic around appending bytes and draining bytes cleaner. Does that make sense?

F3real:
@luser Failures seems unrelated to patch ( If I've seen correctly they are caused by lib). Could you take look at changes ?
https://github.com/luser/sccache2/pull/17/files#diff-4ce93534efc34e923ce01e975eb7ed80L715
I'll remove comment and change it to recv_buffer.clear() tomorrow.

luser:
The tokio refactor in https://github.com/mozilla/sccache/pull/68 made us use `EasyBuf` (which is part of tokio-core), so this is not an issue anymore.

PR #24, opened 2016-08-31
Add recache support

### LINKED ISSUE #18: Add `SCCACHE_RECACHE` support
@glandium points out that we shouldn't start using sccache2 in production without adding support for `SCCACHE_RECACHE`, which forces compiles to rebuild even if a cache entry is available. This is useful for times when bad data gets into the cache.

F3real:
Fixes #18

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -43,6 +43,8 @@ pub enum Cache {
     Hit(CacheRead),
     /// Result was not found in cache.
     Miss,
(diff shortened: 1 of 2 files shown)


luser:
I tweaked this a little and added a test: https://github.com/luser/sccache2/commit/3593d87d1c72570399edc8d2c5bdb3fc340e573f

Thanks for the patch, sorry it took me so long to get back around to it!

PR #26, opened 2016-11-21
Do not cache if Clang modules are specified.

LegNeato:
See https://twitter.com/TedMielczarek/status/800813453430624256.

`ccache` doesn't currently support them either. See this thread:

  https://lists.samba.org/archive/ccache/2016q3/001459.html

Support was disabled because of:

  https://bugzilla.samba.org/show_bug.cgi?id=11048

But looks like `ccache` may start supporting them soon with this PR:

  https://github.com/ccache/ccache/pull/130

...which may inform work that needs to be done for this project.


luser:
Thanks for the patch! Adding real support would be cool as well, although I don't have any current use for it.

PR #27, opened 2016-11-27
[RFC] Add support for custom s3 endpoint url

harshavardhana:
Specify a new endpoint with `SCCACHE_ENDPOINT`
such as Minio https://github.com/minio/minio.

If not defaults to 's3.amazonaws.com'. Current
code for some reason was not working with custom
regions. Introduced a new ENV called `SCCACHE_REGION`
to handle cross region s3 URLs.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -182,8 +182,16 @@ fn parse_size(val: &str) -> Option<usize> {
 /// Get a suitable `Storage` implementation from the environment.
 pub fn storage_from_environment() -> Box<Storage> {
     if let Ok(bucket) = env::var("SCCACHE_BUCKET") {
-        trace!("Trying S3Cache({})", bucket);
-        match S3Cache::new(&bucket) {
+        let endpoint = match env::var("SCCACHE_ENDPOINT") {
+            Ok(endpoint) => format!("{}/{}", endpoint, bucket),
+            _ => match env::var("SCCACHE_REGION") {
+                Ok(ref region) if region != "us-east-1" =>
+                    format!("{}.s3-{}.amazonaws.com", bucket, region),
+                _ => String::from("s3.amazonaws.com"),
+            },
+        };
+        debug!("Trying S3Cache({})", endpoint);
+        match S3Cache::new(&bucket, &endpoint) {
             Ok(s) => {
                 trace!("Using S3Cache");
                 return Box::new(s);
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -50,7 +50,7 @@ pub struct S3Cache {
 
 impl S3Cache {
     /// Create a new `S3Cache` storing data in `bucket`.
-    pub fn new(bucket: &str) -> io::Result<S3Cache> {
+    pub fn new(bucket: &str, endpoint: &str) -> io::Result<S3Cache> {
         let home = try!(env::home_dir().ok_or(Error::new(ErrorKind::Other, "Couldn't find home directory")));
         let profile_providers = vec![
             ProfileProvider::with_configuration(home.join(".aws").join("credentials"), "default"),
@@ -61,7 +61,7 @@ impl S3Cache {
             ];
         let provider = AutoRefreshingProviderSync::with_mutex(ChainProvider::with_profile_providers(profile_providers)).ok().map(Arc::new);
         //TODO: configurable SSL
-        let bucket = Arc::new(Bucket::new(bucket, Ssl::No));
+        let bucket = Arc::new(Bucket::new(bucket, endpoint, Ssl::No));
         Ok(S3Cache {
             bucket: bucket,
             provider: provider,
--- a/src/simples3/s3.rs
+++ b/src/simples3/s3.rs
@@ -26,17 +26,13 @@ pub enum Ssl {
     No,
 }
 
-fn base_url(bucket_name: &str, ssl: Ssl, region: Option<&str>) -> String {
-    format!("{}://{}.s3{}.amazonaws.com/",
+fn base_url(endpoint: &str, ssl: Ssl) -> String {
+    format!("{}://{}/",
             match ssl {
                 Ssl::Yes => "https",
                 Ssl::No => "http",
             },
-            bucket_name,
-            match region {
-                Some(ref r) => format!("-{}", r),
-                None => String::new(),
-            })
+            endpoint)
 }
 
 fn hmac<D: Digest>(d: D, key: &[u8], data: &[u8]) -> Vec<u8> {
@@ -71,8 +67,8 @@ pub enum S3Error {
 }
 
 impl Bucket {
-    pub fn new(name: &str, ssl: Ssl) -> Bucket {
-        let base_url = base_url(&name, ssl, None);
+    pub fn new(name: &str, endpoint: &str, ssl: Ssl) -> Bucket {
+        let base_url = base_url(&endpoint, ssl);
         Bucket {
             name: name.to_owned(),
             base_url: base_url,


luser:
Thanks for the patch! This looks pretty good, but I have a question about the region bits. Per discussion with @glandium when writing this code originally, and from [reading the docs](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro), I'm given to understand that the region is not necessary in the URL when using a vhost URL to access a bucket. Specifically the docs say:

> If you make a request to the http://bucket.s3.amazonaws.com endpoint, the DNS has sufficient information to route your request directly to the region where your bucket resides.

Is there some scenario I'm missing here?

harshavardhana:
Correct but it is better to send the request directly to S3 region endpoint and which doesn't require the routing capabilities of s3 - It also becomes apparent when you bring in support for signature v4.

src/cache/cache.rs
```diff
@@ -203,9 +203,31 @@ pub trait Storage: Send + Sync {
 
 /// Get a suitable `Storage` implementation from the environment.
 pub fn storage_from_environment() -> Box<Storage> {
+    let endpoint_env_res = env::var("SCCACHE_ENDPOINT");
+    let mut endpoint = String::from("s3.amazonaws.com");
```
luser: I tend to prefer chaining with `and_then` or `map` on `Result` and `Option` values instead of using `mut` variables, so I would probably write this like:

```rust
let endpoint = env::var("SCCACHE_ENDPOINT").unwrap_or_else(|_| String::from("s3.amazonaws.com"))
```

And similarly for the region bits below. Also: you might as well move this whole block inside the `SCCACHE_BUCKET` block, since these variables aren't useful if we're not using S3 storage.

src/cache/cache.rs
```diff
@@ -203,9 +203,31 @@ pub trait Storage: Send + Sync {
...
+        Err(e) => warn!("Failed to interpret region: {:?}", e),
+    }
+    if region == "-us-east-1".to_string() {
+        region = String::from("");
+    }
+
     if let Ok(bucket) = env::var("SCCACHE_BUCKET") {
-        trace!("Trying S3Cache({})", bucket);
-        match S3Cache::new(&bucket) {
+        if endpoint.ends_with(".amazonaws.com") {
```
luser: Instead of doing a string comparison here, maybe you should just combine this with the `SCCACHE_ENDPOINT` check, so something like:

```rust
let endpoint = match env::var("SCCACHE_ENDPOINT") {
  Ok(endpoint) => format!("{}/{}", endpoint, bucket),
  _ => format!("{}.s3{}.amazonaws.com", bucket, region),
}
```

That would mean that if someone explicitly set `SCCACHE_ENDPOINT=http://s3.amazonaws.com/` we wouldn't use the bucket vhost, but that seems fine.

src/cache/cache.rs
```diff
@@ -203,9 +203,31 @@ pub trait Storage: Send + Sync {
 
 /// Get a suitable `Storage` implementation from the environment.
 pub fn storage_from_environment() -> Box<Storage> {
+    let endpoint_env_res = env::var("SCCACHE_ENDPOINT");
+    let mut endpoint = String::from("s3.amazonaws.com");
+    match endpoint_env_res {
+        Ok(ep) => endpoint = ep,
+        Err(e) => warn!("Failed to interpret endpoint: {:?}", e),
+    }
+    let region_env_res = env::var("SCCACHE_REGION");
```
luser: I looked again, and I still don't understand why the `GET Bucket/?Location` request in S3 requires you to own the bucket. It would be nice if we could just autodetect that instead. I guess this is just an optimization anyway, since S3 will send your requests to the right place regardless.

luser:
Interestingly, per [this StackOverflow answer](http://stackoverflow.com/a/27092765/69326), S3 sends the bucket region in a HTTP header in response to most queries. It looks like you can reliably HEAD a bucket and get the region in response. I think this would be nicer than requiring it to be specified explicitly as an environment variable, even if it costs a single extra HTTP request on startup:
```
$ curl -I 'http://s3.amazonaws.com/mozilla-releng-s3-cache-us-west-2-prod'
HTTP/1.1 301 Moved Permanently
x-amz-bucket-region: us-west-2
<...>

$ curl -I 'http://s3.amazonaws.com/sccache'
HTTP/1.1 403 Forbidden
x-amz-bucket-region: us-east-1
<...>

$ curl -I 'http://s3.amazonaws.com/taskcluster-level-1-sccache-eu-central-1'
HTTP/1.1 301 Moved Permanently
x-amz-bucket-region: eu-central-1
<...>
```

I think we should pass the bucket and Option<String> for the endpoint down through `S3Cache::new` to `Bucket::new` (where we create a `hyper::Client`, then fetch the bucket region there by doing a `HEAD http://s3.amazonaws.com/bucket`. How does that sound?

src/cache/cache.rs
harshavardhana: You need to that is how API is designed to handle this what is needed is a region cache to be implemented and used. This is how it is done in `minio` SDKs.  But since sccache is only meant to be used with one bucket at any given time it doesn't make sense to make this request instead users can happily provide this value.

luser: I disagree. It's one more configuration value, which is a pain to manage, and it's possible for it to be out-of-sync with the bucket name (I can imagine someone copy/pasting and editing these values and not changing the region properly.)

harshavardhana:
> Interestingly, per this StackOverflow answer, S3 sends the bucket region in a HTTP header in response to most queries. It looks like you can reliably HEAD a bucket and get the region in response. I think this would be nicer than requiring it to be specified explicitly as an environment variable, even if it costs a single extra HTTP request on startup:
> 
> $ curl -I 'http://s3.amazonaws.com/mozilla-releng-s3-cache-us-west-2-prod'
> HTTP/1.1 301 Moved Permanently
> x-amz-bucket-region: us-west-2
> <...>
> 
> $ curl -I 'http://s3.amazonaws.com/sccache'
> HTTP/1.1 403 Forbidden
> x-amz-bucket-region: us-east-1
> <...>
> 
> $ curl -I 'http://s3.amazonaws.com/taskcluster-level-1-sccache-eu-central-1'
> HTTP/1.1 301 Moved Permanently
> x-amz-bucket-region: eu-central-1
> <...>
> I think we should pass the bucket and Option for the endpoint down through S3Cache::new to Bucket::new (where we create a hyper::Client, then fetch the bucket region there by doing a HEAD http://s3.amazonaws.com/bucket. How does that sound?

@luser yes but one cannot assume this since this is an undocumented behavior and possibly might even change in future - implementing using something which is not in the SPEC  and also values returned during an error as the right values for the next operation is a bad practice in my understanding. 

The approach is this patch makes it work for Google Cloud Storage as well. Since they implement Signature V2 compliant S3 front end.  

Specifics related to S3 can be avoided also making it generic enough.

harshavardhana:
> I think we should pass the bucket and Option for the endpoint down through S3Cache::new to Bucket::new (where we create a hyper::Client, then fetch the bucket region there by doing a HEAD http://s3.amazonaws.com/bucket. How does that sound?

The proper way to solve is this is by implementing something called - https://github.com/minio/minio-go/blob/master/bucket-cache.go as i mentioned before which is built around using "GetBucketLocation()" API,  but since the bucket doesn't change in sccache. I thought that taking region wouldn't be such a bad idea.

src/cache/cache.rs
```diff
@@ -204,8 +204,20 @@ pub trait Storage: Send + Sync {
 /// Get a suitable `Storage` implementation from the environment.
 pub fn storage_from_environment() -> Box<Storage> {
     if let Ok(bucket) = env::var("SCCACHE_BUCKET") {
-        trace!("Trying S3Cache({})", bucket);
-        match S3Cache::new(&bucket) {
+        let region = match env::var("SCCACHE_REGION") {
+            Ok(region) => if region != "us-east-1".to_string() {
```
luser: Minor, but you can make this slightly verbose by moving the `if` into a match guard, like:
```rust
let region = match env::var("SCCACHE_REGION") {
    Ok(ref region) if region != "us-east-1" => format!("-{}", region),
    _ => String::new(),
};
```

also `String::from("")` and `String::new()` are functionally equivalent, I don't know if there's a preference there in general Rust usage, but the latter is shorter, at least.

I still think it might be a little nicer to move this down into the match arm where `SCCACHE_ENDPOINT` is unset, since you're not using the result if a custom endpoint is set anyway..

luser:
Also, apologies for the radio silence there, we had a company all-hands last week so I was a bit occupied.

harshavardhana:
> Thanks for the patch and the useful discussion! I take it we'll still need to implement signature v4 support in order to work with minio? If so we should get a followup issue open on that.

No we don't need Minio supports Signature v2.  It is just for the future thing as i mentioned.

harshavardhana:
> After thinking about this some more and discussing it a bit with a colleague I've come to agree with you--S3 just doesn't provide a supported way to get the region without having write access to the bucket, so requiring it to be explicitly set is not unreasonable. AFAICT if it's not set the code will continue to work as it currently does, where requests will be routed through us-east-1, which seems like an OK fallback.
> 
> There's one little code style nit I'd like to fix. If you would like to fix that yourself I will merge your changes afterwards. If you don't, I'll be happy to fix it myself and merge.

Will do right away thanks @luser

harshavardhana:
> I still think it might be a little nicer to move this down into the match arm where SCCACHE_ENDPOINT is unset, since you're not using the result if a custom endpoint is set anyway..

I didn't understand this clearly can you explain more @luser ?

luser:
You've got:
```rust
       let endpoint = match env::var("SCCACHE_ENDPOINT") {
           Ok(endpoint) => format!("{}/{}", endpoint, bucket),
           _ => format!("{}.s3{}.amazonaws.com", bucket, region),
       };
```

You only use `region` inside the second match arm there, so realistically you could move all the code that handles `SCCACHE_REGION` into a block there.

harshavardhana:
> only use region inside the second match arm there, so realistically you could move all the code that handles SCCACHE_REGION into a block there.

Is nested match a common practice? @luser

luser:
Yes, in the almost-everything-is-an-expression world of Rust that sort of thing happens a lot.

harshavardhana:
BTW `cargo test` fails sporadically once in a while, looks like unrelated to my change. @luser

luser:
> BTW cargo test fails sporadically once in a while, looks like unrelated to my change. @luser

Yeah, I saw that on one of my own pushes recently. It must be a recent breakage. I'll try to track it down. It's probably something I broke with my `DiskCache` changes.

harshavardhana:
Made the changes you requested @luser

luser:
Thanks for the patch!

luser:
> BTW cargo test fails sporadically once in a while, looks like unrelated to my change. @luser

Just FYI I fixed the racy test in https://github.com/mozilla/sccache/commit/6d7efe9bdc6937e138ba38eda50b2d6b46045257

PR #32, opened 2016-11-30
Simplify loop in hash_key().

Ms2ger:
This also avoids an underflow if args is empty.


luser:
Thanks for the fix! When it gets released we could use my new itertools `with_position` adapter for this: https://github.com/bluss/rust-itertools/blob/4380d3379ffc7d60103c90e796d48322db75cd22/src/lib.rs#L900

PR #36, opened 2016-11-30
Openstack swift support

leth:
Hi, this is my first foray into Rust, so apologies for the probably shonky code.
This PR adds an Openstack Swift object store cache backend.
I'd appreciate any suggestions for improvement :)

You might notice this relies on building a crate from my github account; I've got a [PR open to incoporate my changes to rust_swiftclient](https://github.com/joel-wright/rust-swiftclient/pull/1).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,7 @@ dependencies = [
  "regex 0.1.73 (registry+https://github.com/rust-lang/crates.io-index)",
  "retry 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rust-crypto 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rust-swiftclient 0.1.0 (git+https://github.com/leth/rust-swiftclient.git)",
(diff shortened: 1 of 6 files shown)


luser:
Cool, thanks for the patch! I may not find time to review it right away, I've got a few things on my plate right now and there's a Mozilla All-Hands next week so I'm going to be pretty busy then. If you don't hear something from me by the week after that please do ping me here!

luser:
Hey, sorry about letting this sit so long! I haven't looked at this too deeply, but I'd like to stick this behind a (cargo feature)[http://doc.crates.io/manifest.html#the-features-section]. If you can do that we can certainly get this landed. Build times are bad enough as it is, I don't think every sccache user wants to pay extra compile time to support N backends. We could actually put the existing s3 backend behind a feature as well, and just have it in `default-features` (you don't have to do that, but we could do it as a follow-up). That way if you're *not* using s3 you could `cargo build --no-default-features --features=swift` and not have to build s3 support.

leth:
No problem. I'm happy if you want to do that; for me this was mostly an exercise in learning a little Rust. Feel free to also not merge it, if that makes more sense.

PR #46, opened 2016-12-12
Support sccache being copied/hardlinked to a different name

### LINKED ISSUE #44: Support running with executable filename set to a compiler filename
ccache supports a mode where if the executable filename is "gcc" or similar it will act as if it were gcc. We should support this in sccache to make it easy to use in some situations.

glandium:
Resolves #44.

There's a bunch of unwrap() in there, but they should be "safe", as in, as far as reading the doc goes, they shouldn't ever be panicking in the code paths they are used under. (For example, exe.file_name().unwrap() shouldn't panic considering env::current_exe succeeded (actually, it could panic if the executable was removed since the process started, but panicking in that case doesn't sound too bad)).

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -20,6 +20,7 @@ use clap::{
 use std::env;
 use std::ffi::OsString;
 use std::path::PathBuf;
+use which::which_in;
 
 /// A specific command to run.
 pub enum Command {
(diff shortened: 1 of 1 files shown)


src/cmdline.rs
```diff
@@ -71,13 +72,48 @@ fn usage() -> Command {
...
+        println!("sccache: Couldn't determine current working directory");
+        return usage();
+    };
     // The internal start server command is passed in the environment.
     let internal_start_server = match env::var("SCCACHE_START_SERVER") {
         Ok(val) => val == "1",
         Err(_) => false,
     };
+    let mut args: Vec<_> = env::args_os().collect();
+    if ! internal_start_server {
```
luser: You could just change things so that the SCCACHE_START_SERVER block above does an early-return. It'd mean it would override any other commands instead of printing the error message from below, but I think that's OK, since it's basically just an implementation detail.

src/cmdline.rs
```diff
@@ -71,13 +72,48 @@ fn usage() -> Command {
...
+                // Otherwise, if it was copied/hardlinked under a different $name, act
+                // as if it were invoked with `sccache $name`, but avoid $name resolving
+                // to ourselves again if it's in the PATH.
+                _ => {
+                    let path = env::var_os("PATH");
+                    match which_in(exe.file_name().unwrap(), path.as_ref(), &cwd) {
+                        Ok(ref full_path) if full_path.canonicalize().unwrap() == exe.canonicalize().unwrap() => {
+                            if let Some(dir) = full_path.parent() {
+                                let path = env::join_paths(env::split_paths(&path.unwrap()).filter(|p| p != dir)).ok();
+                                match which_in(exe.file_name().unwrap(), path, &cwd) {
```
luser: Hum, I thought the `which` crate had exposed an API that gave you an iterator of results, but I guess not. That would be useful here. (You could file that on the upstream which-rs crate if you were so motivated, I rewrote most of it for sccache a while back.)

src/cmdline.rs
```diff
@@ -71,13 +72,48 @@ fn usage() -> Command {
...
+        if let Ok(exe) = env::current_exe() {
+            match exe.file_stem().and_then(|s| s.to_str()).map(|s| s.to_lowercase()) {
+                // If the executable has its standard name, do nothing.
+                Some(ref e) if e == env!("CARGO_PKG_NAME") => {}
+                // Otherwise, if it was copied/hardlinked under a different $name, act
+                // as if it were invoked with `sccache $name`, but avoid $name resolving
+                // to ourselves again if it's in the PATH.
+                _ => {
+                    let path = env::var_os("PATH");
+                    match which_in(exe.file_name().unwrap(), path.as_ref(), &cwd) {
```
luser: I know most of these `unwrap`s are unlikely to ever be a problem, but they do rub me the wrong way a bit. What if you instead made `parse` return a `Result<Command>`, and then tweaked main to accommodate and print errors? Then you could basically write the same code but with `try!`. You might need to create a new `Error` type with conversions from various other `Error`s for this to work properly. I've seen `error-chain` recommended for that, although I haven't used it: https://github.com/brson/error-chain

luser:
Eh, I'll make those changes if I care that much.

PR #50, opened 2016-12-16
Fix to enable cargo install

posix4e:


posix4e:
I can pull https://crates.io/crates/sccache if you want. But we are using it at brave

luser:
> I can pull https://crates.io/crates/sccache if you want. But we are using it at brave

No, that's OK, but can you add me as an owner? Also: what revision was the crate published against?

Issue #30, opened 2016-11-30
Two-tier caching


luser:
We'd like to be able to use a two-tier cache setup, where sccache can use one cache source as a read-only cache, and another as a read+write cache. This would be useful for two scenarios:

1. Allowing try server builds to get cache hits from the tier-3 branch buckets.
2. Allowing local developers to get cache hits from the buckets we use for CI.

Felixoid:
Not sure how many people follow this issue, but I'd like to get a review round in #2581 where I address this particular request.

Issue #12, opened 2016-07-01
Improve formatting of `--show-stats`


luser:
The formatting of the output of `sccache --show-stats` is kind of wonky since I added the `Cache Location` field. On my Linux machine it looks like:

```
$ ./target/debug/sccache --show-stats
Cache location                                    Local disk: "/tmp/sccache_cache"
Compile requests                                                                 1
Compile requests executed                                                        1
Cache hits                                                                       0
Cache misses                                                                     1
Cache read errors                                                                0
Compilation failures                                                             0
Cache errors                                                                     0
Successful compilations which could not be cached                                0
Non-cacheable calls                                                              0
Non-compilation calls                                                            0
Unsupported compiler calls                                                       0
```

On my Windows machine it's even worse:

```
$ ./target/debug/sccache.exe --show-stats
Cache location                                    Local disk: "C:\\Users\\TEDMIE~1\\AppData\\Local\\
Temp\\sccache_cache"
Compile requests
                   0
Compile requests executed
                   0
Cache hits
                   0
Cache misses
                   0
Cache read errors
                   0
Compilation failures
                   0
Cache errors
                   0
Successful compilations which could not be cached
                   0
Non-cacheable calls
                   0
Non-compilation calls
                   0
Unsupported compiler calls
                   0

```

The code that prints the stats output is here:
https://github.com/luser/sccache2/blob/8b8bbc339c7430583a0e48da9e3715084e5b0e77/src/commands.rs#L265

It tries to be fancy and align all of the values, but it'd probably be better to only align values that are a number and not a string (like the cache location). Also, we should probably change the name of `Successful compilations which could not be cached`, since that makes the output a bit longer than it needs to. Those strings are defined here:
https://github.com/luser/sccache2/blob/8b8bbc339c7430583a0e48da9e3715084e5b0e77/src/server.rs#L129

I don't know exactly what ccache does to align its output, but it looks nicer:

```
$ ccache -s
cache directory                     /home/luser/.ccache
cache hit (direct)                 14043
cache hit (preprocessed)           62635
cache miss                        382817
called for link                    55946
called for preprocessing            5417
multiple source files                703
compile failed                      7642
preprocessor error                  5592
bad compiler arguments              1578
unsupported source language         7729
autoconf compile/link              32244
unsupported compiler option          980
no input file                      96810
files in cache                     23731
cache size                           8.9 Gbytes
max cache size                      10.0 Gbytes
```

PR #52, opened 2016-12-18
Fix average write duration and extend server stats

flxo:
The average duration that is taken by a miss or hit is interesting
when testing different cache types. This patch adds the average time
taken by a cache miss and hit to the server stats. This measurements
can be entirely done in the server and do not need any changes to the
cache implementations.
Furthermore the calculation of the average duration for cache writes is
fixed (wrong factor used, that produced micros instead of millis).

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -146,7 +146,7 @@ struct ServerStats {
     pub cache_write_errors: u64,
     /// The number of successful cache writes.
     pub cache_writes: u64,
-    /// The total seconds spent writing cache entries.
+    /// The total time spent writing cache entries.
     pub cache_write_duration: Duration,
     /// The total time spent reading cache hits.
     pub cache_read_hit_duration: Duration,
--- a/src/server.rs
+++ b/src/server.rs
@@ -119,7 +119,6 @@ pub struct SccacheServer<C: CommandCreatorSync + 'static> {
 }
 
 /// Statistics about the cache.
-#[derive(Default)]
 struct ServerStats {
     /// The count of client compile requests.
     pub compile_requests: u64,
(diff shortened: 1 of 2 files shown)


flxo:
Just a note: I [played around with a moving average](https://github.com/flxo/sccache/tree/moving_average) calculation for the durations. The measurements are the same as in this PR.
If the server is constantly running or the stats are persisted this might be worth to consider - I just did't want to PR something that changes the current behavior...

src/server.rs
```diff
@@ -146,15 +145,41 @@ struct ServerStats {
...
+            cache_hits: u64::default(),
+            cache_misses: u64::default(),
+            non_cacheable_compilations: u64::default(),
+            forced_recaches: u64::default(),
+            cache_read_errors: u64::default(),
+            cache_write_errors: u64::default(),
+            cache_writes: u64::default(),
+            cache_write_duration: Duration::new(0, 0),
+            cache_read_hit_duration: Duration::new(0, 0),
+            cache_read_miss_duration: Duration::new(0, 0),
```
luser: I avoided doing this because I was too lazy to spell all this out, and `Duration` didn't have a `Default` implementation. [That'll be fixed in Rust 1.15](https://github.com/rust-lang/rust/issues/37546), but for now this is a reasonable thing to do.

src/server.rs
```diff
@@ -146,15 +145,41 @@ struct ServerStats {
     /// The count of errors writing to cache.
     pub cache_write_errors: u64,
     /// The number of successful cache writes.
-    pub cache_writes: u32,
+    pub cache_writes: u64,
     /// The total seconds spent writing cache entries.
```
luser: You could change the description of these to just "The total time spent..."

src/server.rs
```diff
@@ -166,6 +191,20 @@ impl ServerStats {
             }};
         }
 
+        macro_rules! set_duration_stat {
+            ($vec:ident, $dur:expr, $num:expr, $name:expr) => {{
+                let mut stat = CacheStatistic::new();
+                stat.set_name(String::from($name));
+                if $num > 0 {
+                    let duration = $dur / $num as u32;
+                    stat.set_str(format!("{}.{:03} s", duration.as_secs(), duration.subsec_nanos() / 1000_000));
```
luser: I really wish `Duration` had an `as_fractional_secs` method or something like that. [I filed an issue on that](https://github.com/rust-lang/rust/issues/38475). Also: oops on the conversion bug. I wish there was at least a constant in the stdlib we could use for it!

src/server.rs
```diff
@@ -166,6 +191,20 @@ impl ServerStats {
...
+                } else {
+                    stat.set_str("0.000 s".to_owned());
```
luser: I wonder if there'd be any value in adding a `Duration` case to the `CacheStatistic` enum? We don't currently do anything interesting with it, although I'd like to add a way to have `--show-stats` output a JSON representation of the stats for consumption by other programs. (Granted you're just duplicating what I previously implemented here...)

src/server.rs
```diff
@@ -146,15 +145,41 @@ struct ServerStats {
     /// The count of errors writing to cache.
     pub cache_write_errors: u64,
     /// The number of successful cache writes.
-    pub cache_writes: u32,
+    pub cache_writes: u64,
```
luser: Is there a reason you're changing this to `u64`? You're just doing a conversion in the `set_duration_stat` macro below anyway...

luser:
> Just a note: I played around with a moving average calculation for the durations. The measurements are the same as in this PR.
> If the server is constantly running or the stats are persisted this might be worth to consider - I just did't want to PR something that changes the current behavior...

I think the current behavior is fine for now. When we start persisting stats we might have to do something smarter. We're going to want to rework cache timing stats soon anyway, since we're going to implement support for [fetching from multiple cache sources](https://github.com/mozilla/sccache/issues/30) and other fanciness, so we'll probably want to store cache timing stats per-backend instance.

src/server.rs
flxo: Ups. I missed that and will update the PR.

src/server.rs
flxo: The `Duration` in `CacheStatistic` came to my mind when thinking about #12 in order to have a two column display with value and unit (like in the `ccache` output):

```

max cache size            5.0 GB
some duration             433 ms
other duration          0.324 s
```

luser: Yeah. We can fix that later, it would make fixing that nicer certainly.

src/server.rs
luser: I can't say I have strong feelings either way. I was mostly just curious to make sure I wasn't missing something important about the patch.

luser:
Thanks for the patch!

Issue #33, opened 2016-11-30
Save stats to disk


luser:
Currently the stats are only saved in-memory, so when the server shuts down they're lost. We should save them to disk like the Python sccache did. This is mostly useful for local developers, where they might expect the stats to be persistent like ccache's stats. If we do this we should also add a `--zero-stats` option to match ccache.

elichai:
now that there's `--zero-stats` option, what about persistant stats? it's really helpful to know cache hits/misses over time. Thanks!

poopsicles:
just wondering about this issue, it's bee bothering me for a while
#612 went inactive, but maybe saving it in the config location as maybe `sccache-stats.toml` would suffice?

mathstuf:
Like in https://github.com/mozilla/sccache/issues/1033#issuecomment-1953163097, can a label be used to distinguish these auto-closed inactive issues from actually resolved problems?

Issue #20, opened 2016-07-15
Look into replacing protobuf with serde


luser:
We're using protobuf for the client/server protocol right now. It's kind of a pain. We should look into just using serde instead, and defining a few Rust structs to serialize over the socket. I know serde is a bit difficult to use in non-nightly Rust, but if we get it working with nightly and it looks good we can jump through the hoops to make it work in stable Rust.

luser:
serde uses the "macros 1.1" feature that's on track for stabilization now, so it should be feasible to use this without requiring a nightly compiler soon.

alexcrichton:
Nowadays the way to implement this I believe would be:

* Update the server's [codec](https://github.com/mozilla/sccache/blob/6dba30a3afa512837598f079c5a94bd55468a101/src/server.rs#L814-L871) to not use protobuf, but keep the In/Out types the same
* Update the client's [parsing code](https://github.com/mozilla/sccache/blob/6dba30a3afa512837598f079c5a94bd55468a101/src/client.rs#L53-L77) to use the same logic as the server

luser:
This was fixed by #85.

PR #54, opened 2016-12-19
Zero stats if -z --zero-stats is passed on cmdline

flxo:
Straight forward implementation of clearing statistics when -z or --zero-stats is passed on command line.
This command also outputs the stats upon successful execution in order to provide feedback about the configured cache. The argument names are taken from `ccache`.

The stats are currently not persistent, but for testing this option appears useful to me, because I do not need to restart the server upon every testrun. Once the persistence is implemented, `sccache` should   provide a way to zero them - so take this patch as a preparation.

In order to make the patch less noisy, I'm happy to regenerate with the exact proto/protoc version used the last time, if someone can tell.


luser:
Thanks! This will be more useful when https://github.com/mozilla/sccache/issues/33 gets fixed, but I agree that having this in place is desirable either way. It looks like I used protobuf 2.6.1 to generate protocol.rs in the past, and rust-protobuf 1.0.23 (I'm not sure if that's important or not).

flxo:
I used protobuf `v1.0.19` for this update. Version `v1.0.23` showed some differences when generating on the current `master`...

luser:
I'm actually not sure if that's the version of `rust-protobuf` I have installed, I don't know that there's any way to query the installed binaries for their version.

luser:
Running strings on the binary leads me to believe it was compiled from rust-protobuf 1.0.21.

flxo:
To check the installed version run `cargo install --list`.
cargo also provides a way to force a specific version during install: `cargo install --force --vers 1.0.18 protobuf`. Unfortunately the `--vers` argument is broken (at least on `cargo 0.13.0-nightly (eca9e15 2016-11-01)` that comes with `stable`):
```
~ $ cargo --version
cargo 0.13.0-nightly (eca9e15 2016-11-01)
~ $ cargo install -f --vers 1.0.19 protobuf
Updating registry `https://github.com/rust-lang/crates.io-index`
Compiling protobuf v1.0.24
...
```

On `nightly` it's fine. I did some research in `cargo` but could not find a issue of commit where this was fixed or changed. Not sure whether I searched in the right place...

I tried all version of `protobuf` down to `v1.0.13` and only `v1.0.19` did *not* show any diff in the generated output compared to the current `HEAD`.

Anyway. Shouldn't we generate with the `protobuf` version `v1.0.18` that is required in `Cargo.toml`?
When I install `v1.0.18` and regenerate I see some differences against the current `HEAD`.

How to proceed? I see the following options:

1.  Open a separate PR with a generated `protocol.rs` from `protobuf v1.0.18` and rebase this PR afterwards.
2. I can submit this patch generated with `protobuf v1.0.18`
3. Do not care too much about the `protobuf` version, because it is a subject to removed - see #20 and use `v1.0.19`. This is what I did for `fad15b8`.
4. Use the version used by @luser. Ted, could you please check your setup with `cargo install --list`?

luser:
Thanks, TIL about `cargo install --list`! I do in fact have 1.0.21 installed:
protobuf v1.0.21:
    protobuf-bin-gen-rust-do-not-use
    protoc-gen-rust

I'm not too fussed about it, the generated protocol.rs is pretty ugly anyway, so let's just take this patch and not worry about the specific version.

PR #56, opened 2016-12-28
Add a Redis cache variant

flxo:
This patch adds support for using a [Redis](http://redis.io) instance as cache. The cache is used when `SCCACHE_REDIS` is set to a url that matches `redis://[:<passwd>@]<hostname>[:port][/<db>]`.  The S3 storage options overrule Redis.

For the connection to the Redis a connection pool is used. It is managed by [r2d2](https://github.com/sfackler/r2d2). The size of the connection pool can be controlled by `SCCACHE_REDIS_CONNECTIONS` and default to the number of cpus multiplied by 2. I'm not sure that this fallback is a good idea, but I wanted to have `SCCACHE_REDIS_CONNECTIONS` optional. What do you think?

The `current_size` method of `Storage` uses a lua script to sum the individual key sizes. This might not perform well on larger systems. During my tests this was not a problem for ~50k keys - but I limited the cache size to 10Gb. 

Everything else is quite straight forward. Some parts are borrowed from the `disk` and `S3` cache implementation. Thanks for that!

### DIFF
--- a/src/cache/redis.rs
+++ b/src/cache/redis.rs
@@ -1,0 +1,133 @@
+// Copyright 2016 Mozilla Foundation
+// Copyright 2016 Felix Obenhuber <felix@obenhuber.de>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use cache::{
+    Cache,
+    CacheRead,
+    CacheWrite,
+    Storage,
+};
+use errors::*;
+use futures::Future;
+use futures_cpupool::CpuPool;
+use redis::{
+    cmd,
+    Client,
+    Commands,
+    Connection,
+    InfoDict,
+};
+use std::collections::HashMap;
+use std::io::Cursor;
+use std::time::{
+    Duration,
+    Instant,
+};
+
+/// A cache that stores entries in a Redis.
+#[derive(Clone)]
+pub struct RedisCache {
+    url: String,
+    client: Client,
+    pool: CpuPool,
+}
+
+impl RedisCache {
+    /// Create a new `RedisCache`.
+    pub fn new(url: &str, pool: &CpuPool) -> Result<RedisCache> {
+        Ok(RedisCache {
+            url: url.to_owned(),
+            client: Client::open(url)?,
+            pool: pool.clone(),
+        })
+    }
+
+    /// Returns a connection with configured read and write timeouts.
+    fn connect(&self) -> Result<Connection> {
+        self.client.get_connection()
+            .map_err(|e| e.into())
+            .and_then(|c| {
+                c.set_read_timeout(Some(Duration::from_millis(10_000)))?;
+                c.set_write_timeout(Some(Duration::from_millis(10_000)))?;
+                Ok(c)
+            })
+    }
+}
+
+impl Storage for RedisCache {
+    /// Open a connection and query for a key.
+    fn get(&self, key: &str) -> SFuture<Cache> {
+        let key = key.to_owned();
+        let me = self.clone();
+        self.pool.spawn_fn(move || {
+            let c = me.connect()?;
+            let d = c.get::<&str, Vec<u8>>(&key)?;
+            if d.is_empty() {
+                Ok(Cache::Miss)
+            } else {
+                CacheRead::from(Cursor::new(d))
+                    .map(Cache::Hit)
+            }
+        }).boxed()
+    }
+
+    /// Initiate a cache write. There is nothing special needed
+    /// for the Redis cache.
+    fn start_put(&self, _key: &str) -> Result<CacheWrite> {
+        Ok(CacheWrite::new())
+    }
+
+    /// Open a connecxtion and store a object in the cache.
+    fn finish_put(&self, key: &str, entry: CacheWrite) -> SFuture<Duration> {
+        let key = key.to_owned();
+        let me = self.clone();
+        self.pool.spawn_fn(move || {
+            let start = Instant::now();
+            let c = me.connect()?;
+            let d = entry.finish()?;
+            c.set::<&str, Vec<u8>, ()>(&key, d)?;
+            Ok(start.elapsed())
+        }).boxed()
+    }
+
+    /// Returns the cache location.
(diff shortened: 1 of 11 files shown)


luser:
Hey there! Sorry for letting this sit. Similar to what I said on https://github.com/mozilla/sccache/pull/36, I'd like to stick this behind a [cargo feature](http://doc.crates.io/manifest.html#the-features-section) so that sccache users don't have to start paying extra build time cost for support for a growing number of backends that they aren't using. If you can make that change then merging this will be a lot more straightforward, since people will have to opt-in to building it.

flxo:
Hi. No Problem with the delay - so I had some time to let this patch prove itself in our CI. I prepared a patch to make the Redis (and S3) cache optional but want to wait with the update of the PR until your master is running again.

flxo:
I add the requested cargo features. Unfortunately I had to use `redis-cache` and `s3-cache` instead of `redis` and `s3` because the `redis` crate would have the same name as the dep what is illegal.

The minimum version to build `sccache` increased to `1.13.0` because of the `cfg!` macros.
I extended the CI configuration to build the various `feature` variants, but this is maybe a bit messy and one build with the feature `all` (and `unstable` on nightly) is fine.

The Redis cache also got some read and write timeouts on the connections which seem to be necessary for proper failing on network or server issues. 

The biggest change might be, that the `S3` cache is optional by now - enable it by default if this is not ok.
@luser: Let me know if I can improve the one or other thing. I also don't know whether I got you right about the cargo feature stuff...but at least tried. Thanks!

flxo:
This connection pool thing is IMHO a not so good idea to be used with `Redis` and `sccache`. Being this week out of office with a poor connection to our cache run into problems getting the `sccache` server started before the `sccache` client runs into timeouts and so on, so I started to evaluate timings with a new connection on every `get` and `finish_put` and could not notice any - no noticeable change on the server load as well.I will e valuate this further.
Getting rid of the pool would also reduce dependencies make the `Redis` cache just a couple of trivial lines.

flxo:
I revised this PR. The connection pool is gone because it really felt odd. Especially the determination of the pool size was nothing more than a weak guess. The version without the pool ran for a week in our company CI and overall no impact on the build times is measurable. Removing the pool also reduced the dependencies and the `redis` dependency name got available because the Redis cache module has no other deps than `redis`. So I replaced the dependency names `redis-cache` and `s3-cache` with `redis` and `s3` which feel much better.
Compared to the first version of the Redis patch I also shortened and simplified the cache size calculations. The `INFO` command is used instead of the crazy strlen with lua solution.
Of course the pathes are rebased after the `tokio` patch got applied.

Still the `S3` cache is optional, but as I wrote before I understand that it might be added to the defaults in order to keep the current default feature set.

alexcrichton:
My guess is that https://github.com/mozilla/sccache/pull/70 will have a bunch of merge conflicts with this branch as well, but it should in theory be relatively easy to resolve!

If you run into any issues just let me know and I can try to help out resolving conflicts (or I'll do them all myself if this merges first!)

flxo:
No worries - I think checking your changes at the disk cache will give all I need for the rebase. Thanks for your offer - I will ask if in doubt. Due I'm learning, review comments are always very welcome!

flxo:
Rebased. There are two things I'm not super happy with: First is the wrapper struct around the client that is used to satisfy `'static` and the extended help text. Maybe someone has a better idea than me.

src/cache/redis.rs
```diff
@@ -0,0 +1,147 @@
...
+impl Client {
+    /// Create a new Client with a configured url.
+    fn new(url: &str) -> Result<Client> {
+        Ok(Client {
+            client: ::redis::Client::open(url)?,
+        })
+    }
+
+    /// Returns a connection with configured read and write timeouts.
+    fn connect(&self) -> ::std::result::Result<Connection, RedisError> {
```
alexcrichton: Could just `Result<Connection>` be used here perhaps?

alexcrichton:
This looks great to me!

> First is the wrapper struct around the client that is used to satisfy 'static 

Could you elaborate a bit on this? It wasn't immediately obvious to me at least why the wrapper struct was necessary

src/cache/redis.rs
flxo: Yep. Sure. Looks better!

flxo:
I added this wrapper in order to avoid a clone of the entire `RedisCache` struct before moving. This is probably a premature optimisation, because the overhead is just a `String` and the `CpuPool` which contains a single `Arc`.  Holding `redis::Client` directly in `RedisCache` would make me duplicate the code for setting the read and write timeouts on every connection use... Removing the timeouts is IMHO not an option. Rethinking this, I'd replace the wrapper with a full clone.
 
I noticed that you used `let me = self.clone();` and I wonder whether the `me` is common binding term for `self` clones? Never seen this before...

Thanks for your review!

alexcrichton:
Ah ok, makes sense, thanks for clarifying! I personally have done `let me = self.clone()`, but I've also seen others do `let this = ...`. I wouldn't call it a pattern per se though :)

flxo:
Rebased. The Travis problems with OSX seem to be fixed, so this rebase should give a green ci result.

alexcrichton:
Looks great to me!

Could you also add commands to the Travis configuration for building each feature independently, as well as altogether? Other than that r=me, but @luser may have more comments.

flxo:
Sure! Doing loops in travis scripts is a bad idea because the build doesn't fail if one of the iterations fails. Due to fact that there are currently only two features, I just added build commands in the scripts. Let me know, if you prefer to have something like  `scripts/ci.sh` with some for looping of a list of features.

alexcrichton:
Looks great to me!

flxo:
I rebased and took the opportunity to squash the three `redis` commits. The patch that introduces the `cargo` features is still standalone. Comments are always welcome!
cheers, @flxo

luser:
Thanks! We'll get this landed once @alexcrichton's outstanding PRs get merged. Sorry to let it sit around so long, I've had a lot of work in progress and it's been hard to manage merging PRs during that time.

alexcrichton:
Thanks @flxo!

flxo:
Thanks @luser and @alexcrichton! Your work here is really awesome!

Nercury:
Thanks @flxo for this patch, we will be trying it out on our company CI!

flxo:
Glad to hear that we're not the only ones. ;-). For the Redis config you should consider especially the `maxmemory` and `maxmemory-policy` setting. We use `allkeys-lru` which seems to be a perfect match.

Nercury:
Thanks, it all seems to be working great. However sometimes sccache falls back to using filesystem backend. Is there some way to configure it to always prefer redis? Right now I am invoking it with `SCCACHE_REDIS` env variable, which should in theory start correct daemon on the first call. But it would be better to use redis as default.

luser:
I've been talking about adding support for a config file for a while, but I never opened an issue about it, so I did that: https://github.com/mozilla/sccache/issues/129.

PR #60, opened 2017-01-13
Remove out file from hash generation

flxo:
The name and path of the output file of a compilation shall not be used for the hash calculation. This improves the hit rate when building two modules `A` and `B` and both have a dependency on `C` **and** the build systems composes the object file path from the path of the module that introduces the dependency.

Example:

Three individual modules exist with on source file each:

```
A/a.c
B/b.c
C/c.c
```

where A and B depend on C. A and B are independent projects.

The build systems produces two objects when building A:

```
A/build/a.o
C/build_A/c.o
```

and when building B:

```
B/build/b.o
C/build_B/c.o
``````

The `build_A` and `build_B` is (sadly) the default behaviour of the [build system](https://github.com/esrlabs/bake).

Without this patch `sccache` would end in a cache miss when building `c.c` the second time (as a dep of B). This scenario often happens when `C` is a common thing like e.g `gtest` etc...

This patch also removes the artificial space that is used during iteration over the arguments because it also should not have a effect in the result and the hash.

To check that the output file has no effect on the result do the following (same for gcc):

```
# clang -c test.c -o test0.o
# clang -c test.c -o test1.o
# shasum test0.o test1.o
-> 5999b7872181e9121ad17d33b828757a86318dc2  test0.o
-> 5999b7872181e9121ad17d33b828757a86318dc2  test1.o
```

The `args` can be passed in various ways to `hash_key`. I'm unsure if a `&str`, `&String` would result in better performance.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -308,15 +308,6 @@ mod test {
     }
 
     #[test]
-    fn test_hash_key_out_file_differs() {
-        let f = TestFixture::new();
-        let c = Compiler::new(f.bins[0].to_str().unwrap(), CompilerKind::Gcc).unwrap();
-        const PREPROCESSED : &'static [u8] = b"hello world";
-        assert_neq!(hash_key(&c, vec!["a", "b", "-o", "somefile"], &PREPROCESSED),
-                    hash_key(&c, vec!["a", "b", "-o", "otherfile"], &PREPROCESSED));
-    }
-
-    #[test]
     fn test_hash_key_preprocessed_content_differs() {
         let f = TestFixture::new();
         let c = Compiler::new(f.bins[0].to_str().unwrap(), CompilerKind::Gcc).unwrap();
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -224,16 +224,14 @@ pub const CACHED_ENV_VARS : &'static [&'static str] = &[
 
 /// Compute the hash key of `compiler` compiling `preprocessor_output` with `args`.
 #[allow(dead_code)]
-pub fn hash_key(compiler: &Compiler, args: Vec<&str>, preprocessor_output: &[u8]) -> String {
+pub fn hash_key(compiler: &Compiler, arguments: &str, preprocessor_output: &[u8]) -> String {
     // If you change any of the inputs to the hash, you should change `CACHE_VERSION`.
     let mut m = sha1::Sha1::new();
     m.update(compiler.digest.as_bytes());
     //TODO: drop the compiler filename from the hash
     m.update(compiler.executable.as_bytes());
     m.update(CACHE_VERSION);
-    for arg in args {
-        m.update(arg.as_bytes());
-    }
+    m.update(arguments.as_bytes());
     //TODO: should propogate these over from the client.
     // https://github.com/glandium/sccache/issues/5
     for var in CACHED_ENV_VARS.iter() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by luser** [comment]
I thought that maybe with debug info the object filename would be contained in the output, but I was wrong, so I think this is fine. (I repeated your test with -g and the files are the same either way.)

@glandium: if we're going to change the hash version anyway, maybe we should also change the hash algorithm like we discussed? (Context: we found out that the sha1 crate is way slower than other sha1 implementations, and the sha256+ implementations in ring are at least twice as fast because they're optimized assembly.)

**Comment by flxo** [comment]
Thanks @luser.

There is a TODO in the `hash_key` method to remove the path of the compiler from the hash. I could imagine that changing the hash version is sensible because of global ;-) invalidation of all caches... So maybe this one should be remove as well?

**Comment by luser** [comment]
I had to do some fixup to rebase this on top of the tokio changes that landed, but I just pushed it:
https://github.com/mozilla/sccache/commit/69ca2c1932ef7b79f1e4fce51985f9cbe935fab9

Thanks for the patch!

I took the opportunity to land an additional patch on top of that to fix that TODO about removing the compiler path from the hash, since we were bumping the hash version anyway:
https://github.com/mozilla/sccache/commit/9895290a2144e78a202eb6518840dfb62f98fd48

**Comment by flxo** [comment]
Thanks for merging!

By reading the patch after the merge I noticed something I'm not sure about. I think this PR introduced a bug. Furthermore it does not work at all:

The update [e6fafcb](https://github.com/mozilla/sccache/pull/60/commits/e6fafcb8164f5aa47b0699f0c9eb5ee17cabf140) with the title

`Fix hash calculation by using arguments instead of parsed args`

in fact does something different. It doesn't change the relevant part for the hash calculation and `parsed_args.common_args` is still used. The patch just changed the `hash_key` argument type, but nothing else.  I'm sorry for this embarrassing flaw!

Next, when using `parsed_args`, the patch doesn't work because the `out_file` is not part of the `common_args` with the current compiler implementations. This doesn't break anything, but makes the patch totally useless.

More important is, that the current behaviour of the hash calculation is broken. Using the parsed arguments, omits the input file for the hash calculation. The input file is not part of the `parsed_args` list. At least when compiling with `-g`, `sccache` could produce equals keys for different objects:

```
~/tmp  mkdir a
~/tmp  mkdir b
~/tmp  touch a/test.c b/test.c
~/tmp  clang -c a/test.c -o a/test.o
~/tmp  clang -c b/test.c -o b/test.o
~/tmp  shasum a/test.o b/test.o
42a4083c9de9c3b8b851655de09bcc473143fea1  a/test.o
42a4083c9de9c3b8b851655de09bcc473143fea1  b/test.o
~/tmp  clang -g -c a/test.c -o a/test.o
~/tmp  clang -g -c b/test.c -o b/test.o
~/tmp  shasum a/test.o b/test.o
06ba230ca591c075d4d02bb1914ca56a29795441  a/test.o
d89ccd93bdc7fea571b269ea1534d77d6752856d  b/test.o
```

Fixing this is not as easy as it might look in the first place:
Iterating over `arguments` instead of `parsed_args.common_args` and filtering on `out_file` seems to be obvious. This only works when the object file is a separate argument as it is for clang and gcc. I'm not so familiar with msvc, but looking at the argument parser 
```
119                     v if v.starts_with("-Fo") => {
120                         output_arg = Some(String::from(&v[3..]));
121                     }
```
shows, that the output file is declared here by somethings like `-Fofoo.o`. A stupid string replace might break if someone uses a argument that equals or contains the output file name, e.g `-Dfoo.o` or `-deps=foo.o.d`. Ok, depfile generation is done during the preprocessor run - I just want to point out what I mean. I think that the usage of `sccache` should never ever produce something different from a compilation done without it, so the simple replace is not an option.

What do you think? Do I miss something?

I'd suggest to revert the patches from this PR until this is clarified. Once again sorry!

cheers,

Felix

flxo:
Short note: Using the `parsed_args` might be fine. They contain everything relevant for the compilation but not the object and source file. So adding the source there, before passing to `hash_key` should do the trick. This could look like [this](https://github.com/flxo/sccache/commit/e5031a43e26013926b647f22457a039e9eb36f62).

glandium:
The source path is all over the preprocessed source, which is hashed too. But even better, it's in a normalized form, such that it's the same whether the compiler was invoked with a relative path or an absolute path.

flxo:
Yep. The hash over the preprocessed source is stroke of genius - really like that idea. But when it comes to hits, the source file name baked into the object matters. Tools looking at the debug information might get confused. I've seen this with the `valgrind` plugin on a `Jenkins`, that tries to generate a nice report with included source snippets. Same for [Backward-cpp](https://github.com/bombela/backward-cpp) that we link into some of our C++ projects.
The problematic hits don't need to occur on different machines. Two projects that use a identical source (and compiler args of course) is enough - a common case for things like `gtest.cc` for example. People get confused by pathes from different projects popping up here even if the snippet is identical (and correct - otherwise the lookup would have failed...).

PR #65, opened 2017-01-30
Add support for gcc `@file` options

### LINKED ISSUE #43: support for gcc compiler options being a file via @file option
we are using a file to specify our compiler options to gcc via the @file option. I have tested a trivial compilation and as soon as I do this I see the stats counting this compile as Non-cacheable calls. I believe it's hitting the gcc.rs:109. What I find misleading the the comment in the previous line talking about using a response file. 

any command line parameter starting with @ should be a file with the gcc options. Would be awesome to have support for that.

https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options
@file
Read command-line options from file. The options read are inserted in place of the original @file option. If file does not exist, or cannot be read, then the option will be treated literally, and not removed.
Options in file are separated by whitespace. A whitespace character may be included in an option by surrounding the entire option in either single or double quotes. Any character (including a backslash) may be included by prefixing the character to be included with a backslash. The file may itself contain additional @file options; any such options will be processed recursively.

alexcrichton:
This commit adds support for the `@file` option that gcc/clang supports. This
option means that an `file` should be read and `@file` should be replaced with
all the options specified in `file`.

The online documentation indicates that gcc supports arguments with spaces
through quoting, but this seemed like it may be nontrivial to implement, so I
figured that for now those cases could continue to be un-cacheable.

Closes #43

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -95,7 +95,7 @@ mod test {
     use test::utils::*;
 
     fn _parse_arguments(arguments: &[String]) -> CompilerArguments {
-        gcc::parse_arguments(arguments, argument_takes_value)
+        gcc::parse_arguments(arguments, ".".as_ref(), argument_takes_value)
     }
 
     #[test]
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -69,12 +69,14 @@ pub enum CompilerKind {
 }
 
 impl CompilerKind {
-    pub fn parse_arguments(&self, arguments: &[String]) -> CompilerArguments {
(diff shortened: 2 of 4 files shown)


luser:
Thanks for the patch! [MSVC also supports `@file`](https://msdn.microsoft.com/en-us/library/3te4xt0y(v=vs.100).aspx), it would be nice to add that in a followup. One difference is that the GCC docs explicitly state that recursive use of @ is OK, but the MSVC docs explicitly say it's not. It's not likely to be a problem in practice since I doubt anyone would be using a construct like that.

PR #66, opened 2017-01-30
Remove unused import

nierob:
warning: unused import: `std::time::Duration`, #[warn(unused_imports)] on by default
   --> src/commands.rs:107:9
    |
107 |     use std::time::Duration;
    |         ^^^^^^^^^^^^^^^^^^^


luser:
Oops, must have missed that. Thanks for the patch!

PR #68, opened 2017-01-31
Rewrite the server module with Tokio

alexcrichton:
This commit rewrites the `server` module of sccache to be backed with Tokio. The
previous version was written with `mio`, which Tokio is built on, but is
unfortunately less ergonomic. Tokio is the state-of-the-art for asynchronous
programming in Rust and sccache serves as a great testing ground for ergonomics!

It's intended that the support added here will eventually extend to many other
operations that sccache does as well. For example thread spawning has all been
replaced with `CpuPool` to have a shared pool for I/O operations and such
(namely the filesystem). Eventually the HTTP requests made by the S3 backend can
be integrated with the Tokio branch of Hyper as well to run that on the event
loop instead of in a worker thread. I'd also like to eventually extend this with
`tokio-process` as well to move process spawning off helper threads as well, but
I'm leaving that to a future commit as well.

Overall I found the transition was quite smooth, with the high level
architecture look like:

* The `tokio-proto` crate is used in streaming mode. The streaming part is used
  for the one RPC sccache gets which requires a second response to be sent later
  on. This second response is the "response body" in tokio-proto terms.
* All of sccache's logic is manifested in an implementation of the `Service`
  trait.
* The transport layer is provided with `tokio_core::io::{Framed, Codec}`, and
  simple deserialization/serialization is performed with protobuf.

Some differences in design are:

* The `SccacheService` for now is just a bunch of reference-counted pointers,
  making it cheap to clone. As the futures it returns progress they will each
  retain a reference to a cloned copy of the `SccacheService`. Before all this
  data was just stored and manipulated in a struct directly, but it's now
  directly managed through shared memory.
* The storage backends share a thread pool with the main server instead of
  spawning threads.

And finally, some things I've learned along the way:

* Sharing data between futures isn't a trivial operation. It took an explicit
  decision to use `Rc` and I'm not sure I'm 100% happy with how the ergonomics
  played out.
* Shutdown is pretty tricky here. I've tried to carry over all the previous
  logic but it definitely required not using `TcpServer` in tokio-proto at the
  very least, and otherwise required a few custom futures and such to track the
  various states. I have a hunch that tokio-proto could provide more options out
  of the box for something like this.
* Getting a transport up and running isn't always the easiest. I wanted
  to use sink/stream combinators here but was unable to unfortunately.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -18,8 +18,6 @@ use cache::{
 };
 use compiler::{
     CacheControl,
-    CacheWriteFuture,
-    CacheWriteResult,
     Compiler,
     CompilerArguments,
     CompileResult,
@@ -28,19 +26,17 @@ use compiler::{
     get_compiler_info,
 };
 use filetime::FileTime;
-use futures::Future;
-use mio::*;
-use mio::tcp::{
-    TcpListener,
-    TcpStream,
-};
-use mio::util::Slab;
+use futures::future;
+use futures::sync::mpsc;
+use futures::task::{self, Task};
+use futures::{Stream, Sink, Async, AsyncSink, Poll, StartSend, Future};
+use futures_cpupool::CpuPool;
 use mock_command::{
     CommandCreatorSync,
     ProcessCommandCreator,
 };
 use protobuf::{
-    Message,
+    self,
     ProtobufError,
     RepeatedField,
     parse_length_delimited_from_bytes,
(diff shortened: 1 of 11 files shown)


luser:
> Sharing data between futures isn't a trivial operation. It took an explicit
decision to use Rc and I'm not sure I'm 100% happy with how the ergonomics
played out.

Is this just because some of the futures wind up getting executed on the threadpool, or are there more general lifetime issues around futures? I struggled with this a bit in the existing sccache code, as you can see it was already using `Arc<Box<Storage>>`, for example. I can see what you mean here, though:
https://github.com/mozilla/sccache/pull/68/commits/3b13b682b01bf4628e682d832813fc72f5cd38cf#diff-4ce93534efc34e923ce01e975eb7ed80R539

It seems like there's some tension here between implementing traits on a struct for code clarity and the lifetime entanglement with `&self` that that entails. Maybe there's a pattern that would work better here? The data that the server carries around can definitely be partitioned into "things that need to be accessed on other threads" (like the `Storage` impl) and "things that only need to be accessed on the main thread" (like the listen socket and the stats).

> Shutdown is pretty tricky here

I actually realized that the current shutdown-handling code is not great--it should close the listen socket as soon as shutdown is initiated, but it currently just ignores client connections while shutdown is happening.

alexcrichton:
Yeah so for `Rc` and business the main problem is that all returned futures implement `'static`. This means that none of them persist borrowed references, such as closing over `&self`. They need to share data, however. For example all futures returned from a service need to hold onto a reference to `Stats`. For now those shared references are put behind a `Rc`.

Also yeah the partition works out pretty good today as there's `Rc` for data shared amongst futures on the on the main thread and then there's `Arc` for data like the `Storage` which may cross the thread boundary.

An alternative is to have a global `&'static Data` which contains stuff like `Stats` and such and then we just close over that. It leaks data but we could finagle it perhaps?

> I actually realized that the current shutdown-handling code is not great--it should close the listen socket as soon as shutdown is initiated, but it currently just ignores client connections while shutdown is happening.

Oh interesting! I think the shutdown behavior implemented in this PR should handle this by dropping the TCP listener as soon as the first shutdown notification is received.

luser:
> Yeah so for Rc and business the main problem is that all returned futures implement 'static. This means that none of them persist borrowed references, such as closing over &self. They need to share data, however. For example all futures returned from a service need to hold onto a reference to Stats. For now those shared references are put behind a Rc.

This is because `Future::boxed` has `Self: Sized + Send + 'static`, right? Using non-boxed Futures seemed unergonomic when I last tried it (similar to the issues with Iterator). It does feel like it would be nice to be able to have Futures with a non-static lifetime attached so we could pass around a reference to some shared state for some of them. I'm not sure how that'd shake out with chaining multiple Futures together where some of them might need to run on background threads.

alexcrichton:
Sorta yeah, the biggest part is the `'static` bound. All (or at least most I think) of the `Future` box trait objects in this PR are not `Send` (but that'd just be `Rc` vs `Arc`).

In general the future need access to `&self` at various points along the computation. In other words we need to close over `&self` in various `and_then` callbacks. We can't do that literally though because then the future wouldn't adhere to `'static`. To work around that we subvert this by doing a cheap clone of `&self` (e.g. `let me = self.clone();`) and then we can close over `me` and satisfy the `'static` bound.

Futures with a non-static lifetime is definitely something on our radar, and we're like to work through the details! This'll likely take awhile though, unfortunately, as there may need to be a language change or two accompanied with this.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+            timeout: Timeout::new(timeout, &handle)?,
+            handle: handle.clone(),
+            timeout_dur: timeout,
+        };
+
+        let server = server.select(shutdown_idle);
+        let shutdown = shutdown.map_err(|()| {
+            io::Error::new(io::ErrorKind::Other, "shutdown signal failed")
+        });
+        let server = server.map(|_| ()).map_err(|e| e.0).select(shutdown);
```
luser: I wonder if this wouldn't be slightly easier to read as:
```rust
let server = future::select_all([server, shutdown_idle, shutdown.map_err(...)]);
```
That should be functionally equivalent, right?

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+        let handle = core.handle();
+        let shutdown_idle = ShutdownOrInactive {
+            rx: rx,
+            timeout: Timeout::new(timeout, &handle)?,
+            handle: handle.clone(),
+            timeout_dur: timeout,
+        };
+
+        let server = server.select(shutdown_idle);
+        let shutdown = shutdown.map_err(|()| {
```
luser: These `map_err`s are kinda clunky. Does `error-chain` work well with `Future`s?

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+            .map_err(|p| p.0)?;
+
+        // Once our server has shut down either due to inactivity or a manual
+        // request we still need to give a bit of time for all active
+        // connections to finish. This `wait` future will resolve once all
+        // instances of `SccacheService` have been dropped.
+        //
+        // Note that we cap the amount of time this can take, however, as we
+        // don't want to wait *too* long.
+        core.run(wait.select(Timeout::new(Duration::new(1, 0), &handle)?))
```
luser: One second seems a little short for allowing real-world activity to complete. If we started a compile it could easily take longer than that to finish up and cache the result. I guess this is only an issue when you force a shutdown, so alternately we could just document somewhere (in the `--stop-server` command help?) that outstanding compilations might not finish if you ask the server to shut down.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    /// Handle to the event loop that we're running on.
+    handle: Handle,
+
     /// An object for creating commands.
     ///
     /// This is mostly useful for unit testing, where we
     /// can mock this out.
     creator: C,
+
+    /// Message channel used to learn about requests received by this server.
```
luser: This could stand an additional comment that it's used to drive the idle shutdown timer as well as the explicit shutdown command.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    tx: mpsc::Sender<ServerMessage>,
+
+    /// Information tracking how many services are active.
```
luser: By `services` here you mean `active connected clients`, right?

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    info: ActiveInfo,
+}
+
+type SccacheRequest = Message<ClientRequest, Body<(), io::Error>>;
```
luser: I don't love the `streaming::Message` type, honestly. It feels very HTTP-centric. It would be nicer to be able to just say "a client request is a single `ClientRequest`" here, and then "a server response is one or more `ServerResponse`s" below without the arbitrary head/body distinction.

Maybe that's something we can make easier in `tokio-proto`.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+type SccacheResponse = Message<ServerResponse, Body<ServerResponse, io::Error>>;
+
+pub enum ServerMessage {
```
luser: A little explanatory comment about the purpose of this type would be nice.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    fn zero_stats(&self) -> CacheStats {
+        *self.stats.borrow_mut() = ServerStats::default();
+        self.get_stats()
+    }
+
+
+    /// Handle a compile request from a client.
+    ///
+    /// This will either start compilation and set a `CompileStarted`
+    /// response in `res`, or set an `UnhandledCompile` response in `res`.
```
luser: This comment seems confused. Does it just need copy-edits to match the method's new signature?

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    fn handle_compile(&self, mut compile: Compile)
+                      -> Box<Future<Item = SccacheResponse, Error = io::Error>>
+    {
+        let exe = compile.take_exe();
+        let cmd = compile.take_command().into_vec();
+        let cwd = compile.take_cwd();
+        // See if this compiler is already in the cache.
+        match self.compiler_info_cached(&exe) {
```
luser: Feels like this could be rewritten so the cache lookup just always returned a `Future` to simplify the logic, like:
`self.get_compiler_info(&exe).then(|info| self.check_compiler(info, cmd, cwd))`

When I wrote that code originally I was actually wishing for some sort of `Future`-like abstraction!

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    fn check_compiler(&self,
+                      compiler: Option<Compiler>,
+                      cmd: Vec<String>,
+                      cwd: String)
+                      -> SccacheResponse {
+        let mut res = ServerResponse::new();
+        match compiler {
+            None => {
+                debug!("check_compiler: Unsupported compiler");
+                self.stats.borrow_mut().requests_unsupported_compiler += 1;
```
luser: Having to put `borrow_mut()` everywhere here is not wonderful. I can live with it for now but I'll have to think about the design some more at another point.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+                    }
+                }
+            }
+        }
+
+        res.set_unhandled_compile(UnhandledCompile::new());
+        Message::WithoutBody(res)
+    }
+
+    /// Start running `cmd` in a background task, in `cwd`.
```
luser: Could clarify here--"on a thread in the thread pool".

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+
+                    Ok(None) => {}
+                }
+                Ok(())
+            });
+
+            send.join(cache_write).then(|_| Ok(()))
+        });
+
+        self.handle.spawn(task);
```
luser: This whole method is a lot nicer having tokio available! The existing code with just a `Future` was pretty clunky.

src/server.rs
```diff
@@ -227,855 +725,233 @@ impl ServerStats {
...
     }
+}
 
-    /// Get the port on which this server is listening for connections.
-    #[allow(dead_code)]
-    pub fn port(&self) -> u16 { self.sock.local_addr().unwrap().port() }
+/// Implementation of `Stream + Sink` that tokio-proto is expecting. This takes
+/// a `Framed` instance using `ProtobufCodec` and performs a simple map
+/// operation on the sink/stream halves to translate the protobuf message types
+/// to the `Frame` types that tokio-proto expects.
```
luser: Definitely feels like `tokio-proto` could provide something to make this use case simpler. The whole `SccacheTransport` implementation seems like something we shouldn't have to write here. It'd be nice if `Framed<I, ProtobufCodec<ClientRequest, ServerResponse>>` was enough.

src/server.rs
```diff
@@ -227,855 +725,233 @@ impl ServerStats {
...
-                event_loop.shutdown();
-            }
-        } else {
-            debug!("reset connection; token={:?}", token);
-            self.conns.remove(token);
-            self.check_shutdown(event_loop);
-        }
-    }
+/// Simple tokio-core `Codec` which uses stock protobuf functions to
+/// decode/encode protobuf messages.
```
luser: I love this part of tokio's design! When I eventually rip out protobuf and replace it with something based on serde this will make it a lot easier.

src/server.rs
```diff
@@ -227,855 +725,233 @@ impl ServerStats {
...
-    /// send the client an `UnhandledCompile` message.
-    fn check_compiler(&mut self, compiler: Option<Compiler>, cmd: Vec<String>, cwd: String, token: Token, event_loop: &mut EventLoop<SccacheServer<C>>) {
-        match compiler {
-            None => {
-                debug!("check_compiler: Unsupported compiler");
-                self.stats.requests_unsupported_compiler += 1;
-                self.send_unhandled_compile(token, event_loop);
+        match parse_length_delimited_from_bytes::<Request>(buf.as_slice()) {
+            Ok(req) => {
+                let size = req.write_length_delimited_to_bytes().unwrap().len();
```
luser: The original code was using `req.compute_size()`, which seems more appropriate.

luser:
FYI the failing Travis builds are just the Rust 1.12 builds failing:
https://github.com/mozilla/sccache/blob/4415a1e8df25db114e86d5434ae9519fd83b7a89/.travis.yml#L5

You can just change that to whatever version it was that introduced the `?` operator.

src/test/tests.rs
```diff
@@ -60,33 +59,44 @@ struct ServerOptions {
...
+                 srv.set_idle_timeout(Duration::from_millis(timeout));
             }
         }
-        let port = server.port();
-        let sender = event_loop.channel();
-        let creator = server.command_creator();
-        tx.send((port, sender, creator)).unwrap();
-        run_server(server, event_loop).unwrap()
+        let storage = Arc::new(DiskCache::new(&cache_dir, cache_size, srv.pool()));
+        srv.set_storage(storage.clone());
```
alexcrichton: Ah yeah the next PR queued up pushes futures all the way throughout caches and such, and I could totally plumb them that way. Right now the thread pool is passed into constructors but that could be easily changed I believe.

src/test/tests.rs
```diff
@@ -60,33 +59,44 @@ struct ServerOptions {
...
-    let (port, sender, creator) = rx.recv().unwrap();
-    (port, sender, creator, handle)
+    let (port, creator) = rx.recv().unwrap();
+    (port, shutdown_tx, creator, handle)
 }
 
 #[test]
 fn test_server_shutdown() {
     let f = TestFixture::new();
-    let (port, _, _, child) = run_server_thread(&f.tempdir.path(), None);
```
alexcrichton: Oh this was actually super subtle IIRC in the sense that `let (a, _) = foo()` immediately destroys the second value. Destroying the second value here, `_sender`, would cause the event loop to exit immediately which wouldn't be desired.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+/// Start an sccache server, listening on `port`.
+///
+/// Spins an event loop handling client connections until a client
+/// requests a shutdown.
+pub fn start_server(port: u16) -> io::Result<()> {
+    let res = SccacheServer::<ProcessCommandCreator>::new(port);
+    let notify = env::var_os("SCCACHE_STARTUP_NOTIFY");
+    match res {
+        Ok(srv) => {
+            notify_server_startup(&notify, true)?;
```
alexcrichton: 👍

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+        let core = Core::new()?;
+        let handle = core.handle();
+        let addr = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), port);
+        let listener = TcpListener::bind(&SocketAddr::V4(addr), &handle)?;
+
+        // Prepare the service which we'll use to service all incoming TCP
+        // connections.
+        let (tx, rx) = mpsc::channel(100);
+        let pool = CpuPool::new(20);
+        let storage = storage_from_environment(&pool);
```
alexcrichton: Cool sounds good to me, I'll change to that.

src/server.rs
alexcrichton: Nah unfortunately error-chain isn't *great* with futures today but I've been looking forward to integrating it more into sccache, both error-chain itself and seeing if I can make the error propagation nicer (unsure of what that would look like)

src/server.rs
alexcrichton: Hm good point yeah if there's lingering compilations. Does 10s sound ok?

It looks like documentation may not fit on the command line, but perhaps in the README?

src/server.rs
alexcrichton: Yeah currently `SccacheService` has a 1:1 correspondance with clients as well, but I'll comment to that effect.

src/server.rs
alexcrichton: Ah yeah this is definitely something I could switch up if you'd like. The `Message` type is currently mandated by tokio-proto and yeah is somewhat geared towards HTTP (or other request-then-a-ton-of-bytes protocols).

Ideally the request is in "easy" mode (not streaming) and only the response is in streaming mode (possibly up to two), but I don't think there's a great way to express that. I could change up the `SccacheResponse` though to `()` as the header and send both responses over the channel though if you'd prefer?

src/server.rs
alexcrichton: Eh I'll lift it out for now as I think it makes it at least a little better

src/server.rs
alexcrichton: Yeah that's *almost* enough for this use case except for the one RPC that sends two responses for one request. Otherwise we could use the non-streaming version of tokio-proto but with that we're forced to use the streaming module which brings in some more complexity :(

src/server.rs
alexcrichton: Yeah I tried using that but unfortunately it doesn't account for the header saying how big the request was, so it ended up causing an infinite loop with the codec :(

src/server.rs
alexcrichton: Yeah this is where I felt like futures really started to shine, the overall flow ended up being much easier to follow afterwards.

src/server.rs
```diff
@@ -58,65 +54,567 @@ use protocol::{
...
+    pub fn run<F>(self, shutdown: F) -> io::Result<()>
+        where F: Future,
+    {
+        self._run(Box::new(shutdown.then(|_| Ok(()))))
+    }
+
+    fn _run<'a>(self, shutdown: Box<Future<Item = (), Error = ()> + 'a>)
+                -> io::Result<()>
+    {
+        let SccacheServer { mut core, listener, rx, service, timeout, wait } = self;
```
alexcrichton: Yeah I figured it'd provide a good entry point to configure tests and other such style-shims but I'd also be totally fine just inlining above in `start-server`

alexcrichton:
Ok, updated!

src/server.rs
luser: We can do that as a followup refactor, no need to do it here.

src/test/tests.rs
luser: Ah, that makes sense but is indeed a bit of a subtle footgun!

src/server.rs
luser: Bleh. I would like to switch to something that's not protobuf here, it's not very usable.

src/server.rs
luser: I'm OK with this for now since it's using types from `tokio-proto`, but maybe we can figure out something nicer and fix things upstream to work better here.

alexcrichton:
To get a better idea of the long term impact here I've posted https://github.com/mozilla/sccache/pull/70 which has "futures everywhere" instead of just futures in the server module itself.

luser:
OK, I'm just going to skim through the patch one more time and merge it.

src/server.rs
```diff
@@ -58,65 +54,575 @@ use protocol::{
...
+                                           &c,
+                                           cache_control)
+        });
+
+        let me = self.clone();
+        let task = result.then(move |result| {
+            let mut res = ServerResponse::new();
+            let mut finish = CompileFinished::new();
+            let mut cache_write = None;
+            let mut stats = me.stats.borrow_mut();
```
luser: That's definitely nicer.

luser:
Thanks for the patch!

PR #69, opened 2017-02-02
Fix a regression in us-east-1 buckets

alexcrichton:
It looks like the logic in #27 changed the default host name from
$bucket.s3.amazonaws.com to s3.amazonaws.com, but didn't accompany with changes
to upload to /$bucket in the path name. This restores the original behavior by
uploading to $bucket.s3.amazonaws.com without requiring changes to how paths are
constructed.


luser:
Thanks! We should really have some tests that run against an actual s3 bucket. :-/

PR #70, opened 2017-02-15
Plumb futures through caching/spawning layer, leverage error-chain

alexcrichton:
This PR builds on https://github.com/mozilla/sccache/pull/68 by plumbing futures all the way through the caching layer, effectively future-izing the entire server. The master branch of hyper is used for HTTP requests and tokio-process is used for process management.

This also cleans up a bit along the way related to error handling by leveraging error-chain in as many places as possible. To facilitate that an `SFuture` ("sccache future") alias was introduced to easily pass around and return types of futures.

Note that this should probably hold off on merging until https://github.com/mozilla/sccache/pull/68 lands, but I figured it'd be interesting at least!

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -38,6 +38,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
+use tokio_core::reactor::Handle;
 use zip::{
     CompressionMethod,
     ZipArchive,
@@ -100,7 +101,7 @@ impl CacheRead {
 }
 
 /// A `Future` that may provide a `CacheWriteResult`.
-pub type CacheWriteFuture = Box<Future<Item=Duration, Error=String> + Send>;
+pub type CacheWriteFuture = Box<Future<Item=Duration, Error=String>>;
 
 /// Data to be stored in the compiler cache.
 pub struct CacheWrite {
@@ -177,7 +178,7 @@ fn parse_size(val: &str) -> Option<usize> {
 }
 
 /// Get a suitable `Storage` implementation from the environment.
-pub fn storage_from_environment(pool: &CpuPool) -> Arc<Storage> {
+pub fn storage_from_environment(pool: &CpuPool, handle: &Handle) -> Arc<Storage> {
     if let Ok(bucket) = env::var("SCCACHE_BUCKET") {
         let endpoint = match env::var("SCCACHE_ENDPOINT") {
             Ok(endpoint) => format!("{}/{}", endpoint, bucket),
@@ -188,7 +189,7 @@ pub fn storage_from_environment(pool: &CpuPool) -> Arc<Storage> {
             },
         };
         debug!("Trying S3Cache({})", endpoint);
-        match S3Cache::new(&bucket, &endpoint, pool) {
+        match S3Cache::new(&bucket, &endpoint, handle) {
             Ok(s) => {
                 trace!("Using S3Cache");
                 return Arc::new(s);
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -20,7 +20,6 @@ use app_dirs::{
 use cache::disk::DiskCache;
 use cache::s3::S3Cache;
 use compiler::Compiler;
-use futures::Future;
 use futures_cpupool::CpuPool;
 use regex::Regex;
 use sha1;
@@ -28,8 +27,6 @@ use std::env;
 use std::fmt;
 use std::io::{
     self,
-    Error,
-    ErrorKind,
     Read,
     Seek,
     Write,
@@ -44,6 +41,8 @@ use zip::{
     ZipArchive,
     ZipWriter,
 };
+
+use errors::*;
 
 //TODO: might need to put this somewhere more central
 const APP_INFO: AppInfo = AppInfo {
@@ -85,23 +84,24 @@ pub struct CacheRead {
 
 impl CacheRead {
     /// Create a cache entry from `reader`.
-    pub fn from<R: ReadSeek + 'static>(reader: R) -> io::Result<CacheRead> {
-        let z = try!(ZipArchive::new(Box::new(reader) as Box<ReadSeek>).or(Err(Error::new(ErrorKind::Other, "Failed to parse cache entry"))));
+    pub fn from<R: ReadSeek + 'static>(reader: R) -> Result<CacheRead> {
+        let z = ZipArchive::new(Box::new(reader) as Box<ReadSeek>).chain_err(|| {
+            "Failed to parse cache entry"
+        })?;
         Ok(CacheRead {
             zip: z,
         })
     }
 
     /// Get an object from this cache entry at `name` and write it to `to`.
-    pub fn get_object<T: Write>(&mut self, name: &str, to: &mut T) -> io::Result<()> {
-        let mut file = try!(self.zip.by_name(name).or(Err(Error::new(ErrorKind::Other, "Failed to read object from cache entry"))));
-        try!(io::copy(&mut file, to));
+    pub fn get_object<T: Write>(&mut self, name: &str, to: &mut T) -> Result<()> {
+        let mut file = self.zip.by_name(name).chain_err(|| {
+            "Failed to read object from cache entry"
+        })?;
+        io::copy(&mut file, to)?;
         Ok(())
     }
 }
-
-/// A `Future` that may provide a `CacheWriteResult`.
-pub type CacheWriteFuture = Box<Future<Item=Duration, Error=String>>;
 
 /// Data to be stored in the compiler cache.
 pub struct CacheWrite {
@@ -117,18 +117,19 @@ impl CacheWrite {
     }
 
     /// Add an object containing the contents of `from` to this cache entry at `name`.
-    pub fn put_object<T: Read>(&mut self, name: &str, from: &mut T) -> io::Result<()> {
-        try!(self.zip.start_file(name, CompressionMethod::Deflated).or(Err(Error::new(ErrorKind::Other, "Failed to start cache entry object"))));
-        try!(io::copy(from, &mut self.zip));
+    pub fn put_object<T: Read>(&mut self, name: &str, from: &mut T) -> Result<()> {
+        self.zip.start_file(name, CompressionMethod::Deflated).chain_err(|| {
+            "Failed to start cache entry object"
+        })?;
+        io::copy(from, &mut self.zip)?;
         Ok(())
     }
 
     /// Finish writing data to the cache entry writer, and return the data.
-    pub fn finish(self) -> io::Result<Vec<u8>> {
+    pub fn finish(self) -> Result<Vec<u8>> {
         let CacheWrite { mut zip } = self;
-        zip.finish()
-            .or(Err(Error::new(ErrorKind::Other, "Failed to finish cache entry zip")))
-            .map(|cur| cur.into_inner())
+        let cur = zip.finish().chain_err(|| "Failed to finish cache entry zip")?;
+        Ok(cur.into_inner())
     }
 }
 
@@ -141,16 +142,16 @@ pub trait Storage {
     /// it should return a `Cache::Miss`.
     /// If the entry is successfully found in the cache, it should
     /// return a `Cache::Hit`.
-    fn get(&self, key: &str) -> Box<Future<Item=Cache, Error=io::Error>>;
+    fn get(&self, key: &str) -> SFuture<Cache>;
 
     /// Get a cache entry for `key` that can be filled with data.
-    fn start_put(&self, key: &str) -> io::Result<CacheWrite>;
+    fn start_put(&self, key: &str) -> Result<CacheWrite>;
 
     /// Put `entry` in the cache under `key`.
     ///
     /// Returns a `Future` that will provide the result or error when the put is
     /// finished.
-    fn finish_put(&self, key: &str, entry: CacheWrite) -> CacheWriteFuture;
+    fn finish_put(&self, key: &str, entry: CacheWrite) -> SFuture<Duration>;
 
     /// Get the storage location.
     fn location(&self) -> String;
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -54,8 +54,6 @@ const TEN_GIGS: usize = 10 * 1024 * 1024 * 1024;
 
 /// Result of a cache lookup.
 pub enum Cache {
-    /// Error fetching from cache.
-    Error(Error),
     /// Result was found in cache.
     Hit(CacheRead),
     /// Result was not found in cache.
@@ -67,7 +65,6 @@ pub enum Cache {
 impl fmt::Debug for Cache {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            Cache::Error(ref e) => write!(f, "Cache::Error({:?}", e),
             Cache::Hit(_) => write!(f, "Cache::Hit(...)"),
             Cache::Miss => write!(f, "Cache::Miss"),
             Cache::Recache => write!(f, "Cache::Recache"),
@@ -76,9 +73,9 @@ impl fmt::Debug for Cache {
 }
 
 /// Trait objects can't be bounded by more than one non-builtin trait.
-pub trait ReadSeek : Read + Seek {}
+pub trait ReadSeek : Read + Seek + Send {}
 
-impl<T: Read + Seek> ReadSeek for T {}
+impl<T: Read + Seek + Send> ReadSeek for T {}
 
 /// Data stored in the compiler cache.
 pub struct CacheRead {
@@ -135,7 +132,7 @@ impl CacheWrite {
 }
 
 /// An interface to cache storage.
-pub trait Storage: Send + Sync {
+pub trait Storage {
     /// Get a cache entry by `key`.
     ///
     /// If an error occurs, this method should return a `Cache::Error`.
@@ -143,7 +140,7 @@ pub trait Storage: Send + Sync {
     /// it should return a `Cache::Miss`.
     /// If the entry is successfully found in the cache, it should
     /// return a `Cache::Hit`.
-    fn get(&self, key: &str) -> Cache;
+    fn get(&self, key: &str) -> Box<Future<Item=Cache, Error=io::Error>>;
 
     /// Get a cache entry for `key` that can be filled with data.
     fn start_put(&self, key: &str) -> io::Result<CacheWrite>;
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -57,19 +57,19 @@ fn make_key_path(key: &str) -> PathBuf {
 }
 
 impl Storage for DiskCache {
-    fn get(&self, key: &str) -> Cache {
+    fn get(&self, key: &str) -> Box<Future<Item=Cache, Error=io::Error>> {
         trace!("DiskCache::get({})", key);
-        match self.lru.lock().unwrap().get(make_key_path(key)) {
-            Err(LruError::FileNotInCache) => Cache::Miss,
-            Err(LruError::Io(e)) => Cache::Error(e),
-            Ok(f) => {
-                match CacheRead::from(f) {
-                    Err(e) => Cache::Error(e),
-                    Ok(cache_read) => Cache::Hit(cache_read),
-                }
+        let path = make_key_path(key);
+        let lru = self.lru.clone();
+        self.pool.spawn_fn(move || {
+            let mut lru = lru.lock().unwrap();
+            match lru.get(&path) {
+                Err(LruError::FileNotInCache) => Ok(Cache::Miss),
+                Err(LruError::Io(e)) => Err(e),
+                Ok(f) => CacheRead::from(f).map(Cache::Hit),
+                Err(_) => panic!("Unexpected error!"),
             }
-            Err(_) => panic!("Unexpected error!"),
-        }
+        }).boxed()
     }
 
     fn start_put(&self, key: &str) -> io::Result<CacheWrite> {
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -16,7 +16,6 @@ use cache::{
     Cache,
     CacheRead,
     CacheWrite,
-    CacheWriteFuture,
     Storage,
 };
 use futures::Future;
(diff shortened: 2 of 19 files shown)


luser:
I had started on some work in the error-chain vein recently: https://github.com/luser/sccache/tree/error-chain

I held off landing it because I figured it was going to conflict with the tokio rewrite. The only functional change there is that I changed `notify_server_startup` to send an error message over the socket/pipe if startup fails.

alexcrichton:
Oh yeah no worries! All those changes are just in the last commit, so I don't mind backing that out as well if you'd prefer

luser:
The Rust 1.13 Travis builds all seem to be [failing with errors like](https://travis-ci.org/mozilla/sccache/jobs/201903329#L302) `error[E0252]: a type named ErrorKind has already been imported in this module`.

The Windows builds in AppVeyor [are failing with](https://ci.appveyor.com/project/luser/sccache2/build/1.0.177/job/aijrsrtx8k0ehcpp#L368):
```
error[E0277]: the trait bound `*mut mio::poll::ReadinessNode: std::marker::Sync` is not satisfied
   --> src\mock_command.rs:147:6
```

On other builds there are some test failures. It looks like `test_server_port_in_use` is failing on all builds and `test_compiler_get_cached_or_compile_cached` failed in at least one instance.

alexcrichton:
Oh dear, how embarassing! Looks like I was working with a stale sccache binary locally with up to date tests by accident. I think I've fixed everything up, but I'll check in to make sure

alexcrichton:
Ok, that last commit should do the trick, I can rebase the changes to the right places at the end.

src/compiler/compiler.rs
```diff
@@ -275,124 +296,205 @@ impl Compiler {
...
-        }
+            let mut entry = match storage.start_put(&key) {
+                Ok(entry) => entry,
+                Err(e) => return future::err(e).boxed()
+            };
+            let write = pool.spawn_fn(move || -> io::Result<_> {
+                for (key, path) in &outputs {
+                    let mut f = try!(File::open(&path));
+                    try!(entry.put_object(key, &mut f).map_err(|e| {
+                        let msg = format!("failed to put object `{:?}` in \
```
luser: You lost the `out_file` from the error message here, which is not a huge problem but I put those in all the log messages because otherwise trying to interpret the sccache log from a real parallel build was a nightmare.

src/compiler/compiler.rs
```diff
@@ -275,124 +296,205 @@ impl Compiler {
...
+                  miss_type: MissType)
+                  -> Box<Future<Item=(CompileResult, process::Output),
+                                Error=io::Error>>
+        where T: CommandCreatorSync,
+    {
+        let process::Output { stdout, .. } = preprocessor_result;
+        let start = Instant::now();
+
+        let compile = self.kind.compile(creator, self, stdout, &parsed_args, cwd, &pool);
+        Box::new(compile.and_then(move |(cacheable, compiler_result)| {
```
luser: Just curious: is there a reason to prefer intermediate variables like `compile` here and `write` below vs. just chaining `and_then` off the previous method?

src/compiler/compiler.rs
```diff
@@ -275,124 +296,205 @@ impl Compiler {
...
+                    let mut stdout = &compiler_result.stdout[..];
+                    try!(entry.put_object("stdout", &mut stdout));
+                }
+                if !compiler_result.stderr.is_empty() {
+                    let mut stderr = &compiler_result.stderr[..];
+                    try!(entry.put_object("stderr", &mut stderr));
+                }
+
+                // Try to finish storing the newly-written cache
+                // entry. We'll get the result back elsewhere.
```
luser: This comment should probably be clearer that we don't want to block returning the compile result to the client while we write to the cache.

src/compiler/compiler.rs
```diff
@@ -275,124 +296,205 @@ impl Compiler {
...
+            if !preprocessor_result.status.success() {
+                debug!("[{}]: preprocessor returned error status {:?}",
+                       parsed_args.output_file(),
+                       preprocessor_result.status.code());
+                // Drop the stdout since it's the preprocessor output, just hand back stderr and the exit status.
+                let output = process::Output {
+                    stdout: vec!(),
+                    ..preprocessor_result
+                };
+                return Box::new(future::ok((CompileResult::Error, output))) as Box<Future<Item=_, Error=_>>
```
alexcrichton: Most of the time, yeah, but `.boxed()` has the added requirement of `Send` so here where most trait objects are not `Send` we can't always use it. In this case it inferred the closure return type as then a future which was `Send` causing errors for later returns.

There's other annotations that can happen here, though, so I'll apply those.

src/compiler/compiler.rs
```diff
@@ -275,124 +296,205 @@ impl Compiler {
...
+                            }
+                            Ok(())
+                        });
+                        let output = process::Output {
+                            status: exit_status(0),
+                            stdout: stdout.into_inner(),
+                            stderr: stderr.into_inner(),
+                        };
+                        let result = CompileResult::CacheHit(duration);
+                        return Box::new(write.map(|_| {
```
alexcrichton: Heh yeah it took me awhile to get the hang of them but I've found them to be super useful  once you get the hang

src/compiler/compiler.rs
```diff
@@ -275,124 +296,205 @@ impl Compiler {
...
+                        MissType::Normal
+                    }
+                    Ok(Cache::Recache) => {
+                        debug!("[{}]: Cache recache!", parsed_args.output_file());
+                        MissType::ForcedRecache
+                    }
+                    Err(e) => {
+                        debug!("[{}]: Cache read error: {:?}", parsed_args.output_file(), e);
                         //TODO: store the error in CacheReadError in some way
                         MissType::CacheReadError
```
alexcrichton: Sure yeah, I'll implement that

src/compiler/compiler.rs
alexcrichton: Ah this is just stylistic for me. I prefer to avoid rightward-drift and keep things under 80 cols, so I tend to make more local variables to avoid the drift and indentation. Both work equivalently though of course!

src/compiler/compiler.rs
alexcrichton: Oh good point yeah. I've been trying to leverage error-chain now to ensure that all the errors returned have all necessary context on them, but I'll be sure to leave this in as well.

alexcrichton:
Ok, updated!

src/mock_command.rs
```diff
@@ -142,84 +149,118 @@ impl CommandChild for Child {
...
+    }
+}
+
+impl fmt::Debug for AsyncCommand {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.inner.fmt(f)
     }
 }
 
 /// Unit struct to use `RunCommand` with `std::system::Command`.
```
luser: This doc is no longer correct.

src/commands.rs
```diff
@@ -471,65 +474,62 @@ fn handle_compile_finished<T : Write, U : Write>(response : CompileFinished, std
...
+            let output = try!(core.run(run_input_output(cmd, None)));
+            if !output.stdout.is_empty() {
+                try!(stdout.write_all(&output.stdout));
+            }
+            if !output.stderr.is_empty() {
+                try!(stderr.write_all(&output.stderr));
+            }
+            Ok(output.status.code()
+               .unwrap_or_else(|| {
+                   /* TODO: this breaks type inference, figure out why
```
alexcrichton: Heh, will do!

src/compiler/clang.rs
```diff
@@ -23,6 +23,7 @@ use ::compiler::{
     run_input_output,
 };
 use futures::Future;
+use futures::future;
```
alexcrichton: Oh sure, I'll change to `use futures::future::{self, Future}`

src/compiler/clang.rs
```diff
@@ -57,49 +58,69 @@ pub fn compile<T>(creator: &T,
...
-
-    pool.spawn_fn(move || {
-        trace!("compile");
-        // Clang needs a temporary file for compilation, otherwise debug info
-        // doesn't have a reference to the input file.
+    trace!("compile");
+    // Clang needs a temporary file for compilation, otherwise debug info
+    // doesn't have a reference to the input file.
+    let input = parsed_args.input.clone();
+    let write = pool.spawn_fn(move || {
```
alexcrichton: Excellent idea! I count 4 locations which had to independently remember "oh don't forget to *not* drop `TempDir` early"!

src/compiler/clang.rs
```diff
@@ -57,49 +58,69 @@ pub fn compile<T>(creator: &T,
...
         }
-        let out_file = try!(parsed_args.outputs.get("obj").ok_or(Error::new(ErrorKind::Other, "Missing object file output")));
-        let mut cmd = creator.new_command_sync(&compiler.executable);
+    };
+
+    let compiler2 = compiler.clone();
+    let cwd2 = cwd.to_string();
+    let mut creator2 = creator.clone();
+    let parsed_args2 = parsed_args.clone();
+    let out_file2 = out_file.clone();
```
alexcrichton: Hm let me see if I can clean this up, I've actually got an idea to make this better.

src/simples3/credential.rs
```diff
@@ -309,180 +314,178 @@ fn parse_credentials_file(file_path: &Path) -> Result<HashMap<String, AwsCredent
...
+        }
+    }
+
+    fn iam_role(&self) -> Box<Future<Item=String, Error=CredentialsError>> {
+        // First get the IAM role
+        let address = "http://169.254.169.254/latest/meta-data/iam/security-credentials";
+        let mut req = Request::new(Method::Get, address.parse().unwrap());
+        req.headers_mut().set(Connection::close());
+        let response = self.client.request(req).and_then(|response| {
+            response.body().fold(Vec::new(), |mut body, chunk| {
```
luser: Does hyper not have a helper method for "collect the response body into a Vec of bytes"?

alexcrichton:
Ok I was able to cut down on most of the `let foo2 = foo.clone()`, turns out nothing goes wrong if the `Command` is created *outside* the `and_then` callback! That's just a few internal allocations to hold the description in memory, but the actual spawning is deferred until later.

src/simples3/credential.rs
```diff
@@ -309,180 +314,178 @@ fn parse_credentials_file(file_path: &Path) -> Result<HashMap<String, AwsCredent
...
+                CredentialsError::new(&format!("failed to read http body: {}", e))
             })
-            .or_else(|e| {
-                warn!("Failed to fetch IAM credentials: {}", e);
-                Err(e)
+        });
+        let body = body.map_err(|_e| {
+            CredentialsError::new("Didn't get a parseable response body from instance role details")
+        }).and_then(|body| {
+            String::from_utf8(body).map_err(|_| {
```
luser: These are a little clunky, but I haven't looked at the error-chain patch yet.

src/cache/disk.rs
```diff
@@ -57,41 +57,39 @@ fn make_key_path(key: &str) -> PathBuf {
...
-            entry.finish()
-                 .map_err(|e| format!("{}", e))
-                 .and_then(|v| {
-                     lru.lock().unwrap().insert_bytes(key, &v)
-                        .map(|_| start.elapsed())
-                        .map_err(|e| format!("{}", e))
-                 })
+            let v = entry.finish()?;
+            lru.lock().unwrap().insert_bytes(key, &v)?;
+            Ok(start.elapsed())
```
luser: That is significantly nicer!

src/commands.rs
```diff
@@ -100,94 +100,92 @@ fn get_port() -> u16 {
...
-                        if pfds[0].revents & POLLIN == POLLIN {
-                            // Ready to read
-                            break;
-                        }
-                        if pfds[0].revents & POLLERR == POLLERR {
-                            // Could give a better error here, I suppose?
-                            return Ok(ServerStartup::TimedOut);
-                        }
-                    }
+    let listener = UnixListener::bind(&socket_path)?;
```
luser: I guess we could replace most of this with tokio-uds, right? (It'd be nice to get rid of some unsafe code.)

src/commands.rs
```diff
@@ -206,11 +204,13 @@ fn redirect_stderr(f: File) -> io::Result<()> {
...
-        Ok(filename) => OpenOptions::new().create(true).append(true).open(filename).and_then(redirect_stderr),
-        _ => Ok(()),
-    }
+fn redirect_error_log() -> Result<()> {
+    let name = match env::var("SCCACHE_ERROR_LOG") {
+        Ok(filename) => filename,
+        _ => return Ok(()),
+    };
+    let f = OpenOptions::new().create(true).append(true).open(name)?;
+    redirect_stderr(f)
```
luser: We apparently have inherently different styles about using intermediate variables. :) I think I err on the "way too long functional programming" side of things. Probably all that LISP programming in my formative years!

src/commands.rs
```diff
@@ -235,148 +235,137 @@ fn run_server_process() -> io::Result<ServerStartup> {
...
+    };
+    let mut pi = PROCESS_INFORMATION {
+        hProcess: ptr::null_mut(),
+        hThread: ptr::null_mut(),
+        dwProcessId: 0,
+        dwThreadId: 0,
+    };
+    let mut si: STARTUPINFOW = unsafe { mem::zeroed() };
+    si.cb = mem::size_of::<STARTUPINFOW>() as DWORD;
+    if unsafe { kernel32::CreateProcessW(exe.as_mut_ptr(),
```
luser: I'm looking forward to removing all this code when 1.16 releases...

src/commands.rs
```diff
@@ -235,148 +235,137 @@ fn run_server_process() -> io::Result<ServerStartup> {
...
     //TODO: better error mapping
-    let mut response = try!(conn.request(req).or(Err(Error::new(ErrorKind::Other, "Failed to send zero statistics command to server or failed to receive respone"))));
+    let mut response = conn.request(req).chain_err(|| {
+        "failed to send zero statistics command to server or failed to receive respone"
+    })?;
     if response.has_stats() {
         Ok(response.take_stats())
     } else {
-        Err(Error::new(ErrorKind::Other, "Unexpected server response!"))
+        Err("Unexpected server response!".into())
```
luser: This can just be `bail!("...")` (same thing elsewhere)

src/errors.rs
```diff
@@ -12,8 +12,44 @@
...
+            description("failed to get a successful HTTP status")
+            display("didn't get a successful HTTP status, got `{}`", status)
+        }
+    }
+}
+
+pub type SFuture<T> = Box<Future<Item = T, Error = Error>>;
+
+pub trait FutureChainErr<T> {
+    fn chain_err<F, E>(self, callback: F) -> SFuture<T>
```
luser: This is nice, chain-err should probably just have this built-in.

src/cache/cache.rs
```diff
@@ -87,15 +85,19 @@ pub struct CacheRead {
 impl CacheRead {
     /// Create a cache entry from `reader`.
     pub fn from<R: ReadSeek + 'static>(reader: R) -> io::Result<CacheRead> {
```
luser: We could just make this return `error::Result<CacheRead>` and add `zip::result::ZipResult` to the error-chain.

src/cache/s3.rs
```diff
@@ -51,7 +46,9 @@ pub struct S3Cache {
 impl S3Cache {
     /// Create a new `S3Cache` storing data in `bucket`.
     pub fn new(bucket: &str, endpoint: &str, handle: &Handle) -> io::Result<S3Cache> {
```
luser: Similarly this could be using error-chain.

src/compiler/compiler.rs
```diff
@@ -424,39 +417,33 @@ impl Compiler {
...
-                    return Box::new(future::ok((CompileResult::NotCacheable, compiler_result)))
-                        as SFuture<_>
-                }
-            } else {
+            if !compiler_result.status.success() {
                 debug!("[{}]: Compiled but failed, not storing in cache",
                        parsed_args.output_file());
                 return Box::new(future::ok((CompileResult::CompileFailed, compiler_result)))
+                    as SFuture<_>
+            } else if cacheable != Cacheable::Yes {
```
luser: This doesn't really *need* to be an else since the if block has a return.

src/compiler/compiler.rs
```diff
@@ -487,15 +474,26 @@ impl Compiler {
...
                 let future = Box::new(future);
                 Ok((CompileResult::CacheMiss(miss_type, duration, future), compiler_result))
+            }).chain_err(move || {
+                format!("failed to store `{}` to cache", out_file)
             }))
         }))
     }
 }
 
 /// Write `contents` to `path`.
```
luser: It'd be good to clarify here that this creates a temp dir and just writes a file under *that* path. This is a nice abstraction, though!

src/compiler/compiler.rs
```diff
@@ -504,36 +502,25 @@ pub fn detect_compiler_kind<T>(creator: &T, executable: &str, pool: &CpuPool)
...
 #elif defined(__clang__)
 clang
 #elif defined(__GNUC__)
 gcc
 #endif
-")?;
-        Ok((dir, src))
-    });
-    let write = write.chain_err(|| "failed to write temporary file");
+".to_vec();
```
luser: Having to allocate a `Vec` for a static byte string is sort of a bummer, but I doubt it's a big deal with all the other overhead here.

src/cache/s3.rs
```diff
@@ -36,39 +35,35 @@ use std::io::{
...
             // credentials. We should either match what boto does more directly
             // or make those builders put their credentials in ~/.aws/credentials
             ProfileProvider::with_configuration(home.join(".boto"), "Credentials"),
-            ];
-        let provider = AutoRefreshingProviderSync::with_mutex(ChainProvider::with_profile_providers(profile_providers)).ok().map(Arc::new);
+        ];
+        let provider = AutoRefreshingProvider::new(ChainProvider::with_profile_providers(profile_providers, handle));
         //TODO: configurable SSL
-        let bucket = Arc::new(Bucket::new(bucket, endpoint, Ssl::No));
+        let bucket = Rc::new(Bucket::new(bucket, endpoint, Ssl::No, handle));
```
alexcrichton: Indeed! Hyper now comes with the [`hyper-tls`](https://github.com/hyperium/hyper-tls) crate which depends on [`native-tls`](https://crates.io/crates/native-tls) which uses the "platform appropriate TLS backend" by default.

src/server.rs
```diff
@@ -138,15 +139,15 @@ pub struct SccacheServer<C: CommandCreatorSync> {
...
                storage: Arc<Storage>) -> io::Result<SccacheServer<C>> {
-        let core = Core::new()?;
         let handle = core.handle();
         let addr = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), port);
         let listener = TcpListener::bind(&SocketAddr::V4(addr), &handle)?;
 
         // Prepare the service which we'll use to service all incoming TCP
         // connections.
-        let (tx, rx) = mpsc::channel(100);
+        let (tx, rx) = mpsc::channel(1);
```
alexcrichton: Oh this was actually an error from before, I intended it to be 1 in the first place and set it to 100 in debugging when I thought something was going wrong, I just forgot to change it back until this commit!

src/simples3/credential.rs
alexcrichton: I don't believe so, but https://github.com/hyperium/hyper/issues/953 is related

src/commands.rs
alexcrichton: Oh I certainly don't mind changing back, I think this was primarily just line-length limits for myself which I tend to be a little more conservative on than most.

src/errors.rs
alexcrichton: Certainly, yeah, I opened https://github.com/brson/error-chain/issues/130

src/compiler/compiler.rs
alexcrichton: Ah yeah I assumed this wasn't too bad as well because the compiler info is typically cached anyway (so should hit this rarely)

src/commands.rs
luser: Nah, it's fine, my code gets way *too* long.

src/commands.rs
alexcrichton: Certainly!

src/simples3/credential.rs
luser: Actually, given that `hyper::Body` is just implementing `Stream`, maybe there should be an extension trait in futures to help with this, something like:
```rust
trait ByteStreamExt {
  type Error;
  fn collect_bytes(self) -> Future<Item = Vec<u8>, Error = Self::Error>;
}

impl<T, I, E> ByteStreamExt for T where T: Stream<Item = I, Error = E, I: AsRef<u8> {
// ...
}
```

alexcrichton: That actually sounds pretty nifty to me! I've filed https://github.com/alexcrichton/futures-rs/issues/390

alexcrichton:
Ok I think I've addressed everything and cleaned up the commits (but I don't mind cleaning up further if you'd prefer!)

Should be good to go as soon as CI is green.

src/commands.rs
```diff
@@ -101,66 +100,45 @@ fn get_port() -> u16 {
...
-        Ok(ServerStartup::Ok)
-    } else {
-        //TODO: send error messages over the socket as well.
-        error!("Server startup failed: {}", buffer[0]);
-        Ok(ServerStartup::Err(format!("Server startup failed: {}", buffer[0]).into()))
+        })
+    });
+
+    let timeout = Duration::from_millis(SERVER_STARTUP_TIMEOUT_MS.into());
+    let timeout = Timeout::new(timeout, &handle)?.map(|()| ServerStartup::TimedOut);
```
luser: This feels like something that could wind up rolled into a convenience method on `Future` in tokio-core, like `fn or_timeout(self, timeout: Duration, handle: &Handle) -> impl Future<Item=TimeoutOr<Self::Item>, Error=Self::Error>`, where `TimeoutOr` could just be like `enum TimeoutOr<T> { Timeout, Value(T) }`.

alexcrichton: Definitely yeah.

On the todo list for the 0.2 release of futures/tokio is to work through a more ergonomic timeout story. For example [`tokio-timer` already has this](https://github.com/tokio-rs/tokio-timer/blob/d803e55b64ab1cb89d43b6802c6fd4494a570115/src/timer.rs#L77-L90) (or similar) and we just need to integrate `tokio-timer` and `tokio_core::reactor::Timeout` more.

alexcrichton:
OSX travis seems to be having a lot of issues as of late :(

luser:
I was going to try to see if we could get at least one green OS X build but I'm not going to wait on that. I'll land this and build it on my local mac to sanity check. Thanks for all your work here!

alexcrichton:
Woo! Thanks for being up for reviewing so much :)

luser:
It builds and passes tests on my Mac, so we're fine there.

PR #77, opened 2017-03-07
Add support for caching Rust compilations. Fixes #42

### LINKED ISSUE #42: Add support for caching Rust crate compilation
I [outlined a plan for this](https://github.com/rust-lang/cargo/issues/1997#issuecomment-265859868) in a cargo issue.

The big items here are:
* make sccache able to detect rustc as a compiler
* Write a hash function for rustc compiler invocations that uses the things listed in that plan as input. We'll need to refactor the cache lookup code slightly since it currently expects to run a preprocessor and hash that output. We will probably want to make the "generate a hash key" bit generic per-compiler, and then provide a default implementation that does what we currently do for C++ compilers, and a separate implementation for rustc.

luser:
This is a pretty large set of changes, but it breaks down into a few parts:
* First, I spent some time refactoring the existing logic from `CompilerInfo::get_cached_or_compile`, since it was all tangled with the logic of running the C preprocessor. I split that out into a separate method and moved a few other things around while I was there. This is the "Pre-Rust support refactoring, Part 1" change.
* Next, I added some traits to hide implementation details of the actual compilation process: `Compiler`, `CompilerHasher`, and `Compilation`. I changed `get_cached_or_compile` and everywhere else that was interacting with compilers to only work with boxed trait objects of these types, and made it so that `Compiler::parse_arguments` returns a boxed `CompilerHasher`, and then `CompilerHasher::generate_hash_key` returns a boxed `Compilation`. This way the C compiler implementation and the Rust compiler implementation can each persist some private state through the process. I refactored the C compilation bits further into an implementation of these traits that's generic over a `CCompilerImpl` trait, implementations of which simply call into the existing gcc/clang/msvc functions. This is the "Pre-Rust support refactoring, Part 2" and "Part 3" changes.
* Finally, I added a Rust implementation of the traits. A few refactoring-type changes snuck in to this changeset but I was too lazy to split them out at this point.

There are also a few tiny refactoring changesets mixed in there, I tried to split small, separable changes out when possible to make the whole thing easier to follow.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -15,13 +15,12 @@
 use cache::{
     Cache,
     Storage,
-    hash_key,
 };
-use compiler::{
-    clang,
-    gcc,
-    msvc,
-};
+use compiler::msvc;
+use compiler::c::CCompiler;
+use compiler::clang::Clang;
+use compiler::gcc::GCC;
+use compiler::msvc::MSVC;
 use filetime::FileTime;
 use futures::future;
 use futures::{Future, IntoFuture};
@@ -33,6 +32,8 @@ use mock_command::{
     RunCommand,
     exit_status,
 };
+#[cfg(test)]
+use mock_command::MockCommandCreator;
 use sha1;
 use std::borrow::Cow;
 use std::collections::HashMap;
@@ -48,6 +49,8 @@ use std::path::{Path, PathBuf};
 use std::process::{self,Stdio};
 use std::str;
 use std::sync::Arc;
+#[cfg(test)]
+use std::sync::Mutex;
 use std::time::{
     Duration,
     Instant,
@@ -56,95 +59,68 @@ use tempdir::TempDir;
 
 use errors::*;
 
-/// Result of generating a hash from a compiler command.
-pub enum HashResult {
-    /// Successful.
-    Ok {
-        key: String,
-        //TODO: fix this to be generic so a Rust compiler doesn't have to
-        // provide this.
-        preprocessor_output: Vec<u8>
-    },
-    /// Something failed.
-    Error { output: process::Output },
-}
-
 /// Supported compilers.
 #[derive(Debug, PartialEq, Clone)]
 pub enum CompilerKind {
     /// GCC
-    Gcc,
+    GCC,
     /// clang
     Clang,
     /// Microsoft Visual C++
-    Msvc {
-        /// The prefix used in the output of `-showIncludes`.
-        includes_prefix: String,
-    },
+    MSVC,
 }
 
-impl CompilerKind {
-    pub fn parse_arguments(&self,
-                           arguments: &[String],
-                           cwd: &Path) -> CompilerArguments {
-        match *self {
-            // GCC and clang share the same argument parsing logic, but
-            // accept different sets of arguments.
-            CompilerKind::Gcc => gcc::parse_arguments(arguments, cwd, gcc::argument_takes_value),
-            CompilerKind::Clang => gcc::parse_arguments(arguments, cwd, clang::argument_takes_value),
-            CompilerKind::Msvc { .. } => msvc::parse_arguments(arguments),
-        }
-    }
-
-    pub fn generate_hash_key<T>(&self,
-                                creator: &T,
-                                compiler: &str,
-                                compiler_digest: &str,
-                                parsed_args: &ParsedArguments,
-                                cwd: &str,
-                                pool: &CpuPool)
-                                -> SFuture<HashResult>
-        where T: CommandCreatorSync
-    {
-        match *self {
-            CompilerKind::Gcc | CompilerKind::Clang | CompilerKind::Msvc { .. } => hash_key_from_c_preprocessor_output(creator, self, compiler, compiler_digest, parsed_args, cwd, pool),
-        }
-    }
-
-    pub fn preprocess<T>(&self,
+/// An interface to a compiler.
+pub trait Compiler<T: CommandCreatorSync>: Send + 'static {
+    /// Return the kind of compiler.
+    fn kind(&self) -> CompilerKind;
+    /// Determine whether `arguments` are supported by this compiler.
+    fn parse_arguments(&self,
+                       arguments: &[String],
+                       cwd: &Path) -> CompilerArguments;
+    /// Given information about a compiler command, generate a hash key
+    /// that can be used for cache lookups, as well as any additional
+    /// information that can be reused for compilation if necessary.
+    fn generate_hash_key(&self,
                          creator: &T,
-                         compiler: &str,
+                         executable: &str,
+                         executable_digest: &str,
                          parsed_args: &ParsedArguments,
                          cwd: &str,
                          pool: &CpuPool)
-                         -> SFuture<process::Output>
-        where T: CommandCreatorSync
-    {
-        match *self {
-            CompilerKind::Gcc | CompilerKind::Clang => {
-                // GCC and clang use the same preprocessor invocation.
-                gcc::preprocess(creator, compiler, parsed_args, cwd, pool)
-            },
-            CompilerKind::Msvc { ref includes_prefix } => msvc::preprocess(creator, compiler, parsed_args, cwd, includes_prefix, pool),
-        }
-    }
+                         -> SFuture<HashResult<T>>;
+    fn box_clone(&self) -> Box<Compiler<T>>;
+}
 
-    pub fn compile<T>(&self,
-                      creator: &T,
-                      compiler: &str,
-                      preprocessor_output: Vec<u8>,
-                      parsed_args: &ParsedArguments,
-                      cwd: &str,
-                      pool: &CpuPool)
-                      -> SFuture<(Cacheable, process::Output)>
-        where T: CommandCreatorSync,
-    {
-        match *self {
-            CompilerKind::Gcc => gcc::compile(creator, compiler, preprocessor_output, parsed_args, cwd, pool),
-            CompilerKind::Clang => clang::compile(creator, compiler, preprocessor_output, parsed_args, cwd, pool),
-            CompilerKind::Msvc { .. } => msvc::compile(creator, compiler, preprocessor_output, parsed_args, cwd, pool),
-        }
-    }
+impl<T: CommandCreatorSync> Clone for Box<Compiler<T>> {
+    fn clone(&self) -> Box<Compiler<T>> { self.box_clone() }
+}
+
+pub trait Compilation<T: CommandCreatorSync> {
+    /// Given information about a compiler command, execute the compiler.
+    fn compile(self: Box<Self>,
+               creator: &T,
+               executable: &str,
+               parsed_args: &ParsedArguments,
+               cwd: &str,
+               pool: &CpuPool)
+               -> SFuture<(Cacheable, process::Output)>;
+}
+
+/// Result of generating a hash from a compiler command.
+pub enum HashResult<T: CommandCreatorSync> {
+    /// Successful.
+    Ok {
+        /// The hash key of the inputs.
+        key: String,
+        /// An object to use for the actual compilation, if necessary.
+        compilation: Box<Compilation<T> + 'static>,
+    },
+    /// Something failed.
+    Error {
+        /// The error output and return code.
+        output: process::Output,
+    },
 }
 
 /// The results of parsing a compiler commandline.
@@ -184,15 +160,15 @@ pub enum CompilerArguments {
 
 /// Information about a compiler.
 #[derive(Clone)]
-pub struct CompilerInfo {
+pub struct CompilerInfo<T: CommandCreatorSync> {
     /// The path to the compiler binary.
     pub executable: String,
     /// The last modified time of `executable`.
     pub mtime: FileTime,
     /// The sha-1 digest of `executable`, as a hex string.
     pub digest: String,
-    /// The kind of compiler, from the set of known compilers.
-    pub kind: CompilerKind,
+    /// The actual compiler implementation.
+    pub compiler: Box<Compiler<T>>,
 }
 
 /// Specifics about cache misses.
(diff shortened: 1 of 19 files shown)


luser:
The Travis builds are broken on 1.13 for some reason, I will take a look at that.

src/compiler/compiler.rs
```diff
@@ -556,41 +560,44 @@ gcc
...
-    Box::new(detect_compiler_kind(creator, &executable, &pool).and_then(move |kind| {
-        match kind {
-            Some(kind) => {
-                pool.spawn_fn(move || Compiler::new(&executable, kind))
-                    .chain_err(|| "failed to learn compiler metadata")
+    let detect = detect_compiler(creator, &executable, &pool);
+    Box::new(detect.and_then(move |compiler| {
+        match compiler {
+            Some(compiler) => {
+                Box::new(sha1_digest(executable.clone(), &pool)
```
alexcrichton: This is actually somewhat interesting in the rustc case. I'd imagine that quite commonly this is actually the rustup version of `rustc.exe` rather than the rustc version of `rustc.exe`. I wonder if we could automatically detect rustup and go fetch the actual compiler instead?

I believe that would work best by looking for `rustup.exe` in the directory next to `rustc.exe` (if found). If that's there we can run `rustup which rustc` which should print out the full path to the actual compiler.

Also as a side note here, I wonder if we should handle the dynamic libraries that rustc itself links to? I can't really imagine a use case though where the dylibs are changed and the rustc executable *wouldn't* change, though, so it's perhaps not an issue.

src/compiler/rust.rs
```diff
@@ -0,0 +1,761 @@
...
+                    trace!("[{}]: got {} source files from dep-info in {}", crate_name,
+                           files.len(), fmt_duration_as_secs(&start.elapsed()));
+                    // Just to make sure we capture temp_dir.
+                    drop(temp_dir);
+                    hash_all(files, &pool)
+                }
+                Err(e) => return Box::new(future::err(e)),
+            }
+        } else {
+            Box::new(future::err("Failed run rustc --dep-info".into()))
```
alexcrichton: Could this contain the rustc errors logs (e.g. `output` above)?

src/compiler/rust.rs
```diff
@@ -0,0 +1,761 @@
...
+        if output.status.success() {
+            match parse_dep_file(&dep_file, &cwd) {
+                Ok(files) => {
+                    trace!("[{}]: got {} source files from dep-info in {}", crate_name,
+                           files.len(), fmt_duration_as_secs(&start.elapsed()));
+                    // Just to make sure we capture temp_dir.
+                    drop(temp_dir);
+                    hash_all(files, &pool)
+                }
+                Err(e) => return Box::new(future::err(e)),
```
alexcrichton: Perhaps this could use `chain_err` to say "failed to parse dep info at {}"

src/compiler/rust.rs
```diff
@@ -0,0 +1,761 @@
...
+    if log_enabled!(Trace) {
+        trace!("get_compiler_outputs: {:?}", cmd);
+    }
+    let outputs = run_input_output(cmd, None);
+    Box::new(outputs.and_then(move |output| -> Result<_> {
+        if output.status.success() {
+            let outstr = String::from_utf8(output.stdout).chain_err(|| "Error parsing rustc output")?;
+            Ok(outstr.lines().map(|l| l.to_owned()).collect())
+        } else {
+            bail!("Failed to run `rustc --print file-names`");
```
alexcrichton: I wonder if this may actually make a good helper as part of `run_input_output`, having a method to run a command and auto-check `output.status.success()` and bundle up all the relevant error info in the case of failure.

src/compiler/rust.rs
```diff
@@ -0,0 +1,761 @@
...
+            "--help" | "-V" | "--version" => return CompilerArguments::NotCompilation,
+            v if v.starts_with("--print") || v.starts_with("--explain") || v.starts_with("--pretty") || v.starts_with("--unpretty") => return CompilerArguments::NotCompilation,
+            // Could support `-o file` but it'd be more complicated.
+            "-o" => return CompilerArguments::CannotCache,
+            //TODO: support linking against native libraries. This
+            // will require replicating the linker search strategy
+            // so we can *find* them.
+            "-l" => return CompilerArguments::CannotCache,
+            v if v.starts_with("--emit") => {
+                //XXX: do we need to handle --emit specified more than once?
```
alexcrichton: Ah yeah I think this does happen from time to time (I think I type it in at least every once in a while!)

src/compiler/rust.rs
```diff
@@ -0,0 +1,761 @@
...
+
+/// Return true if `arg` is in the set of arguments `set`.
+fn arg_in(arg: &str, set: &HashSet<&str>) -> bool
+{
+    set.contains(arg) || set.iter().any(|a| arg.starts_with(a))
+}
+
+/// Calculate the SHA-1 digest of each file in `files` on background threads
+/// in `pool`.
+fn hash_all(files: Vec<String>, pool: &CpuPool) -> SFuture<Vec<String>>
```
alexcrichton: Should these files be sorted to ensure that they're always hashed in the same order? I think the dep-info is relatively deterministic (but that's likely not guaranteed) and I wouldn't be surprised if the order of `--extern` from Cargo was also slightly nondeterministic

src/compiler/rust.rs
```diff
@@ -0,0 +1,761 @@
...
+            // normalize them so we can get cross-machine cache hits.
+            let args = arguments.iter().map(|s| s.as_str()).collect::<String>();
+            m.update(args.as_bytes());
+            // 4. The sha-1 digests of all source files (this includes src file from cmdline).
+            // 5. The sha-1 digests of all files listed on the commandline (self.externs)
+            for h in source_hashes.into_iter().chain(extern_hashes) {
+                m.update(h.as_bytes());
+            }
+            // 6. TODO: Environment variables:
+            //    RUSTFLAGS, maybe CARGO_PKG_*?
```
alexcrichton: RUSTFLAGS is safe to ignore as it's just used by Cargo, not rustc.

Other env vars though yeah like CARGO_* should probably be hashed. That being said I don't think the env from the child ships over to the server, right? (this is a case where that may actually be quite important to implement!)

In any case I think it's safe to assume that all of the child's `CARGO_*` env vars should be hashed here (although I guess arguably *all* env vars should be hashed due to `env!`)

alexcrichton:
Overall looks great to me, nice!

I wonder if we should perhaps document somewhere the limitations of rustc caching? Off the top of my head what I can think of is:

* Only some flag combinations work (e.g. `--emit dep-info,link`)
* Values from `env!` may not be tracked in caching
* Procedural macros that read files from the filesystem may not be cached correctly (yet)
* The system linker isn't cached (e.g. changing gcc/ld won't invalidate)
* Target specs aren't hashed (e.g. custom target specs I believe)

luser:
We should definitely document those! I wrote down a bunch of things in the `parse_arguments` implementation as comments, but it'd be good to have them somewhere else.

> The system linker isn't cached (e.g. changing gcc/ld won't invalidate)

This is a good point. Is there a way to ask rustc what linker it would use? If this isn't simple to fix I wonder if we shouldn't avoid caching dylib/bin crates for now.

> Target specs aren't hashed (e.g. custom target specs I believe)

I'm not clear on what this means--do you mean that we don't hash the target when it's not passed on the commandline?

src/compiler/rust.rs
luser: I sort the list of source files from the dep-info, but not the list of externs, they're input to the hash in the order they're listed on the commandline.

If cargo produces nondeterministic commandlines we've got a bigger problem, since we're feeding the entire commandline as hash input! Can we verify that assumption?

src/compiler/rust.rs
luser: I'm not overly concerned with caching all possible commandlines (clearly), but it'd be good to ensure that we're at least not producing incorrect behavior in that case.

src/compiler/rust.rs
luser: I'll have to look at the various places where I call this--I do think I wrote this pattern more than once.

src/compiler/rust.rs
luser: That's a good idea. We return the preprocessor stderr for C compilation when it fails, we should do something smarter here. The error propagation in this code isn't great in general, I should make sure we're providing enough information that users can attempt to diagnose failures.

src/compiler/compiler.rs
luser: That's a really good point and I can't believe I didn't realize it! I don't have a non-rustup install handy, but would it be enough to run `rustc --print sysroot` and hash `$sysroot/bin/rustc`? I'd have to sanity check how that works with the tooltool-downloaded rustc we use in Firefox CI.

We discussed the dylib issue in Hawaii and we kind of hand-waved it away. I definitely don't want to be parsing the output of `ldd` or reading the binary headers to figure this out, but maybe it's enough to hash rustc + everything in `$sysroot/lib/`? Actually given that rustc is just a stub, maybe we skip hashing rustc entirely, hash `$sysroot/lib` and that's good enough?

alexcrichton: > but would it be enough to run rustc --print sysroot and hash $sysroot/bin/rustc

Oh right yeah that'd be perfect. (and also solve the rustup case)

This may also throw a wrench into sccache's own caching of the compiler as well unfortunately b/c rustup can have a different rustc depending on which directory it's invoked in (and that can change over time). I think sccache assumes the main compiler doesn't change much, right?

Also yeah I *think* we can handwave away b/c every release has different hashes in the dylib filenames which I believe implies that the contents of `rustc` will be different (e.g. the file somehow references those dylibs). That being said hashing `$sysroot/lib` would put a nail in the coffin and work perfectly as well.

The only catch there is that dylibs are in `$sysroot/bin` on Windows and `$sysroot/lib` everywhere else.

src/compiler/rust.rs
alexcrichton: Yeah I think it's fine to just return "not cacheable" for multiple `--emit` calls for now.

src/compiler/rust.rs
alexcrichton: Ah yeah just confirmed, Cargo's order of `--extern` is nondeterministic.

If you'd prefer to change in Cargo I can also do that!

src/compiler/rust.rs
alexcrichton: > I wonder if it'd be possible to teach rustc to tell us about the usage of env!, similar to emitting dep-info

I was thinking the same thing yeah. I filed https://github.com/rust-lang/rust/issues/40364 to track this.

I also definitely agree that there's tons of junk in env vars typically. We could try to maintain a blacklist of "don't worry about caching this" but otherwise I think blanket hashing `CARGO_*` should work fine for now.

alexcrichton:
> This is a good point. Is there a way to ask rustc what linker it would use? If this isn't simple to fix I wonder if we shouldn't avoid caching dylib/bin crates for now.

Not currently, unfortunately. Let's stick w/ just caching rlibs for now which is conservative and most of the benefit I think anyway.

> I'm not clear on what this means--do you mean that we don't hash the target when it's not passed on the commandline?

Oh sorry! So for more esoteric use cases you can actually pass `--target foo` where `foo.json` is in the cwd. The `foo.json` file then says all the various options for your "custom target" (aka these are called "custom target specs").

Note that this is a pretty niche use case though. We can probably protect against this by running `rustc --print target-list` and then checking if the `--target` value is not in there. If it's not then it's a custom target and we probably shouldn't cache for now. (again though, seems like a prime candidate for a TODO somewhere as opposed to blocking this on fixing)

src/compiler/compiler.rs
luser: > This may also throw a wrench into sccache's own caching of the compiler as well unfortunately b/c rustup can have a different rustc depending on which directory it's invoked in (and that can change over time). I think sccache assumes the main compiler doesn't change much, right?

Also a good point, we'll have to do something smarter here. sccache caches compiler info by (path, mtime), so rustup will totally break this expectation. I'll have to think about that for a bit, I don't want to have an extra rustc invocation for every single compile if we can avoid it.

src/compiler/rust.rs
luser: Fixing that in cargo would certainly make my life easier, and we're likely to need to wait for your dep-info speedup for this to all be useful anyway, so we can probably wait for a fix for that.

Is anything else about the commandline nondeterministic? If so, we'll have to fix sccache to not use the commandline as a direct hash input.

alexcrichton: Ok I just checked, everything should be deterministic except the order of `--extern` which is nondeterministic.

luser:
OK, I think I addressed all the review comments.

luser:
I rebased on top of master and built it locally on Linux/Mac/Windows, so everything should be OK, but I figured I'd push the rebased changes here for sanity's sake. If the CI is green I'll merge this and then we can get on with rebasing and merging all your other PRs.

src/compiler/rust.rs
```diff
@@ -0,0 +1,859 @@
...
+
+/// Parse dependency info from `file` and return a Vec of files mentioned.
+/// Treat paths as relative to `cwd`.
+fn parse_dep_file<T, U>(file: T, cwd: U) -> Result<Vec<String>>
+    where T: AsRef<Path>,
+          U: AsRef<Path>,
+{
+    let mut f = File::open(file)?;
+    let mut deps = String::new();
+    f.read_to_string(&mut deps)?;
```
alexcrichton: FWIW it looks like this happens on the main event loop thread

alexcrichton:
I've pushed a commit that parses dep-info files on a cpu pool, but otherwise looks great! r=me w/ green CI

src/compiler/rust.rs
luser: Good catch, thanks!

Issue #61, opened 2017-01-16
Support MSVC compiler options starting with "/"


fxb:
There already is a TODO in ```msvc.rs``` mentioning this, but I think it is good to have an issue open for better visibility.

It seems quite straight-forward to add support for options starting with ```/``` as well.

See: https://msdn.microsoft.com/en-us/library/610ecb4h.aspx

> Options are specified by either a forward slash (/) or a dash (–).

Additionally, it might be good to also support options that accept arguments after a space.

> If an option takes an argument, the option's description documents whether a space is allowed between the option and the arguments

For example, I have encountered defines being set like ```/D WIN32``` instead of ```/DWIN32```. Maybe this should have a separate issue.

Issue #47, opened 2016-12-13
Support caching MSVC compiles with -showIncludes


luser:
We currently treat `-showIncludes` as a non-cacheable MSVC option. I ported this directly from the Python implementation. Unfortunately a lot of build tools use this to produce dependency info, and it may not be possible to work around that. @alexcrichton [ran into this](https://github.com/rust-lang/rust/issues/38119#issuecomment-266631585) trying to get sccache working for the LLVM build in Rust.

I suspect the issue here is that the output from `-showIncludes` is generated when we run the preprocessor, but we cache the output of running the compiler on the already-preprocessed output. It might be a little weird, but I think we could save all the extra output lines and just prepend them to the compiler output if the original command included `-showIncludes`.

@glandium, is there anything I'm missing there?

fxb:
Using CMake with the Ninja generator will add ```/showIncludes``` to all compiler invocations it seems, making it impossible to use with ```sccache``` at the moment.

luser:
Right, that was the original motivation for this bug. Rust builds LLVM using CMake, so their compilations include `/showIncludes`.

fxb:
It seems like a common way other compiler caches work, is to always add ```/showIncludes``` to the compiler command line, effectively including it in the cache key (instead of preprocessing). Obviously this will likely break shared caching, unless paths are the same on all machines.

I guess a straight forward fix, as you said, would just be to strip printed includes from the preprocessor output for caching purposes, and then prepend them to the output again. Just to clarify we are on the same page. I'd be happy to help working on a fix for this issue.

luser:
> Obviously this will likely break shared caching, unless paths are the same on all machines.

This is already an issue with sccache, https://github.com/mozilla/sccache/issues/35 covers fixing it.

> I guess a straight forward fix, as you said, would just be to strip printed includes from the preprocessor output for caching purposes, and then prepend them to the output again. Just to clarify we are on the same page. I'd be happy to help working on a fix for this issue.

Yeah, that should work fine. sccache caches the *compiler* output, and `-showIncludes` is part of the *preprocessor* output. We always invoke the preprocessor, so we can always get the proper local `-showIncludes` output. If you'd like to work on this feel free! Most of the work should be contained to `msvc.rs`, but we'd have to create some way to save the bits from the preprocessor output to return with the compiler output.

[...]

Issue #72, opened 2017-02-21
Preprocessor warnings are hidden


flxo:
`sccache` hides preprocessor warnings. In the following snippet a two line c file is compiled with and without `sccache` using `clang` (`Apple LLVM version 8.0.0 (clang-800.0.42.1)`) and the current `HEAD` of `sccache`.  As you can see the warning from the `#warning` directive is not on stdout/stderr. The second run with `sccache` shows that the behaviour is the same with or without a hit. I did the same test with a double `#define` - same result.

```
~/tmp  cat test.c
#warning foo
int bar() {}

~/tmp  clang -c -o test.o test.c
test.c:1:2: warning: foo [-W#warnings]
#warning foo
 ^
test.c:2:12: warning: control reaches end of non-void function [-Wreturn-type]
int bar() {}
           ^
2 warnings generated.
~/tmp  sccache clang -c -o test.o test.c
test.c:2:12: warning: control reaches end of non-void function [-Wreturn-type]
int bar() {}
           ^
1 warning generated.
~/tmp  sccache clang -c -o test.o test.c
test.c:2:12: warning: control reaches end of non-void function [-Wreturn-type]
int bar() {}
           ^
1 warning generated.
```

A possible solution could be to concat `stderr` from the preprocessor run with `stdou/stderr` from the compilation step. I'm not 100% sure whether this would give the exact same output as compiling the original file.

luser:
Specifically this would have been fixed by https://github.com/mozilla/sccache/pull/178

PR #78, opened 2017-03-08
Changes for sccache+msvc+LLVM

alexcrichton:
This few commits were what I found was necessary to get sccache fully working on MSVC when compiling LLVM in rust-lang/rust. It basically boiled down to:

Fixes https://github.com/mozilla/sccache/issues/61
Fixes https://github.com/mozilla/sccache/issues/47

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -64,6 +64,8 @@ pub struct ParsedArguments {
     pub preprocessor_args: Vec<String>,
     /// Commandline arguments for the preprocessor or the compiler.
     pub common_args: Vec<String>,
+    /// Whether or not the `-showIncludes` argument is passed on MSVC
+    pub msvc_show_includes: bool,
 }
 
 impl ParsedArguments {
@@ -77,7 +79,7 @@ struct CCompilation<I: CCompilerImpl> {
     parsed_args: ParsedArguments,
     executable: String,
     /// The output from running the preprocessor.
-    preprocessor_output: Vec<u8>,
+    preprocessor_result: process::Output,
     compiler: I,
 }
 
@@ -114,7 +116,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
     fn compile<T>(&self,
                   creator: &T,
                   executable: &str,
-                  preprocessor_output: Vec<u8>,
+                  preprocessor_result: process::Output,
                   parsed_args: &ParsedArguments,
                   cwd: &str,
                   env_vars: &[(OsString, OsString)],
@@ -217,7 +219,7 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
                 compilation: Box::new(CCompilation {
                     parsed_args: parsed_args,
                     executable: executable,
-                    preprocessor_output: preprocessor_result.stdout,
+                    preprocessor_result: preprocessor_result,
                     compiler: compiler,
                 }),
             })
@@ -244,8 +246,8 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compilation<T> for CCompilation<I>
                -> SFuture<(Cacheable, process::Output)>
     {
         let me = *self;
-        let CCompilation { parsed_args, executable, preprocessor_output, compiler } = me;
-        compiler.compile(creator, &executable, preprocessor_output, &parsed_args, cwd, env_vars,
+        let CCompilation { parsed_args, executable, preprocessor_result, compiler } = me;
+        compiler.compile(creator, &executable, preprocessor_result, &parsed_args, cwd, env_vars,
                          pool)
     }
 
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -68,7 +68,7 @@ impl CCompilerImpl for Clang {
     fn compile<T>(&self,
                   creator: &T,
                   executable: &str,
-                  preprocessor_output: Vec<u8>,
+                  preprocessor_result: process::Output,
                   parsed_args: &ParsedArguments,
                   cwd: &str,
                   env_vars: &[(OsString, OsString)],
@@ -76,7 +76,7 @@ impl CCompilerImpl for Clang {
                   -> SFuture<(Cacheable, process::Output)>
         where T: CommandCreatorSync
     {
-        compile(creator, executable, preprocessor_output, parsed_args, cwd, env_vars, pool)
+        compile(creator, executable, preprocessor_result, parsed_args, cwd, env_vars, pool)
     }
 }
 
@@ -95,7 +95,7 @@ pub fn argument_takes_value(arg: &str) -> bool {
 
 fn compile<T>(creator: &T,
               executable: &str,
-              preprocessor_output: Vec<u8>,
+              preprocessor_result: process::Output,
               parsed_args: &ParsedArguments,
               cwd: &str,
               env_vars: &[(OsString, OsString)],
@@ -111,7 +111,7 @@ fn compile<T>(creator: &T,
             Some(name) => name,
             None => return future::err("missing input filename".into()).boxed(),
         };
-        write_temp_file(pool, filename.as_ref(), preprocessor_output)
+        write_temp_file(pool, filename.as_ref(), preprocessor_result.stdout)
     };
     let input = parsed_args.input.clone();
     let out_file = match parsed_args.outputs.get("obj") {
@@ -193,6 +193,7 @@ mod test {
         }
     }
 
+
     #[test]
     fn test_parse_arguments_simple() {
         let a = parses!("-c", "foo.c", "-o", "foo.o");
@@ -236,13 +237,14 @@ mod test {
             outputs: vec![("obj", "foo.o".to_owned())].into_iter().collect::<HashMap<&'static str, String>>(),
             preprocessor_args: vec!(),
             common_args: vec!(),
+            msvc_show_includes: false,
         };
         let compiler = f.bins[0].to_str().unwrap();
         // Compiler invocation.
         next_command(&creator, Ok(MockChild::new(exit_status(0), "", "")));
         let (cacheable, _) = compile(&creator,
                                      &compiler,
-                                     vec!(),
+                                     empty_output(),
                                      &parsed_args,
                                      f.tempdir.path().to_str().unwrap(),
                                      &[],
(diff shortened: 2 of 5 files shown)


src/compiler/clang.rs
```diff
@@ -133,36 +133,53 @@ mod test {
...
-                assert!(preprocessor_args.is_empty());
-                assert!(common_args.is_empty());
-            }
-            o @ _ => assert!(false, format!("Got unexpected parse result: {:?}", o)),
-        }
+        let args = stringvec!["-c", "foo.c", "-o", "foo.o"];
+        let ParsedArguments {
+            input,
+            extension,
+            depfile: _depfile,
```
KalitaAlexey: You can write `_` instead of `_depfile`.
https://is.gd/7nHDiH

src/compiler/clang.rs
```diff
@@ -133,36 +133,53 @@ mod test {
...
-                assert_eq!("cxx", extension);
-                assert_map_contains!(outputs, ("obj", "foo.o"));
-                //TODO: fix assert_map_contains to assert no extra keys!
-                assert_eq!(1, outputs.len());
-                assert!(preprocessor_args.is_empty());
-                assert_eq!(stringvec!["-arch", "xyz", "-fabc", "-I", "include", "-include", "file"], common_args);
-            }
-            o @ _ => assert!(false, format!("Got unexpected parse result: {:?}", o)),
-        }
+        let args = stringvec!["-c", "foo.cxx", "-arch", "xyz", "-fabc","-I", "include", "-o", "foo.o", "-include", "file"];
```
KalitaAlexey: Missing space after `"-afbc".

src/compiler/clang.rs
```diff
@@ -133,36 +133,53 @@ mod test {
...
+        let ParsedArguments {
+            input,
+            extension,
+            depfile: _depfile,
```
KalitaAlexey: The same as I suggested above.

src/compiler/gcc.rs
```diff
@@ -326,138 +327,210 @@ mod test {
...
-                assert!(preprocessor_args.is_empty());
-                assert!(common_args.is_empty());
-            }
-            o @ _ => assert!(false, format!("Got unexpected parse result: {:?}", o)),
-        }
+        let args = stringvec!["-c", "foo.c", "-o", "foo.o"];
+        let ParsedArguments {
+            input,
+            extension,
+            depfile: _depfile,
```
KalitaAlexey: The same as I suggested above.

src/compiler/msvc.rs
```diff
@@ -108,56 +109,58 @@ pub fn parse_arguments(arguments: &[String]) -> CompilerArguments {
     let mut debug_info = false;
     let mut pdb = None;
     let mut depfile = None;
+    let mut show_includes = false;
 
     //TODO: support arguments that start with / as well.
```
KalitaAlexey: I suppose the comment is no longer needed.

src/compiler/msvc.rs
```diff
@@ -108,56 +109,58 @@ pub fn parse_arguments(arguments: &[String]) -> CompilerArguments {
...
+                common_args.push(arg.clone());
+                if let Some(arg_val) = it.next() {
+                    common_args.push(arg_val.clone());
                 }
             }
-            None => break,
+            v @ _ if v.starts_with("-deps") => {
+                depfile = Some(v[5..].to_owned());
+            }
+            // Arguments we can't handle.
```
KalitaAlexey: The comment should be placed below because we will be able to handle `"-showIncludes"`.

src/compiler/msvc.rs
```diff
@@ -152,9 +152,15 @@ pub fn parse_arguments(arguments: &[String]) -> CompilerArguments<ParsedArgument
     let mut depfile = None;
 
     //TODO: support arguments that start with / as well.
-    let mut it = arguments.iter();
+    let mut it = arguments.iter().map(|i| {
+        if i.starts_with("/") {
+            format!("-{}", &i[1..])
```
luser: This is a clever hack!

src/compiler/msvc.rs
```diff
@@ -153,53 +153,48 @@ pub fn parse_arguments(arguments: &[String]) -> CompilerArguments<ParsedArgument
...
+            "-showIncludes" => return CompilerArguments::CannotCache,
+            a if a.starts_with('@') => return CompilerArguments::CannotCache,
+            // Arguments we can't handle because they output more files.
+            // TODO: support more multi-file outputs.
+            "-FA" | "-Fa" | "-Fe" | "-Fm" | "-Fp" | "-FR" | "-Fx" => return CompilerArguments::CannotCache,
+            "-Zi" => {
+                debug_info = true;
+                common_args.push(arg.clone());
+            }
+            v if v.starts_with("-Fd") => {
```
luser: I was thinking it might be nice to refactor some of the other compiler argument parsing code to use the `ArgsIter` pattern I implemented for Rust: https://github.com/mozilla/sccache/blob/7ae4c259873d6bcbb60ad58bd9735581f7ddbe7a/src/compiler/rust.rs#L366

It's probably not as useful here, but it was nice to be able to write all the match arms as strings instead of having `v.starts_with(...)` all over the place.

src/compiler/c.rs
```diff
@@ -77,7 +79,7 @@ struct CCompilation<I: CCompilerImpl> {
     parsed_args: ParsedArguments,
     executable: String,
     /// The output from running the preprocessor.
-    preprocessor_output: Vec<u8>,
+    preprocessor_result: process::Output,
```
luser: This change ought to make it easier to fix https://github.com/mozilla/sccache/issues/72 as well!

src/compiler/clang.rs
```diff
@@ -170,36 +170,53 @@ mod test {
...
+        let ParsedArguments {
+            input,
+            extension,
+            depfile: _,
+            outputs,
+            preprocessor_args,
+            msvc_show_includes,
+            common_args,
+        } = match _parse_arguments(&args) {
+            CompilerArguments::Ok(args) => args,
```
luser: That is a much more sensible way to write this.

src/compiler/msvc.rs
```diff
@@ -430,6 +432,19 @@ fn compile<T>(creator: &T,
...
+    // If the `-showIncludes` command line option was originally passed we need
+    // to be sure to ship the output from the preprocessor as the actual
+    // result of this compilation.
+    let mut extra_stderr = Vec::new();
+    if parsed_args.msvc_show_includes {
+        extra_stderr = preprocessor_result.stderr;
+    }
+    Box::new(ret.map(|(cacheable, mut output)| {
+        let prev = mem::replace(&mut output.stderr, extra_stderr);
+        output.stderr.extend(prev);
```
luser: I don't think this is actually correct, and this is probably what breaks the configure test in the Firefox build. See my comment in `detect_showincludes_prefix`:
https://github.com/mozilla/sccache/blob/7ae4c259873d6bcbb60ad58bd9735581f7ddbe7a/src/compiler/msvc.rs#L104

I just did a simple local test with MSVC2015 and the `-showIncludes` output goes to stdout when compiling:
```
ted@DESKTOP-32V9ND8 /c/build
$ cl -Fonul -nologo -showIncludes -c test.c > /tmp/cl.stdout 2>/tmp/cl.stderr

ted@DESKTOP-32V9ND8 /c/build
$ cat /tmp/cl.stdout
test.c
Note: including file: C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\ucrt\stdio.h
Note: including file:  C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\ucrt\corecrt.h
Note: including file:   C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\vcruntime.h
Note: including file:    C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\sal.h
Note: including file:     C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\ConcurrencySal.h
Note: including file:    C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\vadefs.h
Note: including file:  C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\ucrt\corecrt_wstdio.h
Note: including file:   C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\ucrt\corecrt_stdio_config.h
```

When running the preprocessor it definitely goes to stderr.

Unfortunately I think this means we'll have to do something slightly more complex here, like parse the preprocessor stderr to get just the showIncludes lines, and then put just those in the compiler stdout.

alexcrichton:
Thanks for the review! It'll take a bit I think to get time to cover the comments, but I'll dig in as soon as I get a chance.

alexcrichton:
According to https://msdn.microsoft.com/en-us/library/hdkef6tk.aspx 

> The /showIncludes option emits to stderr, not stdout.

This seems empirically false...

src/compiler/msvc.rs
```diff
@@ -444,6 +445,25 @@ fn compile<T>(creator: &T,
...
+
+    // If the `-showIncludes` command line option was originally passed we need
+    // to be sure to ship the output from the preprocessor as the actual
+    // result of this compilation.
+    //
+    // Note, though, that when we ran the preprocessor we passed `-E` which
+    // means that the "show includes" business when to stderr. Normally, though,
+    // the compiler emits `-showIncludes` output to stdout. To handle that we
+    // take the stderr of the preprocessor and prepend it to the stdout of the
+    // compilation.
```
luser: I think at some point we'll want to do something like this for all compilers to fix #72, but we'll probably have to put the warnings on stderr for compatibility.

alexcrichton:
Oh sorry I pushed an update but then confused myself as to how it could work at all! I'm not actually sure how this patch works but maybe I'm missing something?

We [recreate `stderr_bytes`](https://github.com/mozilla/sccache/blob/335f191205f41970d3c79494071bd22cb20023d3/src/compiler/msvc.rs#L336) but [don't add include lines to it](https://github.com/mozilla/sccache/blob/335f191205f41970d3c79494071bd22cb20023d3/src/compiler/msvc.rs#L345-L346), so I'm not sure how this reaches the final compiler's output. When manually executing `sccache cl foo.c -c -showIncludes` though it seems to work!

alexcrichton:
Oh so it turns out my `cl.exe` was totally broken for other resons for totally unrelated reasons. I'll send a follow-up patch.

alexcrichton:
Ok, I've sent a PR which I believe fixes the issue: https://github.com/mozilla/sccache/pull/106

alexcrichton:
Sorry for the runaround!

PR #79, opened 2017-03-13
Add timeouts to cache lookups and handle server eofs

alexcrichton:
We've unfortunately been having a lot of trouble over at https://github.com/rust-lang/rust/issues/40240 with sccache causing spurious failures. A number of logs are linked from that issue and to try to diagnose these issues I've enabled sccache logging which we cat to the Travis logs on failures.

Today we got a [new failure](https://github.com/rust-lang/rust/pull/40329#issuecomment-286054302) whose [failing build](https://travis-ci.org/rust-lang/rust/jobs/210484803) pointed to:

```
caused by: IoError(Error { repr: Custom(Custom { kind: Other, error: StringError("unexpected EOF") }) })
caused by: unexpected EOF
```

Upon inspection of the sccache server debug logs (at the end of the build log) I noticed the failing object was `ELFDumper.cpp.o`, we attempted to fetch this object from the cache (I think), and no other messages about `ELFDumper.cpp.o` (success or not) happened on the server. My theory for what happened looks like:

* We sent off an S3 request 
* This S3 request then takes *forever*, presumably an "infinite" amount of time
* The sccache server decided to shut down due to being idle
* sccache waited 10 seconds for the S3 connection to finish
* It didn't, so we shut down the whole server, closing active connections
* The client then receives EOF unexpectedly, causing the error

To handle this failure mode this PR makes a few modifications

* The client now handles `UnexpectedEof` the same way it handles "unhandled compilation", it just compiles the object locally. This should protect against this in the future and other spurious races like this which are generally benign.
* I've added logging to shutdown of the server just to know when it happens
* I've added a timeout to cache lookups (60s currently) which should help avoiding waiting on S3 for too long.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -562,7 +562,11 @@ fn handle_compile_response<T>(mut creator: T,
                 // get a response then we just forge ahead locally and run the
                 // compilation ourselves.
                 Err(ProtobufError::IoError(ref e))
-                    if e.kind() == io::ErrorKind::UnexpectedEof => {}
+                    if e.kind() == io::ErrorKind::UnexpectedEof => {
+                    writeln!(io::stderr(),
+                             "warning: sccache server looks like it shut down \
+                              unexpectedly, compiling locally instead").unwrap();
+                }
 
                 Err(e) => return Err(e).chain_err(|| {
                     //TODO: something better here?
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -32,7 +32,7 @@ use number_prefix::{
     Prefixed,
     Standalone,
 };
-use protobuf::RepeatedField;
+use protobuf::{RepeatedField, ProtobufError};
 use protocol::{
     CacheStats,
     ClientRequest,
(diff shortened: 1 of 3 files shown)


arielb1:
Could you write something to stderr/somewhere else visible on buildbot when we get unexpected EOF errors, so we can have some monitoring over them?

alexcrichton:
@arielb1 yeah unfortunately architecturally sccache is slightly difficult to debug since it primarily runs as a daemon (without stderr/stdout). Note though that it supports [redirection of stderr](https://github.com/mozilla/sccache/blob/1e6c6fc1e7cfdde6571215b1e03aa88a4ef93686/src/commands.rs#L184-L192) in the daemon which [we're leveraging](https://github.com/rust-lang/rust/blob/fd182c4010b5aee72d070b15e90c98cb0fdc3776/.travis.yml#L86-L87) and [printing](https://github.com/rust-lang/rust/blob/fd182c4010b5aee72d070b15e90c98cb0fdc3776/.travis.yml#L145-L146).

(those logs are how I discovered this issue)

alexcrichton:
@luser FWIW I've deployed this to rust-lang/rust and it seems to have at least erased that particular failure mode!

luser:
@alexcrichton That's great to hear! I have been trying to get my Rust PR landable, I will finish that off so we can get everything else rebased and landed.

src/commands.rs
```diff
@@ -458,43 +458,57 @@ fn handle_compile_response<T>(mut creator: T,
...
+                    }
+                }
+
+                // Currently the shutdown behavior of the remote sccache server
+                // is to wait at most N seconds for all active connections to
+                // finish and then close everything. If we get unlucky and don't
+                // get a response then we just forge ahead locally and run the
+                // compilation ourselves.
+                Err(ProtobufError::IoError(ref e))
+                    if e.kind() == io::ErrorKind::UnexpectedEof => {}
```
luser: Can you put a log message here so it's easy to figure out when this happens? Actually I think even straight printing something to stderr would be OK in this case, it feels like a bug if we hit this in practice.

src/compiler/compiler.rs
```diff
@@ -355,6 +360,20 @@ impl Compiler {
                 storage.get(&key)
             };
 
+            // Wait at most a minute for the cache to respond before we forge
+            // ahead ourselves with a compilation.
+            let timeout = Duration::new(60, 0);
```
luser: As a followup, it might be nice to have a way to configure this so we can tweak it in practice.

src/server.rs
```diff
@@ -575,6 +576,9 @@ impl<C> SccacheService<C>
                                     stats.cache_misses += 1;
                                     stats.forced_recaches += 1;
                                 }
+                                MissType::TimedOut => {
+                                    stats.cache_misses += 1;
```
luser: Might be good to add a stat for this as well so we can track it?

PR #81, opened 2017-03-17
Fix non-GMT time stamp in HTTP Date header

na-g:
HTTP/1.1 [requires](https://tools.ietf.org/html/rfc2616#section-3.3.1) that all time stamps created be in GMT. Posts to S3 were using local time with a numerical offset instead.  Ceph's radosgw is [picky](http://tracker.ceph.com/issues/3973) enough to care, so this was preventing sccache from POSTing to ceph's S3 interface.


luser:
Thanks for the fix!

PR #83, opened 2017-03-21
Remove dependency on `named_pipe`

alexcrichton:
I was sporadically receiving a segfault locally when trying to debug issues on
Windows and in tracking this down I discovered blackbeam/named_pipe#3 which
leads to segfaults locally on startup.

This switches the one use case to the relevant functionality in
`mio-named-pipes` (already pulled in as part of `tokio-process`) and then
otherwise mirrors the same logic as the Unix version, just waiting for a byte
with a timeout.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 cache: cargo
 
 rust:
-  - 1.13.0
+  - 1.16.0
   - stable
   - beta
   - nightly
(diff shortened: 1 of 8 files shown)


alexcrichton:
Note that I took this as an opportunity as well to remove the raw call to `CreateProcess` now that the associated extension method is now stable.

luser:
This looks great (thanks for the CreateProcess cleanup!) There are just two things that I'd like fixed before you merge this:
1) That `windows_process_extensions` is also in use [in the process spawning code](https://github.com/mozilla/sccache/blob/6dba30a3afa512837598f079c5a94bd55468a101/src/mock_command.rs#L195), so you'll break the nightly Travis build with this right now. Just strip out the feature tests there (and remove the no-op version).
2) Bump the minimum Rust version in .travis.yml and the README.

alexcrichton:
Certainly, fixed!

PR #85, opened 2017-03-22
Move from protobuf to bincode for a wire format 

alexcrichton:
This commit migrates away from the `protobuf` crate to instead just working with
bincode on the wire as a serialization format. This is done by leveraging a few
different crates:

* The `bincode` and `serde_derive` crates are used to define serialization
  for Rust structures as well as provide a bincode implementation.
* The `tokio_io::codec::length_delimited` module implements framing via length
  prefixes to transform an asynchronous stream of bytes into a literal `Stream`
  of `BytesMut`.
* The `tokio_serde_bincode` crate is then used to tie it all together, parsing
  these `BytesMut` as the request/response types of sccache.

Most of the changes here are related to moving away from the protobuf API
throughout the codebase (e.g. `has_foo` and `take_foo`) towards a more
rustic-ish API that just uses enums/structs. Overall it felt quite natural (as
one would expect) to just use the raw enum/struct values.

This may not be quite as performant as before but that doesn't really apply to
sccache's use case where perf is hugely dominated by actually compiling and
hashing, so I'm not too too worried about that.

My personal motivation for this is twofold:

1. Using `protobuf` was a little clunky throughout the codebase and definitely
   had some sharp edges that felt good to smooth out.
2. There's currently what I believe some mysterious segfault and/or stray write
   happening in sccache and I'm not sure where. The `protobuf` crate had a lot
   of `unsafe` code and in lieu of actually auditing it I figured it'd be good
   to kill two birds with one stone. I have no idea if this fixes my segfault
   problem (I never could reproduce it) but I figured it's worth a shot.


luser:
Excellent! I've wanted to drop protobuf for serde for a while now (https://github.com/mozilla/sccache/issues/20). I'm not sure that the specific wire format matters all that much in terms of perf, and being able to packet-sniff the traffic and read the JSON could prove useful.

The only thing I worried about with using JSON was the compile response where a large part of the body is `Vec<u8>`. Presumably that will blow up to enormous size in JSON? It still may not matter in the grand scheme of things.

alexcrichton:
Oh jeez that's a very good point. It is true that swapping a different serde protocol should be pretty easy and another possible option is [bincode](https://github.com/TyOverby/bincode) which I believe Servo uses (but would be harder to inspect on the wire).

I'll do some testing locally to see what happens.

alexcrichton:
Yeah looks like `b"foo"` is serialized as `[100,111,111]`, from 3 to 13 bytes. I'll switch to bincode!

alexcrichton:
Ok switched over to bincode!

luser:
Wire inspection is not a deal breaker (given that we control both ends of the connection and already have logging), it was just a possible nice-to-have. :)

src/client.rs
```diff
@@ -52,28 +45,29 @@ impl ServerConnection {
     }
 
     /// Send `request` to the server, read and return a `ServerResponse`.
```
luser: nit: the doc comment still mentions `ServerResponse`

src/client.rs
```diff
@@ -52,28 +45,29 @@ impl ServerConnection {
...
+        let mut len = [0; 4];
+        BigEndian::write_u32(&mut len, bytes.len() as u32);
+        self.writer.write_all(&len)?;
+        self.writer.write_all(&bytes)?;
+        self.writer.flush()?;
         trace!("ServerConnection::request: sent request");
         self.read_one_response()
     }
 
     /// Read a single `ServerResponse` from the server.
```
luser: same nit as above

src/client.rs
```diff
@@ -52,28 +45,29 @@ impl ServerConnection {
...
-    pub fn read_one_response(&mut self) -> ProtobufResult<ServerResponse> {
+    pub fn read_one_response(&mut self) -> Result<Response> {
         trace!("ServerConnection::read_one_response");
-        //FIXME: wish `parse_length_delimited_from` worked here!
-        let len = try!({
-            let mut is = CodedInputStream::from_buffered_reader(&mut self.reader);
-            is.read_raw_varint32()
-        });
+        let mut bytes = [0; 4];
+        self.reader.read_exact(&mut bytes)?;
```
luser: Would things look nicer overall if we used tokio here instead of blocking IO? Originally I used blocking IO in the client because the mio stuff was so complicated.

src/commands.rs
```diff
@@ -455,28 +431,20 @@ fn print_stats(stats: CacheStats) -> Result<()> {
...
-    }
-    let mut req = ClientRequest::new();
-    let mut compile = Compile::new();
-    compile.set_exe(exe.to_owned());
-    compile.set_cwd(cwd.to_owned());
-    compile.set_command(RepeatedField::from_vec(args));
-    trace!("request_compile: {:?}", compile);
-    req.set_compile(compile);
+    let req = Request::Compile(Compile {
+        exe: exe.as_ref().to_str().ok_or("bad exe")?.to_owned(),
```
luser: as a follow-up, we could now just use `OsStr` or `OsString` all the way for the values from the client commandline.

src/commands.rs
```diff
@@ -504,18 +472,14 @@ fn handle_compile_finished(response: CompileFinished,
...
     // interleaved streams to really make it work.
-    if response.has_stdout() {
-        try!(stdout.write_all(response.get_stdout()));
-    }
-    if response.has_stderr() {
-        try!(stderr.write_all(response.get_stderr()));
-    }
-    if response.has_retcode() {
-        let ret = response.get_retcode();
+    try!(stdout.write_all(&response.stdout));
```
luser: Switch these to `?` while you're here?

src/commands.rs
```diff
@@ -542,39 +506,21 @@ fn handle_compile_response<T>(mut creator: T,
...
-                    if res.has_compile_finished() {
-                        trace!("Server sent CompileFinished");
-                        return handle_compile_finished(res.take_compile_finished(),
-                                                       stdout, stderr)
-                    } else {
-                        bail!("unexpected response from server")
-                    }
-                }
-
-                // Currently the shutdown behavior of the remote sccache server
```
luser: I don't see that you replaced this behavior anywhere (compile locally on EOF from server). Presumably you do want to keep this in some form?

src/server.rs
```diff
@@ -603,13 +579,24 @@ impl<C> SccacheService<C>
...
-                    status.code()
-                        .map_or_else(
-                            || trace!("CompileFinished missing retcode"),
-                            |s| { trace!("CompileFinished retcode: {}", s); finish.set_retcode(s) });
+                    trace!("CompileFinished retcode: {}", status);
+                    match status.code() {
+                        Some(code) => res.retcode = Some(code),
+                        None => res.signal = Some(get_signal(status)),
+                    };
                     //TODO: sort out getting signal return on Unix
```
luser: You can remove this TODO that you fixed. :)

src/server.rs
```diff
@@ -718,24 +704,25 @@ impl ServerStats {
     fn to_cache_statistics(&self) -> Vec<CacheStatistic> {
```
luser: We could probably do better here now just by defining the `CacheStats` type to use the `CacheStat` type for its fields directly, but we can sort that out in a followup.

src/server.rs
```diff
@@ -718,24 +704,25 @@ impl ServerStats {
...
+                    fmt_duration_as_secs(&duration)
                 } else {
-                    stat.set_str("0.000s".to_owned());
-                }
-                $vec.push(stat);
+                    "0.000s".to_owned()
+                };
+                $vec.push(CacheStatistic {
+                    name: String::from($name),
+                    value: CacheStat::String(s),
```
luser: I probably should have added a `CacheStat::Duration` for clarity, eh?

src/client.rs
alexcrichton: Nah I think they'd basically be equivalent. As soon as we'd want to do anything else, though, such as timeouts, then we'd definitely want to use Tokio as it'd look quite a bit better.

The tokio version would basically just do the futures equivalent of this, using `tokio_io` combinators, and a little less ergonomically.

src/commands.rs
alexcrichton: Yeah I tried to do that but `Serialize`/`Deserialize` wasn't implemented for one of them :(

alexcrichton: Actually... I'll send a PR!

alexcrichton: Ok this is now tracked at https://github.com/serde-rs/serde/issues/823

src/commands.rs
alexcrichton: Oops, updated!

PR #86, opened 2017-03-23
Go back to raw process spawning on Windows

alexcrichton:
Unfortunately a crucial detail I missed is that the previous incarnation of this
function set `bInheritHandles` to `FALSE` which caused the server process to not
inherit any handles from the parent process. The Rust standard library, however,
unconditionally passes `TRUE` for this variable right now.

I discovered, though, that with CMake the `cmake` process itself would hang
until all of its child pipes were closed, and one of those pipes was one that
leaked all the way into the server process. I unfortunately could not figure out
a way to stop the leak, so I've opted to just revert the part to use the
standard library's spawning mechanism back to calling `CreateProcess` manually.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -195,13 +195,18 @@ fn redirect_error_log() -> Result<()> {
 #[cfg(windows)]
 fn run_server_process() -> Result<ServerStartup> {
     use futures::Future;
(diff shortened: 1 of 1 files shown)


luser:
That's unfortunate. :-(

PR #90, opened 2017-03-23
Try to get more info out of server crashes

alexcrichton:
Apply two separate techniques:

* Ensure that core dumps come out of the server by calling `setrlimit` manually
* Install some signal handlers which print out the signal that was received

Hopefully this'll help debugging issues by at least confirming what happens.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -146,15 +146,83 @@ fn run_server_process() -> Result<ServerStartup> {
 #[cfg(not(windows))]
 fn daemonize() -> Result<()> {
     use daemonize::Daemonize;
-    if match env::var("SCCACHE_NO_DAEMON") {
-            Ok(val) => val == "1",
-            Err(_) => false,
-    } {
-        Ok(())
-    } else {
-        Daemonize::new().start().chain_err(|| {
-            "failed to daemonize"
-        })
+    use libc;
+    use std::mem;
+
+    match env::var("SCCACHE_NO_DAEMON") {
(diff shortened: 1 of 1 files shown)


src/commands.rs
```diff
@@ -146,15 +146,78 @@ fn run_server_process() -> Result<ServerStartup> {
...
+    // which means that no one's probably listening for our exit status.
+    // In order to assist with debugging crashes of the server we configure our
+    // rlimit to allow core dumps and we also install a signal handler for
+    // segfaults which at least prints out what just happened.
+    unsafe {
+        let rlim = libc::rlimit {
+            rlim_cur: libc::RLIM_INFINITY,
+            rlim_max: libc::RLIM_INFINITY,
+        };
+        libc::setrlimit(libc::RLIMIT_CORE, &rlim);
```
luser: This worries me a little bit, in that for non-ephemeral build environments (like OS X) we could pile up coredumps without noticing. Could we put just the call to `setrlimit` behind an environment variable?

luser:
Otherwise this looks fine, feel free to merge it, I don't think it will conflict with my Rust work.

src/commands.rs
alexcrichton: Sure yeah I can do that. I figured that `sccache` itself segfaulting would be pretty bad, but I think this also applies to child processes, and I guess gcc segfaulting is "less bad" in the sense of being "more expected" at least :)

PR #92, opened 2017-03-27
Send environment variables from client, hash CARGO_ env vars in Rust

luser:
This fixes a longstanding issue about not using the client's environment for compilation, and then uses that fix to include CARGO_ env vars in Rust hash generation.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -18,12 +18,12 @@ use futures_cpupool::CpuPool;
 use mock_command::CommandCreatorSync;
 use sha1;
 use std::borrow::Cow;
-use std::collections::HashMap;
-use std::env;
+use std::collections::{HashMap, HashSet};
+use std::ffi::{OsStr, OsString};
 use std::fmt;
 use std::path::Path;
 use std::process;
-use util::sha1_digest;
+use util::{os_str_bytes, sha1_digest};
 
 use errors::*;
 
@@ -106,6 +106,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
                      executable: &str,
                      parsed_args: &ParsedArguments,
                      cwd: &str,
+                     env_vars: &[(OsString, OsString)],
                      pool: &CpuPool)
                      -> SFuture<process::Output> where T: CommandCreatorSync;
     /// Run the C compiler with the specified set of arguments, using the
@@ -116,6 +117,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
                   preprocessor_output: Vec<u8>,
                   parsed_args: &ParsedArguments,
                   cwd: &str,
+                  env_vars: &[(OsString, OsString)],
                   pool: &CpuPool)
                   -> SFuture<(Cacheable, process::Output)>
         where T: CommandCreatorSync;
@@ -167,13 +169,15 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
     fn generate_hash_key(self: Box<Self>,
                          creator: &T,
                          cwd: &str,
+                         env_vars: &[(OsString, OsString)],
                          pool: &CpuPool)
                          -> SFuture<HashResult<T>>
     {
         let me = *self;
         let CCompilerHasher { parsed_args, executable, executable_digest, compiler } = me;
-        let result = compiler.preprocess(creator, &executable, &parsed_args, cwd, pool);
+        let result = compiler.preprocess(creator, &executable, &parsed_args, cwd, env_vars, pool);
         let out_file = parsed_args.output_file().into_owned();
+        let env_vars = env_vars.to_vec();
         let result = result.map_err(move |e| {
             debug!("[{}]: preprocessor failed: {:?}", out_file, e);
             e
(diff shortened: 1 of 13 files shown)


src/compiler/c.rs
```diff
@@ -106,6 +105,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
                      executable: &str,
                      parsed_args: &ParsedArguments,
                      cwd: &str,
+                     env_vars: &Vec<(String, String)>,
```
alexcrichton: Perhaps `&[(String, String)]`?

src/compiler/rust.rs
```diff
@@ -215,12 +219,15 @@ fn parse_dep_info<T>(dep_info: &str, cwd: T) -> Vec<String>
...
-fn get_compiler_outputs<T>(creator: &T, executable: &str, arguments: &[String], cwd: &str) -> SFuture<Vec<String>>
+fn get_compiler_outputs<T>(creator: &T, executable: &str, arguments: &[String], cwd: &str,
+                           env_vars: &Vec<(String, String)>) -> SFuture<Vec<String>>
     where T: CommandCreatorSync,
 {
     let mut cmd = creator.clone().new_command_sync(executable);
     cmd.args(&arguments)
         .args(&["--print", "file-names"])
+        .env_clear()
+        .envs(env_vars.iter().map(|&(ref k, ref v)| (k, v)))
```
alexcrichton: This seems like it's somewhat tricky to remember (clear the env before setting more env vars). I wonder if sccache could benefit from a [`Command` look alike](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/process_builder.rs#L11-L17) which allows introspection (we still need to add this to libstd...) to maybe package up arguments like the executable, cwd, env, and base arguments? 

(just a random note for a future PR/refactoring)

src/compiler/rust.rs
```diff
@@ -540,15 +549,26 @@ impl<T> CompilerHasher<T> for RustHasher
...
             }
-            // 6. TODO: Environment variables: CARGO_* at the very least.
-            // https://github.com/mozilla/sccache/issues/89
+            // 6. Environment variables. Ideally we'd use anything referenced
+            // via env! in the program, but we don't have a way to determine that
+            // currently, and hashing all environment variables is too much, so
+            // we'll just hash the CARGO_ env vars and hope that's sufficient.
+            // Upstream Rust issue tracking getting information about env! usage:
+            // https://github.com/rust-lang/rust/issues/40364
+            for &(ref var, ref val) in env_vars.iter() {
```
alexcrichton: I think we'll want to be sure to sort these before hashing, right?

src/compiler/rust.rs
luser: Yeah, that's a sensible idea. I actually went back and changed this to add `envs` on `RunCommand` to make it slightly less verbose, but encapsulating all this commonality seems like a good idea.

src/compiler/rust.rs
luser: Oops, good catch! Might have even caught that with the unit test I added if I had used more than one env var...

src/protocol.rs
```diff
@@ -23,35 +33,55 @@ pub enum CompileResponse {
...
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Compile {
+    /// The full path to the compiler executable.
     pub exe: String,
+    /// The current working directory in which to execute the compile.
     pub cwd: String,
+    /// The commandline arguments passed to the compiler.
     pub args: Vec<String>,
+    /// The environment variables present when the compiler was executed, as (var, val).
+    pub env_vars: Vec<(String, String)>,
```
alexcrichton: With the new serde version I think this could also start to be a OsString as well

src/compiler/rust.rs
luser: I filed https://github.com/mozilla/sccache/issues/95 on this.

src/compiler/c.rs
luser: ...yeah that makes more sense. Fixed.

src/protocol.rs
luser: I pushed the serde bump to master and then fell down this rabbit hole. The only thing that I didn't have a great solution for was adding `OsString`s to the hash, so [I wrote a crappy equivalent of `OsStringExt::as_bytes` for Windows](https://github.com/luser/sccache/blob/27d719873d3e7ffa3de69f8d387f10e89dd18875/src/util.rs#L62) that just `mem::transmute`s. I suspect it will not ever be a problem in practice (and I added a unit test that should fail if it ever breaks) but it sure *feels* wrong.

src/compiler/c.rs
```diff
@@ -255,19 +261,21 @@ pub const CACHED_ENV_VARS : &'static [&'static str] = &[
...
-    //TODO: should propogate these over from the client.
-    // https://github.com/glandium/sccache/issues/5
-    for var in CACHED_ENV_VARS.iter() {
-        if let Ok(val) = env::var(var) {
-            m.update(var.as_bytes());
+    //TODO: use lazy_static.
+    let cached_env_vars: HashSet<OsString> = CACHED_ENV_VARS.iter().map(|v| OsStr::new(v).to_os_string()).collect();
+    for &(ref var, ref val) in env_vars.iter() {
+        if cached_env_vars.contains(var) {
+            m.update(os_str_bytes(var));
```
alexcrichton: This runs a risk of breaking in future rust versions, but could `Hash for OsStr` be used here instead?

PR #94, opened 2017-03-28
Remove line break in release script

flxo:
With `2349791` a line break in the release script is introduced that breaks
it. This patch removes that one. I noticed that when rereading the patch yesterday.

### DIFF
--- a/scripts/build-release.sh
+++ b/scripts/build-release.sh
@@ -24,8 +24,7 @@ case $system in
     Linux)
         # Build using rust-musl-builder
         rm -rf target/x86_64-unknown-linux-musl/release
-        docker run --rm -it -v "$(pwd)":/home/rust/src -v ~/.cargo/git:/home/rust/.cargo/git -v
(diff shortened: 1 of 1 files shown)


luser:
Thanks for the fix! I'm sure I would have banged my head against that soon enough.

flxo:
Thanks for merging. 
Would you mind updating `luser/rust-musl-builder`? Seems that there is a quite outdated `rustc 1.10.0` included. Maybe switching to [upstream rust-musl-builder](https://hub.docker.com/r/ekidd/rust-musl-builder/) is also a option...

luser:
I can't remember why I needed that, exactly, but I will look into that today.

flxo:
Ok. I did a short smoke test with the original `ekidd/rust-musl-builder`. Worked fine.

luser:
Yeah, that seems to work fine. I don't remember why I changed it in the first place, but I pushed https://github.com/mozilla/sccache/commit/bbb850245430cfb444b4f4d222ce3e6dff2c5f59 to change it back. Thanks! I'm hoping to figure out a better system for building releases in the near future where I can just tag a release and have some CI build release binaries for me.

PR #98, opened 2017-04-05
Re-enable s3 by default, build with --features=all in CI

luser:
I think I missed this when the redis storage patch landed. Opening a PR to make sure the CI builds are OK before I land this.


luser:
I don't know why AppVeyor didn't schedule a build, but Travis is green and we were already building `--features=all` on AppVeyor anyway, so I'll just land this.

PR #105, opened 2017-04-11
Treat error exit codes of compilation as compile fail

### LINKED ISSUE #103: Compile errors hidden because of return code not equal zero of (C) compiler invocation
When a compile error with e.g `gcc` happens it returns with something != 0. The `clang` and `gcc` implementations use `utils::run_input_output` for compilation. The check for the result is 

```
    Box::new(future::result(child)
             .and_then(|child| {
                 wait_with_input_output(child, input).and_then(|output| {
                     if output.status.success() {
                         f_ok(output)
                     } else {
                         f_err(ErrorKind::ProcessError(output))
                     }
                 })
             }))
```

which ends up in treating the result as an error in `server::start_compile_task` (the match in [server.rs:538](https://github.com/mozilla/sccache/blob/master/src/server.rs#L538)). The `stdout` and `stderr` are [discarded and never shown somewhere](https://github.com/mozilla/sccache/blob/master/src/server.rs#L592-L603). The return code is also discarded and the client exits with -2.

flxo:
If the invocation of a compilation fails with a non zero exit
code the compilation failed. This patch forwards stdout and
stderr with the compiler exit code to the client.

This fixes #103.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -94,6 +94,16 @@ fn notify_server_startup(name: &Option<OsString>, success: bool) -> io::Result<(
     notify_server_startup_internal(pipe, success)
 }
 
+#[cfg(unix)]
+fn get_signal(status: ExitStatus) -> i32 {
+    use std::os::unix::prelude::*;
+    status.signal().expect("must have signal")
+}
+#[cfg(windows)]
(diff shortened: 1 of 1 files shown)


luser:
Thanks for the fix!

PR #106, opened 2017-04-11
Actually support -showIncludes on MSVC

alexcrichton:
Various problems locally meant that I wasn't able to thoroughly test before
pushing up a patch, but now I've fixed those issues and this is tested locally!
Namely:

* Pass `-showIncludes` to the preprocessor even if there's no dep file output
  requested.
* When a dep file is requested and `-showIncludes` is passed then ship all
  output of dep files to the compiler's stdout.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -509,9 +509,9 @@ gcc
     cmd.stdout(Stdio::piped())
        .stderr(Stdio::null());
     let output = write.and_then(move |(tempdir, src)| {
-        let args = vec!(OsString::from("-E"), OsString::from(&src));
+        cmd.arg("-E").arg(src);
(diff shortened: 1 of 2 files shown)


luser:
This does work for me in local testing, it just has one quirk. MSVC outputs the source file name for every compiler invocation, and since we're running the compiler twice we get the source filename output twice:
```
$ cl -nologo -Fohello.obj -showIncludes -c hello.c -I.
hello.c
Note: including file: .\hello.h
$ /c/build/sccache/target/debug/sccache.exe cl -nologo -Fohello.obj -showIncludes -c hello.c -I.
hello.c
Note: including file: .\hello.h
hello.c
```
It's not a terribly big deal, just a quirk.

PR #109, opened 2017-04-21
Switch from SHA-1 to SHA-512

### LINKED ISSUE #108: Faster hash algorithm?
Running sccache against a fully cached LLVM build on S3 I've found that 82% of the runtime of the server is consumed in sha1 hashing. That's quite a lot! Knowing C++ it is indeed hashing megabytes and megabytes of output...

I wonder if perhaps other hash algorithms have been considered? So far some pieces I've learned are:

* The OpenSSL implementation of SHA1 is ~4x faster than the `sha1` crate. (presumably due to fancy simd bits)
* The *ring* implementation is the same speed as the `sha1` crate (presumably because both are written in stable rust without simd)
* The OpenSSL SHA256 implementation is ~2x faster than the `sha1` crate
* The *ring* implementation is also 2x faster than the `sha1` crate 
* Various blake2b crates can get to about 3x faster than the `sha1` crate
* I'm sure md5 can be even faster still!

Are there various thoughts on the hashing algorithm here? I'd be wary of using OpenSSL due to its difficult-to-get-working Windows support. I think *ring* could be a good option but it's not *that* much faster in sha256 mode. Maybe blake2b is fast enough? (and I think still written in Rust).

alexcrichton:
Local benchmarking showed that the implementation of SHA-512 in the *ring* crate
is 3x faster than the implementation of SHA-1 in the `sha1` crate. I've also
noticed that 80%+ of sccache's runtime on a fully cached build is spent hashing.
With this change I noticed a decrease from 108s to 92s when building a fully
cached LLVM from the network. Not a huge win but if the network were faster
could perhaps add up!

Closes #108


src/util.rs
```diff
@@ -27,26 +27,60 @@ use std::time::Duration;
...
+pub struct Digest {
+    inner: Context,
+}
+
+impl Digest {
+    pub fn new() -> Digest {
+        Digest { inner: Context::new(&SHA512) }
+    }
+
+    /// Calculate the SHA-1 digest of the contents of `path`, running
```
est31: It should read SHA-512 here.

src/compiler/rust.rs
```diff
@@ -842,7 +840,6 @@ bar.rs:
         drop(env_logger::init());
         let f = TestFixture::new();
         // SHA-1 digest of an empty file.
```
est31: Comment forgotten to be removed.

alexcrichton:
Ah just realized I should update the *ring* dependency, so this now depends on *ring* 0.9

alexcrichton:
Ok I've rebased over https://github.com/mozilla/sccache/pull/112, but something I realized was that *ring* 0.9 depends on Rust 1.17 so I bumped that dep, if that's a problem though I can change it back.

est31:
@alexcrichton if you bump the required rust version in the .travis.yml, you should also bump it in the README.

luser:
> Ok I've rebased over #112, but something I realized was that ring 0.9 depends on Rust 1.17 so I bumped that dep, if that's a problem though I can change it back.

I don't love constantly bumping our version requirement to chase latest stable, but as long as it's mentioned in the README and enforced in our CI config it's not a dealbreaker for anything.

alexcrichton:
@luser as a heads up I updated the "install rust" code in AppVeyor to use rustup, I think the old manifest may have not made their way to 1.17.0

luser:
Thanks! I copied that script from somewhere and haven't looked at it in a while. :)

PR #110, opened 2017-04-26
Retry compilation with gcc if -Werror is passed and the compilation of preprocessed source fails

flxo:
In some (`gcc`) compiler and source combinations the compilations of the preprocessed source issues warnings that are present when compiling the original source file. This patch adds a compile retry to `gcc` in the same manner as `clang` does if `-Werror` is passed.

I needed this extension to use `sccache` for `Android` builds. Somewhere in `bionic` (which is built with `-Werror`) the compilation of the preprocessed source throws a warning and the build fails.
The difference to the `clang` implementation here in `sccache` is that first attempt uses the preprocessed  source on `stdin` along with the `-x` argument.

Sidenote: To use `sccache` when building `Android`, export
```
CC_WRAPPER=sccache
CXX_WRAPPER=sccache
```
before building. See `build/core/ccache.mk` for reference.

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -109,64 +109,21 @@ fn compile<T>(creator: &T,
     let write = {
         let filename = match Path::new(&parsed_args.input).file_name() {
             Some(name) => name,
-            None => return future::err("missing input filename".into()).boxed(),
+            None => return future::err("Missing input filename".into()).boxed(),
         };
-        write_temp_file(pool, filename.as_ref(), preprocessor_result.stdout)
+
+        Box::new(
+            write_temp_file(pool, filename.as_ref(), preprocessor_result.stdout.clone())
+                 .and_then(move |(tempdir, input)| {
+                     match input.into_os_string().into_string() {
+                         Ok(p) => future::ok((None, vec!(p), Some(tempdir))),
+                         Err(_) => future::err("Failed to write input file".into()),
+                     }
+                 })
+        )
     };
-    let input = parsed_args.input.clone();
-    let out_file = match parsed_args.outputs.get("obj") {
-        Some(obj) => obj,
-        None => {
-            return future::err("Missing object file output".into()).boxed()
-        }
-    };
 
-    let mut attempt = creator.clone().new_command_sync(executable);
-    attempt.arg("-c")
-        .arg("-o")
-        .arg(&out_file)
-        .args(&parsed_args.common_args)
-        .env_clear()
-        .envs(env_vars.iter().map(|&(ref k, ref v)| (k, v)))
-        .current_dir(&cwd);
-    let output = write.and_then(move |(tempdir, input)| {
-        attempt.arg(&input);
-        run_input_output(attempt, None).map(|output| {
-            drop(tempdir);
-            (Cacheable::Yes, output)
-        })
-    });
-
-    // clang may fail when compiling preprocessor output with -Werror,
-    // so retry compilation from the original input file if it fails and
-    // -Werror is in the commandline.
-    //
-    // Otherwise if -Werror is missing we can just use the first instance.
-    if !parsed_args.common_args.iter().any(|a| a.starts_with("-Werror")) {
-        return Box::new(output)
-    }
-
-    let mut cmd = creator.clone().new_command_sync(executable);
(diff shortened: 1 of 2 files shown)


luser:
Can we just merge the clang and gcc compile implementations here, since we already have this behavior for clang?

flxo:
You're right. I'll give it a try. For the compilation step the only difference is, that `clang` needs to write the input to a temporary file and `gcc` reads `stdin` - thats all.

Extracting general functionality of the compiler adapters might make sense for future support for more compiler kinds. We use a `sccache` [extension for Windriver Diab](https://github.com/flxo/sccache/tree/diab) in our CI. Beside a couple of minor tweaks (especially for the `-@X` args) this is a boring copy of `gcc/clang` and it could also benefit from some infrastructure. All those C/C++ compiler behave more or less the same from the usage perspective...

luser:
During the refactoring I did for the Rust rewrite I tried to factor out a lot of the common logic and wrap it around `CCompilerImpl`, but we could probably take that further. Ideally the C compiler implementations would just be lists of arguments that merit special handling.

flxo:
I merged the `clang` and `gcc` compilation, not sure whether passing this pre compilation step is a good idea.

In parallel I tried to do a single `compile` method in `CCompilation` that handles to compiles for all compiler variants (and removed `compile`). This is not so easy because of the small differences, e.g. `mscv` `showinclude` option and so on. I extended `CCompilerImpl` with something like `pre_compilation` and `post_compilation` that return futures that are chained before and after the actual two compilation steps. After a while I decided to just update this PR with the merge of `gcc` and `clang` because the combination of all compiler variants will end up in bigger changesets.

flxo:
Ups. I'll check the CI. Sorry.

luser:
> In parallel I tried to do a single compile method in CCompilation that handles to compiles for all compiler variants (and removed compile).

Thanks for trying! It's a nice idea but probably would wind up being a bigger refactor, we can think about that another time.

src/compiler/gcc.rs
```diff
@@ -279,30 +279,60 @@ fn compile<T>(creator: &T,
...
-    let mut cmd = creator.clone().new_command_sync(executable);
-    cmd.args(&["-c", "-x"])
-        .arg(match parsed_args.extension.as_ref() {
-            "c" => "cpp-output",
-            "cc" | "cpp" | "cxx" => "c++-cpp-output",
-            e => {
-                error!("gcc::compile: Got an unexpected file extension {}", e);
-                return future::err("Unexpected file extension".into()).boxed()
-            }
+    let extension = match parsed_args.extension.as_ref() {
```
luser: We should probably change `parsed_args.extension` into something like `enum SourceType { C, Cxx }`, shouldn't we? (That can be a followup.)

src/compiler/clang.rs
```diff
@@ -109,64 +109,21 @@ fn compile<T>(creator: &T,
...
-    // Otherwise if -Werror is missing we can just use the first instance.
-    if !parsed_args.common_args.iter().any(|a| a.starts_with("-Werror")) {
-        return Box::new(output)
-    }
+        Box::new(
+            write_temp_file(pool, filename.as_ref(), preprocessor_result.stdout.clone())
+                 .and_then(move |(tempdir, input)| {
+                     match input.into_os_string().into_string() {
+                         Ok(p) => future::ok((None, vec!(p), Some(tempdir))),
+                         Err(_) => future::err("Failed to write input file".into()),
```
luser: FYI there are helper functions in error.rs to shorten these: `f_ok` and `f_err`.

luser:
Thanks for the patches!

src/compiler/clang.rs
flxo: Ah ok. Thanks, I didn't know.

flxo:
Thanks for merging!

Merging the C compiler stuff is a way bigger thing. I'm not even sure if there's so much benefit from that because of all those specialties.

luser:
> Merging the C compiler stuff is a way bigger thing. I'm not even sure if there's so much benefit from that because of all those specialties.

I agree. The refactoring I did in advance of adding Rust support was pretty invasive. If we decide it will be beneficial in the future we can revisit it.

PR #111, opened 2017-04-27
Attempt to debug fatal errors more easily

alexcrichton:
This is an attempt to assist in debuggin "Error 254" situations which represent
when the sccache server itself has a fatal error when executing the compiler.
This serializes the entire error context into a string and ships it back as
stderr so it gets printed out by the calling context, hopefully providing
contextual information about the error happening.


luser:
We should have done this ages ago, thanks!

PR #112, opened 2017-04-28
Leverage OsString much more deeply througout

### LINKED ISSUE #93: Use serde's OsStr and OsString impls
Following up on https://github.com/mozilla/sccache/pull/85#discussion_r108140271:

> **\<@luser>** as a follow-up, we could now just use OsStr or OsString all the way for the values from the client commandline.
> **\<@alexcrichton>** Yeah I tried to do that but Serialize/Deserialize wasn't implemented for one of them :(
> Actually... I'll send a PR!
> Ok this is now tracked at serde-rs/serde#823

I published Serde 0.9.12 containing Alex's impls for OsStr and OsString.

alexcrichton:
This initially started out tackling #93 and ended up plumbing the `OsStr` type
basically everywhere! This knocked out a TODO or two along the way and should
remove some pesky unwraps and/or error handling around handling of arguments and
paths that may not be unicode.

Some notable other points are:

* A few utilities were added to `src/util.rs` to assist with handling os strings
* The Rust implementation doesn't deal with os strings in argument parsing at
  all, but otherwise internally stores os strings for maximal compatibility
  (more comments in the code).
* Some unsafe transmutes were removed in util in favor of a more stable
  implementation.
* The `output_file` function was renamed to `output_pretty` as it turns out it's
  not actually the actual output file but rather just a pretty description of it
  (for logging and such).

Closes #93

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -21,9 +21,10 @@ use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 use std::fmt;
-use std::path::Path;
+use std::hash::Hash;
+use std::path::{Path, PathBuf};
 use std::process;
-use util::{os_str_bytes, sha1_digest};
+use util::{HashToSha1, sha1_digest};
 
 use errors::*;
 
@@ -32,7 +33,7 @@ use errors::*;
 pub struct CCompiler<I>
     where I: CCompilerImpl,
 {
-    executable: String,
+    executable: PathBuf,
     executable_digest: String,
     compiler: I,
 }
@@ -43,7 +44,7 @@ pub struct CCompilerHasher<I>
     where I: CCompilerImpl,
 {
     parsed_args: ParsedArguments,
-    executable: String,
+    executable: PathBuf,
     executable_digest: String,
     compiler: I,
 }
@@ -53,31 +54,34 @@ pub struct CCompilerHasher<I>
 #[derive(Debug, PartialEq, Clone)]
 pub struct ParsedArguments {
     /// The input source file.
-    pub input: String,
+    pub input: PathBuf,
     /// The file extension of the input source file.
     pub extension: String,
     /// The file in which to generate dependencies.
-    pub depfile: Option<String>,
+    pub depfile: Option<PathBuf>,
     /// Output files, keyed by a simple name, like "obj".
-    pub outputs: HashMap<&'static str, String>,
+    pub outputs: HashMap<&'static str, PathBuf>,
     /// Commandline arguments for the preprocessor.
-    pub preprocessor_args: Vec<String>,
+    pub preprocessor_args: Vec<OsString>,
     /// Commandline arguments for the preprocessor or the compiler.
-    pub common_args: Vec<String>,
+    pub common_args: Vec<OsString>,
     /// Whether or not the `-showIncludes` argument is passed on MSVC
     pub msvc_show_includes: bool,
 }
 
 impl ParsedArguments {
-    pub fn output_file(&self) -> Cow<str> {
-        self.outputs.get("obj").and_then(|o| Path::new(o).file_name().map(|f| f.to_string_lossy())).unwrap_or(Cow::Borrowed("Unknown filename"))
+    pub fn output_pretty(&self) -> Cow<str> {
+        self.outputs.get("obj")
+            .and_then(|o| o.file_name())
+            .map(|s| s.to_string_lossy())
+            .unwrap_or(Cow::Borrowed("Unknown filename"))
     }
 }
 
 /// A generic implementation of the `Compilation` trait for C/C++ compilers.
 struct CCompilation<I: CCompilerImpl> {
     parsed_args: ParsedArguments,
-    executable: String,
+    executable: PathBuf,
     /// The output from running the preprocessor.
     preprocessor_result: process::Output,
     compiler: I,
@@ -100,14 +104,14 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
     fn kind(&self) -> CCompilerKind;
     /// Determine whether `arguments` are supported by this compiler.
     fn parse_arguments(&self,
-                       arguments: &[String],
+                       arguments: &[OsString],
                        cwd: &Path) -> CompilerArguments<ParsedArguments>;
     /// Run the C preprocessor with the specified set of arguments.
     fn preprocess<T>(&self,
                      creator: &T,
-                     executable: &str,
+                     executable: &Path,
                      parsed_args: &ParsedArguments,
-                     cwd: &str,
+                     cwd: &Path,
                      env_vars: &[(OsString, OsString)],
                      pool: &CpuPool)
                      -> SFuture<process::Output> where T: CommandCreatorSync;
@@ -115,10 +119,10 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
     /// previously-generated `preprocessor_output` as input if possible.
     fn compile<T>(&self,
                   creator: &T,
-                  executable: &str,
+                  executable: &Path,
                   preprocessor_result: process::Output,
                   parsed_args: &ParsedArguments,
-                  cwd: &str,
+                  cwd: &Path,
                   env_vars: &[(OsString, OsString)],
                   pool: &CpuPool)
                   -> SFuture<(Cacheable, process::Output)>
@@ -128,7 +132,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
 impl <I> CCompiler<I>
     where I: CCompilerImpl,
 {
-    pub fn new(compiler: I, executable: String, pool: &CpuPool) -> SFuture<CCompiler<I>>
+    pub fn new(compiler: I, executable: PathBuf, pool: &CpuPool) -> SFuture<CCompiler<I>>
     {
         Box::new(sha1_digest(executable.clone(), &pool).map(move |digest| {
             CCompiler {
@@ -143,7 +147,7 @@ impl <I> CCompiler<I>
 impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
     fn kind(&self) -> CompilerKind { CompilerKind::C(self.compiler.kind()) }
     fn parse_arguments(&self,
-                       arguments: &[String],
+                       arguments: &[OsString],
                        cwd: &Path) -> CompilerArguments<Box<CompilerHasher<T> + 'static>> {
         match self.compiler.parse_arguments(arguments, cwd) {
             CompilerArguments::Ok(args) => {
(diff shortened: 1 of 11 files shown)


gumpt:
This would be so nice!  I ran into a lot of conversions when working on [#104](https://github.com/mozilla/sccache/pull/104) between PathBuf <-> OsString <-> String.

src/compiler/msvc.rs
```diff
@@ -173,53 +173,63 @@ pub fn parse_arguments(arguments: &[String]) -> CompilerArguments<ParsedArgument
...
     });
     while let Some(arg) = it.next() {
-        match &arg[..] {
-            "-c" => compilation = true,
-            v if v.starts_with("-Fo") => {
-                output_arg = Some(String::from(&v[3..]));
-            }
-            // Arguments that take a value.
-            "-FI" => {
+        if let Some(arg) = arg.split_prefix("-Fo") {
```
luser: I wonder if this code would be nicer if we refactored it similarly to the Rust code that splits it out into `(arg, val)` tuples? We'd have to do some additional fiddling to handle the goofy MSVC argument styles.

src/compiler/msvc.rs
```diff
@@ -173,53 +173,63 @@ pub fn parse_arguments(arguments: &[String]) -> CompilerArguments<ParsedArgument
...
+            depfile = Some(arg);
+            continue
+        }
+        if let Some(p) = arg.split_prefix("-Fd") {
+            pdb = Some(p);
+            common_args.push(arg.clone());
+            continue
+        }
+        match arg.to_str() {
+            Some("-c") => compilation = true,
```
luser: Same comment here as in the gcc code--this is a little clunky.

src/compiler/msvc.rs
```diff
@@ -340,13 +350,15 @@ pub fn preprocess<T>(creator: &T,
...
+    let cwd = cwd.to_owned();
 
     Box::new(run_input_output(cmd, None).and_then(move |output| {
         let parsed_args = &parsed_args;
         if let (Some(ref objfile), &Some(ref depfile)) = (parsed_args.outputs.get("obj"), &parsed_args.depfile) {
-            let mut f = File::create(Path::new(&cwd).join(depfile))?;
-            write!(f, "{}: {} ", objfile, parsed_args.input)?;
+            let mut f = File::create(cwd.join(depfile))?;
+
+            // TODO: these calls to `display` are lossy ...
```
luser: This isn't really worse than what we had before, where a source file name that wasn't valid Unicode would just fail somewhere along the way, but it's a little worrying that we could be writing bad data here and silently breaking things. In any event, you don't have to fix this right now. I was going to say you could use [`local_encoding`](https://docs.rs/local-encoding/0.2.0/local_encoding/) like we're using to decode the MSVC output, but that only deals with turning strings into bytes and that doesn't actually help. I guess we could open an issue on `local_encoding` to support encoding `&[u16]` or some such to bytes, and then pass it `input.to_wide()`? It'd basically just need to call `WideCharToMultiByte`, and I'm sure it's already doing something very similar.

src/compiler/rust.rs
```diff
@@ -348,8 +358,20 @@ impl<'a> Iterator for ArgsIter<'a> {
     }
 }
 
-fn parse_arguments(arguments: &[String], _cwd: &Path) -> CompilerArguments<ParsedArguments>
+fn parse_arguments(arguments: &[OsString], _cwd: &Path) -> CompilerArguments<ParsedArguments>
 {
+    // While we could go the extra mile here and handle non-utf8 `OsString`
+    // instances the rustc compiler certainly does not. With that knowledge
```
luser: Hah!

src/util.rs
```diff
@@ -134,10 +120,151 @@ pub fn run_input_output<C>(mut command: C, input: Option<Vec<u8>>)
...
 
-#[test]
-fn test_os_str_bytes() {
-    // Just very basic sanity checks in case anyone changes the underlying
-    // representation of OsStr on Windows.
-    assert_eq!(os_str_bytes(OsStr::new("hello")), b"hello");
-    assert_eq!(os_str_bytes(OsStr::new("你好")), "你好".as_bytes());
+pub trait OsStrExt {
+    fn starts_with(&self, s: &str) -> bool;
+    fn split_prefix(&self, s: &str) -> Option<OsString>;
```
luser: I keep thinking that these sorts of things would be really nice in std.

src/util.rs
```diff
@@ -134,10 +120,151 @@ pub fn run_input_output<C>(mut command: C, input: Option<Vec<u8>>)
...
+#[cfg(windows)]
+use std::os::windows::ffi::{OsStrExt as _OsStrExt, OsStringExt};
+
+#[cfg(windows)]
+impl OsStrExt for OsStr {
+    fn starts_with(&self, s: &str) -> bool {
+        // Attempt to interpret this OsStr as utf-16. This is a pretty "poor
+        // man's" implementation, however, as it only handles a subset of
+        // unicode characters in `s`. Currently that's sufficient, though, as
+        // we're only calling `starts_with` with ascii string literals.
```
luser: This is clever, but not so clever that it's hard to understand. :)

src/util.rs
```diff
@@ -134,10 +120,151 @@ pub fn run_input_output<C>(mut command: C, input: Option<Vec<u8>>)
...
+    pub sha: &'a mut sha1::Sha1,
+}
+
+impl<'a> Hasher for HashToSha1<'a> {
+    fn write(&mut self, bytes: &[u8]) {
+        self.sha.update(bytes)
+    }
+
+    fn finish(&self) -> u64 {
+        panic!("not supposed to be ");
```
luser: This panic string feels unfinished?

luser:
Oh! But you should fix whatever caused the CI bustage, obviously. Possibly just an issue being rebased over one of your other PRs?
```
error[E0425]: cannot find value `output` in this scope
   --> src\server.rs:597:54
    |
597 |                     error!("[{:?}] fatal error: {}", output, err);
    |                                                      ^^^^^^ not found in this scope
error[E0425]: cannot find value `output` in this scope
   --> src\server.rs:602:47
    |
602 |                         error!("[{:?}] \t{}", output, e);
    |                                               ^^^^^^ not found in this scope
error: aborting due to 2 previous errors
```

src/util.rs
alexcrichton: Definitely! We on the libs team have often wanted some form of string-like API added to `OsStr` as well, we just never figured out the best way to do so... For now less efficient implementations :(

src/compiler/c.rs
```diff
@@ -267,21 +268,25 @@ pub const CACHED_ENV_VARS : &'static [&'static str] = &[
 ];
 
 /// Compute the hash key of `compiler` compiling `preprocessor_output` with `args`.
-pub fn hash_key(compiler_digest: &str, arguments: &str, env_vars: &[(OsString, OsString)],
+pub fn hash_key(compiler_digest: &str,
+                arguments: &[OsString],
+                env_vars: &[(OsString, OsString)],
                 preprocessor_output: &[u8]) -> String
 {
     // If you change any of the inputs to the hash, you should change `CACHE_VERSION`.
```
alexcrichton: Nah I think in practice this is going to end up disturbing the hashes a bit. The motivation of `HashToSha1` was to avoid transmuting `OsStr` to bytes (which while valid today isn't something we'd like to commit to in libstd). 

Is changing the hash though a problem? I could back that out if it's best left for another time.

src/compiler/c.rs
```diff
@@ -202,17 +206,14 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
-            // Remove object file from arguments before hash calculation
             let key = {
-                let out_file = parsed_args.output_file();
-                let arguments = parsed_args.common_args.iter()
-                    .filter(|a| **a != out_file)
-                    .map(|a| a.as_str())
-                    .collect::<String>();
-                hash_key(&executable_digest, &arguments, &env_vars, &preprocessor_result.stdout)
+                hash_key(&executable_digest,
+                         &parsed_args.common_args,
```
alexcrichton: Oh right sorry I meant to ask you about this. I wasn't really sure if this was still necessary. The test here is against `output_file` which is what prompted me to rename the relevant method to `output_pretty` because the return value of `output_file` is just the file name, not the full path.

In that sense I'm not sure that this was 100% working before, and I otherwise couldn't figure out what it was intended to be doing. I thought the output file was already missing from the `common_args`?

src/compiler/c.rs
```diff
@@ -170,26 +174,26 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
+                         cwd: &Path,
                          env_vars: &[(OsString, OsString)],
                          pool: &CpuPool)
                          -> SFuture<HashResult<T>>
     {
         let me = *self;
         let CCompilerHasher { parsed_args, executable, executable_digest, compiler } = me;
         let result = compiler.preprocess(creator, &executable, &parsed_args, cwd, env_vars, pool);
-        let out_file = parsed_args.output_file().into_owned();
+        let out_pretty = parsed_args.output_pretty().into_owned();
```
alexcrichton: Yeah one thing I was considering in the logging PR was to add the output file to the "local logger's scope" and that way it'd automatically get prepended to all log messages. What would you think of that approach? (I can change that PR over there)

src/compiler/msvc.rs
alexcrichton: Oh looks like `WideCharToMultiByte` may actually [be supported](https://docs.rs/local-encoding/0.2.0/i686-pc-windows-gnu/local_encoding/windows/fn.wide_char_to_multi_byte.html), I'll play around with calling that (never tried it before!)

alexcrichton:
Ok I think I've updated with all comments, but I'll hold off on merging until I hear back on [removing the object file bit](https://github.com/mozilla/sccache/pull/112#discussion_r114597859) changing [the hash value](https://github.com/mozilla/sccache/pull/112#discussion_r114598122).

If you wanna take a peek at what I did to `parse_argument` in gcc/msvc you can also tell me if I'm crazy!

src/compiler/c.rs
luser: Changing the hash is fine, but if you're doing so please bump the cache version as well so we're explicit about it.

alexcrichton:
@luser just to confirm, is [removing the object file handling](https://github.com/mozilla/sccache/pull/112#discussion_r114597859) ok there?

luser:
> just to confirm, is removing the object file handling ok there?

Given that the code as written was broken, that's fine. We can reopen the issue that made that change originally.

src/compiler/c.rs
luser: That would make things a lot nicer vs. all the cloning we do just to have useful log messages.

luser:
> If you wanna take a peek at what I did to parse_argument in gcc/msvc you can also tell me if I'm crazy!

Those changes look reasonable. You've got CI bustage, maybe it's just from rebasing over https://github.com/mozilla/sccache/commit/70b4426c44c08f0891245eca4fbe5e671f9efc85 ? That added some new tests that probably need the same fixup you did for other tests.

Feel free to land this once you fix that.

alexcrichton:
Ok thanks for the info! I've opened an issue at https://github.com/mozilla/sccache/issues/121

PR #113, opened 2017-04-28
Read SCCACHE_RECACHE from the client

### LINKED ISSUE #96: Use SCCACHE_RECACHE from client's environment, not server
Currently the server looks for `SCCACHE_RECACHE` in its environment [on startup](https://github.com/mozilla/sccache/blob/bbb850245430cfb444b4f4d222ce3e6dff2c5f59/src/server.rs#L383) and then [uses that value](https://github.com/mozilla/sccache/blob/bbb850245430cfb444b4f4d222ce3e6dff2c5f59/src/server.rs#L531) for every compile request. Since fixing https://github.com/mozilla/sccache/issues/48 we now have the client's environment available in `start_compile_task`, so we should use the value of `SCCACHE_RECACHE` from there instead. This should make it easier to use since you won't have to restart the server to force a recache.

alexcrichton:
Now that client env vars are sent to the server, this was much easier! This
allows forcing a recache from the client rather than having to set it for the
entire server.

Closes #96

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -40,7 +40,7 @@ use protocol::{Compile, CompileFinished, CompileResponse, Request, Response};
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::env;
-use std::ffi::OsString;
+use std::ffi::{OsStr, OsString};
 use std::fs::metadata;
 use std::io::{self, Write};
 use std::net::{SocketAddr, SocketAddrV4, Ipv4Addr};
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -526,7 +514,10 @@ impl<C> SccacheService<C>
                           cwd: String,
                           env_vars: Vec<(OsString, OsString)>,
                           tx: mpsc::Sender<Result<Response>>) {
-        let cache_control = if self.force_recache {
+        let force_recache = env_vars.iter().any(|&(ref k, ref _v)| {
+            k.as_os_str() == OsStr::new("SCCACHE_RECACHE")
```
luser: `OsStr` has a `PartialEq` impl that handles comparing to '&str', so you don't need the `OsStr::new`, I don't think.

luser:
I thought we had unit tests that would need fixing, but turns out they're in `compiler.rs` and call `get_cached_or_compile` directly.

PR #115, opened 2017-05-02
Ship all log messages to the client

alexcrichton:
This commit updates the sccache server to capture all log messages and ship them
to the client. Captured log messages are then printed out by the client. The
intention of this is to assist with debugging sccache by setting the `RUST_LOG`
env var on the client and having it propagate to the server and back.

With this PR you'd start up the server with a maximum log level set, e.g. with

    RUST_LOG=foo=debug

and then all clients would use their own `RUST_LOG` env var when executing to
get logs message shipped back. Note that log messages show up on the server by
default as well, but log messages are typically cast into the void as
stdout/stderr are closed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,6 @@ dependencies = [
  "daemonize 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "fern 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "filetime 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures-cpupool 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


alexcrichton:
Some notes about this:

* I'm not 100% sure this is exactly what we'd want to land, so I'm open to discussion!
* I strived to make this as unintrusive as possible, so only the edges of the connection are touched, nothing in the `src/compile/*.rs` files and such. (or in caching)
* The `log` crate has some behavior to be aware of here. When you first install a logger a global log level is configured. This global log level is *unconditionally* respected and cannot be changed after initialization.  This means that if you start up the server without a `RUST_LOG` env var then nothing will ever be logger. That's why the example above says `foo=debug`, basically just configuring anything with the `debug` log level to make logging possible. We could instead unconditionally configure a `Trace` log level by default and let clients filter out more (for example), but this doesn't interact with existing log messages unfortunately as there's still low-level log messages not captured here.
* I removed the dep on `fern` to avoid having multiple logging impls, but I can add it back in if you'd like!

luser:
I'll have to take a closer look at this, but I like the idea, for sure!

> I removed the dep on fern to avoid having multiple logging impls, but I can add it back in if you'd like!

This is fine, I was thinking about removing it anyway since `SCCACHE_ERROR_LOG` gives the same utility of logging to a file with the finer-grained log levels that env_logger supports.

alexcrichton:
Ok I've rebased over https://github.com/mozilla/sccache/pull/112. I wanted to add the "set a TLS context of the current local object" feature here, but unfortunately couldn't due to a missing feature in the `log` crate, but I suspect that will be fixed relatively soon!

In the meantime it may be the case that the log messages won't actually be that necessary if we're shipping all the relevant messages to the client!

luser:

PR #118, opened 2017-05-03
Adds ability to change storage compression scheme

na-g:
I would like the ability to allow the filesystem (ZFS, BTRFS, etc.) to handle compression. This adds an environment variable to control the compression method. The Zip crate has optional support for Bzip2 (not that it would be a fast choice), though that feature doesn't seem to be enabled. For right now the compression methods available are "deflate" and "none". The default behavior remains unchanged.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -125,7 +125,19 @@ impl CacheWrite {
     pub fn put_object<T>(&mut self, name: &str, from: &mut T, mode: Option<u32>) -> Result<()>
         where T: Read,
     {
-        let opts = FileOptions::default().compression_method(CompressionMethod::Deflated);
+        let method = if let Ok(ref method) = env::var("SCCACHE_COMPRESSION_METHOD") {
+            match method.to_lowercase().trim() {
+                "deflate" => CompressionMethod::Deflated,
+                "none" => CompressionMethod::Stored,
+                _ => {
+                    warn!("Unknown compression method {} using 'deflate'.", method);
+                    CompressionMethod::Deflated
+                },
+            }
+        } else {
+          CompressionMethod::Deflated
+        };
+        let opts = FileOptions::default().compression_method(method);
         let opts = if let Some(mode) = mode { opts.unix_permissions(mode) } else { opts };
         self.zip.start_file(name, opts).chain_err(|| {
             "Failed to start cache entry object"


src/cache/cache.rs
```diff
@@ -125,7 +125,19 @@ impl CacheWrite {
     pub fn put_object<T>(&mut self, name: &str, from: &mut T, mode: Option<u32>) -> Result<()>
         where T: Read,
     {
-        let opts = FileOptions::default().compression_method(CompressionMethod::Deflated);
+        let method = if let Ok(ref method) = env::var("SCCACHE_COMPRESSION_METHOD") {
+            match method.to_lowercase().trim() {
+                "deflate" => CompressionMethod::Deflated,
+                "none" => CompressionMethod::Stored,
+                _ => {
+                    warn!("Unknown compression method {} using 'deflate'.", method);
```
flxo: This would silently fail on typos etc. Maybe it's better to return `Err` to show the user that something has gone wrong

src/cache/cache.rs
```diff
@@ -125,7 +125,19 @@ impl CacheWrite {
...
+                    CompressionMethod::Deflated
+                },
+            }
+        } else {
+          CompressionMethod::Deflated
```
flxo: Here are two spaces missing.

flxo:
Did you think about making the compression method part of the hash calculation? Technically it should not be needed but I could imagine that you want to (with non local caches).

src/cache/cache.rs
na-g: I am not entirely sure what you mean by silent failure, could you clarify? My thinking is that it is logging a warning and falling back to deflate in the case of typos. If it is to be a hard error (prevents caching), I would think the proper place to do this check would be startup. It doesn't seem that sccache makes those kinds of configuration checks yet.

na-g:
@flxo Yes, and I don't think it should be part of the hash for a few reasons.

1. it doesn't effect the compilation results
2. if you are just a consumer of a remote cache you shouldn't need to know its compression scheme to be able to get cache hits
3. the compression type does not need to be set on read, it is automatically detected, so there is not a problem with having heterogeneously compressed entries

na-g:
Another major motivation for this patch, which I should have stated above, is that when I am building Firefox NVME bandwith is much more abundant than CPU time. At least with ccache I have found there is a non-trivial CPU time cost to compressing results.

src/cache/cache.rs
flxo: Sorry for being imprecise. The default logging setup doesn't show any logs. Furthermore until #115 is merged it's hard for the client to get the logs at all. On a CI system you also might have the case that it's unclear which specific job the starts the server. Restarting to ensure log settings might break the one or other job running in parallel. Personally I'd prefer to see cache errors in the stats if I did something wrong in the cache configuration.

flxo:
@na-g Understand. I just thought about the case where one doesn't care about bandwidth and sets `stored`.  Someone else works with a poor connection, gets hits and is forced to pull the big blobs. For the later the only option would be to force a recache and update the shared cache with `deflate`.

na-g:
@flxo alternatively, could the problem you described be solved with `Accept-Encoding: gzip` (or deflate, etc.) support, so it is handled at the HTTP layer?

If so, I could limit the option to local cache only until that is implemented.

flxo:
It probably can be solved by configuration: Using dedicated `S3` buckets or select a specific database index for the `Redis` backend, depending on the compression method - but then it's up to the user to do it right.

na-g:
I am not sure what the path forward is here. If the worry is that sub-optimal configurations may be possible, I am not sure that much can be done. If in the future other compression algorithms are supported, e.g. LZ4, then it doesn't make much sense to have the algorithm be part of the hash.

luser:

PR #119, opened 2017-05-09
Handle storage get failure

luser:
We're seeing failures reading cache entries in Firefox CI. I'm not sure if we're getting a truncated response from S3 or what, but it's failing to parse the zip file. This change makes storage failures like that non-fatal, they're just handled as cache misses. I believe this was the original behavior of sccache and it got changed during the tokio refactor.

I've also added another change to check that the response body matches the `Content-Length` header for S3 requests, and a diagnostic log for the response body in all cases.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -147,7 +147,7 @@ pub trait CompilerHasher<T>: fmt::Debug + Send + 'static
             });
 
             // Check the result of the cache lookup.
-            Box::new(cache_status.and_then(move |result| {
+            Box::new(cache_status.then(move |result| {
                 let duration = start.elapsed();
                 let pwd = Path::new(&cwd);
                 let outputs = compilation.outputs()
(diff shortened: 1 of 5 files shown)


alexcrichton:
:+1:

alexcrichton:
I think we just hit this in rust-lang/rust at https://travis-ci.org/rust-lang/rust/jobs/231599352

> ERROR:sccache::server: [Error(Msg("Failed to parse cache entry"), State { next_error: Some(InvalidArchive("Could not find central directory end")) })] 	HexagonSplitDouble.cpp.o

I'm updating to sccache's master branch now so we can get the better logging from this.

luser:
Yeah, that's the same error I had seen in Firefox CI:
> DEBUG:sccache::server: [Error(Msg("Failed to parse cache entry"), State { next_error: Some(InvalidArchive("Could not find central directory end")) })] compilation failed: "Unified_cpp_layout_painting0.o"

This patch will make it non-fatal, although I'd love to figure out what the root cause is.

PR #120, opened 2017-05-11
Get rid of Storage::start_put and rename finish_put

### LINKED ISSUE #71: Get rid of `Storage::start_put`
The `start_put` / `finish_put` API is a lot less useful than it used to be. Originally `DiskCache` would return a `CacheWrite` that was writing directly to a `File`, so it didn't have to buffer the zip in-memory first. When I added the `LRUDiskCache` I wound up making `DiskCache` just hand out a `Cursor` from `start_put` instead, so all the cache implementations are returning the same thing from `start_put` now.

We should probably just get rid of `start_put` and make the code that currently calls it just call `CacheWrite::new()` directly.

flxo:
This patch removes the start_put function from the `Storage` trait. The `finish_put` function is renamed
to `put` get a nice `get`/`put` pair in `Storage.

This fixes #71.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -154,14 +154,11 @@ pub trait Storage {
     /// return a `Cache::Hit`.
     fn get(&self, key: &str) -> SFuture<Cache>;
 
-    /// Get a cache entry for `key` that can be filled with data.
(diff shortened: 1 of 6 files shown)


flxo:
Rebased. Thanks for the quick review of this simple cleanup!

Issue #123, opened 2017-05-18
Support caching Rust compilation with `-l static=foo`


luser:
I punted on caching any Rust compilation with `-l` in it in the original implementation, but [@alexcrichton pointed out](https://github.com/mozilla/sccache/issues/88#issuecomment-288729305) that supporting just `-l static=foo` should be straightforward.

luser:
Just for reference, I built servo with sccache on a slightly out-of-date servo rev (https://github.com/servo/servo/commit/bb4ea82b236ebcd7cc6416262045a91438beb089) and the sccache logs showed 127 `CannotCache` lines. Doing a little grep-and-sed on the log showed:
```
$ rg CannotCache /tmp/sccache.log | sed -re 's/.*CannotCache\((.+)\).*/\1/' | sort | uniq -c
     22 -l
     36 multiple input files
     50 no output file
      1 stdin
     18 unknown source extension
```
So there are 22 rust compilations that weren't cacheable due to passing `-l`.

luser:
I wrote a little Python script to filter those down to just invocations with `-l static=something`, and those account for 12/22.

alexcrichton:
Ah yeah `-lstatic` sould be pretty easy to support. The logic basically looks like:

* Take a look at all `-L` arguments
* Strip out everything that's not `-L /foo` and `-L native=/foo`, e.g. strip `dependency=` and `framework=`, etc.
* For all of those paths iterate them in order
* For linking the staticlib `foo`, find the first of `libfoo.a`, `foo.lib`, `foo.a`, and return that.

Or at least IIRC that's the rough algorithm rustc uses!

luser:
Fixed in https://github.com/mozilla/sccache/pull/125

Issue #88, opened 2017-03-23
Support Rust compilations that link against native libraries


luser:
I'm punting on this for now by returning `CompilerArguments::CannotCache` for any compilation containing `-l`, so it shouldn't break anything, sccache will just refuse to cache any compilations that would link native libraries.

The complicated part is that we would have to replicate the linker's search strategy to fix this. If we can get rustc to tell us what linker it would use, maybe the best way to do that would be to interrogate it as to what its search paths are?

alexcrichton:
I think this may actually be relatively easy in practice. If we assume that sccache isn't handling linker invocations yet (e.g. creation of dylibs, cdylib, executables, etc), then all we're left with is rlibs. Rlibs in general don't process `-l` flags except for `-l static=foo`, and those have a pretty naive lookup strategy (easy to reimplement).

Punting for now seems good regardless.

Dushistov:
It would be great to have caching of code that link againts with native.

For example if I rebuild my code twice with sccache enabled, the second time I got:

 Finished release [optimized + debuginfo] target(s) in 480.33 secs

So source code not changed, it already build once and sccache cache all compiler invocation,
but still it takes 480 seconds, stats:
```sh
evgeniy@localhost ~ $ sccache -s
Compile requests                217
Compile requests executed       136
Cache hits                      110
Cache misses                     26
Cache timeouts                    0
Forced recaches                   0
Cache write errors                0
Compilation failures              0
Cache errors                      0
Non-cacheable compilations        0
Non-cacheable calls              16
Non-compilation calls            65
Unsupported compiler calls        0
Average cache write           0.003 s
Average cache read miss      44.699 s
Average cache read hit        0.001 s
Cache location             Local disk: "/home/evgeniy/.cache/sccache"
Cache size                      924 MiB
Max cache size                   10 GiB
```

clang-sys, rusqlite and others require a lot of cpu power.

luser:
> I think this may actually be relatively easy in practice. If we assume that sccache isn't handling linker invocations yet (e.g. creation of dylibs, cdylib, executables, etc), then all we're left with is rlibs. Rlibs in general don't process -l flags except for -l static=foo, and those have a pretty naive lookup strategy (easy to reimplement).

I don't think I read this carefully at the time, but I think supporting just `-l static=foo` as a first pass would be a good step.

alexcrichton:
[...]

PR #125, opened 2017-05-19
Support Rust compilation with -l and stop caching non-rlib/staticlib crate output

luser:
As discussed in https://github.com/mozilla/sccache/issues/88 and https://github.com/mozilla/sccache/issues/123 .

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -22,6 +22,7 @@ use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::env::consts::DLL_EXTENSION;
 use std::ffi::OsString;
+use std::fmt;
 use std::fs::{self, File};
 use std::hash::Hash;
 use std::io::Read;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me! To confirm, different `-l` flags on the command line still generate different hashes, right? In that `rustc -l foo` is a different hash than `rustc -l bar`? I didn't *immediately* see where that happens, but so long as it happens somewhere r=me and feel free to merge.

luser:
> Looks good to me! To confirm, different -l flags on the command line still generate different hashes, right? In that rustc -l foo is a different hash than rustc -l bar? I didn't immediately see where that happens, but so long as it happens somewhere r=me and feel free to merge.

Correct. Since we're no longer rejecting `-l` in `parse_arguments`, they're just included in `ParsedArguments::arguments` and we include all of that in the hash calculation.

alexcrichton:
🍨

Issue #128, opened 2017-05-19
Handle default object file name for C compilation


luser:
We currently return `CannotCache(no output file)` if a C compilation doesn't include `-o output` or the MSVC equivalent. Compilers use pretty simple rules to generate the output file here, we should just support that case. This shows up a lot in sccache logs from building servo because autoconf does this a lot. I'm not sure it will help much, but it should be easy to support and there's no good reason not to.

[GCC's documentation](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options) says "By default, the object file name for a source file is made by replacing the suffix ‘.c’, ‘.i’, ‘.s’, etc., with ‘.o’. "  [MSVC's documentation](https://docs.microsoft.com/en-us/cpp/build/reference/fo-object-file-name) says "If you do not use this option, the object file uses the base name of the source file and the .obj extension." Related: @vvuk mentioned to me that he had hit an issue because he was building a project with MSVC+sccache that used `-Fosome\dir\`, which I didn't know was a thing you could do (even though it's mentioned in the docs I just linked), which causes MSVC to generate the object file in `some\dir` with the default object file name.

F3real:
Hey Ted,
I'll look into this.

vvuk:
The windows `-Fo` fix is at https://github.com/vvuk/sccache/commit/9ed9a78d09105ff70b6bebeb94ce1c7f34200785#diff-885eee6e54f6200dcc65474f2a44989fR293 -- should be easy to also handle a null output_arg there and do the replacement.

PR #130, opened 2017-05-24
Use default name if -o not given

F3real:
Solves #128

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -216,21 +216,21 @@ fn _parse_arguments(arguments: &[OsString],
         None => return CompilerArguments::CannotCache("no input file"),
     };
     let mut outputs = HashMap::new();
-    match output_arg {
+    let output = match output_arg {
         // We can't cache compilation that doesn't go to a file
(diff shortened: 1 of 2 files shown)


luser:
This looks great, thanks! I have some little nitpicky suggestions but I'll just clean them up and push this for you. I'll add some comments just so you can see what I would do.

src/compiler/gcc.rs
```diff
@@ -218,7 +218,7 @@ fn _parse_arguments(arguments: &[OsString],
     let mut outputs = HashMap::new();
     let output = match output_arg {
         // We can't cache compilation that doesn't go to a file
-        None => OsString::from(input.to_str().unwrap().to_owned() + ".o"),
+        None => OsString::from(Path::new(&input).with_extension("o")),
```
luser: You just wind up converting this back to a `Path` and `PathBuf` down below, so might as well leave it a `PathBuf` here, and make the `Some` case return a `PathBuf` as well.

luser:
Squashed and landed as https://github.com/mozilla/sccache/commit/04136395f252b4a30cd82136ef417da880736a51 . Thanks again!

PR #133, opened 2017-05-25
Use base64 instead of rustc-serialize

saneyuki:
[rustc-serialize has been deprecated](https://github.com/rust-lang-deprecated/rustc-serialize).
Instead, this pull request switchs to [base64](https://crates.io/crates/base64)


luser:
Thanks for the fix!

PR #134, opened 2017-05-25
Add clang arguments to the list of arguments that take values.

russellmcc:
These arguments are passed by xcode during normal builds, so should be supported by sccache.


src/compiler/gcc.rs
```diff
@@ -83,7 +83,7 @@ pub const ARGS_WITH_VALUE: &'static [&'static str] = &[
     "-iframework", "-imacros", "-imultilib", "-include",
     "-install_name", "-iprefix", "-iquote", "-isysroot",
     "-isystem", "-iwithprefix", "-iwithprefixbefore",
-    "-u",
+    "-u", "-x", "-arch", "--serialize-diagnostics",
```
luser: Thanks for the patch! This looks reasonable, but shouldn't these go in clang.rs if they're clang-specific?

russellmcc:
Oops!  Thanks, I will fix.

PR #137, opened 2017-06-05
Add GCS Support

cramertj:


cramertj:
Rebased. Ping @luser-- can this be merged? Do you have any changes you'd like to see?

src/cache/gcs.rs
```diff
@@ -0,0 +1,353 @@
...
+        let auth_request_jwt = jwt::encode(
+            &jwt::Header::new(jwt::Algorithm::RS256),
+            &jwt_claims,
+            &binary_key,
+        )?;
+
+        Ok(auth_request_jwt)
+    }
+
+    pub fn credentials(&self, client: &Client<HttpsConnector>) -> SFuture<GCSCredential> {
```
luser: I think you should split this function up a bit--it's got a lot going on and it's hard to follow it all.

src/cache/gcs.rs
```diff
@@ -143,16 +143,8 @@ pub struct GCSCredentialProvider {
 
 #[derive(Debug, Deserialize)]
 struct ServiceAccountKey {
-    #[serde(rename = "type")]
```
luser: It's not clear to me from the serde-json docs--will serde deserialize a struct OK if there are additional fields in the JSON? If so it might be worth a comment here mentioning that there are additional fields that we don't need.

src/cache/cache.rs
```diff
@@ -230,23 +230,48 @@ pub fn storage_from_environment(pool: &CpuPool, _handle: &Handle) -> Arc<Storage
...
+                          "No SCCACHE_GCS_KEY_PATH specified-- no authentication will be used.");
+                        None
+                    }
+                };
+
+                let rw_mode = match env::var("SCCACHE_RW_MODE").as_ref().map(String::as_str) {
+                    Ok("READ_ONLY") => RWMode::ReadOnly,
+                    Ok("READ_WRITE") => RWMode::ReadWrite,
+                    Ok(_) => {
+                        warn!("Invalid SCCACHE_RW_MODE-- defaulting to READ_ONLY.");
```
luser: I would like to provide a way for storage backends to error during initialization and have it produce an immediate error on the client. It would be much nicer to get a clear error message in these cases rather than digging through logs to try to figure it out. (Nothing you need to do here, just a note.)

src/cache/cache.rs
```diff
@@ -230,23 +230,48 @@ pub fn storage_from_environment(pool: &CpuPool, _handle: &Handle) -> Arc<Storage
...
+                let base_url = match env::var("SSCACHE_GCS_BASE_URL") {
+                    Ok(base_url) => base_url,
+                    _ => {
+                        let default = "https://www.googleapis.com";
+                        warn!("No SCCACHE_GCS_BASE_URl specified, using default: {}", default);
+                        default.to_owned()
+                    }
+                };
+
+                let cred_path = match env::var("SCCACHE_GCS_KEY_PATH") {
```
luser: I would probably write this more like:
```
let cred_path = env::var("SCCACHE_GCS_KEY_PATH").ok();
if cred_path.is_none() {
 ...
}
```

src/cache/cache.rs
```diff
@@ -230,23 +230,48 @@ pub fn storage_from_environment(pool: &CpuPool, _handle: &Handle) -> Arc<Storage
...
+                    Ok(cred_location) => Some(cred_location),
+                    _ => {
+                        warn!(
+                          "No SCCACHE_GCS_KEY_PATH specified-- no authentication will be used.");
+                        None
+                    }
+                };
+
+                let rw_mode = match env::var("SCCACHE_RW_MODE").as_ref().map(String::as_str) {
```
luser: This variable is not very clearly named, and since it's only honored by this storage backend it should probably have GCS in the name.

src/cache/gcs.rs
```diff
@@ -96,19 +104,27 @@ impl Bucket {
...
+            future::Either::B(future::ok(None))
+        };
+
+        Box::new(creds_opt_future.and_then(move |creds_opt| {
             let mut request = Request::new(Method::Post, url.parse().unwrap());
             {
                 let mut headers = request.headers_mut();
-                headers.set(Authorization(Bearer { token: creds.token }));
+                if let Some(creds) = creds_opt {
+                    headers.set(Authorization(Bearer { token: creds.token }));
```
luser: I assume a put request isn't going to succeed without credentials, right? Should you make that explicitly fail here? I see that `GCSCache::put` will early-exit if it's configured in read-only mode.

One thing I've been meaning to add support for is better configuration of storage backends, so that you could have an explicitly read-only storage. I want this to support two-tier caching, where we can try to read from bucket A, and then fall back to reading from bucket B, and store new objects in bucket B.

src/cache/cache.rs
```diff
@@ -230,23 +230,48 @@ pub fn storage_from_environment(pool: &CpuPool, _handle: &Handle) -> Arc<Storage
...
+                    _ => {
+                        warn!("No SCCACHE_RW_MODE specified-- defaulting to READ_ONLY.");
+                        RWMode::ReadOnly
+                    }
+                };
+
                 trace!("Using GCSCache");
-                let gcs_cred_provider = GCSCredentialProvider::new(false, key_path);
-                return Arc::new(GCSCache::new(bucket, base_url, gcs_cred_provider, _handle));
+                let gcs_cred_provider =
```
luser: You should probably have some sort of error here if we're configured in `RWMode::ReadWrite` but don't have credentials, since that isn't going to work right.

README.md
```diff
@@ -3,7 +3,7 @@
 sccache - Shared Compilation Cache
 ==================================
 
-Sccache is a [ccache](https://ccache.samba.org/)-like tool. It is used as a compiler wrapper and avoids compilation when possible, storing a cache in a remote storage using the S3 API.
+Sccache is a [ccache](https://ccache.samba.org/)-like tool. It is used as a compiler wrapper and avoids compilation when possible, storing a cache in a remote storage using the S3 API, the GCS API, or Redis.
```
luser: I know we don't spell out S3 here, but could you expand GCS? I didn't know what it was at first glance. It'd be fine to explicitly say "Amazon S3" as well. Making those names into links to their respective home pages would be a good addition at this point.

README.md
```diff
@@ -51,6 +51,10 @@ If you want to use S3 storage for the sccache cache, you need to set the `SCCACH
 
 Set `SCCACHE_REDIS` to a [Redis](https://redis.io/) url in format `redis://[:<passwd>@]<hostname>[:port][/<db>]` to store the cache in a Redis instance.
 
+To use [GCS](https://cloud.google.com/storage/), you need to set the `SCCACHE_GCS_BUCKET` environment variable to the name of the GCS bucket.
+If you're using authentication, set `SCCACHE_GCS_KEY_PATH` to the location of your JSON service account credentials.
+By default, SCCACHE on GCS will be read-only. To change this, set `SCCACHE_GCS_RW_MODE` to either `READ_ONLY` or `READ_WRITE`.
```
luser: We don't default to read-only on any other backend, that seems like an odd decision.

luser:
Sorry for the delay on reviewing this, it was just big enough that I needed to find time to sit down and read through it.

Cargo.toml
```diff
@@ -28,11 +28,13 @@ futures = "0.1.11"
 futures-cpupool = "0.1"
 hyper = { git = "https://github.com/hyperium/hyper", optional = true }
 hyper-tls = { git = "https://github.com/hyperium/hyper-tls", optional = true }
+jsonwebtoken = { version = "2.0", optional = true }
 libc = "0.2.10"
 local-encoding = "0.2.0"
 log = "0.3.6"
 lru-disk-cache = { path = "lru-disk-cache" }
 number_prefix = "0.2.5"
+openssl = { version = "0.9", optional = true }
```
cramertj: I didn't know about that crate-- sounds perfect!

cramertj:
No worries! Thanks for the feedback.

src/cache/cache.rs
cramertj: You can actually specify Read/Write permissions for the allusers group in GCS.

README.md
cramertj: Our vision for this is to populate the cache using bots, which are given a key.  Anyone else should be able to just point sccache at the public bucket and use it as a read-only cache without having to use any credentials.

However, if you think that's not going to be the common case, I'd be happy to discuss how to accommodate other scenarios. One option would be to default to READ_WRITE if credentials are provided, and READ_ONLY otherwise. What do you think?

README.md
cramertj: +1

src/cache/gcs.rs
cramertj: Put requests _can_ succeed without credentials if the system has been configured properly (the allusers group has been given read/write permissions).

src/cache/cache.rs
cramertj: Agreed-- that was something that frustrated me when debugging.

Cargo.toml
cramertj: The last commit was two years ago, and there's a note on the repo: "Disclaimer: this is a personal project that may eat your laundry." Converting from PEM to DER seems like a pretty stable operation, but the inactivity of the repo makes me wonder if this is the best idea. What do you think?

src/cache/gcs.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn auth_request_jwt(&self, expire_at: &chrono::DateTime<chrono::UTC>) -> Result<String> {
+        let metadata = fs::metadata(&self.credentials_path).chain_err(|| {
+            "Couldn't stat GCS credentials file"
+        })?;
+        if !metadata.is_file() {
+            bail!("Couldn't open GCS credentials file.");
+        }
+        let mut file = File::open(&self.credentials_path)?;
+        let mut service_account_json = String::new();
+        file.read_to_string(&mut service_account_json)?;
```
cramertj: SGTM-- I only did it this way because I was trying to match the behavior for S3, which reads in the files during the event loop, just before it goes to use them. See [here](https://github.com/mozilla/sccache/blob/master/src/simples3/credential.rs#L187).

src/cache/gcs.rs
```diff
@@ -0,0 +1,353 @@
...
+            cached_credentials: RefCell::new(None),
+        }
+    }
+
+    fn auth_request_jwt(&self, expire_at: &chrono::DateTime<chrono::UTC>) -> Result<String> {
+        let metadata = fs::metadata(&self.credentials_path).chain_err(|| {
+            "Couldn't stat GCS credentials file"
+        })?;
+        if !metadata.is_file() {
+            bail!("Couldn't open GCS credentials file.");
```
cramertj: I was trying to match the S3 behavior [here](https://github.com/mozilla/sccache/blob/master/src/simples3/credential.rs#L197). If you don't care about them matching, I'm happy to change it.

src/cache/gcs.rs
```diff
@@ -0,0 +1,353 @@
...
+                        trace!("PUT failed with error: {:?}", e);
+                        Err(e.into())
+                    }
+                }
+            })
+        }))
+    }
+}
+
+pub struct GCSCredentialProvider {
```
cramertj: Done.

src/cache/gcs.rs
```diff
@@ -0,0 +1,353 @@
...
+
+                let mut request = Request::new(Method::Post, url.parse().unwrap());
+                {
+                    let mut headers = request.headers_mut();
+                    headers.set(ContentType("application/x-www-form-urlencoded".parse().unwrap()));
+                    headers.set(ContentLength(params.len() as u64));
+                }
+                request.set_body(params);
+
+                client.request(request).map_err(Into::into)
```
cramertj: Unfortunately, you have to manually map the error because otherwise the error types don't match-- `IntoFuture` has error as an associated type, so a given future can only implement `IntoFuture` for one error type. `hyper::FutureResponse` has its error type as `hyper::error::Error`, so we need to coerce it to our manual error type before we can return it.

src/cache/gcs.rs
cramertj: Added a comment.

src/cache/cache.rs
cramertj: Done.

src/cache/cache.rs
cramertj: Okay. I changed it to `gcs_read_write_mode`. Not perfect, but it's something.

cramertj:
@luser WRT IAM roles-- Google does offer something similar, and it might indeed be good to support it, but I'd personally prefer to defer that to a later PR if possible.

Cargo.toml
cramertj: I'm adding a commit that removes openssl and uses pem-parser instead, although I noticed when doing so that it doesn't return an error if the PEM is invalid base64-- it `unwrap()`s it. Probably not a big deal-- I think it's unlikely that people will supply files that are the appropriate format but don't have a valid PEM.

cramertj: Also, when removing openssl I discovered that it's actually already a dependency-- sccache depends on hyper-tls, which depends on native-tls, which depends on openssl (tokio-tls also depends on openssl).

cramertj:
Rebased.

Cargo.toml
luser: FWIW, native-tls only depends on OpenSSL on Linux. It uses system libraries on Windows/OSX. We could probably figure out a way to use rustls (which uses ring) instead, but I'm not overly worried about it.

luser: That's unfortunate that pem-parser doesn't handle invalid files properly. If you feel sufficiently motivated you could submit a PR to that project to fix it. :) I understand if you want to skip that, I've been on the other end of this process and it can feel like you wind up fixing every part of the stack just to fix the one thing you want.

luser:
> @luser WRT IAM roles-- Google does offer something similar, and it might indeed be good to support it, but I'd personally prefer to defer that to a later PR if possible.

That's totally fine. Open a new issue about it so we don't forget?

cramertj:
@luser Done. See #144.

cramertj:
@luser Are there other changes you'd like to see on this?

luser:
> Are there other changes you'd like to see on this?

I think this is good, let me just give it another quick once-over.

luser:
Thanks for the patch!

PR #138, opened 2017-06-07
added missing args with values on clang osx

andreparodi:
when compiling on osx I had to add these two extra values to the missing args with values list. should be harmless I think.

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -83,7 +83,8 @@ pub const ARGS_WITH_VALUE: &'static [&'static str] = &[
     "-iframework", "-imacros", "-imultilib", "-include",
     "-install_name", "-iprefix", "-iquote", "-isysroot",
     "-isystem", "-iwithprefix", "-iwithprefixbefore",
-    "-u", "-x", "-arch",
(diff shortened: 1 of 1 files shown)


src/compiler/gcc.rs
```diff
@@ -83,7 +83,8 @@ pub const ARGS_WITH_VALUE: &'static [&'static str] = &[
     "-iframework", "-imacros", "-imultilib", "-include",
     "-install_name", "-iprefix", "-iquote", "-isysroot",
     "-isystem", "-iwithprefix", "-iwithprefixbefore",
-    "-u", "-x", "-arch",
+    "-u", "-x", "-arch", "--serialize-diagnostics",
+    "--sysroot"
```
luser: `--sysroot` is supported by gcc, but `--serialize-diagnostics` seems to be clang-specific. Can you move that to clang.rs instead?

andreparodi: looks like --serialize-diagnostics is already in clang.rs

luser:
I squashed these commits and landed them: https://github.com/mozilla/sccache/commit/132ef5d048be5e4d6df81c2dfd2fe20c19e48123

Thanks!

PR #139, opened 2017-06-07
Additional s3 documentation

andreparodi:
minor addition for undocumented s3 env variables needed to use a private s3 compatible object store like minio.


andreparodi:
Accidentally included the changes from my other PR. I have reverted them in my latest commit so this PR should just be a couple of extra lines of documentation.

luser:
Rebased and landed as https://github.com/mozilla/sccache/commit/c59615004500199c325f758b8eeae6198bbcc2a0

Thanks for the patch! At some point I think I'd like to break the documentation for each storage backend out to a separate .md file, but documenting undocumented things is always an improvement.

PR #143, opened 2017-06-14
Fix build with --no-default-features --features=redis

aperezdc:
The `serde_json` crate is always needed by `src/commands.rs` even when building only with support for Redis, but `serde_json` would be left out when the S3 feature is disabled. This make it possible to build `sccache` only with Redis support.

Building with S3 disabled trims down a stripped release build from 5.1 MiB down to 4.4 MiB on x86_64.


luser:
Thanks for the patch!

PR #145, opened 2017-06-28
Fix compilation with GCC to properly handle preprocessor arguments

luser:
... when re-compiling when the first compile fails with -Werror.

This patch isn't complete, I just don't have GCC handy locally and want to verify that the new test I wrote currently fails.


luser:
OK, here's a Travis run from my previous push with the added test but without the fix, showing that the test was failing with the existing code: https://travis-ci.org/mozilla/sccache/jobs/248492992

The tests should pass now with the latest push, which contains the fixes.

src/compiler/gcc.rs
```diff
@@ -169,6 +169,7 @@ fn _parse_arguments(arguments: &[OsString],
                 v if v.starts_with('@') => return CompilerArguments::CannotCache("@file"),
                 "-M" |
                 "-MM" |
+                "-MP" |
```
glandium: You should mention in the commit message that this not being recognized as a preprocessor flag made gcc error out without preprocessor flags because gcc wants other -M flags alongside.

src/compiler/gcc.rs
```diff
@@ -331,6 +332,7 @@ pub fn compile<T>(creator: &T,
     cmd.arg("-c")
         .arg(&parsed_args.input)
         .arg("-o").arg(&out_file)
+        .args(&parsed_args.preprocessor_args)
         .args(&parsed_args.common_args)
```
glandium: I wonder if it would make sense to just use the original arguments. IIRC that's what the python implementation was doing.

FWIW, I dug in the code history, and this problem has existed ever since the rust implementation has had the "retry on error building the preprocessed output" code path for clang (092417b4)

src/test/system.rs
```diff
@@ -179,6 +183,16 @@ fn run_sccache_command_test(sccache: &Path, compiler: Compiler, tempdir: &Path)
         let expected_lines: Vec<_> = expected.lines().collect();
         assert_eq!(lines, expected_lines);
     }
+    if name == "gcc" {
```
glandium: You /could/ test with clang too.

src/test/system.rs
```diff
@@ -179,6 +183,16 @@ fn run_sccache_command_test(sccache: &Path, compiler: Compiler, tempdir: &Path)
...
+        trace!("test -MP with -Werror");
```
glandium: I think it would be better to do something less relying on multiple things going wrong at the same time, which this test does, and instead, test the actual underlying problem.

So, something like, put a warning producing code in a #ifdef FOO, and pass -Wall -Werror -DFOO=1 on the command line.

staktrace:
Can we please get this landed asap? No compiler error messages in build failures is surprisingly detrimental to my productivity.

luser:
Sorry, being at an all hands was not super conducive to actually working, and I dropped the ball on this this week.

src/test/system.rs
luser: I mean, the whole point here is simply that we fail the first compilation, so we retry, and then we muck up the commandline for the second compilation. It's sort of hard to make a non-contrived test case here.

src/test/system.rs
luser: clang doesn't actually fail in this situation though, so it would be harder to write a useful test for it.

src/compiler/gcc.rs
luser: Not sure if this was a bug I inherited from the Python version or one I invented all on my own. We probably should just use the original arguments. I filed https://github.com/mozilla/sccache/issues/147 on doing that.

PR #149, opened 2017-07-12
Miscellaneous changes

glandium:


src/compiler/gcc.rs
```diff
@@ -306,6 +306,7 @@ pub fn compile<T>(creator: &T,
         let extension = match extension.as_ref() {
             "c" => "cpp-output".to_owned(),
             "cc" | "cpp" | "cxx" => "c++-cpp-output".to_owned(),
+            "m" | "mm" => "objc-cpp-output".to_owned(),
```
luser: I was thinking at some point that it'd be nice to have an enum for C compilation modes instead of matching on strings here.

PR #155, opened 2017-07-20
Allow to override the default idle timeout with an environment variable

glandium:
While the default is reasonable for developers, CI environment may want
to choose larger timeouts (or virtually infinite ones).

Note that "virtually infinite", in practice, may not really mean infinite: I'm not sure what rayon does under the hood, but with u64::MAX, it does setup epoll_wait for i32::MAX milliseconds, which is, in fact, 24 days and some.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -49,6 +49,7 @@ use std::process::{Output, ExitStatus};
 use std::rc::Rc;
 use std::sync::Arc;
 use std::time::Duration;
+use std::u64;
 use tokio_core::net::TcpListener;
 use tokio_core::reactor::{Handle, Core, Timeout};
 use tokio_io::codec::length_delimited::Framed;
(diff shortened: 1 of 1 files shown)


glandium:
FWIW, we have the problem on Firefox CI, on PGO builds, where linking xul.dll takes more than 20 minutes, during which sccache shuts down.

na-g:
@glandium, when the timeout is 0 is the intent that it will immediately go into the shutdown state, and if so will it accept new work (extending the shutdown duration)?

glandium:
That's a good question... OTOH, considering it's milliseconds, any really small value will have weird useless effects... so 0 is not entirely special there...

glandium:
Come to think of it, we could also make 0 mean u64::MAX...

glandium:
So interesting, while on Linux, this just works with u64::MAX, on Windows, this leads to:
```
thread 'main' panicked at 'overflow when adding duration to time', src\libcore\option.rs:794
```

luser:
Sorry for letting this sit!

PR #164, opened 2017-07-28
Support -B argument for GCC

### LINKED ISSUE #126: Add -B to gcc::ARGS_WITH_VALUE
I did this for clang recently but I missed the fact that gcc takes it as well. I saw this show up in sccache logs from a servo build. It might not actually help anything, since they seem to be all using gcc for final linking, but we should do this anyway for correctness.

aperezdc:
Fixes #126 

Can it really be that this patch can be this easy? :thinking:


luser:
I think so, yeah! Thanks!

PR #170, opened 2017-08-09
Refactored argument parsing

glandium:
See the descriptive commit messages. I verified that this fixes the issues with openvr on OSX with the new SDK (with @luser's patch from bug 1324892).

Unfortunately, this doesn't fix the large number of warnings, and it looks like the only solution for those is to build from the actual source instead of from the preprocessed output :(

This actually makes the refactor largely unnecessary, because fixing that will make distinguishing between preprocessor and other args much less useful. I do think it makes things clearer, though, so I didn't scrape it all.

### DIFF
--- a/src/compiler/args.rs
+++ b/src/compiler/args.rs
@@ -1,0 +1,915 @@
+use std::cmp::Ordering;
+use std::ffi::OsString;
+use std::path::PathBuf;
+
+pub type Delimiter = Option<u8>;
+
+/// Representation of a parsed argument
+#[derive(PartialEq, Clone, Debug)]
+pub enum Argument {
+    /// Unknown non-flag argument ; e.g. "foo"
+    Raw(OsString),
+    /// Unknown flag argument ; e.g. "-foo"
+    UnknownFlag(OsString),
+    /// Known flag argument ; e.g. "-bar"
+    Flag(&'static str),
+    /// Known argument with a value ; e.g. "-qux bar", where the way the
+    /// value is passed is described by the ArgDisposition type.
+    WithValue(&'static str, ArgumentValue, ArgDisposition),
+}
+
+/// How a value is passed to an argument with a value.
+#[derive(PartialEq, Clone, Debug)]
+pub enum ArgDisposition {
+    /// As "-arg value"
+    Separated,
+    /// As "-arg value", but "-arg<delimiter>value" would be valid too
+    CanBeConcatenated(Delimiter),
+    /// As "-arg<delimiter>value", but "-arg value" would be valid too
+    CanBeSeparated(Delimiter),
+    /// As "-arg<delimiter>value"
+    Concatenated(Delimiter),
+}
+
+pub enum NormalizedDisposition {
+    Separated,
+    Concatenated,
+}
+
+impl Argument {
+    /// For arguments that allow both a concatenated or separated disposition,
+    /// normalize a parsed argument to a prefered disposition.
+    pub fn normalize(self, disposition: NormalizedDisposition) -> Self {
+        match self {
+            Argument::WithValue(s, v, ArgDisposition::CanBeConcatenated(d)) |
+            Argument::WithValue(s, v, ArgDisposition::CanBeSeparated(d)) => {
+                Argument::WithValue(
+                    s,
+                    v,
+                    match disposition {
+                        NormalizedDisposition::Separated => ArgDisposition::Separated,
+                        NormalizedDisposition::Concatenated => ArgDisposition::Concatenated(d),
+                    },
+                )
+            }
+            a => a,
+        }
+    }
+
+    pub fn to_os_string(&self) -> OsString {
+        match *self {
+            Argument::Raw(ref s) |
+            Argument::UnknownFlag(ref s) => s.clone(),
+            Argument::Flag(ref s) |
+            Argument::WithValue(ref s, _, _) => s.into(),
+        }
+    }
+
+    pub fn to_str(&self) -> Option<&'static str> {
+        match *self {
+            Argument::Flag(s) |
+            Argument::WithValue(s, _, _) => Some(s),
+            _ => None,
+        }
+    }
+
+    pub fn get_value(&self) -> Option<ArgumentValue> {
+        match *self {
+            Argument::WithValue(_, ref v, _) => Some(v.clone()),
+            _ => None,
+        }
+    }
+}
+
+pub struct IntoIter {
+    arg: Argument,
+    emitted: usize,
+}
+
+/// Transforms a parsed argument into an iterator.
+impl IntoIterator for Argument {
+    type Item = OsString;
+    type IntoIter = IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        IntoIter {
+            arg: self,
+            emitted: 0,
+        }
+    }
+}
+
+impl Iterator for IntoIter {
+    type Item = OsString;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let result = match self.arg {
+            Argument::Raw(ref s) |
+            Argument::UnknownFlag(ref s) => {
+                match self.emitted {
+                    0 => Some(s.clone()),
+                    _ => None,
+                }
+            }
+            Argument::Flag(s) => {
+                match self.emitted {
+                    0 => Some(s.into()),
+                    _ => None,
+                }
+            }
+            Argument::WithValue(s, ref v, ref d) => {
+                match (self.emitted, d) {
+                    (0, &ArgDisposition::CanBeSeparated(d)) |
+                    (0, &ArgDisposition::Concatenated(d)) => {
+                        let mut s = OsString::from(s);
+                        if let Some(d) = d {
+                            s.push(OsString::from(String::from_utf8(vec![d]).expect(
+                                "delimiter should be ascii",
+                            )));
+                        }
+                        s.push(OsString::from(v.clone()));
+                        Some(s)
+                    }
+                    (0, &ArgDisposition::Separated) |
+                    (0, &ArgDisposition::CanBeConcatenated(_)) => Some(s.into()),
+                    (1, &ArgDisposition::Separated) |
+                    (1, &ArgDisposition::CanBeConcatenated(_)) => Some(v.clone().into()),
+                    _ => None,
+                }
+            }
+        };
+        if let Some(_) = result {
+            self.emitted += 1;
+        }
+        result
+    }
+}
+
+/// The value associated with a parsed argument
+#[derive(PartialEq, Clone, Debug)]
+pub enum ArgumentValue {
+    String(OsString),
+    PathVal(PathBuf),
+}
+
+impl From<ArgumentValue> for OsString {
+    fn from(a: ArgumentValue) -> OsString {
+        match a {
+            ArgumentValue::String(s) => s,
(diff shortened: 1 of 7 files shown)


glandium:
A try build with the patch from bug 1324892 and using this version of sccache:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=99994c55212238ed5a49753a35c465ce7ba31a16

luser:
> Unfortunately, this doesn't fix the large number of warnings, and it looks like the only solution for those is to build from the actual source instead of from the preprocessed output :(

That's a bummer. :-(

src/compiler/args.rs
```diff
@@ -0,0 +1,820 @@
...
+    pub fn to_os_string(&self) -> OsString {
+        match *self {
+            Argument::Raw(ref s) |
+            Argument::UnknownFlag(ref s) => s.clone(),
+            Argument::Flag(ref s) |
+            Argument::WithValue(ref s, _, _) => s.into(),
+        }
+    }
+
+    pub fn to_str(&self) -> Option<&'static str> {
```
luser: It's probably more conventional to name this `as_str`, since it's not allocating anything.

glandium: OTOH, `as_str` doesn't usually return an Option (never in libstd, at least)

glandium: FWIW, as a data point `OsStr::to_str` doesn't allocate on unix (and returns an Option<&str>)

src/compiler/gcc.rs
```diff
@@ -75,142 +76,188 @@ impl CCompilerImpl for GCC {
...
-                "-MQ" => {
-                    preprocessor_args.push(arg.clone());
-                    if let Some(arg_val) = it.next() {
-                        preprocessor_args.push(arg_val);
+
+    for item in ArgsIter::new(it, arg_info) {
+        // Refuse to cache arguments such as "-include@foo" because they're a
+        // mess. See https://github.com/mozilla/sccache/issues/150#issuecomment-318586953
+        if let Argument::WithValue(_, ref v, CanBeSeparated(_)) = item.arg {
+            if OsString::from(v.clone()).to_string_lossy().starts_with("@") {
```
luser: You can use `OsStrEx::starts_with` here as well.

src/compiler/gcc.rs
```diff
@@ -75,142 +76,188 @@ impl CCompilerImpl for GCC {
...
+                return CompilerArguments::CannotCache("@");
+            }
+        }
+        match item.data {
```
luser: I like how this code worked out!

src/compiler/gcc.rs
```diff
@@ -75,142 +76,188 @@ impl CCompilerImpl for GCC {
...
+            Some(Output) => output_arg = item.arg.get_value().map(|s| s.unwrap_path()),
+            Some(NeedDepTarget) => need_explicit_dep_target = true,
+            Some(DepTarget) => dep_target = item.arg.get_value().map(OsString::from),
+            Some(PreprocessorArgument) |
+            Some(PassThrough) |
+            Some(Language) => {}
+            None => {
+                match item.arg {
+                    Argument::Raw(ref val) => {
+                        if input_arg.is_some() {
```
luser: Doesn't this regress your change to make linking invocations produce `NotCompilation`?

src/compiler/msvc.rs
```diff
@@ -208,88 +246,48 @@ pub fn parse_arguments(arguments: &[OsString]) -> CompilerArguments<ParsedArgume
...
-                    common_args.push("-FI".into());
-                    if let Some((arg_val, flag)) = it.next() {
-                        if flag.is_some() {
-                            return CompilerArguments::CannotCache("extra -FI arg")
-                        }
-                        common_args.push(arg_val);
+    for item in ArgsIter::new(it, &ARGS[..]) {
+        match item.data {
+            Some(TooHard) => {
+                return CompilerArguments::CannotCache(item.arg.to_str().expect(
```
luser: You could make `TooHard` contain a `&'static str` and just return that in the `CannotCache`, but it's not particularly pressing.

src/compiler/c.rs
```diff
@@ -77,6 +77,19 @@ impl ParsedArguments {
     }
 }
 
+pub fn normalized_extension(file: &Path) -> Option<&'static str> {
```
luser: I had actually mentioned this on a different PR a while ago, but it might be nicer to have an enum of C dialects to use here.

src/compiler/c.rs
```diff
@@ -276,6 +291,7 @@ pub fn hash_key(compiler_digest: &str,
     let mut m = Digest::new();
     m.update(compiler_digest.as_bytes());
     m.update(CACHE_VERSION);
+    m.update(extension.as_bytes());
```
luser: I guess you should rev `CACHE_VERSION` here, since you're changing the hash inputs.

src/compiler/gcc.rs
```diff
@@ -208,8 +209,19 @@ where
...
             Some(DepTarget) => dep_target = item.arg.get_value().map(OsString::from),
             Some(PreprocessorArgument) |
-            Some(PassThrough) |
-            Some(Language) => {}
+            Some(PassThrough) => {}
+            Some(Language) => {
+                // Set extension to one matching the language set with -x
+                let lang = item.arg.get_value().map(OsString::from);
+                let lang = lang.as_ref().map(|a| a.to_string_lossy());
+                extension = match lang.as_ref().map(|a| a.as_ref()) {
```
luser: Maybe this wants an `ArgumentValue::as_str` or something like that? I guess you'd have to return a `Cow<str>` because of the `to_string_lossy` in there.

src/compiler/gcc.rs
```diff
@@ -291,7 +299,7 @@ where
 
     CompilerArguments::Ok(ParsedArguments {
         input: input.into(),
-        extension: extension,
+        extension: extension.expect("extension should always be set"),
```
luser: You can drop this if you rewrite the bit above to be like:
```
let extension = match extension {
  None => return CompilerArguments::CannotCache("unknown source extension"),
  Some(e) => e,
};
```

I wrote a macro to do this for required options in the Rust compiler implementation.
```

**Comment by luser** [comment]
Oh heh, I forgot that the "review changes" button applies to the entire PR. Oh well. The second patch looks fine modulo those few little comments, but I'd really like to see the change to not use the enum variant names in the first patch.

File: src/compiler/args.rs [hunk 10]
```diff
@@ -0,0 +1,820 @@
...
+    /// Known flag argument ; e.g. "-bar"
+    Flag(&'static str),
+    /// Known argument with a value ; e.g. "-qux bar", where the way the
+    /// value is passed is described by the ArgDisposition<Delimiter> type.
+    WithValue(&'static str, ArgumentValue, ArgDisposition<Delimiter>),
+}
+
+/// How a value is passed to an argument with a value.
+#[derive(PartialEq, Clone, Debug)]
+pub enum ArgDisposition<T> {
```
glandium: It makes normalize() non awkward to call.

src/compiler/args.rs
```diff
@@ -0,0 +1,820 @@
...
+    Separated,
+    /// As "-arg value", but "-arg<delimiter>value" would be valid too
+    CanBeConcatenated(T),
+    /// As "-arg<delimiter>value", but "-arg value" would be valid too
+    CanBeSeparated(T),
+    /// As "-arg<delimiter>value"
+    Concatenated(T),
+}
+
+pub use self::ArgDisposition::*;
```
glandium: Is it? That's technically how you get Ok(), Err(), Some() and None for Result and Option...

src/compiler/args.rs
```diff
@@ -0,0 +1,820 @@
...
+            PathVal(p) => p.into(),
+        }
+    }
+}
+
+impl ArgumentValue {
+    pub fn unwrap_path(self) -> PathBuf {
+        match self {
+            PathVal(p) => p,
+            StringVal(_) => panic!("Can't unwrap_path an ArgumentValue::StringVal"),
```
glandium: The idea is that you only get a Path if the option was meant to be one.

src/compiler/args.rs
```diff
@@ -0,0 +1,820 @@
...
+/// Allow to search over a couple of arrays of ArgInfo, where the second
+/// complements or overrides the first one.
+impl<T> SearchableArgInfo for (&'static [T], &'static [T])
+where
+    T: 'static + EmbedsArgInfo,
+{
+    type Info = T;
+
+    fn search(&self, key: &str) -> Option<&Self::Info> {
+        match (self.0.search(key), self.1.search(key)) {
```
glandium: Yes, and there's no way around that.

src/compiler/args.rs
luser: I can't find a strong guideline about it anywhere, but [clippy has a lint](https://github.com/rust-lang-nursery/rust-clippy/wiki#enum_glob_use) for glob importing enum variants.

src/compiler/args.rs
luser: Ok, then it seems fine as it is.

src/compiler/args.rs
```diff
@@ -0,0 +1,820 @@
...
+                    Argument::WithValue(s, t.process(a), Separated)
+                } else {
+                    Argument::Flag(s)
+                }
+            }
+            TakeArg(s, t, Concatenated(d)) => {
+                let mut len = s.len();
+                debug_assert_eq!(&arg[..len], s);
+                if let Some(d) = d {
+                    debug_assert_eq!(arg.as_bytes()[len], d);
```
glandium: Except `arg` is a `&str` and `split_prefix` doesn't exist there.

src/compiler/gcc.rs
glandium: That still requires 3 lines thanks to the borrow checker.

src/compiler/gcc.rs
glandium: Doh, yes (why didn't this comment appear on the main PR page?)

src/compiler/gcc.rs
```diff
@@ -209,8 +210,18 @@ where
             Some(NeedDepTarget) => need_explicit_dep_target = true,
             Some(DepTarget) => dep_target = item.arg.get_value().map(OsString::from),
             Some(PreprocessorArgument) |
-            Some(PassThrough) |
-            Some(Language) => {}
+            Some(PassThrough) => {}
+            Some(Language) => {
+                let lang = item.arg.get_value().map(OsString::from);
+                let lang = lang.as_ref().map(|a| a.to_string_lossy());
+                language = match lang.as_ref().map(|a| a.as_ref()) {
```
luser: That's a heck of a pile of `as_ref`s.

src/compiler/gcc.rs
```diff
@@ -311,14 +322,22 @@ pub fn preprocess<T>(creator: &T,
     where T: CommandCreatorSync
 {
     trace!("preprocess");
+    let language = match parsed_args.language {
+        Language::C => "c",
+        Language::Cxx => "c++",
+        Language::ObjectiveC => "objective-c",
+        Language::ObjectiveCxx => "objective-c++",
```
luser: Sort of a bummer that there are so many ways to spell these that you can't just have a common method on `Language`.

luser:
Thanks for the nice refactor!

PR #172, opened 2017-08-11
Add a Memcached cache variant

barrbrain:
On top of the original python implementation of sccache, I implemented a [minimal memcache backend](https://github.com/barrbrain/sccache/commit/a370c755). It has quietly served a billion or so objects over the past 2 years and saved thousands of engineering hours. This is a simple attempt at bringing that to Rust.
I'm hoping for an early review to help get this to a state suitable for merging. I started learning Rust just 1 day before the first draft of this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,6 +26,7 @@ dependencies = [
  "local-encoding 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "lru-disk-cache 0.1.0",
+ "memcached-rs 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "mio-named-pipes 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "native-tls 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "number_prefix 0.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


src/cache/memcache.rs
```diff
@@ -0,0 +1,95 @@
...
+            .map_err(|e| e.into())
+    }
+}
+
+impl Storage for MemCache {
+    fn get(&self, key: &str) -> SFuture<Cache> {
+        let key = key.to_owned();
+        let me = self.clone();
+        self.pool.spawn_fn(move || {
+            let mut c = me.connect().unwrap();
```
barrbrain: A major shortcoming of this first draft is that a new Client is created for every request. This has the side-effect that a new TCP connection is also opened.

flxo: I had the same thoughts with the Redis cache and [tried](https://github.com/flxo/sccache/commit/25118045c477fe76bfdcd9b0f6649644df543e64) out  [r2d2](https://github.com/sfackler/r2d2). 

We have had problems with the thousands of opened/closed connections when running sccache within docker. It's not yet clear in detail why - still investigating. That's why I added the pooling in our internal test branch.

barrbrain:
@luser I think I'm done iterating on this now. Implementation is concise; latency is minimal; errors are contained. Misconfiguration may possibly be handled more gracefully.

luser:
Sorry for letting this sit for so long! I rebased this patch on top of master and landed it: https://github.com/mozilla/sccache/commit/d2c7f925adf0b105a9b87200129d57cf39439ada

Thanks!

PR #173, opened 2017-08-15
Add a note about using sccache with Jenkins CI

flxo:
I'd like to share a note about `sccache` in a Jenkins environment. We had sporadic fails in our CI until we spawned a dedicated `sccache` servers for each job. Using virtualisation might also have helped.
Native speakers feedback is welcome!

### DIFF
--- a/docs/Jenkins.md
+++ b/docs/Jenkins.md
@@ -1,0 +1,20 @@
+sccache on Jenkins
+==================
+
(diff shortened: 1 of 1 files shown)


luser:
Useful info, thanks! Do you want to link this from the README? We should probably have a more coherent docs story at some point.

flxo:
Sure - I opened #177 
Thanks!

PR #178, opened 2017-08-17
Never compile from the preprocessed output

glandium:
Testing this on try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d97d1ade23fd363042312e63fc5d456034fa0fa9

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -492,6 +492,8 @@ mod test {
 
     use super::*;
     use ::compiler::*;
+    use mock_command::*;
+    use test::utils::*;
     use tempdir::TempDir;
 
     fn _parse_arguments(arguments: &[String]) -> CompilerArguments<ParsedArguments> {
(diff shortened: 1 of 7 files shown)


glandium:
New try with the updated cache version.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=92a45f408db07d4a25b9273ef22c0824abb0ee8c

src/compiler/gcc.rs
```diff
@@ -364,62 +362,26 @@ pub fn compile<T>(creator: &T,
...
-        Language::C => "cpp-output",
-        Language::Cxx => "c++-cpp-output",
-        Language::ObjectiveC => "objective-c-cpp-output",
-        Language::ObjectiveCxx => "objective-c++-cpp-output",
+        Language::C => "c",
+        Language::Cxx => "c++",
+        Language::ObjectiveC => "objective-c",
+        Language::ObjectiveCxx => "objective-c++",
     };
     let mut attempt = creator.clone().new_command_sync(executable);
```
luser: nit: we could probably rename this something other than `attempt` now that it's the only command.

PR #180, opened 2017-09-01
Add Windriver Diab compiler support

flxo:
This patch adds support for using `sccache` in combination with the `Windriver Diab compiler` (dcc). The implementation is heavily borrowed from gcc/clang and tested on `Linux` and `Windows` with diab `5.8.0.0` and `5.9.4.8`. I do not have access to other versions. Diab is deteted by `sccache` via the `__DCC__` define. The command line parsing is quite the same as for gcc/clang except the weird `@` argument which also accepts environment arguments and if that one doesn't exists diab looks for a file with the name - probably nobody outside `Windriver` ever used this feature ;-).

A full reference of `diab` is [here.](http://www.embedded-tools.de/sites/default/files/pdf/wr_diab_compiler_options_reference_5.9.3-R2.pdf)

We (have to) use `diab` for various automotive embedded projects. The `diab` support in `sccache` runs for a year now stable in our CI and on some developer workstations. I rebased it from time to time on the mozilla master.

@glandium: Thanks for the refactoring of the command line parsing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ or use it with rust, like so:
 
 > $ RUSTC_WRAPPER=[path to sccache] cargo build
 
-Sccache (tries to) support gcc, clang and MSVC. If you don't [specify otherwise](#storage-options), sccache will use a local disk cache.
+Sccache (tries to) support gcc, clang, [diab](https://www.windriver.com/products/development-tools/#diab_compiler) and MSVC. If you don't [specify otherwise](#storage-options), sccache will use a local disk cache.
 
 You can run `sccache --start-server` to start the background server process without performing any compilation.
 
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -15,6 +15,7 @@
 use cache::{Cache, CacheWrite, Storage};
 use compiler::c::{CCompiler, CCompilerKind};
 use compiler::clang::Clang;
+use compiler::diab::Diab;
 use compiler::gcc::GCC;
 use compiler::msvc;
 use compiler::msvc::MSVC;
@@ -883,6 +884,8 @@ msvc
 clang
 #elif defined(__GNUC__)
 gcc
+#elif defined(__DCC__)
+diab
 #endif
 ".to_vec();
     let write = write_temp_file(&pool, "testfile.c".as_ref(), test);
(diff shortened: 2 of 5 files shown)


flxo:
@luser I'd like to ask if there is a chance to get this one merged? It's shouldn't have any impact on existing stuff and I'm getting tired of maintaining my fork...

flxo:
Refactored and rebased on the latest arg parsing changes.

@luser Would be great if you can have a look. If it's impossible to get it merged I can stop updating this PR. Thanks!

src/compiler/compiler.rs
```diff
@@ -742,6 +745,10 @@ gcc
                 debug!("Found clang");
                 return Box::new(CCompiler::new(Clang, executable, &pool)
                                 .map(|c| Some(Box::new(c) as Box<Compiler<T>>)));
+            } else if line == "diab" {
```
luser: This if/else if chain probably wants to be a match, eh? (I must have written this while I was still fairly new to Rust!)

src/compiler/diab.rs
```diff
@@ -0,0 +1,694 @@
...
+use std::fs::File;
+use std::io::Read;
+use std::path::{Path, PathBuf};
+use std::process;
+use util::{run_input_output, OsStrExt};
+
+#[derive(Clone, Debug)]
+pub struct Diab;
+
+impl CCompilerImpl for Diab {
```
luser: I wonder if we shouldn't ought to just write a macro that generates these `impl CCompilerImpl` bits, since they all look mostly the same and they just call out to module-local functions? (Room for improvement in the future.)

luser:
Hey! Sorry about this, I lost track of it and never got back around to it.

src/compiler/diab.rs
```diff
@@ -0,0 +1,694 @@
...
+            // Return options that produce additional output and are not cacheable
+            if arg.starts_with("E=") || arg.starts_with("O=") || arg.starts_with("@") {
+                return Some(arg);
+            }
+
+            // According to diab [1], @file means:
+            // Read command line options from either a file or an environment
+            // variable. When -@name is encountered on the command line, the
+            // driver first looks for an environment variable with the given
+            // name and substitutes its value. If an environment variable is
```
luser: Wow, I thought gcc/msvc's behavior was difficult to deal with, but reading commandline options from an *environment variable* is something else!

src/compiler/diab.rs
```diff
@@ -0,0 +1,694 @@
...
+            // not found then the driver tries to open a file with given name
+            // and substitutes the contents of the file. If neither an
+            // environment variable or a file can be found, an error message
+            // is issued and the driver terminates.
+            //
+            // [1]: http://www.vxdev.com/docs/vx55man/diab5.0ppc/c-invoke.htm#3000619
+
+            if let Ok(env_arg) = env::var(&arg) {
```
luser: This isn't going to do the right thing--this will read the environment from the sccache server process. If you do:
```
sccache --start-server
FOO="-c foo.c -o foo.o" sccache diab -@FOO
```
The server process won't have the `FOO` environment variable. The server does have access to the client's environment variables at the time it calls `parse_arguments`:
https://github.com/mozilla/sccache/blob/3d9f934ae94759e6aef4b3c02bf0779df4fe2cc6/src/server.rs#L565

So you could change the signature of `parse_arguments` to pass them through, and change all the implementations to match.

luser: I don't know if it's workable for you, but an alternate tack would be to simply return `CannotCache` for `-@FOO` where "FOO" is not a file.

README.md
```diff
@@ -102,7 +102,7 @@ or use it with rust, like so:
 
 > $ RUSTC_WRAPPER=[path to sccache] cargo build
 
-Sccache (tries to) support gcc, clang and MSVC. If you don't [specify otherwise](#storage-options), sccache will use a local disk cache.
+Sccache (tries to) support gcc, clang, diab and MSVC. If you don't [specify otherwise](#storage-options), sccache will use a local disk cache.
```
luser: I think making diab a link to the diab compiler site would be useful, since it's not as well-known as the others. (I personally had never heard of it.)

src/compiler/compiler.rs
flxo: Yes. I just tend to make patches as less invasive as possible for clean diffs in a PR. But in this case...anyway I'll change that.

src/compiler/diab.rs
flxo: I really doubt that anybody uses that feature. I will check how `ccache` handles that case (yes they support `diab`).

src/compiler/diab.rs
flxo: Thanks for pointing the fail. At our place nobody uses the env var feature so I didn't get any reports for that. The tests are fine but I should have written a integration test to cover it completely.

Adding the environment to `Compiler::parse_arguments` would obviously affect *all* compiler implementations and add something unused there.

As commented before I've never seen anybody using this feature. When thinking about it I imagine totally confused colleagues who wonder why some piece of software behaves different depending on the machine built.

I'll follow your `CannotCache` suggestion. If anybody really needs this we can add it later.

README.md
flxo: Lucky you!

I'll add a link...

flxo:
@luser Thanks for the review. I'll incorporate the changes and update the PR.

src/compiler/diab.rs
flxo: I couldn't find anybody here using that feature. Furthermore checked `ccache` which also doesn't support it. 
I remove the corresponding code and flag the compilation as non cacheable.

flxo:
@luser I rebased and updated the patch according to your suggestions. Let me know if there is something more I can do. Thanks!

drahnr:
@luser besides rebasing, is there anything that stops this from being merged?

drahnr:
Is there anything preventing this from being merged besides a rebase? If so, could you point me in the right direction so this can be merged :)

flxo:
@drahnr Thanks for the ping about the needed rebase. Done.

src/compiler/compiler.rs
flxo: I replaced the `if` `else` stuff with a match.

src/compiler/diab.rs
flxo: As described above this case is now treated as a `CannotCache`.

src/compiler/diab.rs
flxo: Done.

README.md
flxo: I added link.

flxo:
Hi @luser,

anything left to do?

Thanks!

luser:
Sorry to have let this drag on so long, thanks!

flxo:
@luser Thanks! I'll monitor this for issues, but please feel free to ping me for anything related.

PR #185, opened 2017-09-27
Add jobserver support to sccache

alexcrichton:
This commit alters the main sccache server to operate and orchestrate its own
GNU make style jobserver. This is primarily intended for interoperation with
rustc itself.

The Rust compiler currently has a multithreaded mode where it will execute code
generation and optimization on the LLVM side of things in parallel. This
parallelism, however, can overload a machine quickly if not properly accounted
for (e.g. if 10 rustcs all spawn 10 threads...). The usage of a GNU make style
jobserver is intended to arbitrate and rate limit all these rustc instances to
ensure that one build's maximal parallelism never exceeds a particular amount.

Currently for Rust Cargo is the primary driver for setting up a jobserver. Cargo
will create this and manage this per compilation, ensuring that any one `cargo
build` invocation never exceeds a maximal parallelism. When sccache enters the
picture, however, the story gets slightly more odd.

The jobserver implementation on Unix relies on inheritance of file descriptors
in spawned processes. With sccache, however, there's no inheritance as the
actual rustc invocation is spawned by the server, not the client. In this case
the env vars used to configure the jobsever are usually incorrect.

To handle this problem this commit bakes a jobserver directly into sccache
itself. The jobserver then overrides whatever jobserver the client has
configured in its own env vars to ensure correct operation. The settings of each
jobserver may be misconfigured (there's no way to configure sccache's jobserver
right now), but hopefully that's not too much of a problem for the forseeable
future.

The implementation here was to provide a thin wrapper around the `jobserver`
crate with a futures-based interface. This interface was then hooked into the
mock command infrastructure to automatically acquire a jobserver token when
spawning a process and automatically drop the token when the process exits.
Additionally, all spawned processes will now automatically receive a configured
jobserver.

cc rust-lang/rust#42867, the original motivation for this commit

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -31,6 +31,7 @@ use futures::sync::mpsc;
 use futures::task::{self, Task};
 use futures::{Stream, Sink, Async, AsyncSink, Poll, StartSend, Future};
 use futures_cpupool::CpuPool;
+use jobserver::Client;
 use mock_command::{
     CommandCreatorSync,
     ProcessCommandCreator,
@@ -121,10 +122,11 @@ fn get_signal(_status: ExitStatus) -> i32 {
 /// requests a shutdown.
 pub fn start_server(port: u16) -> Result<()> {
     trace!("start_server");
+    let client = unsafe { Client::new() };
     let core = Core::new()?;
     let pool = CpuPool::new(20);
     let storage = storage_from_environment(&pool, &core.handle());
-    let res = SccacheServer::<ProcessCommandCreator>::new(port, pool, core, storage);
+    let res = SccacheServer::<ProcessCommandCreator>::new(port, pool, core, client, storage);
     let notify = env::var_os("SCCACHE_STARTUP_NOTIFY");
     match res {
         Ok(srv) => {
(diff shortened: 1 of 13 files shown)


src/server.rs
```diff
@@ -389,6 +397,7 @@ impl<C> SccacheService<C>
 {
     pub fn new(storage: Arc<Storage>,
                handle: Handle,
+               client: &Client,
```
tamird: looks like a few test calls to this function weren't updated, so tests are failing to compile.

tomprince:
> The jobserver implementation on Unix relies on inheritance of file descriptors
in spawned processes.

It would probably be possible to pass the appropriate file descriptors, if sccache used unix sockets and https://github.com/rust-lang-nursery/unix-socket/issues/4.

luser:
I rebased this patch, fixed the compile error in the tests, added a blurb to the README about jobserver behavior, and pushed it to master: https://github.com/mozilla/sccache/commit/adaa4ef190f7de41715f1c08bbc764ac7eb4df23

Thanks! Sorry for letting this sit for so long!

alexcrichton:
👍 Thanks!

If there's any issues w/ this though feel free to cc me!

PR #187, opened 2017-10-05
Update iovec to fix a soundness hole on Windows

alexcrichton:
Also update a few other dependencies that transitively needed fixing


luser:
Not the best set of CI runs, here.

alexcrichton:
Oops, turned out I forgot to upgrade it as far as it needed to go...

Issue #107, opened 2017-04-18
Describe how to use it with MSVC


KalitaAlexey:
Hello everyone.
I tried to find any information about coupling MSVC with sccache, but I failed.
It would be great to have a page describing how to setup it, debug it and etc.

luser:
Hi there! sccache works very similarly to ccache, in that you simply take a compile commandline and prepend sccache to it, like:
`cl.exe -c -Fotest.obj test.c`
becomes: `sccache.exe cl.exe -c -Fotest.obj test.c`

How you would do this in a specific project's build system depends on the build system in use.

KalitaAlexey:
@luser,
Have you used it with DevEnv or with Visual Studio?
This information can be helpful more than only for me.

luser:
I have not, sorry! I don't know if it's even possible to make Visual Studio wrap compilations like that.

KalitaAlexey:
I tried to find any information and I failed.
That's why I am asking about tutorial about setting sccache up for MSVC.

KalitaAlexey:
I found some information about it.
Right now I am testing it.
I will open a PR if it works.

kwaegel:
Thanks @KalitaAlexey. I'd be curious to see what you're trying out, even if it doesn't end up working.

Absent a better solution, we could implement a "stand-in" mode like [cclash](https://github.com/inorton/cclash) does, replacing cl.exe with the sccache binary and forwarding calls to a renamed version of the actual compiler (e.g. cl_real.exe). It's an ugly hack, but it should work with Visual Studio even when nothing else does.

I'd also suggest changing the docs to differentiate MSVC support from Visual Studio support, since that's likely to confuse people.

luser:
> Absent a better solution, we could implement a "stand-in" mode like cclash does, replacing cl.exe with the sccache binary and forwarding calls to a renamed version of the actual compiler (e.g. cl_real.exe). It's an ugly hack, but it should work with Visual Studio even when nothing else does.

sccache does support this already--if you copy the sccache.exe binary to cl.exe it will run as if you ran `sccache.exe cl.exe`, and use the next cl.exe available in your path.

> I'd also suggest changing the docs to differentiate MSVC support from Visual Studio support, since that's likely to confuse people.

[...]

PR #192, opened 2017-10-19
Support MSVC response files

jasonwhite:
Adds support for `@`-files appearing on the MSVC command line.

The response file is expanded in-place on the command line, but not recursively like in the gcc implementation in order to conform to the spec. I also implemented a Windows command line parser for this instead of relying on [`CommandLineToArgvW`](https://msdn.microsoft.com/en-us/library/windows/desktop/bb776391(v=vs.85).aspx) just in case people somehow use the MSVC compiler on non-Windows platforms.

This was one of the blockers for MSBuild support as mentioned in https://github.com/mozilla/sccache/issues/107#issuecomment-337343925. After this is merged, I'll look into adding support for handling multiple source files on one command line.

Please rip it apart in a code review. :feelsgood:

### DIFF
--- a/docs/MSBuild.md
+++ b/docs/MSBuild.md
@@ -1,0 +1,46 @@
+**Note**: Using `sccache` with `msbuild` is not fully supported yet. Supporting
+multiple inputs on one command line is the last remaining issue. See
+https://github.com/mozilla/sccache/issues/107 for more information.
+
+Instruct `msbuild` to use our own `cl.exe`:
+
+    msbuild blah.sln /p:CLToolExe=cl.exe /p:CLToolPath=C:\tools
+
+Where `C:\tools` contains `cl.exe`, the renamed `sccache.exe`.
+
+Since there is no way to tell `msbuild` to run `sccache cl [args...]`, `sccache`
+must determine which compiler to immitate based upon the name of the executable
+alone. Thus, the `sccache` executable must be renamed to `cl.exe`.
+
+## Debugging
+
+If you want to see what the `sccache` server is doing, start it up manually
+within a Visual Studio command prompt (`cl.exe` will not find the DLLs it needs
+otherwise):
+
+    set SCCACHE_LOG_LEVEL=info
+    set SCCACHE_START_SERVER=1
+    set SCCACHE_NO_DAEMON=1
+    sccache
+
+## Cache Misses
+
+Common causes of cache misses are detailed in this section.
+
+### Multiple source files per command line
+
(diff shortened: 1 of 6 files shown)


jasonwhite:
I fixed a few problems in my implementation:
 1. While MSBuild encodes response files as UTF-16le, users may write them in other encodings such as latin1 or UTF-8. To handle this, I decided to add a new dependency on the `encoding` crate. This will decode the BOM at the begging of the response file (if any) and decide how to decode it based on that.
 2. MSBuild passes preprocessor defines as separate arguments (e.g., `/D DEBUG` instead of `/DDEBUG`). A simple change in the argument parser fixes it.
 3. Added a unit test to actually test everything.

I'm more confident that this implementation is solid now, but a code review would be great.

jasonwhite:
@luser Could you review this when you have a chance? I'm not in a rush, but it'd be nice to get this in before it rots for too long.

I'll follow up with another pull request to add support for multiple inputs per command line. It'll be a while before that part is done though.

StevenEBarbaro:
Hi all!

Any chance the above PR will be accepted/rejected sometime soon?

I've been attempting to leverage `sccache` to build Chromium on Windows with mixed results.
These changes relieve some of the problems. I believe more will be resolved by [#107 (comment)](https://github.com/mozilla/sccache/issues/107#issuecomment-337343925)

jasonwhite:
I rebased to resolve the merge conflict. This is good to go.

StevenEBarbaro:
Hi again,

Is this one projected to merge sometime soon?

luser:
Sorry, this is *just* large enough that I need to clear headspace to review it and I haven't managed.

docs/MSBuild.md
```diff
@@ -0,0 +1,46 @@
...
+
+### PDB output
+
+Invididual compilations are not allowed to output to a shared PDB. This is the
+default behavior when `/Zi` (Program Databse) or `/ZI` (Program Database for
+Edit and Continue) is specified.
+
+Instead, you must use `/Z7` where the debug information is embedded in the
+object file itself. The main trade off is that you cannot use Minimal Rebuild
+(`/Gm`) or Edit and Continue.
```
payload: Also consider the `/Fd` option to name your PDB file. When you compile a single input and write to an exclusiv PDB file, explicitly named with `/Fd` compilation, hashing and caching is safe.

froydnj:
@jasonwhite, do you have space/time to rebase this?

jasonwhite:
Closing this as I currently don't use sccache. I may come back to this in the future, but someone else is free to pick this back up where I left off.

PR #194, opened 2017-10-25
Fix rust caching

luser:
It's broken on Windows as of Rust 1.20.


alexcrichton:
Looks great! The only tweak I'd make is to also set `CARGO_TARGET_DIR` into the temporary directory for the test, but otherwise r=me

SimonSapin:
Should `CARGO_TARGET_DIR` and `CARGO_HOME` also be excluded from cache keys?

luser:
> Should CARGO_TARGET_DIR and CARGO_HOME also be excluded from cache keys?

That seems totally reasonable. File a new issue for those? Should be easy enough to fix.

PR #195, opened 2017-10-26
Replace Future::boxed with Box::new (redis, gcs)

barrbrain:
This follows up on commit 0.2.2~5 which replaced deprecated use of
Future::boxed everywhere except the redis and gcs modules.


luser:
Thanks for the patch! Sorry we missed these in the original fixup! (The Travis Mac failures are my fault, the new test I added is currently failing on Rust 1.20+ on Mac.)

PR #197, opened 2017-10-26
Use different hashing functions for different file types

metajack:
This adds a new special hasher for static libraries, which contain timestamps
and other info that prevent them from being cachable in some cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,7 +21,7 @@ dependencies = [
 
 [[package]]
 name = "ar"
-version = "0.6.1"
+version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byteorder 1.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


luser:
I did a bit of rework on this patch. It's passing tests now. I'd like to add a real-world cargo test for it alongside [the existing cargo tests](https://github.com/mozilla/sccache/blob/master/tests/sccache_cargo.rs) we have and I also need to verify that it works for windows-msvc, but otherwise I think this is pretty close to landable.

luser:
I tested this on windows-msvc and it works (apparently MSVC .lib files are the same archive format, TIL), but there's a slight complication in that MSVC generates timestamps in object files unless you pass the undocumented `/Brepro` flag. I opened an issue on the `cc` crate (which you can see above) to suggest doing so by default. Adding that to [my test crate's build script](https://github.com/luser/foo/blob/645e20a323b7f53d95880fc7ad15cdfd67c9031c/build.rs#L5-L8) made things work such that I was able to get 100% cache hits on Windows.

metajack:
What's left to land this then?

luser:
I forget what I was hung up on with this. Maybe writing tests? We had disabled the cargo integration tests on Mac because they were flaky:
https://github.com/mozilla/sccache/blob/bf4158875d5690294f4d4b2b190097b18c8d0a21/tests/sccache_cargo.rs#L21

...and I think I wanted to fix that before landing this, since otherwise we wouldn't have test coverage of the changes on the platform where we actually care about them?

In any event I'm no longer actively working on sccache. If you want to get this landed feel free to take my changes and work with @chmanchester to get it done.

drahnr:
@chmanchester what's required to get this landed :)

luser:
FYI the `cc` crate added support for invoking Apple's `ar` tool with the undocumented flag that makes it zero out timestamps, so with up-to-date dependencies in your crate graph this change is likely no longer necessary:
https://github.com/alexcrichton/cc-rs/commit/555e7737237dda29e39308c6fc6c88cf93bc5853

mitchhentges:
Hey Jack, this patch has withstanded the weather out here for quite some time, woah :open_mouth: 
Would you mind rebasing this off `master`, then we'll get this reviewed and landed?

FooBarWidget:
Here is a fork that rebases the PR on top of the latest main. It compiles, but the `test_generate_hash_key` test does not pass, not yet sure why. https://github.com/FooBarWidget/sccache/tree/feature/ar-hash

FooBarWidget:
All tests now pass. Turns out I was using iterators in a wrong way.

FooBarWidget:
I opened a separate PR in #1135 for my rebased version.

metajack:
Thanks for picking this up! I can't believe this is 4.5 years old now.

PR #198, opened 2017-10-26
Fix some rust argument definitions

glandium:
I don't know how I ended up with the definitions that landed in cae21a2,
but many had the wrong attribute, and the wrong target type.

Also, short options allow the forms:
  -Cfoo
  -C foo

while long options allow the forms:
  --codegen=foo
  --codegen foo

But for some reason short options had been made to support:
  -C=foo
  -C foo

This should effectively fix the bug found in https://bugzilla.mozilla.org/show_bug.cgi?id=1412044

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -64,7 +64,7 @@ pub struct RustHasher {
     parsed_args: ParsedArguments,
 }
 
-#[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


indygreg:
Strangely, I can't reproduce the failure I encountered earlier today. Therefore I can't confirm the fix.

I'm confused :/

luser:
Thanks!

PR #202, opened 2017-10-30
parse the -gcc-toolchain option for clang

froydnj:
Not handling this means that argument parsing interprets -gcc-toolchain's
argument as a second input file, which leads to uncacheable compilations.


luser:
Thanks! I keep thinking it'd be nice to figure out how to parse the gcc/clang docs so we could exhaustively handle all their arguments.

PR #205, opened 2017-11-12
Request rustc to always colorize output.

### LINKED ISSUE #99: Enable color output in rustc when it normally would be
@jrmuizel was testing sccache for local Rust compilation and he noted that it made rustc's errors lose their normal color formatting. This is because rustc defaults to `--color auto`, which checks if stdout is a tty. Since sccache sends rustc's stdout/stderr to pipes, it doesn't output in color. We could append `--color always` to our rustc invocations if `--color` isn't already present on the commandline. Ideally we'd check if the sccache client's stdout is a tty and only do so in that case. That would present an edge case where we could cache color rustc output, then someone could run another compilation and pipe the output, and we'd use the cached output with the colors, but it doesn't seem to be likely to be a big problem in practice. The only colorful output rustc produces in successful compilations is warnings, AFAIK. We could make that edge case work by stripping ANSI color codes from the output when stdout is not a tty. If we did that, then we could just unIlaterally pass `--color always` to rustc and strip out the color codes when they're not wanted.

versusvoid:
Strip styling when not at terminal and not explicitly asked by user.

Fix #99

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -36,6 +36,8 @@ use std::io::{
 };
 #[cfg(unix)]
 use std::os::unix::process::ExitStatusExt;
+#[cfg(unix)]
+use std::os::unix::io::AsRawFd;
 use std::path::{
     Path,
 };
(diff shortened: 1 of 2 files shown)


luser:
Hey there! Thanks for the patch, sorry for not acting on it for so long. I like the ideas here, I was just a little wary about some of the implementation details, so I rewrote it in https://github.com/mozilla/sccache/pull/224 . I wrote and published a [`strip-ansi-escapes`](https://github.com/luser/strip-ansi-escapes) crate that uses the `vte` crate for the underlying parsing, which I think should be more resilient to different inputs. I also used the `atty` crate for tty detection, and moved the detection of `--color` options into the existing rustc commandline parsing, with the server passing that info back to the client in the `CompileFinished` message.

Thanks for doing this work, even if I didn't wind up using your patch!

PR #206, opened 2017-11-12
Strip absolute path to compiled rust package to share crates cache when possible.

### LINKED ISSUE #196: Cache misses when different projects shares same dependencies
I have Rust 1.22 nightly installed. I installed sccache with `cargo install sccache` (sccache 0.2.2). I followed instruction to set up RUSTC_WRAPPER=sccache. I noticed that even multiple projects shares same dependencies, sccache will not have cache hits as expected. My OS is Windows 10.

Steps to reproduce:
1. Create a snippet project `snippet1` with one dependency and no transitive dependency:
```
[dependencies]
itoa = "0.3.4"
```
Cargo.lock may look like
```
[root]
name = "snippet"
version = "0.1.0"
dependencies = [
 "itoa 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
]

[[package]]
name = "itoa"
version = "0.3.4"
source = "registry+https://github.com/rust-lang/crates.io-index"

[metadata]
"checksum itoa 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)" = "8324a32baf01e2ae060e9de58ed0bc2320c9a2833491ee36cd3b4c414de4db8c"
```
2. Copy `snippet1` to `snippet2` with no change.
3. `cargo build` the `snippet1` project. Then `cargo clean` and `cargo build` again. Observe 1 cache hit in `sccache -s`.
4. `cargo build` the `snippet2` project and `sccache -s`.

Expected result:
sccache should now report 2 cache hits.
Actual result:
sccache still reports 1 cache hit, indicate itoa 0.3.4 from `snippet2` was not a cache hit.

Now, create `.cargo/config` file in the parent directory and point a common `target-dir` for the two projects. Repeat steps 1-4. sccache will report 2 cache hits.

This shows that sccache probably uses the cargo parameters `--out-dir` or even `-L dependency` as part of the hash key. I'm not 100% sure if there is any corner case, but this behavior defeats the purpose of sharing common library across projects with sccache.

Related cargo output for `snippet1`:
```
   Compiling itoa v0.3.4
     Running `sccache rustc --crate-name itoa D:\Application\Rust\Cargo\registry\src\github.com-1ecc6299db9ec823\itoa-0.3.4\src\lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=fd1fa8f333ce5a78 -C extra-filename=-fd1fa8f333ce5a78 --out-dir C:\workspace\snippet1\target\debug\deps -L dependency=C:\workspace\snippet1\target\debug\deps --cap-lints allow`
```
`snippet2`:
```
   Compiling itoa v0.3.4
     Running `sccache rustc --crate-name itoa D:\Application\Rust\Cargo\registry\src\github.com-1ecc6299db9ec823\itoa-0.3.4\src\lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=fd1fa8f333ce5a78 -C extra-filename=-fd1fa8f333ce5a78 --out-dir C:\workspace\snippet2\target\debug\deps -L dependency=C:\workspace\snippet2\target\debug\deps --cap-lints allow`
```

versusvoid:
Fix #196

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -589,6 +589,25 @@ impl<T> CompilerHasher<T> for RustHasher
         let cwd = cwd.to_owned();
         let env_vars = env_vars.to_vec();
         let hashes = source_hashes.join3(extern_hashes, staticlib_hashes);
+
+        // Absolute path to package (crate) being compiled. Usually contains "Cargo.toml" and "target".
+        let mut package_absolute_path = None;
+        // If output directory specified as absolute path
+        if output_dir.has_root() {
+            let mut i = None;
+            // .. find last component with name "target"
+            for (j, p) in output_dir.iter().enumerate() {
+                if p == r"target" {
+                    i = Some(j);
+                }
+            }
+            if i.is_some() {
+                // .. and take "target"'s parent as path to package
+                let p : PathBuf = output_dir.iter().take(i.unwrap()).collect();
+                package_absolute_path = p.to_str().map(|s| s.to_owned());
+            }
+        }
+
         Box::new(hashes.and_then(move |(source_hashes, extern_hashes, staticlib_hashes)|
                                         -> SFuture<_> {
             // If you change any of the inputs to the hash, you should change `CACHE_VERSION`.
@@ -601,8 +620,7 @@ impl<T> CompilerHasher<T> for RustHasher
                 m.update(d.as_bytes());
             }
             // 3. The full commandline (self.arguments)
-            // TODO: there will be full paths here, it would be nice to
-            // normalize them so we can get cross-machine cache hits.
+            // TODO: normalize paths in arguments so we can get cross-machine cache hits.
             // A few argument types are not passed in a deterministic order
             // by cargo: --extern, -L, --cfg. We'll filter those out, sort them,
             // and append them to the rest of the arguments.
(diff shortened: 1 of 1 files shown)


mmun:
I ran into this as well.

luser:
Hey, sorry for letting this sit for so long, I would really like to fix this! I took a look at your patch, and I think we could do things in a slightly simpler way (just remove `-L`, `--extern`, `--out-dir` args from the hash calculation entirely). However, from local testing unfortunately the build directory winds up in the generated rlib in a number of places, so I don't think we can actually do this without figuring something out there. ☹️  There is [a rustc issue](https://github.com/rust-lang/rust/issues/41555) that tracks adding a compiler option to remap paths, but it hasn't stabilized yet. It's available in nightly Rust as `-Zremap-path-prefix-from=... -Zremap-path-prefix-to=...`, but even with that I don't seem to get bit-identical rlibs out for dependent crates when building in different directories.

mitchmindtree:
@luser just thought I'd mention that it looks like the blocking issue you linked has [since been stabilised in rustc v1.26](https://github.com/rust-lang/rust/issues/41555#issuecomment-404822711).

@freesig and I are currently trying to reduce some massive build times nannou-org/nannou#178 and it looks like this might be a nice win, at least until cargo gets some sort of built-in caching. That said, I haven't actually tested this PR to know if it contributes significantly - today's the first day I've taken a look at sccache and I'm still getting my head around what it is and how it works!

luser:
I've got an alternate patch for this that seems to work reasonably well. I'm going to close this out and open a separate PR for that. Thanks for the patch, I'm sorry it didn't get merged!

Issue #207, opened 2017-12-05
Make hashes less dependent on source and build paths


mbitsnbites:
Currently I was not able to properly use a shared cache between two machines without them using the exact same source and build folders, which seems quite unnecessary.

Apparently (according to #4 ), there are already some thoughts for how to fix this?

~~My gut feeling is that since the preprocessed result is what's being hashed (together with the command line flags that affect how the compilation is performed), things like file names (input, output) and include paths should not be important at all, so you should be able to remove them from the hash completely, right? (this includes `-I`, `-o`, `-MT`, `-MF`, `-include` etc, plus the source file name)~~

mbitsnbites:
...come to think of it: flags for preprocessor defines (`-D`, `-U`, `-undef`, and similar) should also be possible to exclude from the hash. That would be a huge win for projects that use different compile-time configurations, since many compilation units are pre-processed to the same code regardless of configuration.

I don't know how these things are treated in sccache today, though.

mbitsnbites:
Forget everything I've said about compiler flags and paths. I did some debugging of `generate_hash_key()`, and `parsed_args.common_args` seems to be very clean, containing only flags that affect the post-preprocessing compilation step.

Same thing for the environment flags, where only `MACOSX_DEPLOYMENT_TARGET` and `IPHONEOS_DEPLOYMENT_TARGET` affect the hash.

However, the **preprocessor output** contains lots of paths, at least for gcc. These are typically of the form:
```c++
# 1 "/home/alice/git/coolproject/modules/foo/src/cool_source.cpp"
# 1 "/home/alice/git/coolproject-build/Release//"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 1 "<command-line>" 2
# 1 "/home/alice/git/coolproject/modules/foo/src/cool_source.cpp"
...
```
A relatively simple solution (that hopefully works) would be to strip the preprocessor output of all lines starting with `# `.

mbitsnbites:
PR #208 addresses this issue for GCC and Clang.

I'm not sure if the corresponding problem exists for MSVC.

mqudsi:
Does this issue affect other languages like rust?

mqudsi:
[...]

PR #208, opened 2017-12-06
Make hashes path agnostic for GCC

mbitsnbites:
We now pass -P to the preprocessor, which inhibits the generation
of linemarkers. This has two positive side effects:

  1. Absolute file paths and file names are excluded from the
     preprocessor output, which makes the hash independent of
     the location of the source and build directories.

  2. The preprocessing step is slightly faster (in one case it
     was 20% faster) and the preprocessor output is smaller (in
     the aforementioned case the output was 23% smaller) which
     should translate into faster hashing too.

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -325,6 +325,7 @@ pub fn preprocess<T>(creator: &T,
     let mut cmd = creator.clone().new_command_sync(executable);
     cmd.arg("-x").arg(language)
         .arg("-E")
+        .arg("-P")
         .arg(&parsed_args.input)
         .args(&parsed_args.preprocessor_args)
         .args(&parsed_args.common_args)


mbitsnbites:
According to my testing, this PR fixes #207 for GCC (and hopefully Clang too).

luser:
Great find! Since we no longer compile from the preprocessor output this shouldn't harm anything and it fixes this longstanding issue in a simple way. I think there still might be some quirks about file paths in debug info, but we can sort that out later.

It turns out MSVC does have an equivalent option: [`-EP`](https://docs.microsoft.com/en-us/cpp/build/reference/ep-preprocess-to-stdout-without-hash-line-directives) (vs. just [`-E`](https://docs.microsoft.com/en-us/cpp/build/reference/e-preprocess-to-stdout)). I'll switch the MSVC code to use that in a followup.

mbitsnbites:
Sweet! This makes sccache much more attractive for our slightly heterogeneous machine park.

luser:
We punted and just do all our CI builds in the same Docker image in the same paths. :)

PR #211, opened 2017-12-29
Properly shutdown when requested, if the idle timeout is zero.

tomprince:
The current code currently ignores the shutdown request in that case.


luser:
Thanks for the patch!

Issue #131, opened 2017-05-24
Change type of TEN_GIGS constants to avoid overflow on  32bit OS


F3real:
Running cargo test on 32bit OS gives:

> warning: constant evaluation error: attempt to multiply with overflow. This will become a HARD ERROR in the future
  --> src/cache/cache.rs:51:25
   |
51 | const TEN_GIGS: usize = 10 * 1024 * 1024 * 1024;
   |                         ^^^^^^^^^^^^^^^^^^^^^^^
   |

This will also require code changes in:
https://github.com/luser/lru-cache/blob/cache-meter-trait/src/lib.rs#L208

posix4e:
Any reason why not just make this 1GB in 32 bit mode?

luser:
It's not a great default, but it probably doesn't matter all that much. If we don't fix this to use a u64 all the way down then local disk caching is not going to be very useful in a 32-bit sccache, in any event. Do you have a real need for a 32-bit sccache binary?

na-g:
While this isn't my problem, I will point out that many of the SoC linux boards run 32 Bit OSes, e.g. Raspberry Pi (even on the 64 bit hardware version).

posix4e:
My use case isn't that important, but... *shrug*

luser:
That's fair. So as per the first comment here, we'd need to change the type of that constant from `usize` to `u64` and then similarly change the types down through `LruDiskCache` and our forked copy of `LruCache`.

posix4e:
Happy to make it happen

luser:
Feel free! I am not opposed to making that change, especially since I decided to fork lru-cache.

luser:
Fixed by https://github.com/mozilla/sccache/pull/212

PR #212, opened 2017-12-29
Change size and capacity types to u64 to support 32-bit Windows

johnklai1:
Currently compiling for 32-bit Windows (`--target=i686-pc-windows-msvc`) errors on the following line:

```
error[E0080]: constant evaluation error
  --> src\cache\cache.rs:57:25
   |
57 | const TEN_GIGS: usize = 10 * 1024 * 1024 * 1024;
   |                         ^^^^^^^^^^^^^^^^^^^^^^^ attempt to multiply with overflow

error: aborting due to previous error

error: Could not compile `sccache`.
```

These changes convert size and capacity to u64 instead of usize to support 32-bit targets.

### DIFF
--- a/lru-disk-cache/src/lib.rs
+++ b/lru-disk-cache/src/lib.rs
@@ -128,7 +128,7 @@ impl LruDiskCache {
     ///
     /// The cache is not observant of changes to files under `path` from external sources, it
     /// expects to have sole maintence of the contents.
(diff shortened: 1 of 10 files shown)


luser:
This fixes https://github.com/mozilla/sccache/issues/131

luser:
Thanks!

PR #213, opened 2017-12-30
Add option to not autostart server.

tomprince:


luser:
What's the intended use case here?

src/commands.rs
```diff
@@ -71,6 +71,14 @@ fn get_port() -> u16 {
         .unwrap_or(DEFAULT_PORT)
 }
 
+/// Get whether the server should be started automatically.
+fn should_autostart() -> bool {
+    env::var("SCCACHE_AUTOSTART")
+        .ok()
+        .and_then(|s| s.parse().ok())
+        .unwrap_or(false)
```
payload: Does this mean the default is false?

src/commands.rs
```diff
@@ -343,16 +351,18 @@ fn connect_or_start_server(port: u16) -> Result<ServerConnection> {
     trace!("connect_or_start_server({})", port);
     match connect_to_server(port) {
         Ok(server) => Ok(server),
-        Err(ref e) if e.kind() == io::ErrorKind::ConnectionRefused ||
-                      e.kind() == io::ErrorKind::TimedOut => {
+        Err(ref e)
+            if should_autostart() &&
+                   (e.kind() == io::ErrorKind::ConnectionRefused ||
+                        e.kind() == io::ErrorKind::TimedOut) => {
```
payload: The indention doesn't look as clean as before.

totph:
> What's the intended use case here?

Currently the server starts in the first directory where a compilation happens. This process might block an NFS unmount or confuse scripts which try to determine if a build job is still running in a certain directory.

This variable would require a clean and explicit `cd /  && sccache --start-server`. An alternative for *these* use cases would be that the server start directory can be specified or defaults to root.

luser:
That's a totally sensible use case, thanks! Indeed we've hit similar issues in Firefox CI in the past where the sccache server shut down due to inactivity and then additional compiles started later which didn't get the proper logging we configure when we run `sccache --start-server`.

luser:

PR #214, opened 2018-01-07
Update dependencies

kornelski:
Versions < 0.13 are not compatible with current Rust beta or nightly.


luser:
I'm not super excited about using an -alpha version of ring, but I don't see that we have any other choice at the moment. The Travis failure just looks like an infra failure on their end, most of the builds succeeded.

PR #215, opened 2018-01-07
Fix warnings

kornelski:


luser:
The CI failures look to all be failing to build ring on beta/nightly which should be fixed by your other PR. Thanks!

PR #231, opened 2018-03-05
Update to schannel version 0.1.11

### LINKED ISSUE #228: Intermittent TLS failures on Windows when using GCS cache.
Hello,

I wonder if anyone else has observed the following intermittent error using GCS caching on Windows:

`A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider.`

The error does not appear consistently, but when it starts happening, any GCS connections stop working for a few minutes (cache write errors start increasing), and after some time it usually disappears again. I'm guessing this must have something to do with `hyper-tls` or rather the underlying `schannel` implementation in Windows.

Searching for this problem online obviously gives results that are not very helpful, since it mostly comes down to someone not having root certificates set up correctly, or using self-signed certificates.

As a workaround, I've switched the code to use `hyper-openssl` for TLS instead [1] (see code snippet below) and have not seen any problems with that. I wonder if it would be useful to add support for this officially. I've seen that you rather would want to get rid of the dependency on the openssl create...

```
    let mut ssl = SslConnector::builder(SslMethod::tls()).unwrap();
    ssl.set_ca_file(&Path::new("roots.pem")).unwrap();
    let mut http_connector = HttpConnector::new(1, &handle);
    http_connector.enforce_http(false);
    let https_connector = HttpsConnector::with_connector(http_connector, ssl);

    let client = Client::configure()
                    .connector(https_connector.unwrap())
                    .build(handle);
```

Of course, the path to the root certificates would need to be configurable and not hardcoded.

[1]: Using Google's root certficiates from [here](https://pki.goog/roots.pem)

fxb:
Fixes #228

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -669,7 +669,7 @@ dependencies = [
  "lazy_static 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
  "openssl 0.9.23 (registry+https://github.com/rust-lang/crates.io-index)",
- "schannel 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "schannel 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


luser:
Thanks!

PR #237, opened 2018-03-15
Add documentation on how to statically link OpenSSL.

fxb:
When uing the GCS feature and distributing binaries to build machines it is easy to run into problems with OpenSSL shared libraries not being available on Linux and macOS, as well as MSVC CRT DLLs missing on older Windows versions.

Here are some instructions on how to statically link against OpenSSL on Linux and macOS, and a static CRT on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,58 @@ Build Requirements
 
 Sccache is a [Rust](https://www.rust-lang.org/) program. Building it requires `cargo` (and thus `rustc`). sccache currently requires **Rust 1.21**.
 
-We recommend you install Rust via [Rustup](https://rustup.rs/). The generated binaries can be built so that they are very portable, see [scripts/build-release.sh](scripts/build-release.sh). By default `sccache` supports a local disk cache. To build `sccache` with support for `S3` and/or `Redis` cache backends, add `--features=all` or select a specific feature by passing `s3`, `gcs`, and/or `redis`. Refer the [Cargo Documentation](http://doc.crates.io/manifest.html#the-features-section) for details.
+We recommend you install Rust via [Rustup](https://rustup.rs/). The generated binaries can be built so that they are very [portable](#building-portable-binaries)). By default `sccache` supports a local disk cache. To build `sccache` with support for `S3` and/or `Redis` cache backends, add `--features=all` or select a specific feature by passing `s3`, `gcs`, and/or `redis`. Refer the [Cargo Documentation](http://doc.crates.io/manifest.html#the-features-section) for details.
 
-## Build
+Build
+-----
 
 > $ cargo build [--features=all|redis|s3|gcs] [--release]
+
+### Building portable binaries
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,12 +32,59 @@ Build Requirements
...
 
 > $ cargo build [--features=all|redis|s3|gcs] [--release]
 
+### Building portable binaries
+
+When building with the `gcs` feature, `sccache` will depend on OpenSSL, which can be an annoyance if you want to distribute portable binaries. It is possible to statically link against OpenSSL using the steps below before building with `cargo`.
+
+#### Linux
+
+You will need to download and build OpenSSL with `-fPIC` in order to statically link against it.
```
luser: We might want to recommend building with the [`rust-musl-builder`](https://github.com/emk/rust-musl-builder/) Docker image, which is what I was previously using to build release binaries:
https://github.com/mozilla/sccache/blob/1a82a4783d360adb6fad2ede3d14c80b177912c4/scripts/build-release.sh

README.md
```diff
@@ -32,12 +32,59 @@ Build Requirements
...
+It is possible to statically link against the CRT using a `.cargo/config` file with the following contents.
+
+```
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-Ctarget-feature=+crt-static"]
+```
+
+Build with `cargo` and use `dumpbin /dependents` to check that the resulting binary does not depend on MSVC CRT DLLs anymore.
+
+In order to statically link against both the CRT and OpenSSL, you will need to build OpenSSL with a statically linked CRT, which is left as an exercise for the reader. Generally it is simpler to just ship the OpenSSL DLLs.
```
luser: I wonder if anyone has a Docker image that uses the mingw toolchain to make this easy for building Rust binaries? I have done some similar hoop-jumping to cross-compile a Windows binary with mingw32 in the past:
https://github.com/luser/breakpad-taskcluster/blob/b9578a3187ad500122378169b30ef882fa2d64bf/build-prereqs.sh#L27
https://dxr.mozilla.org/mozilla-central/rev/6ff60a083701d08c52702daf50f28e8f46ae3a1c/taskcluster/scripts/misc/minidump_stackwalk.sh#83

README.md
```diff
@@ -32,12 +32,59 @@ Build Requirements
...
+
+#### macOS
+
+Just setting the below environment variable will enable static linking.
+
+```
+export OPENSSL_STATIC=yes
+```
+
+Build with `cargo` and use `otool -L` to check that the resulting binary does not depend on OpenSSL anymore.
```
luser: We should probably also mention that `MACOSX_DEPLOYMENT_TARGET` can be set to ensure a minimum version of macOS on which the resulting binary will run.

fxb:
Yes, I've also thought about maybe adding a more detailed `BUILDING.md` file that is linked from the main `README.md`

README.md
fxb: Yes, that makes sense. I see the now removed `scripts/build-release.sh` used to set that.

luser:
I should have just merged this earlier. We can make those fixes as followups. If you feel motivated I'm happy to merge another PR, otherwise I might remember to do it at some point. :)

PR #238, opened 2018-03-18
Set env vars for compiler detection

alexcrichton:
This commit updates the C compiler detection logic to be sure to set the
environment variables on the compile request for the compiler detection as well.
This fixes a bug we discovered in rust-lang/rust where the environment variables
weren't configured when the sccache server was started but were configured when
the client instances were run.

Our bug manifested where a 64->32 cross-compilation required `PATH` to look a
particular way on MSVC. Without `PATH` correct the selected `cl.exe` compiler
was failing due to a missing dll, so the detection logic was always being
aborted an no compilations were being cached!

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -473,7 +473,10 @@ pub fn write_temp_file(pool: &CpuPool, path: &Path, contents: Vec<u8>)
 }
 
 /// If `executable` is a known compiler, return `Some(Box<Compiler>)`.
-fn detect_compiler<T>(creator: &T, executable: &Path, pool: &CpuPool)
+fn detect_compiler<T>(creator: &T,
+                      executable: &Path,
+                      env: &[(OsString, OsString)],
+                      pool: &CpuPool)
                       -> SFuture<Option<Box<Compiler<T>>>>
     where T: CommandCreatorSync
 {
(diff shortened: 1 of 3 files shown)


luser:
It looks like this is broken because the compiler tests didn't get fixed up to match the new `detect_compiler` signature:
```
error[E0061]: this function takes 4 parameters but 3 parameters were supplied
   --> src/compiler/compiler.rs:644:17
    |
476 | / fn detect_compiler<T>(creator: &T,
477 | |                       executable: &Path,
478 | |                       env: &[(OsString, OsString)],
479 | |                       pool: &CpuPool)
...   |
527 | |     }))
528 | | }
    | |_- defined here
...
644 |           let c = detect_compiler(&creator, &f.bins[0], &pool).wait().unwrap().unwrap();
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 4 parameters
```

alexcrichton:
Oops sorry about that! Should be fixed now

src/compiler/compiler.rs
```diff
@@ -542,6 +549,9 @@ gcc
     let mut cmd = creator.clone().new_command_sync(&executable);
     cmd.stdout(Stdio::piped())
        .stderr(Stdio::null());
+    for &(ref k, ref v) in env.iter() {
```
luser: We do have [`RunCommand::envs`](https://github.com/mozilla/sccache/blob/master/src/mock_command.rs#L110) FYI.

alexcrichton: Sure thing, updated!

alexcrichton:
ping r? @luser

luser:
Oh, oops! FYI I'm fine with you merging your own PRs, especially if I've already reviewed them.

PR #243, opened 2018-04-16
Replace uses of try!() with ?

kornelski:


luser:
Thanks! I had thought about doing this but I was too lazy to ever get around to it. Did you do this manually or did you use a tool?

kornelski:
I've used rerast.

Issue #102, opened 2017-04-10
Support clang-cl


luser:
We use clang-cl in some of our CI builds now, it would be good to support it in sccache.

luser:
Should be fixed by https://github.com/mozilla/sccache/pull/245/commits/a1a13cde9d5b99e5c35b8240b768da52945087e2

PR #245, opened 2018-04-24
Fix compatibility with `clang-cl.exe`

alexcrichton:
I've been giving clang-cl a spin lately to test a few things and I've noticed
that it's not currently cached by sccache. Looks like it's a pretty minor fix
though thankfully!

It looks like MSVC may print out absolute paths to included header files but
clang-cl prints out relative ones, so when testing the compiler look up the
relative paths rather than the absolute ones.


luser:
Oh, nice! This fixes https://github.com/mozilla/sccache/issues/102 . cc @aidanhs

PR #249, opened 2018-05-17
Experiments with distributed compilation

aidanhs:
An initial version that works and can actually be used to at least try it out.


Cargo.toml
```diff
@@ -45,7 +46,10 @@ rust-crypto = { version = "0.2.36", optional = true }
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
+# TODO: publish to crates.io
+sccache-dist = { path = "sccache-dist" }
```
aidanhs: I don't know how you'd prefer to deal with this - I can publish and update this PR? Or you can do it next time you release.

src/compiler/gcc.rs
```diff
@@ -349,10 +348,6 @@ pub fn preprocess<T>(creator: &T,
     let mut cmd = creator.clone().new_command_sync(executable);
     cmd.arg("-x").arg(language)
         .arg("-E");
-    // With -fprofile-generate line number information is important, so don't use -P.
-    if !parsed_args.profile_generate {
-        cmd.arg("-P");
-    }
```
aidanhs: With distributed compilation, line number information is always important. More generally, it's important if you want to use a preprocessed file as input.

src/compiler/compiler.rs
```diff
@@ -286,20 +309,110 @@ pub trait CompilerHasher<T>: fmt::Debug + Send + 'static
     fn box_clone(&self) -> Box<CompilerHasher<T>>;
 }
 
+#[cfg(not(feature = "dist"))]
```
aidanhs: This feature is disabled by default, therefore so is distributed compilation

aidanhs:
@luser this is ready for review - it adds the basic structure of two binaries for distributed compilation, the way they communicate and the initial support for C, C++ and Linux. Although this unconditionally shuffles some things around (e.g. the generation of compilation command rather than executing them), distributed compilation is hidden behind a feature flag.

Followups coming soon.

src/compiler/gcc.rs
luser: Hm. This will regress the change in https://github.com/mozilla/sccache/pull/208. Would it be possible to change things so that when distributed compilation is not enabled we can continue to use this?

Cargo.toml
luser: Is there a specific reason you're adding this code as a separate crate? It's not really usable standalone, right?

sccache-dist/src/cache.rs
```diff
@@ -0,0 +1,158 @@
...
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::fs::{self, File};
+use std::io::{self, BufReader, Read, Write};
+use std::path::{Path, PathBuf};
+use std::sync::Mutex;
+
+use errors::*;
+
+// TODO: possibly shouldn't be public
```
luser: I'm fine with TODO comments, but I'd prefer if they either get removed or if you file follow-up issues and put the issue URL in the comment.

sccache-dist/src/cache.rs
```diff
@@ -0,0 +1,158 @@
...
+        let count = reader.read(&mut buffer[..])?;
+        if count == 0 {
+            break;
+        }
+        m.update(&buffer[..count]);
+    }
+    Ok(hex(m.finish().as_ref()))
+}
+
+fn hex(bytes: &[u8]) -> String {
```
luser: This feels like something we could pull in from a crate, like the `hex` crate or similar.

sccache-dist/src/cache.rs
```diff
@@ -0,0 +1,158 @@
...
+        debug!("Weak key {} appears to be new", weak_key);
+        // TODO: don't use this as a poor exclusive lock on this global file
+        let mut cache = self.cache.lock().unwrap();
+        let file = fs::OpenOptions::new()
+            .create(true)
+            .write(true)
+            .truncate(true)
+            .open("/tmp/toolchain_cache.tar")?;
+        create(file)?;
+        // TODO: after, if still exists, remove it
```
luser: This sounds like something that should definitely be fixed.

sccache-dist/src/cache.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+    // TODO: It's more correct to have a FnBox or Box<FnOnce> here
+    // If the toolchain doesn't already exist, create it and insert into the cache
+    pub fn put_toolchain_cache(&self, weak_key: &str, create: &mut FnMut(fs::File) -> io::Result<()>) -> Result<String> {
+        if let Some(strong_key) = self.weak_to_strong(weak_key) {
+            debug!("Using cached toolchain {} -> {}", weak_key, strong_key);
+            return Ok(strong_key)
+        }
+        debug!("Weak key {} appears to be new", weak_key);
+        // TODO: don't use this as a poor exclusive lock on this global file
```
luser: Does this need to be an exclusive lock? You've already got a mutex around the `LRUDiskCache` here. If not, using a tempfile would be much nicer.

sccache-dist/src/cache.rs
```diff
@@ -0,0 +1,158 @@
...
+use std::path::{Path, PathBuf};
+use std::sync::Mutex;
+
+use errors::*;
+
+// TODO: possibly shouldn't be public
+pub struct ClientToolchainCache {
+    cache_dir: PathBuf,
+    cache: Mutex<TcCache>,
+    // Local machine mapping from 'weak' hashes to strong toolchain hashes
```
luser: Some exposition on what weak and strong hashes are would be helpful. I gather that the strong hash is a SHA512 of the toolchain tar file.

sccache-dist/src/cache.rs
aidanhs: Refined caching and scheduling are a slightly later PR.

sccache-dist/src/cache.rs
aidanhs: Added a tempfile (for now)

sccache-dist/src/cache.rs
aidanhs: Added the explanation that 'weak' is just what sccache uses as part of the cache key.

src/compiler/gcc.rs
aidanhs: Oh, this is disappointing, I should've done a bit more digging.

I'm not sure if it's better to say "distributed compilation *may* happen, your cache is completely invalid", or just preprocess a second time on-demand when sending for remote compilation. Looking at a reasonably large file of C++ code I have easily to hand, it's spending 10-20% of wall-clock time in preprocessing so running twice isn't ideal.

What about: if we *might* be distributing, do the full preprocess with line numbers but strip out the line numbers with a regexp when considering the cache key.

luser: That sounds fine. Since users have to explicitly opt-in to using distributed compilation, I don't think it's unreasonable for them to get this behavior in that case. Running the preprocessor twice seems like too much overhead just for this purpose. We can make a note somewhere if necessary that cache hits using different source paths are not possible with distributed compilation enabled.

Cargo.toml
luser: Yeah, a separate binary within the same crate is what I had envisioned. I get the compile argument (although incremental compilation should help with that nowadays), but managing crates.io releases for two intertwined crates like that is a PITA. (I've dealt with this with another project of mine.)

sccache-dist/src/cache.rs
luser: Oh, hah! If you move this code into the existing crate then you can just reuse that code, I guess. I'm certainly not going to hold you to a higher standard than the existing code. :)

src/server.rs
```diff
@@ -134,8 +136,21 @@ pub fn start_server(port: u16) -> Result<()> {
...
     let core = Core::new()?;
     let pool = CpuPool::new(20);
+    let dist_client: Arc<dist::Client> = match CONFIG.dist.scheduler_addr {
+        #[cfg(feature = "dist")]
+        Some(addr) => Arc::new(dist::http::Client::new(&core.handle(), addr, &CONFIG.dist.cache_dir.join("client"), CONFIG.dist.toolchain_cache_size)),
+        #[cfg(not(feature = "dist"))]
+        Some(_) => {
+            warn!("Scheduler address configured but dist feature disabled, disabling distributed sccache");
+            Arc::new(dist::NoopClient)
+        },
```
aidanhs: This bit has added in light of the new preprocessing rules, to use `-P` in more cases when it's possible.

src/compiler/gcc.rs
```diff
@@ -349,8 +350,11 @@ pub fn preprocess<T>(creator: &T,
...
     cmd.arg("-x").arg(language)
         .arg("-E");
+    // When performing distributed compilation, line number info is important for error
+    // reporting and to not cause spurious compilation failure (e.g. no exceptions build
+    // fails due to exceptions transitively included in the stdlib).
     // With -fprofile-generate line number information is important, so don't use -P.
-    if !parsed_args.profile_generate {
+    if !may_dist && !parsed_args.profile_generate {
         cmd.arg("-P");
     }
```
aidanhs: Added this back in and explained why it's important for a distributed compile

aidanhs:
`sccache-dist` has been moved back in-crate as a separate binary, enabled by the `dist-server` feature.

Are you ok to wait for the rest of my PRs for the TODOs to be resolved?

luser:
> Are you ok to wait for the rest of my PRs for the TODOs to be resolved?

Yes. The separate crate part was the only thing I was really hung up on. I'll make sure I've given the rest of the code here a once-over today and we should be good to merge this.

Cargo.toml
```diff
@@ -13,10 +13,20 @@ keywords = ["ccache"]
...
 appveyor = { repository = "mozilla/sccache" }
 
+[[bin]]
+name = "sccache"
+path = "src/main.rs"
+
+[[bin]]
+name = "sccache-dist"
+required-features = ["dist-server"]
+path = "sccache-dist/main.rs"
```
luser: It's a bit unconventional to have source in a path that's not under `src/`, isn't it?

lru-disk-cache/src/lib.rs
```diff
@@ -202,6 +202,23 @@ impl LruDiskCache {
             })
     }
 
+    /// Add a file by calling `with` with the open `File` corresponding to the cache at path `key`.
+    pub fn insert_with<K: AsRef<OsStr>, F: FnOnce(File) -> io::Result<()>>(&mut self, key: K, with: F) -> Result<()> {
```
luser: I'm not quite sure on the precise differences of this vs. `insert_by`. Can this just call `insert_by`, passing a closure that opens the file and calls `with(f)`?

aidanhs:
Added two new commits, one to put sccache-dist in a conventional location (and remove explicit paths to the main.rs files), one to simplify insert_with (this had some minor knock-on effects in insert_by as we don't have the size immediately available).

I also took the opportunity to clean up a couple of things in add_file, as well as add a `rel_to_abs_path` helper that means we no longer need to store the absolute path in the LRU (since it can be derived from the key).

Issue #251, opened 2018-05-31
Issue with gcc -M option


CoRfr:
The sccache output happens to be empty.

I was trying to build http://www.pjsip.org/ and it has issues with generation of dependency files.

*How to reproduce*:
```
git clone https://github.com/pjsip/pjproject
cd pjproject
CC="sccache gcc" ./configure.sh
cd pjlib/src
sccache gcc -c -M -Wall -DPJ_AUTOCONF=1 -O2 -DPJ_IS_BIG_ENDIAN=0 -DPJ_IS_LITTLE_ENDIAN=1    -I../include ../src/pjlib-test/ssl_sock.c
```
Response from sccache is completely empty.
Response from gcc or ccache:
```
ssl_sock.o: ../src/pjlib-test/ssl_sock.c /usr/include/stdc-predef.h \
 ../src/pjlib-test/test.h ../include/pj/types.h ../include/pj/config.h \
 ../include/pj/compat/cc_gcc.h ../include/pj/compat/os_auto.h \
 ../include/pj/compat/m_auto.h ../include/pj/compat/size_t.h \
 /usr/lib/gcc/x86_64-linux-gnu/6/include/stddef.h \
 ../include/pj/config_site.h \
<truncated>
```

What I can see in the log is:
```
DEBUG:sccache::server: check_compiler: Supported compiler
DEBUG:sccache::server: parse_arguments: Ok: ["-c", "-M", "-Wall", "-DPJ_AUTOCONF=1", "-O2", "-DPJ_IS_BIG_ENDIAN=0", "-DPJ_IS_LITTLE_ENDIAN=1", "-I../include", "../src/pj/file_access_unistd.c"]
DEBUG:sccache::compiler::compiler: [file_access_unistd.o]: get_cached_or_compile: ["-c", "-M", "-Wall", "-DPJ_AUTOCONF=1", "-O2", "-DPJ_IS_BIG_ENDIAN=0", "-DPJ_IS_LITTLE_ENDIAN=1", "-I../include", "../src/pj/file_access_unistd.c"]
DEBUG:tokio_proto::streaming::pipeline::advanced: not ready
DEBUG:tokio_proto::streaming::pipeline::advanced: write in body not done
...
DEBUG:sccache::compiler::compiler: [file_access_unistd.o]: generate_hash_key took 0.007 s
...
DEBUG:sccache::compiler::compiler: [file_access_unistd.o]: Compiled in 0.007 s, storing in cache
[...]

PR #252, opened 2018-05-31
Mark gcc -M and -MM as TooHard

CoRfr:
It appears that there are issues with these two as can be
seen in issue #251 .

Synchronizing with https://github.com/ccache/ccache/blob/master/src/compopt.c

With this patch it the same output as calling gcc directly.


luser:
Thanks for the patch! I apologize for letting this sit for so long without even a comment. ☹️

PR #253, opened 2018-06-04
Allow configuration of caches with a config file

aidanhs:
Notable difference to the variant in https://github.com/vvuk/sccache is that this branch retains the behaviour of trying each cache in turn.

~It does this by trying caches configured in the environment, then caches configured in the config file, then the disk cache from the environment (if any), then the disk cache configured in the config file (if any), then finally a fallback disk cache in the default location. An alternative to falling back to the configurations on the different levels would be instead to treat them as overrides, i.e. there is only ever one S3 connection attempted, with any environment settings overriding config file settings.~

It does this by accumulating a list of all configured caches (treating the environment as overriding file-based config) and working through them.

I also fixed an issue where tests would fail with `--no-default-features`, and added the run to travis so it won't be missed in future.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -53,11 +53,11 @@ fn parse_size(val: &str) -> Option<u64> {
         })
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 #[derive(Deserialize)]
 pub struct AzureCacheConfig;
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 #[derive(Deserialize)]
 pub struct DiskCacheConfig {
     pub dir: PathBuf,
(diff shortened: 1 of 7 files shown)


aidanhs:
> An alternative to falling back to the configurations on the different levels would be instead to treat them as overrides, i.e. there is only ever one S3 connection attempted, with any environment settings overriding config file settings.

I'm coming round to this idea being the better approach - if I set a redis url in my environment that didn't work, I wouldn't expect sccache to fall back to the url set in my sccache config file.

I'll update this PR appropriately.

aidanhs:
Updated PR

src/lib.rs
```diff
@@ -114,6 +116,8 @@ use std::io::Write;
 
 pub fn main() {
     init_logging();
+    // Initialise config
+    let _ = config::CONFIG.caches.len();
```
luser: You could just call `lazy_static::initialize(&CONFIG.caches)` here.

src/config.rs
```diff
@@ -0,0 +1,350 @@
...
+    let mut file = File::open(path)
+        .map_err(|e| debug!("Couldn't open config file: {}", e))
+        .ok()?;
+
+    let mut string = String::new();
+    file.read_to_string(&mut string)
+        .map_err(|e| warn!("Failed to read config file: {}", e))
+        .ok()?;
+
+    let toml: toml::Value = toml::from_str(&string)
```
luser: You could just make this `let cfg: FileConfig = toml::from_str(&string)...` and skip a step, right?

PR #260, opened 2018-07-13
Add support for --coverage

### LINKED ISSUE #258: Support --coverage
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1334940.

ccache supports it: https://lists.samba.org/archive/ccache/2015q1/001273.html.

gabriel-v:
closes #258

This initial implementation is following [this comment](https://github.com/mozilla/sccache/issues/258#issuecomment-402265866).

Here's a problem: `-fprofile-generate` on its own does not generate `.gcno` files, but this patch behaves like it did.

I think we need to split the `ProfileGenerate` flag into `ProfileArcs` and `TestCoverage`, with `ProfileArcs` having the meaning of `ProfileGenerate` and `TestCoverage` meaning we'll find a `.gcno` file in the outputs. But that would mean `--coverage` needs to be flagged with both of these. Can the `flag!` macro take two flags?

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -75,7 +75,7 @@ pub struct ParsedArguments {
     pub common_args: Vec<OsString>,
     /// Whether or not the `-showIncludes` argument is passed on MSVC
     pub msvc_show_includes: bool,
-    /// Whether the compilation is generating profiling data.
+    /// Whether the compilation is generating profiling or coverage data.
     pub profile_generate: bool,
 }
 
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -83,13 +83,16 @@ pub enum GCCArgAttribute {
     Language,
     SplitDwarf,
     ProfileGenerate,
+    TestCoverage,
+    Coverage,
 }
 
 use self::GCCArgAttribute::*;
 
 // Mostly taken from https://github.com/ccache/ccache/blob/master/src/compopt.c#L32-L84
-pub static ARGS: [(ArgInfo, GCCArgAttribute); 61] = [
+pub static ARGS: [(ArgInfo, GCCArgAttribute); 64] = [
     flag!("-", TooHard),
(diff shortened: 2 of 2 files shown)


luser:
> Can the flag! macro take two flags?

I don't believe so. I looked into making the enum a bitflags sort of thing recently but I couldn't find a crate that did exactly what I wanted. It might be easier to just add a `Coverage` flag and use it for `--coverage`, and adjust things accordingly.

gabriel-v:
Will we need to suppress [the `-P` flag here](https://github.com/mozilla/sccache/blob/512826537da482e3f94709d372193eb5221808e0/src/compiler/gcc.rs#L335) for --coverage too? @marco-c any idea on this?

marco-c:
If -P can mess line number information, then I'd say let's suppress it.

src/compiler/c.rs
```diff
@@ -75,7 +75,7 @@ pub struct ParsedArguments {
     pub common_args: Vec<OsString>,
     /// Whether or not the `-showIncludes` argument is passed on MSVC
     pub msvc_show_includes: bool,
-    /// Whether the compilation is generating profiling data.
+    /// Whether the compilation is generating profiling or coverage data.
     pub profile_generate: bool,
```
gabriel-v: I'd rename this from `profile_generate` to `keep_line_numbers`, because it's only used to suppress the `-P` flag.

luser: That seems reasonable.

luser:
The AppVeyor failure looks to be a spurious network issue FYI.

luser:
Thanks for the patch!

Jennymebage:
console.log('foo')

return (
 <div>
  <h1></h1>
 </div>
)

src/compiler/c.rs
Jennymebage: Alph applo

Issue #262, opened 2018-07-18
Handle -Xclang for clang-cl


luser:
We added basic support for clang-cl recently, but we don't handle`-Xclang`, which can be used to pass options down to the clang frontend (vs. the clang-cl frontend which parses MSVC options). We use this in Firefox builds to pass down GCC-style dependency options, like:
`-Xclang -MP -Xclang -dependency-file -Xclang .deps/lz4.obj.pp -Xclang -MT -Xclang lz4.obj`

Currently sccache refuses to cache these compiles because it thinks the arguments to `-Xclang` that do not start with a dash are input source files, and it will not cache a compile with multiple source files. At the very least we need it to understand that `-Xclang` takes an argument so that it parses the commandline correctly.

A more thorough followup would be to correctly detect clang-cl as clang-cl, and parse its commandlines with MSVC arguments but also accumulate `-Xclang` arguments into a separate list and then parse them with the clang argument parser, which might be a bit hairy. Otherwise we won't properly cache compiles like the example above, where we're instructing clang-cl to output a dependency file.

luser:
The changes in https://github.com/mozilla/sccache/pull/271 add support for this.

luser:
I finally got around to testing this locally now that #271 has merged and it does indeed fix the issue we were hitting:
```
$ /c/build/sccache/target/debug/sccache.exe ~/.mozbuild/clang/bin/clang-cl.exe -Fohello.obj -c hello.c -I.
$ /c/build/sccache/target/debug/sccache.exe ~/.mozbuild/clang/bin/clang-cl.exe -Fohello.obj -c hello.c -I.
$ /c/build/sccache/target/debug/sccache.exe --show-stats
Compile requests                 2
Compile requests executed        2
Cache hits                       1
Cache misses                     1
<...>
$ /c/build/sccache/target/debug/sccache.exe ~/.mozbuild/clang/bin/clang-cl.exe -Fohello.obj -c hello.c -I. -Xclang -MP -Xclang -dependency-file -Xclang hello.obj.pp -Xclang -MT -Xclang hello.obj
$ /c/build/sccache/target/debug/sccache.exe --show-stats
Compile requests                 3
Compile requests executed        3
Cache hits                       2
Cache misses                     1
<...>
```

emerick:
That's fantastic, thanks!

PR #271, opened 2018-08-06
Support distributed cross-compilation on Windows

aidanhs:
For review after #249

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -73,6 +73,8 @@ enum BuilderType {
 
 enum Void {}
 
+// Only supported on x86_64 Linux machines
+#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 fn main() {
     init_logging();
     std::process::exit(match parse() {
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -296,7 +296,7 @@ impl Server {
 impl ServerIncoming for Server {
     type Error = Error;
     fn handle_assign_job(&self, job_id: JobId, tc: Toolchain) -> Result<AssignJobResult> {
-        let need_toolchain = !self.cache.lock().unwrap().contains_key(&tc.archive_id);
+        let need_toolchain = !self.cache.lock().unwrap().contains_toolchain(&tc);
         assert!(self.job_toolchains.lock().unwrap().insert(job_id, tc).is_none());
         if !need_toolchain {
             // TODO: can start prepping the container now
@@ -313,10 +313,10 @@ impl ServerIncoming for Server {
         };
         let mut cache = self.cache.lock().unwrap();
         // TODO: this returns before reading all the data, is that valid?
-        if cache.contains_key(&tc.archive_id) {
+        if cache.contains_toolchain(&tc) {
             return Ok(SubmitToolchainResult::Success)
         }
-        Ok(cache.insert_with(&tc.archive_id, |mut file| io::copy(&mut {tc_rdr}, &mut file).map(|_| ()))
+        Ok(cache.insert_with(&tc, |mut file| io::copy(&mut {tc_rdr}, &mut file).map(|_| ()))
             .map(|_| SubmitToolchainResult::Success)
             .unwrap_or(SubmitToolchainResult::CannotCache))
     }
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -13,14 +13,13 @@
 // limitations under the License.
 
 use compiler::{Cacheable, ColorMode, Compiler, CompilerArguments, CompileCommand, CompilerHasher, CompilerKind,
-               Compilation, HashResult};
+               pkg::CompilerPackager, Compilation, HashResult};
 use dist;
 use futures::Future;
 use futures_cpupool::CpuPool;
 use mock_command::CommandCreatorSync;
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
-use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fmt;
 use std::fs::{self, File};
@@ -257,30 +256,7 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
             // the toolchain will not contain the correct path to invoke the compiler! Add the compiler
             // executable path to try and prevent this
             let weak_toolchain_key = format!("{}-{}", executable.to_string_lossy(), executable_digest);
-            let env_executable = executable.clone();
-            #[cfg(windows)]
-            let toolchain_creator = Box::new(|_| Err(io::Error::new(io::ErrorKind::Other, "packaging not supported on Windows")));
-            // TODO: detect OSX
-            #[cfg(unix)]
-            let toolchain_creator = Box::new(move |f| {
-                use std::os::unix::ffi::OsStrExt;
-                info!("Packaging C compiler");
-                // TODO: write our own, since this is GPL
-                let curdir = env::current_dir().unwrap();
-                env::set_current_dir("/tmp").unwrap();
-                let output = process::Command::new("icecc-create-env").arg(&env_executable).output().unwrap();
-                if !output.status.success() {
-                    println!("{:?}\n\n\n===========\n\n\n{:?}", output.stdout, output.stderr);
-                    panic!("failed to create toolchain")
-                }
-                let file_line = output.stdout.split(|&b| b == b'\n').find(|line| line.starts_with(b"creating ")).unwrap();
-                let filename = &file_line[b"creating ".len()..];
-                let filename = OsStr::from_bytes(filename);
-                io::copy(&mut File::open(filename).unwrap(), &mut {f}).unwrap();
-                fs::remove_file(filename).unwrap();
-                env::set_current_dir(curdir).unwrap();
-                Ok(())
-            });
+            let toolchain_creator = Box::new(CCompilerPackager { executable: executable.clone() });
             Ok(HashResult {
                 key: key,
                 compilation: Box::new(CCompilation {
@@ -356,6 +332,40 @@ impl<I: CCompilerImpl> Compilation for CCompilation<I> {
     fn outputs<'a>(&'a self) -> Box<Iterator<Item=(&'a str, &'a Path)> + 'a>
     {
         Box::new(self.parsed_args.outputs.iter().map(|(k, v)| (*k, &**v)))
+    }
+}
+
+struct CCompilerPackager {
+    executable: PathBuf,
+}
+
+impl CompilerPackager for CCompilerPackager {
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+    fn write_pkg(self: Box<Self>, f: File) -> io::Result<()> {
+        use std::env;
+        use std::os::unix::ffi::OsStrExt;
+
+        info!("Packaging C compiler");
+        // TODO: write our own, since this is GPL
+        let curdir = env::current_dir().unwrap();
+        env::set_current_dir("/tmp").unwrap();
+        let output = process::Command::new("icecc-create-env").arg(&self.executable).output().unwrap();
+        if !output.status.success() {
+            println!("{:?}\n\n\n===========\n\n\n{:?}", output.stdout, output.stderr);
+            panic!("failed to create toolchain")
+        }
+        let file_line = output.stdout.split(|&b| b == b'\n').find(|line| line.starts_with(b"creating ")).unwrap();
+        let filename = &file_line[b"creating ".len()..];
+        let filename = OsStr::from_bytes(filename);
+        io::copy(&mut File::open(filename).unwrap(), &mut {f}).unwrap();
+        fs::remove_file(filename).unwrap();
+        env::set_current_dir(curdir).unwrap();
+        Ok(())
+    }
+
+    #[cfg(not(all(target_os = "linux", target_arch = "x86_64")))]
+    fn write_pkg(self: Box<Self>, f: File) -> io::Result<()> {
+        Err(io::Error::new(io::ErrorKind::Other, "Automatic packaging not supported on this platform"))
     }
 }
 
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -157,6 +157,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
     /// Generate a command that can be used to invoke the C compiler to perform
     /// the compilation.
     fn generate_compile_commands(&self,
+                                path_transformer: &mut dist::PathTransformer,
                                 executable: &Path,
                                 parsed_args: &ParsedArguments,
                                 cwd: &Path,
@@ -290,19 +291,24 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
 }
 
 impl<I: CCompilerImpl> Compilation for CCompilation<I> {
-    fn generate_compile_commands(&self)
+    fn generate_compile_commands(&self, path_transformer: &mut dist::PathTransformer)
                                 -> Result<(CompileCommand, Option<dist::CompileCommand>, Cacheable)>
     {
         let CCompilation { ref parsed_args, ref executable, ref compiler, preprocessed_input: _, ref cwd, ref env_vars } = *self;
-        compiler.generate_compile_commands(executable, parsed_args, cwd, env_vars)
+        compiler.generate_compile_commands(path_transformer, executable, parsed_args, cwd, env_vars)
     }
 
-    fn into_inputs_creator(self: Box<Self>) -> Result<Box<FnMut(&mut io::Write)>> {
+    fn into_dist_inputs_creator(self: Box<Self>, path_transformer: &mut dist::PathTransformer) -> Result<Box<FnMut(&mut io::Write)>> {
+        let input_path = self.cwd.join(&self.parsed_args.input);
+        let dist_input_path = path_transformer.to_dist(&input_path).unwrap();
+        // tar-rs imposes that `set_path` takes a relative path
+        assert!(dist_input_path.starts_with("/"));
+        let dist_input_path = dist_input_path.trim_left_matches("/").to_owned();
+        assert!(!dist_input_path.starts_with("/"));
+
         Ok(Box::new(move |wtr| {
             let mut builder = tar::Builder::new(wtr);
-            let input_path = self.cwd.join(&self.parsed_args.input);
             let metadata_res = fs::metadata(&input_path);
-            let input_path = input_path.strip_prefix("/").unwrap();
 
             let mut file_header = tar::Header::new_ustar();
             // TODO: test this works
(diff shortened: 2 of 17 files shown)


src/compiler/msvc.rs
```diff
@@ -286,12 +301,58 @@ pub fn parse_arguments(arguments: &[OsString]) -> CompilerArguments<ParsedArgume
...
+                Some(s) if s.len() == 2 => NormalizedDisposition::Concatenated,
+                _ => NormalizedDisposition::Separated,
+            };
+            for arg in item.arg.normalize(norm) {
+                args.push("-Xclang".into());
+                args.push(arg)
+            }
+        }
+    }
+
```
aidanhs: For the purposes of distributed compilation it's not sufficient to just 'pass through' any `-Xclang` options as suggested in #262 - it's important to filter out the preprocessor-only args, which the dependency stuff falls under.

I'm not sure what the best solution long-term is - by the time you've created a clang (or rather, gcc) `ParseArguments` you've lost a lot of information (e.g. "there is another input file" turns into "here's an arg") which reduces our accuracy. I think you almost want an interim `LosslessParsedArgs` so you can inspect it e.g. within this function to see how it affects the msvc compilation.

For now it just grabs the pieces that can be easily partitioned into common or preprocessor args without any side effects and bails otherwise.

src/compiler/msvc.rs
```diff
@@ -328,12 +390,13 @@ pub fn parse_arguments(arguments: &[OsString]) -> CompilerArguments<ParsedArgume
...
         };
     }
+
     CompilerArguments::Ok(ParsedArguments {
         input: input.into(),
         language: language,
         depfile: depfile.map(|d| d.into()),
         outputs: outputs,
-        preprocessor_args: vec!(),
+        preprocessor_args: preprocessor_args,
```
aidanhs: I was a little unclear why all of these used to go in common args - it's pessimistic so wouldn't be a problem, and I couldn't see anything in the git history indicating why it's like this.

src/compiler/clang.rs
```diff
@@ -60,29 +62,31 @@ impl CCompilerImpl for Clang {
...
 }
 
-static ARGS: [(ArgInfo, gcc::GCCArgAttribute); 8] = [
+pub static ARGS: [(ArgInfo, gcc::GCCArgAttribute); 8] = [
     take_arg!("--serialize-diagnostics", String, Separated, PassThrough),
     take_arg!("--target", String, Separated, PassThrough),
-    take_arg!("-Xclang", String, Separated, PassThrough),
+    // TODO: should be extracted and reprocessed, though bear in mind some
+    // flags are not valid under a -Xclang
+    take_arg!("-Xclang", String, Separated, TooHard),
```
aidanhs: It felt a bit dubious to me to just pass these through blindly.

src/config.rs
```diff
@@ -180,13 +180,22 @@ impl CacheConfigs {
...
 }
 
+#[derive(Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize)]
+pub struct CustomToolchain {
+    pub compiler_executable: PathBuf,
+    pub archive: PathBuf,
+    pub archive_compiler_executable: String,
+}
+
```
aidanhs: This is the way you specify a cross compiler toolchain - the compiler executable you want to override, an archive and a path within the archive you use to override the path of the compiler executable.

aidanhs:
Oh, I hadn't realised this needed a rebase - odd, I was already rebasing on top of the last PR.

src/bin/sccache-dist/main.rs
```diff
@@ -73,6 +73,8 @@ enum BuilderType {
 
 enum Void {}
 
+// Only supported on x86_64 Linux machines
+#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
```
luser: Is there a way to control this at the cargo level so that we don't build it on other platforms?

src/compiler/c.rs
```diff
@@ -314,19 +291,24 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
     {
         let CCompilation { ref parsed_args, ref executable, ref compiler, preprocessed_input: _, ref cwd, ref env_vars } = *self;
-        compiler.generate_compile_commands(executable, parsed_args, cwd, env_vars)
+        compiler.generate_compile_commands(path_transformer, executable, parsed_args, cwd, env_vars)
     }
 
-    fn into_inputs_creator(self: Box<Self>) -> Result<Box<FnMut(&mut io::Write)>> {
+    fn into_dist_inputs_creator(self: Box<Self>, path_transformer: &mut dist::PathTransformer) -> Result<Box<FnMut(&mut io::Write)>> {
+        let input_path = self.cwd.join(&self.parsed_args.input);
+        let dist_input_path = path_transformer.to_dist(&input_path).unwrap();
```
luser: Any particular reason this has an `unwrap`?

src/compiler/c.rs
```diff
@@ -314,19 +291,24 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
+        // tar-rs imposes that `set_path` takes a relative path
+        assert!(dist_input_path.starts_with("/"));
```
luser: These shouldn't be actual `assert!`s, should they? Those will take down the whole server.

src/compiler/c.rs
```diff
@@ -359,6 +341,40 @@ impl<I: CCompilerImpl> Compilation for CCompilation<I> {
...
+impl CompilerPackager for CCompilerPackager {
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+    fn write_pkg(self: Box<Self>, f: File) -> io::Result<()> {
+        use std::env;
+        use std::os::unix::ffi::OsStrExt;
+
+        info!("Packaging C compiler");
+        // TODO: write our own, since this is GPL
+        let curdir = env::current_dir().unwrap();
+        env::set_current_dir("/tmp").unwrap();
```
luser: Wouldn't it be cleaner to set the cwd of the `Command` instead?

src/compiler/clang.rs
luser: I didn't realize this was even here, honestly. I don't think I've ever seen this used outside of clang-cl.

src/compiler/compiler.rs
```diff
@@ -332,32 +334,45 @@ fn dist_or_local_compile<T>(dist_client: Arc<dist::Client>,
...
                             -> SFuture<(Cacheable, process::Output)>
         where T: CommandCreatorSync {
+    use boxfnonce::BoxFnOnce;
     use futures::future;
 
     debug!("[{}]: Attempting distributed compilation", out_pretty);
     let compile_out_pretty = out_pretty.clone();
     let compile_out_pretty2 = out_pretty.clone();
-    let (compile_cmd, dist_compile_cmd, cacheable) = compilation.generate_compile_commands().unwrap();
+    let compile_out_pretty3 = out_pretty.clone();
```
luser: I guess it would be nice to figure out something smarter here, eh?

src/compiler/compiler.rs
```diff
@@ -332,32 +334,45 @@ fn dist_or_local_compile<T>(dist_client: Arc<dist::Client>,
...
         .and_then(move |dist_compile_cmd| {
-            let output_paths = compilation.outputs().map(|(_key, path)| cwd.join(path)).collect();
-            // TODO: does this need to be in future::result?
-            compilation.into_inputs_creator()
-                .map(|inputs_creator| (dist_compile_cmd, inputs_creator, output_paths))
+            debug!("[{}]: Creating distributed compile request", compile_out_pretty);
+            let dist_output_paths = compilation.outputs()
+                .map(|(_key, path)| path_transformer.to_dist_abs(&cwd.join(path)))
+                .collect::<Option<_>>()
+                .unwrap();
```
luser: Another unwrap.

src/compiler/compiler.rs
```diff
@@ -332,32 +334,45 @@ fn dist_or_local_compile<T>(dist_client: Arc<dist::Client>,
...
+            compilation.into_dist_inputs_creator(&mut path_transformer)
+                .map(|dist_inputs_creator| (path_transformer, dist_compile_cmd, dist_inputs_creator, dist_output_paths))
         })
-        .and_then(move |(dist_compile_cmd, inputs_creator, output_paths)| {
-            debug!("[{}]: Distributed compile request created, requesting allocation", compile_out_pretty);
+        .and_then(move |(path_transformer, mut dist_compile_cmd, dist_inputs_creator, dist_output_paths)| {
+            debug!("[{}]: Identifying toolchain", compile_out_pretty2);
+            let toolchain_creator_cb = BoxFnOnce::from(move |f| toolchain_creator.write_pkg(f));
             // TODO: put on a thread
```
luser: If you don't have a fix for this TODO in a later patch, can you file a follow-up issue for it so we don't forget?

src/compiler/compiler.rs
```diff
@@ -655,7 +671,9 @@ fn detect_c_compiler<T>(creator: T,
 {
     trace!("detect_c_compiler");
 
-    let test = b"#if defined(_MSC_VER)
+    let test = b"#if defined(_MSC_VER) && defined(__clang__)
+msvc-clang
```
luser: I would use `clang-cl` here since that's how it's generally referred to.

src/compiler/msvc.rs
```diff
@@ -219,24 +224,28 @@ static ARGS: [(ArgInfo, MSVCArgAttribute); 20] = [
...
     take_arg!("-Fr", Path, Concatenated, TooHard),
     flag!("-Fx", TooHard),
-    take_arg!("-I", Path, Concatenated, PreprocessorArgument),
+    take_arg!("-I", Path, CanBeSeparated, PreprocessorArgument),
     take_arg!("-U", String, Concatenated, PreprocessorArgument),
+    take_arg!("-Xclang", String, Separated, XClang),
     flag!("-Zi", DebugInfo),
     flag!("-c", DoCompilation),
     take_arg!("-deps", Path, Concatenated, DepFile),
+    take_arg!("-o", Path, Separated, Output), // Deprecated but valid
```
luser: Is this needed for something in particular?

src/compiler/msvc.rs
```diff
@@ -287,12 +301,58 @@ pub fn parse_arguments(arguments: &[OsString]) -> CompilerArguments<ParsedArgume
...
         match item.data {
-            Some(PreprocessorArgument) |
+            Some(PreprocessorArgument) => preprocessor_args.extend(item.arg.normalize(NormalizedDisposition::Concatenated)),
             Some(ProgramDatabase) |
             Some(DebugInfo) => common_args.extend(item.arg.normalize(NormalizedDisposition::Concatenated)),
             _ => {}
         }
     }
+
+    // TODO: doing this here reorders the arguments, hopefully that doesn't affect the meaning
```
luser: We'd probably have to look at the clang-cl source to figure this out, but hopefully it's not an issue. Maybe we can just make a note of it in the docs somewhere?

src/compiler/msvc.rs
luser: This seems like a reasonable way to handle things for now. It might be worth refactoring things in the future so that we could use something less leaky like `parse_arguments`, but I won't lose any sleep over it. clang-cl is an odd duck, and dealing with `-Xclang` is even odder. I'd probably want to read the clang-cl source to figure things out before anyone spent any more time on this.

src/compiler/msvc.rs
luser: I couldn't figure this out either, TBH. This is definitely better, though.

src/compiler/msvc.rs
```diff
@@ -493,26 +557,60 @@ fn generate_compile_commands(executable: &Path,
...
         executable: executable.to_owned(),
         arguments: arguments,
         env_vars: env_vars.to_owned(),
         cwd: cwd.to_owned(),
-    }, None, cacheable))
+    };
+
+    let dist_command = (|| {
+        // http://releases.llvm.org/6.0.0/tools/clang/docs/UsersManual.html#clang-cl
+        // TODO: Use /T... for language?
```
luser: This would be a useful followup. It looks like you could just replace `-c` with `-Tc` or `-Tp`:
https://docs.microsoft.com/en-us/cpp/build/reference/tc-tp-tc-tp-specify-source-file-type

src/compiler/pkg.rs
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io;
+use std::fs::File;
+
+pub trait CompilerPackager {
+    fn write_pkg(self: Box<Self>, f: File) -> io::Result<()>;
```
luser: Is there a particular reason this uses `io::Result` instead of our custom `Error/Result`?

src/config.rs
luser: OK, we can see how people like this in practice. I don't suppose it's worse than the current way people create cross-toolchains for icecream.

src/dist/cache.rs
```diff
@@ -14,23 +17,33 @@ use util;
 
 use errors::*;
 
+#[derive(Clone, Debug)]
+pub struct CustomToolchain {
```
luser: It seems a little unfortunate to have two `CustomToolchain` structs that differ by only one member. I gather that since this one gets stored in a `HashMap` the other member is the key, but it's still not great.

src/dist/cache.rs
```diff
@@ -49,42 +62,88 @@ impl ClientToolchainCache {
         let tc_cache_dir = cache_dir.join("tc");
         let cache = Mutex::new(TcCache::new(&tc_cache_dir, cache_size).unwrap());
 
+        let mut custom_toolchain_paths = HashMap::new();
+        for ct in config_custom_toolchains.into_iter() {
+            if custom_toolchain_paths.contains_key(&ct.compiler_executable) {
+                panic!("Multiple toolchains for {:?}", ct.compiler_executable)
```
luser: This doesn't seem good.

src/dist/cache.rs
```diff
@@ -49,42 +62,88 @@ impl ClientToolchainCache {
...
 
     // Get the bytes of a toolchain tar
-    pub fn get_toolchain_cache(&self, key: &str) -> Option<Vec<u8>> {
-        let mut toolchain_reader = match self.cache.lock().unwrap().get(key) {
-            Ok(rdr) => rdr,
-            Err(LruError::FileNotInCache) => return None,
-            Err(e) => panic!("{}", e),
+    // TODO: by this point the toolchain should be known to exist
+    pub fn get_toolchain(&self, tc: &Toolchain) -> Option<Vec<u8>> {
+        // TODO: be more relaxed about path casing and slashes on Windows
```
luser: File a followup for this? Maybe just running the paths through `fs::canonicalize` would be good enough.

src/dist/cache.rs
```diff
@@ -49,42 +62,88 @@ impl ClientToolchainCache {
...
+        let mut rdr = if let Some(custom_tc) = self.custom_toolchains.lock().unwrap().get(tc) {
+            Box::new(fs::File::open(&custom_tc.archive).unwrap())
+        } else {
+            match self.cache.lock().unwrap().get(tc) {
+                Ok(rdr) => rdr,
+                Err(LruError::FileNotInCache) => return None,
+                Err(e) => panic!("{}", e),
```
luser: This is also not good.

src/dist/cache.rs
```diff
@@ -49,42 +62,88 @@ impl ClientToolchainCache {
...
+            }
         };
         let mut ret = vec![];
-        toolchain_reader.read_to_end(&mut ret).unwrap();
+        rdr.read_to_end(&mut ret).unwrap();
         Some(ret)
     }
     // TODO: It's more correct to have a FnBox or Box<FnOnce> here
```
luser: Well, you can remove this TODO now!

src/dist/cache.rs
```diff
@@ -49,42 +62,88 @@ impl ClientToolchainCache {
...
     // If the toolchain doesn't already exist, create it and insert into the cache
-    pub fn put_toolchain_cache(&self, weak_key: &str, create: &mut FnMut(fs::File) -> io::Result<()>) -> Result<String> {
-        if let Some(strong_key) = self.weak_to_strong(weak_key) {
-            debug!("Using cached toolchain {} -> {}", weak_key, strong_key);
-            return Ok(strong_key)
+    pub fn put_toolchain(&self, compiler_path: &Path, weak_key: &str, create: BoxFnOnce<(fs::File,), io::Result<()>>) -> Result<(Toolchain, Option<String>)> {
+        if let Some(tc_and_compiler_path) = self.get_custom_toolchain(compiler_path) {
+            debug!("Using custom toolchain for {:?}", compiler_path);
+            let (tc, compiler_path) = tc_and_compiler_path.unwrap();
```
luser: Another `unwrap`.

src/dist/mod.rs
```diff
@@ -30,6 +32,124 @@ pub mod http;
 #[cfg(test)]
 mod test;
 
+// TODO: paths (particularly outputs, which are accessed by an unsandboxed program)
+// should be some pre-sanitised AbsPath type
```
luser: Maybe as a first pass you could just have a newtype around `PathBuf` that implements `AsRef`?

src/dist/mod.rs
```diff
@@ -30,6 +32,124 @@ pub mod http;
...
+#[cfg(unix)]
+mod path_transform {
+    use std::path::{Path, PathBuf};
+
+    pub struct PathTransformer;
+
+    impl PathTransformer {
+        pub fn new() -> Self { PathTransformer }
+        pub fn to_dist_abs(&mut self, p: &Path) -> Option<String> {
+            if !p.is_absolute() { panic!("non absolute path {:?}", p) }
```
luser: This shouldn't panic.

src/dist/mod.rs
```diff
@@ -30,6 +32,124 @@ pub mod http;
...
+    }
+
+    impl PathTransformer {
+        pub fn new() -> Self {
+            PathTransformer {
+                dist_to_local_path: HashMap::new(),
+            }
+        }
+        pub fn to_dist_abs(&mut self, p: &Path) -> Option<String> {
+            if !p.is_absolute() { panic!("non absolute path {:?}", p) }
```
luser: This shouldn't panic.

src/dist/mod.rs
```diff
@@ -30,6 +32,124 @@ pub mod http;
...
+                None
+            };
+
+            let mut dist_suffix = String::new();
+            for component in components {
+                let part = match component {
+                    Component::Prefix(_) |
+                    Component::RootDir => panic!("unexpected part in path {:?}", p),
+                    Component::Normal(osstr) => osstr.to_str()?,
+                    // TODO: should be forbidden
```
luser: Did you run into cases where these were necessary in practice? If not, it seems like it'd be better to forbid them now before anyone relies on this working.

PR #272, opened 2018-08-10
replace openssl dependency with pem

### LINKED ISSUE #223: Look into removing the openssl dependency for gcs
I realized that the gcs backend wasn't enabled in `--features=all`, so [I fixed that](https://github.com/mozilla/sccache/commit/de61c8b48efb9ed0596e3318703e26ccd56f6967), but it broke the AppVeyor build because it requires OpenSSL, which is a PITA to configure on Windows, so [I added an `all-windows` feature](https://github.com/mozilla/sccache/commit/1a82a4783d360adb6fad2ede3d14c80b177912c4) that includes everything but gcs for now.

We did used to [install OpenSSL for the AppVeyor build](https://github.com/mozilla/sccache/blob/fef9b5464614c7fcb24f0084ddd1afe2a0f25cb7/appveyor_openssl_install.ps1) but I'm not eager to go back to that.

The only usage of OpenSSL is for parsing PEM files. @cramertj [tried](https://github.com/mozilla/sccache/commit/39fdfd83a84ccf1f1a8df58588c33953bebe702d#diff-ebbecd80e06e3ef6b721ccc76a30e524) using the [`pem-parser`](https://crates.io/crates/pem-parser) crate during the initial landing of the gcs backend, but it's not usable enough unfortunately. It looks like the [`pem`](https://crates.io/crates/pem) crate might be suitable. Some simple testing with random .pem key files on my system shows that it produces the same data as the openssl code that's currently in sccache.

nornagon:
fixes #223

### DIFF
--- a/src/cache/gcs.rs
+++ b/src/cache/gcs.rs
@@ -34,7 +34,7 @@ use hyper::Method;
 use hyper::client::{Client, HttpConnector, Request};
 use hyper_tls::HttpsConnector;
 use jwt;
-use openssl;
+use pem;
 use serde_json;
 use tokio_core::reactor::Handle;
 use url::form_urlencoded;
@@ -229,14 +229,14 @@ impl GCSCredentialProvider {
             issued_at: chrono::UTC::now().timestamp(),
         };
 
-        let binary_key = openssl::rsa::Rsa::private_key_from_pem(
+        let pem_cert = pem::parse(
             self.sa_key.private_key.as_bytes()
-        )?.private_key_to_der()?;
+        )?;
 
         let auth_request_jwt = jwt::encode(
             &jwt::Header::new(jwt::Algorithm::RS256),
             &jwt_claims,
-            &binary_key,
+            &pem_cert.contents,
         )?;
 
         Ok(auth_request_jwt)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -725,6 +725,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "pem"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "base64 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "error-chain 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "percent-encoding"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


nornagon:
NB. I haven't tested this against actual GCS. But it compiles :D

luser:
Thanks for the patch! I'm in the middle of trying to get some large pull requests merged, but once those are taken care of I will merge this. It would be good to have someone test against GCS before landing this, though. @cramertj: do you have time for that?

luser:
If this is landable we could get rid of the silly "all-windows" feature set:
https://github.com/mozilla/sccache/blob/55210fe1bb61615fdc3aaede47dbe399ddc05794/Cargo.toml#L108
https://github.com/mozilla/sccache/blob/55210fe1bb61615fdc3aaede47dbe399ddc05794/appveyor.yml#L19

That's only there because it was too much work to build openssl for Windows, so we're not building gcs in Windows CI.

nornagon:
Nice! Sounds like a good follow-up PR.

luser:
I tested this out locally and it doesn't seem to work, unfortunately. ☹️ I created a GCS bucket and first tested sccache master with it, where it successfully stored a cache entry and was able to retrieve it as a cache hit. Then I built this PR and tried compiling the same source with the same GCS bucket and got the following error in the log:
```
TRACE:sccache::compiler::compiler: [hello.o]: Hash key: 3be4f5a0edf0f289a81522dde151d2a08d2bbe412ee4af60d48c83f6c7e93c72e5910ce9074232e5fa2b157f2eeeeeaeb3876f3d9f7b008e6e2593da9ca447a7
WARN:sccache::cache::gcs: Got GCS error: Error(Msg("Invalid Key"), State { next_error: None, backtrace: Some(stack backtrace:

<...>

   8:        0x10ef5cdc3 - sccache::cache::gcs::GCSCredentialProvider::credentials::{{closure}}::h4b9a4ddff862ad98
                        at src/cache/gcs.rs:309
```

Line 309 is here: https://github.com/mozilla/sccache/blob/b086c3883e1fd2810391f9863c020e7a54be4ce9/src/cache/gcs.rs#L309

src/cache/gcs.rs
```diff
@@ -229,14 +229,14 @@ impl GCSCredentialProvider {
...
+        let pem_cert = pem::parse(
             self.sa_key.private_key.as_bytes()
-        )?.private_key_to_der()?;
+        )?;
 
         let auth_request_jwt = jwt::encode(
             &jwt::Header::new(jwt::Algorithm::RS256),
             &jwt_claims,
-            &binary_key,
+            &pem_cert.contents,
```
fxb: It looks to me like the private key is never converted to DER format, like it was above.

fxb: Or is that just the same as decoding the base64 to binary?

drahnr: @fxb that's right (assuming you stripped the `--- BEGIN... ` `---END..` pre/post ambles first), there is a good overview at https://knowledge.digicert.com/generalinformation/INFO4448.html about the differences

luser:
I merge #367 which fixes this in a different way. Sorry we couldn't get this PR to work!

PR #274, opened 2018-08-13
Add authentication throughout sccache-dist, including simple token auth

aidanhs:
This adds
 - authenticated scheduler<->server communication in its (mostly) final form
 - the infrastructure for client<->scheduler communication, as well as a basic auth method (token)
 - job tokens to authenticate whatever you do with a job after the initial job creation

Issue #263, opened 2018-07-24
Expose CannotCache reasons somewhere outside of logs


luser:
Currently if sccache determines it can't cache a compile it will log that info and the client will run the compiler, but it's not surfaced outside of the logs anywhere. It would be nice if we saved this info and exposed it somewhere user-visible. Including it in the output of `--show-stats` is probably sensible. [There's an existing TODO on this](https://github.com/mozilla/sccache/blob/df13487fad5874f144ce1316c999170ac3803237/src/server.rs#L519) in the code.

The simplest thing to do would be to maintain a `HashMap<&'static str, int>` (probably in the `ServerStats` struct) and do like:
```
CompilerArguments::CannotCache(why) => {
    *stats.not_cached.entry(why).or_insert(0) += 1
}
```

Then add some code to print the contents of the HashMap in [`ServerStats::print`](https://github.com/mozilla/sccache/blob/df13487fad5874f144ce1316c999170ac3803237/src/server.rs#L747).

luser:
I've got a patch for this, I made it output as part of `--show-stats`. Here's what it looks like from a Firefox build:

```
Non-cacheable reasons:
-E                              27
crate-type                      29
unknown source language         11
incremental                     37
-                                6
```

PR #277, opened 2018-08-22
Richer argument handling for compilers

aidanhs:
This now correctly pulls apart arguments like `rustc -l static=a` into their constituent parts at parse time. There should be no functional changes in this PR.

This is the majority of the major argument handling rework I mentioned, pulled out of the rust distributed work for ease of review. Each of the commits can be considered in isolation.

Should be reviewed after #274 (or just examine the commits after the scheduler auth one).

### DIFF
--- a/src/compiler/args.rs
+++ b/src/compiler/args.rs
@@ -5,30 +5,29 @@ use std::fmt::{self, Debug, Display};
 use std::marker::PhantomData;
 use std::path::PathBuf;
 use std::result::Result as StdResult;
+use std::str;
 
-type ArgResult<T> = StdResult<T, ArgError>;
+pub type ArgParseResult<T> = StdResult<T, ArgParseError>;
 
 #[derive(Debug, PartialEq)]
-pub enum ArgError {
+pub enum ArgParseError {
     UnexpectedEndOfArgs,
+    InvalidUnicode(OsString),
+    Other(&'static str),
 }
 
-impl ArgError {
-    pub fn static_description(&self) -> &'static str {
-        match self {
-            ArgError::UnexpectedEndOfArgs => "Unexpected end of args",
-        }
-    }
-}
-
-impl Display for ArgError {
+impl Display for ArgParseError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let s = self.static_description();
+        let s = match self {
+            ArgParseError::UnexpectedEndOfArgs => "Unexpected end of args".into(),
+            ArgParseError::InvalidUnicode(s) => format!("String {:?} contained invalid unicode", s),
+            ArgParseError::Other(s) => format!("Arg-specific parsing failed: {}", s),
+        };
         write!(f, "{}", s)
     }
 }
 
-impl Error for ArgError {
+impl Error for ArgParseError {
     fn cause(&self) -> Option<&Error> { None }
 }
 
@@ -120,7 +119,7 @@ impl<T: ArgumentValue> Argument<T> {
     }
 }
 
-pub struct Iter<'a, T: ArgumentValue + 'a> {
+pub struct Iter<'a, T: 'a> {
     arg: &'a Argument<T>,
     emitted: usize,
 }
@@ -129,7 +128,7 @@ impl<'a, T: ArgumentValue> Iterator for Iter<'a, T> {
     type Item = OsString;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let result = match self.arg {
+        let result = match *self.arg {
             Argument::Raw(ref s) |
             Argument::UnknownFlag(ref s) => {
                 match self.emitted {
@@ -149,7 +148,7 @@ impl<'a, T: ArgumentValue> Iterator for Iter<'a, T> {
                     (0, &ArgDisposition::Concatenated(d)) => {
                         let mut s = OsString::from(s);
                         if let Some(d) = d {
-                            s.push(OsString::from(String::from_utf8(vec![d]).expect(
+                            s.push(OsString::from(str::from_utf8(&[d]).expect(
                                 "delimiter should be ascii",
                             )));
                         }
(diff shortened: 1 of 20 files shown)


src/compiler/args.rs
```diff
@@ -1,21 +1,51 @@
...
 #[derive(PartialEq, Clone, Debug)]
-pub enum Argument {
+pub enum Argument<T> {
     /// Unknown non-flag argument ; e.g. "foo"
     Raw(OsString),
     /// Unknown flag argument ; e.g. "-foo"
     UnknownFlag(OsString),
     /// Known flag argument ; e.g. "-bar"
-    Flag(&'static str),
+    Flag(&'static str, T),
```
luser: It took me a lot more code reading to figure out why `Flag` needed to store the additional value here. I think the comment above `Argument<T>` could stand an additional sentence or two explaining what the type parameter is used for.

src/compiler/args.rs
```diff
@@ -437,82 +430,63 @@ where
...
+            })
         } else {
             None
         }
     }
 }
 
 /// Helper macro used to define ArgInfo::Flag's.
+/// Variant is an enum variant, e.g. enum ArgType { Variant(()) }
 ///     flag!("-foo")
```
luser: You removed the single argument macro rule below, so the comment here should be removed as well.

src/compiler/gcc.rs
```diff
@@ -72,91 +72,94 @@ impl CCompilerImpl for GCC {
...
-    NeedDepTarget,
-    DepTarget,
-    Language,
-    SplitDwarf,
-    ProfileGenerate,
-    TestCoverage,
-    Coverage,
+ArgData!{ pub
+    TooHardFlag(()),
+    TooHard(OsString),
```
luser: This is a little awkward, but it's not really terrible.

src/compiler/gcc.rs
```diff
@@ -72,91 +72,94 @@ impl CCompilerImpl for GCC {
...
+    PreprocessorArgumentPath(PathBuf),
+    DoCompilation(()),
+    Output(PathBuf),
+    NeedDepTarget(()),
+    DepTarget(OsString),
+    Language(OsString),
+    SplitDwarf(()),
+    ProfileGenerate(()),
+    TestCoverage(()),
+    Coverage(()),
```
luser: I don't know if it's feasible, but it would be nicer if the macro could take `$x:ident` without a type and handle it as `()`.

src/compiler/gcc.rs
```diff
@@ -193,16 +196,17 @@ where
     // and interpreting it as a list of more arguments.
     let it = ExpandIncludeFile::new(cwd, arguments);
 
-    for item in ArgsIter::new(it, arg_info) {
+    for arg in ArgsIter::new(it, arg_info) {
+        let arg = try_arg!(arg.map_err(|e| e.static_description()));
```
luser: You seem to have this same `map_err` clause in all the places you use `try_arg!`, so why not just bake it into the macro?

src/compiler/gcc.rs
luser: To answer my own question: I don't think it's possible with current stable Rust. [RFC 2298](https://github.com/rust-lang/rfcs/blob/master/text/2298-macro-at-most-once-rep.md) adds a `?` repetition operator to macros which would make this feasible. This code works on Nightly (although I couldn't get it to work on the playground for some reason):
https://gist.github.com/luser/0284cecdec8f7a2f789e37ffc52f963e

luser: Oh, it needed Edition: 2018 to work on the playground. (My local nightly is a few months out of date.)

src/compiler/msvc.rs
```diff
@@ -195,44 +195,46 @@ fn encode_path(dst: &mut Write, path: &Path) -> io::Result<()> {
...
-    ShowIncludes,
-    Output,
-    DepFile,
-    ProgramDatabase,
-    DebugInfo,
-    XClang,
+ArgData!{
+    TooHardFlag(()),
+    TooHard(OsString),
+    TooHardPath(PathBuf),
```
luser: Since we don't really need the argument for `TooHardPath`, can we just drop this and use `TooHard(OsString)` for those arguments?

Further musing has me wonder if we could combine them all into `TooHard(Option<OsString>)` and just impl the right traits for `Option<OsString>`.

src/compiler/compiler.rs
```diff
@@ -462,17 +462,26 @@ pub enum CompilerArguments<T>
     /// Commandline can be handled.
     Ok(T),
     /// Cannot cache this compilation.
-    CannotCache(&'static str),
+    CannotCache(&'static str, Option<String>),
```
luser: Does this buy us a whole lot? Do we wind up with anything but "failed to parse argument value" in the second field? I have a WIP patch for https://github.com/mozilla/sccache/issues/263 that saves counts of the `CannotCache` reasons in the server stats. I don't think this change will be a problem for it, but I'm not sure how I'd expose this new field.

src/compiler/rust.rs
```diff
@@ -72,7 +72,7 @@ pub struct RustHasher {
 #[derive(Debug, Clone, PartialEq)]
 pub struct ParsedArguments {
     /// The full commandline, with arguments and their values as pairs.
```
luser: Change the comment here to match?

src/compiler/rust.rs
```diff
@@ -712,6 +712,9 @@ impl<T> CompilerHasher<T> for RustHasher
                     outputs.insert(dep_info.to_string_lossy().into_owned(), p);
                 }
                 let toolchain_creator = Box::new(RustCompilerPackager { sysroot: sysroot.clone() });
+                let mut arguments = arguments;
+                // Always request color output, the client will strip colors if needed.
+                arguments.push(Argument::WithValue("--color", ArgData::Color("always".into()), ArgDisposition::Separated));
```
luser: Would it be simpler to move this into `generate_compile_commands`? I don't remember exactly why I put it here in the first place.

src/compiler/rust.rs
```diff
@@ -317,61 +317,193 @@ impl<T> Compiler<T> for Rust
...
+    name: String,
+}
+impl FromArg for ArgLinkLibrary {
+    fn process(arg: OsString) -> ArgParseResult<Self> {
+        let (kind, name) = match split_os_string_arg(arg, "=")? {
+            (kind, Some(name)) => (kind, name),
+            // If no kind is specified, the default is dylib.
+            (name, None) => ("dylib".to_owned(), name),
+        };
+        Ok(ArgLinkLibrary { kind: kind, name: name })
```
luser: nit: this could just be `{ kind, name }`.

src/compiler/rust.rs
```diff
@@ -317,61 +317,193 @@ impl<T> Compiler<T> for Rust
...
+        let mut s = OsString::new();
+        $(
+            s.push($v);
+        )*
+        s
+    }};
+}
+
+#[derive(Clone, Debug, PartialEq)]
+struct ArgLinkLibrary {
```
luser: All of these could stand to have a one-line descriptive comment, like "Represents -l [KIND=]PATH link arguments to rustc".

src/compiler/rust.rs
```diff
@@ -317,61 +317,193 @@ impl<T> Compiler<T> for Rust
...
+        // a directory, i.e. resulting in /my/dir/.json), does the path exist?
+        let mut path = arg.clone();
+        path.push(".json");
+        if Path::new(&path).is_file() {
+            // Unfortunately, we're now not sure what will happen without having
+            // a list of all the built-in targets handy, as they don't get .json
+            // auto-added for target json discovery
+            return Ok(ArgTarget::Unsure(arg))
+        }
+        // The file doesn't exist so it can't be a path, safe to assume it's a name
```
luser: I wonder if we'd be better off running `rustc --print=target-list` in the initial compiler detection and matching against that?

src/compiler/compiler.rs
aidanhs: Ah, I think you're inverting the way you're looking at this - "argument parse" is the consistent value that always gets put in the first field (exactly because of the cannotcache reason counting), but it didn't seem ideal to just throw away all nuance of that. Specifically, take a look at `ArgParseError` - `Display` creates a description of what went wrong with what part of the argument, which isn't possible in the bounds of a static string. This nice description will get printed in the logs.

src/compiler/rust.rs
```diff
@@ -608,10 +607,13 @@ impl<T> CompilerHasher<T> for RustHasher
         let me = *self;
         let RustHasher { executable, sysroot, compiler_shlibs_digests, parsed_args: ParsedArguments { arguments, output_dir, externs, staticlibs, crate_name, dep_info, color_mode: _ } } = me;
         trace!("[{}]: generate_hash_key", crate_name);
+        // TODO: this doesn't produce correct arguments if they should be concatenated - should use iter_os_strings
```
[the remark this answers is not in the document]
aidanhs: I will observe that this behaviour has always been the case (it's just being flagged now) so I suspect in practice it does not cause issues :)

src/compiler/gcc.rs
```diff
@@ -73,25 +73,25 @@ impl CCompilerImpl for GCC {
 }
 
 ArgData!{ pub
-    TooHardFlag(()),
+    TooHardFlag,
     TooHard(OsString),
     PassThrough(OsString),
     PassThroughPath(PathBuf),
-    PreprocessorArgumentFlag(()),
+    PreprocessorArgumentFlag,
```
luser: That's much nicer! I fiddled with this for a bit but couldn't find a working solution except with the unstable `?` macro repetition operator.

luser:
The AppVeyor failures look to be network failures, but the stable and beta builds were successful so I'm going to merge this anyway.

luser:
Rebased locally and merged: https://github.com/mozilla/sccache/compare/105ab9199e8a0ed66a3bf6463f6471ea712bb65d...080e1a770d665abb6a50225c33d060baee72cb9e

PR #279, opened 2018-08-27
Use Reqwest to access s3 to follow redirects

Marwes:
Fixes #265 

(I forgot to disable rustfmt so it unfortunately formatted the file, I can try and remove it if necessary).

PR #280, opened 2018-08-27
Use redis-0.9 for async requests 

Marwes:
While fixing #279 I noticed that the redis implementation used a cpu pool to emulate futures whke using the blocking implementation so I figured I could try and update it to use the async API in 0.9.

### DIFF
--- a/src/cache/redis.rs
+++ b/src/cache/redis.rs
@@ -20,14 +20,14 @@ use cache::{
     Storage,
 };
 use errors::*;
+use futures::{future, Future};
 use futures_cpupool::CpuPool;
 use redis::{
     cmd,
     Client,
-    Commands,
-    Connection,
     InfoDict,
 };
+use redis::async::Connection;
 use std::collections::HashMap;
 use std::io::Cursor;
 use std::time::{
@@ -54,14 +54,9 @@ impl RedisCache {
     }
 
     /// Returns a connection with configured read and write timeouts.
-    fn connect(&self) -> Result<Connection> {
-        self.client.get_connection()
+    fn connect(&self) -> impl Future<Item = Connection, Error = Error> {
+        self.client.get_async_connection()
             .map_err(|e| e.into())
-            .and_then(|c| {
-                c.set_read_timeout(Some(Duration::from_millis(10_000)))?;
-                c.set_write_timeout(Some(Duration::from_millis(10_000)))?;
-                Ok(c)
-            })
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/cache/redis.rs
```diff
@@ -54,14 +54,9 @@ impl RedisCache {
...
 
     /// Returns a connection with configured read and write timeouts.
-    fn connect(&self) -> Result<Connection> {
-        self.client.get_connection()
+    fn connect(&self) -> impl Future<Item = Connection, Error = Error> {
+        self.client.get_async_connection()
             .map_err(|e| e.into())
-            .and_then(|c| {
-                c.set_read_timeout(Some(Duration::from_millis(10_000)))?;
-                c.set_write_timeout(Some(Duration::from_millis(10_000)))?;
```
flxo: Do we have something comparable with in the async version?

flxo:
I smoke tested the patches and the redis access happy path looks fine.

src/cache/redis.rs
Marwes: I looked it up and it seems it is primarly a sync concept https://github.com/tokio-rs/tokio/issues/510. The linked crate would need to be integrated in redis itself which seems awkward as well.

What is likely a better solution is to wrap the futures in `get`/`put` etc with https://docs.rs/tokio-timer/0.2.6/tokio_timer/struct.Timeout.html but it is hard to say what timeout duration is good there.

flxo: I totally agree. The 10s is the current code is just a rough guess.
Adding a `Timeout` seems reasonable to me anyway. The cache backends currently map errors to a `Cache::Miss`.

Marwes: The 10s in this case checks that tcp reads/writes happen at least every 10s, the full transmission may still take more than that, say if one uploads a large file. Probably needs to be configurable as well but I don't have any good guess towards what the largest reasonable upload/download time is for these methods :/.

Marwes: I guess each `get`/`put` is one object file/rlib etc? Maybe 60s would be a reasonable default?

luser:
Thanks for the patch! If @flxo is OK with this then I'm happy to merge it once #274 and #277 merge.

src/cache/redis.rs
flxo: The compiler mod already has a 60s timeout for [every cache access]( https://github.com/mozilla/sccache/blob/55210fe1bb61615fdc3aaede47dbe399ddc05794/src/compiler/compiler.rs#L162-L165), so adding a 60s timeout in the Redis mod might not be good and is not needed since we get a `CacheMiss` anyway.

I'd suggest to not add the timeout since we behave like the other storage mods. Maybe adding a (global) option to treat any cache error or timeout as compilation error might be considerable since sccache runs unattended in some CI most of the time.

Marwes: Ah, in that case there is no need. Since it is all async the connection will just get disconnected at that point then. The read_timeout were only needed to ensure that the thread pool didn't get exhausted by blocking connection.

luser:
I rebased and pushed these changes:
https://github.com/mozilla/sccache/commit/8bb8a9070d2b2e9850fbd7ba9404395e6eb3ead6
https://github.com/mozilla/sccache/commit/e048db0eac5aba68284f03c15241dae543697a1d

Thanks!

PR #285, opened 2018-08-28
Set env vars correctly for building rustc

tmandry:
This adds env variables during the compiler detection phase for rust. The rustc build script uses a bootstrap compiler that requires some of those env variables to run, even for `--version`.

Also, the `RUSTC_COLOR` env variable is filtered out, because it interferes with the automatic color handling in sccache.

I wanted to experiment with using this to build rustc, and immediately hit a few errors. This fixes those enough to build a working stage1 compiler. There are still other problems, though...
- more env vars should be included in the hash key
- no support for incremental compilation means this might not be that useful, yet

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -292,6 +292,11 @@ impl<'a> Hasher for HashToDigest<'a> {
     }
 }
 
+/// Turns a slice of environment var tuples into the type expected by Command::envs.
+pub fn ref_env(env: &[(OsString, OsString)]) -> impl Iterator<Item = (&OsString, &OsString)> {
+    env.iter().map(|&(ref k, ref v)| (k, v))
+}
+
 #[cfg(test)]
 mod tests {
     use std::ffi::{OsStr, OsString};
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -628,6 +628,8 @@ fn detect_compiler<T>(creator: &T,
         let child = creator.clone().new_command_sync(&executable)
             .stdout(Stdio::piped())
             .stderr(Stdio::null())
+            .env_clear()
+            .envs(env.iter().map(|&(ref k, ref v)| (k, v)))
             .args(&["--version"])
             .spawn();
         let output = child.and_then(move |child| {
@@ -655,7 +657,8 @@ fn detect_compiler<T>(creator: &T,
     Box::new(is_rustc.and_then(move |is_rustc| {
         if is_rustc {
             debug!("Found rustc");
-            Box::new(Rust::new(creator, executable, pool).map(|c| Some(Box::new(c) as Box<Compiler<T>>)))
+            Box::new(Rust::new(creator, executable, &env, pool)
+                .map(|c| Some(Box::new(c) as Box<Compiler<T>>)))
         } else {
             detect_c_compiler(creator, executable, env, pool)
         }
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -51,7 +51,7 @@ use std::time::{
 };
 use tempdir::TempDir;
 use tempfile::NamedTempFile;
-use util::{fmt_duration_as_secs, run_input_output};
+use util::{fmt_duration_as_secs, ref_env, run_input_output};
 use tokio_core::reactor::{Handle, Timeout};
 
 use errors::*;
@@ -629,7 +629,7 @@ fn detect_compiler<T>(creator: &T,
             .stdout(Stdio::piped())
             .stderr(Stdio::null())
             .env_clear()
-            .envs(env.iter().map(|&(ref k, ref v)| (k, v)))
+            .envs(ref_env(env))
             .args(&["--version"])
             .spawn();
         let output = child.and_then(move |child| {
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -239,15 +239,21 @@ fn get_compiler_outputs<T>(creator: &T,
 impl Rust {
     /// Create a new Rust compiler instance, calculating the hashes of
     /// all the shared libraries in its sysroot.
-    pub fn new<T>(mut creator: T, executable: PathBuf, pool: CpuPool) -> SFuture<Rust>
+    pub fn new<T>(mut creator: T,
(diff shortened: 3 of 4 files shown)


src/util.rs
```diff
@@ -292,6 +292,11 @@ impl<'a> Hasher for HashToDigest<'a> {
     }
 }
 
+/// Turns a slice of environment var tuples into the type expected by Command::envs.
```
luser: Nice, this bothered the heck out of me. :)

luser:
Thanks for the patch! There are a few other in-flight PRs that I'd like to get merged first, but this looks fine.

> more env vars should be included in the hash key

Do you have a sense for what these are? As the comment says we're left guessing since Rust code can use `env!` to depend on arbitrary env vars, but we can't hash the entire environment or nothing would ever get a cache hit.

> no support for incremental compilation means this might not be that useful, yet

Yeah. :-/ This might help *a little* now that rustc uses crates from crates.io though, right? Presumably you could at least get cache hits for those. With recent fixes using sccache for a compile with incremental compilation enabled should at least not have huge overhead. Longer-term I do want to sit down with some folks that understand rustc's internals and figure out a better story here. I think ideally we'd be able to do a quick check for "are there existing incremental compilation artifacts" and refuse to cache in that case (to let rustc do its thing), but if there aren't (like in a clean build) we could ideally cache the output *including* the incremental compilation artifacts so that a rebuild would be fast as well.

tmandry:
> Do you have a sense for what these are?

Yep. Here's a list of everything that looks like it was used in the build process. Go crazy!

```
"BOOTSTRAP_PARENT_ID"="39351"
"BUILD"="x86_64-apple-darwin"
"BUILD_DIR"="/Users/tyler/code/rust/rust/build"
"CARGO"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0/bin/cargo"
"CARGO_INCREMENTAL"="1"
"CARGO_MAKEFLAGS"="--jobserver-fds=3,4 -j --jobserver-auth=3,4 -j"
"CARGO_MANIFEST_DIR"="/Users/tyler/.cargo/registry/src/github.com-1ecc6299db9ec823/cc-1.0.18"
"CARGO_PKG_AUTHORS"="Alex Crichton <alex@alexcrichton.com>"
"CARGO_PKG_DESCRIPTION"="A build-time dependency for Cargo build scripts to assist in invoking the native\\nC compiler to compile native C code into a static archive to be linked into Rust\\ncode.\\n"
"CARGO_PKG_HOMEPAGE"="https://github.com/alexcrichton/cc-rs"
"CARGO_PKG_NAME"="cc"
"CARGO_PKG_VERSION"="1.0.18"
"CARGO_PKG_VERSION_MAJOR"="1"
"CARGO_PKG_VERSION_MINOR"="0"
"CARGO_PKG_VERSION_PATCH"="18"
"CARGO_PKG_VERSION_PRE"=""
"CARGO_TARGET_DIR"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0-codegen"
"CFG_CODEGEN_BACKENDS_DIR"="codegen-backends"
"CFG_COMPILER_HOST_TRIPLE"="x86_64-apple-darwin"
"CFG_LIBDIR_RELATIVE"="lib"
"CFG_PREFIX"=""
"CFG_RELEASE"="1.30.0-dev"
"CFG_RELEASE_CHANNEL"="dev"
"CFG_VERSION"="1.30.0-dev"
"DYLD_LIBRARY_PATH"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0-codegen/release/deps:/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0/lib"
"RUSTBUILD_NATIVE_DIR"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/native"
"RUSTC"="/Users/tyler/code/rust/rust/build/bootstrap/debug/rustc"
"RUSTC_BOOTSTRAP"="1"
"RUSTC_CODEGEN_UNITS"="8"
"RUSTC_DEBUGINFO"="true"
"RUSTC_DEBUGINFO_LINES"="true"
"RUSTC_DEBUG_ASSERTIONS"="true"
"RUSTC_DENY_WARNINGS"="1"
"RUSTC_ERROR_METADATA_DST"="/Users/tyler/code/rust/rust/build/tmp/extended-error-metadata"
"RUSTC_FORCE_UNSTABLE"="1"
"RUSTC_LIBDIR"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0-sysroot/lib"
"RUSTC_METADATA_SUFFIX"="rustc"
"RUSTC_REAL"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0/bin/rustc"
"RUSTC_RPATH"="true"
"RUSTC_SNAPSHOT"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0/bin/rustc"
"RUSTC_SNAPSHOT_LIBDIR"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0/lib"
"RUSTC_STAGE"="0"
"RUSTC_SYSROOT"="/Users/tyler/code/rust/rust/build/x86_64-apple-darwin/stage0-sysroot"
"RUSTC_VERBOSE"="2"
"RUSTC_WRAPPER"="sccache"
"RUSTDOC"="/Users/tyler/code/rust/rust/build/bootstrap/debug/rustdoc"
"RUSTDOC_CRATE_VERSION"="1.30.0-dev"
"RUSTDOC_REAL"="/path/to/nowhere/rustdoc/not/required"
"RUST_TEST_THREADS"="8"
"SRC"="/Users/tyler/code/rust/rust"
"TEST_MIRI"="false"
"WINAPI_NO_BUNDLED_LIBRARIES"="1"
"__CARGO_DEFAULT_LIB_METADATA"="bootstrap"
```

It looks to me like simply adding a `RUST` prefix could go a long way.

There are also `RUSTFLAGS` which get passed to the compiler. And I left out `PATH` which might matter in some situations.

> This might help a little now that rustc uses crates from crates.io though, right?

Yes, I was getting cache hits for those!

I'm not an expert on incremental compilation, but it sounds like caching its artifacts would work.

PR #286, opened 2018-08-28
Update most dependencies

Marwes:
Just went through and updated most dependencies of sccache, mainly small cosmetic changes but I also used reqwest instead of `hyper::Client` in more places to be able to remove the explicit hyper-tls usage.

## cargo outdated -R
```
Name     Project  Compat  Latest  Kind    Platform
----     -------  ------  ------  ----    --------
bincode  0.9.2    ---     1.0.1   Normal  ---
hyper    0.11.27  ---     0.12.8  Normal  ---
```

Avoided updating `bincode` since there maybe be versioning issues(?).

`hyper-0.11` is also needed `reqwest` (no version has been release with `hyper-0.12` yet) so I left it alone.


luser:
Thanks for the patch! This seems to be broken on Windows though, all the AppVeyor builds are busted:
```
error[E0432]: unresolved import `url`
  --> src\azure\blobstore.rs:25:5
   |
25 | use url::Url;
   |     ^^^ Did you mean `rouille::url`?
warning: unused import: `std::str::FromStr`
```

Note that Windows CI builds use `--features="all-windows"` which is "everything but gcs":
https://github.com/mozilla/sccache/blob/55210fe1bb61615fdc3aaede47dbe399ddc05794/Cargo.toml#L108

Marwes:
Had to add the url feature for azure. Also updated winapi to 0.3.

luser:
> Avoided updating bincode since there maybe be versioning issues(?).

I don't think this is a problem. sccache only uses bincode for communication between the client and server on the same machine, so they should always be the same version.

Marwes:
👍 Going through the rebase atm

luser:
> 👍 Going through the rebase atm

Great, thanks! I started looking at this but it was sort of a PITA.

Marwes:
Rebased and updated rand and bincode as well

luser:
Thanks again! I should probably set up dependabot now that you've done most of the hard work. :)

PR #287, opened 2018-08-29
Add a note about using Redis as LRU cache

### LINKED ISSUE #275: Add expiration on redis keys
Currently when using redis as the backing storage for sccache values won't be deleted.
Every key that is added to redis will just stay there for good, even if it is not longer relevant.
Redis offers the expires command (https://redis.io/commands/expire). By using this mechanism one can make sure that the cache server doesn't keep stuff forever.

In our current setup we added a cronjob that scans redis for keys with TTL=-1 (i.e no expiration) and sets expiration on them. A better solution would be that sccache sets the expiration of cache keys upon usage.

flxo:
Redis can be configured as a LRU cache. This patch extends the Readme with a note about those possibilities and refer to the excellent Redis configuration.

Partially fixes #275


luser:
Thanks!

PR #290, opened 2018-09-04
Distributed sccache compilation for Rust on Linux x86_64

aidanhs:


.travis.yml
```diff
@@ -9,6 +9,8 @@ matrix:
 
     # rustc version compat
     - rust: 1.27.0 # oldest supported version, keep in sync with README.md
+    - rust: 1.27.0
+      env: DIST_SCCACHE=1
```
luser: I think it would be simpler to just build dist-client in all builds, and dist-server in all the Linux builds. I can fix that in a followup, I had a WIP patch for that locally.

luser:
I manually rebased this and pushed it.

PR #291, opened 2018-09-05
Fix various trivial warnings on macos

flxo:
This patch fixes various warnings:

* Allow unused fields in CCompilerPackager and RustCompilerPackager
* Fix some unused warning in the sccache_cargo mod by moving use directives

This patch contains no functional change.

I'm not sure whether the `dead_code` annotation is fine but I couldn't figure out how to the unused warning with a `cfg` directive.


luser:
Thanks! I think most of these will be fixed by #290 but this patch is small enough to not cause merge issues so I'm fine with just landing it.

PR #294, opened 2018-09-13
Sccache support for building Rust on Windows

aidanhs:
The RlibDepDecoder is the most important piece here - in short, you can't distribute dylibs cross-platform (without something like Wine), so we either need to be pessimistic (and disallow distributed compilation if there's every a dylib anywhere in the deps folder, which is almost certain after first compilation) or have some way of telling what is a dependency of a particular crate rlib (we know the direct rlib dependencies of the thing we're compiling as they're passed in explicitly, we're interested in going a step down). `-Z ls` does this.

As a nice side effect, it makes other compiles more efficient too as it reduces what you send over the wire.

I did some informal measurements distributing over Wireless G from Windows 4core i3 non-SSD laptop to an 8core i7 SSD laptop: Firefox clobber build went from ~100mins to ~55mins. YMMV.

Unfortunately, while I was playing I saw an oddity with generated `.d` files - it looks like remapping prefixes is perhaps not working in some situations. The firefox build works, so I need to look at the implications.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -618,6 +618,36 @@ pub fn run_command(cmd: Command) -> Result<i32> {
             let stats = request_shutdown(server)?;
             stats.print();
         }
+        Command::ZeroStats => {
+            trace!("Command::ZeroStats");
+            let conn = connect_or_start_server(get_port())?;
+            let stats = request_zero_stats(conn).chain_err(|| {
+                "couldn't zero stats on server"
+            })?;
+            stats.print();
+        }
+        #[cfg(feature = "dist-client")]
+        Command::PackageToolchain(executable, out) => {
+            use compiler;
+            use futures_cpupool::CpuPool;
+
+            trace!("Command::PackageToolchain({})", executable.display());
+            let mut core = Core::new()?;
+            let jobserver = unsafe { Client::new() };
+            let creator = ProcessCommandCreator::new(&core.handle(), &jobserver);
+            let env: Vec<_> = env::vars_os().collect();
+            let pool = CpuPool::new(1);
+            let out_file = File::create(out)?;
+
+            let compiler = compiler::get_compiler_info(&creator, &executable, &env, &pool);
+            let packager = compiler.map(|c| c.get_toolchain_packager());
+            let res = packager.and_then(|p| p.write_pkg(out_file));
+            core.run(res)?
+        }
+        #[cfg(not(feature = "dist-client"))]
+        Command::PackageToolchain(_executable, _out) => {
+            bail!("Toolchain packaging not compiled in")
+        }
         Command::Compile { exe, cmdline, cwd, env_vars } => {
             trace!("Command::Compile {{ {:?}, {:?}, {:?} }}", exe, cmdline, cwd);
             let jobserver = unsafe { Client::new() };
@@ -636,14 +666,6 @@ pub fn run_command(cmd: Command) -> Result<i32> {
             return res.chain_err(|| {
                 "failed to execute compile"
             })
-        }
-        Command::ZeroStats => {
-            trace!("Command::ZeroStats");
-            let conn = connect_or_start_server(get_port())?;
-            let stats = request_zero_stats(conn).chain_err(|| {
-                "couldn't zero stats on server"
-            })?;
-            stats.print();
         }
     }
 
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -646,7 +646,7 @@ pub fn run_command(cmd: Command) -> Result<i32> {
         }
         #[cfg(not(feature = "dist-client"))]
         Command::PackageToolchain(_executable, _out) => {
-            bail!("Toolchain packaging not compiled in")
+            bail!("Toolchain packaging not compiled in, please rebuild with the dist-client feature")
         }
         Command::Compile { exe, cmdline, cwd, env_vars } => {
             trace!("Command::Compile {{ {:?}, {:?}, {:?} }}", exe, cmdline, cwd);
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -402,6 +402,9 @@ pub fn generate_compile_commands(path_transformer: &mut dist::PathTransformer,
                                 env_vars: &[(OsString, OsString)])
                                 -> Result<(CompileCommand, Option<dist::CompileCommand>, Cacheable)>
 {
+    #[cfg(not(feature = "dist-client"))]
+    let _ = path_transformer;
+
     trace!("compile");
 
     let out_file = match parsed_args.outputs.get("obj") {
@@ -434,6 +437,9 @@ pub fn generate_compile_commands(path_transformer: &mut dist::PathTransformer,
         cwd: cwd.to_owned(),
     };
 
+    #[cfg(not(feature = "dist-client"))]
+    let dist_command = None;
+    #[cfg(feature = "dist-client")]
     let dist_command = (|| {
         // https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Overall-Options.html
         let language = match parsed_args.language {
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -58,6 +58,8 @@ const LIBS_DIR: &str = "bin";
 pub struct Rust {
     /// The path to the rustc executable.
     executable: PathBuf,
+    /// The host triple for this rustc.
+    host: String,
     /// The path to the rustc sysroot.
     sysroot: PathBuf,
     /// The SHA-1 digests of all the shared libraries in rustc's $sysroot/lib (or /bin on Windows).
@@ -69,6 +71,8 @@ pub struct Rust {
 pub struct RustHasher {
     /// The path to the rustc executable.
     executable: PathBuf,
+    /// The host triple for this rustc.
+    host: String,
     /// The path to the rustc sysroot.
     sysroot: PathBuf,
     /// The SHA-1 digests of all the shared libraries in rustc's $sysroot/lib (or /bin on Windows).
@@ -103,6 +107,8 @@ pub struct ParsedArguments {
 pub struct RustCompilation {
     /// The path to the rustc executable.
     executable: PathBuf,
+    /// The host triple for this rustc.
+    host: String,
     /// The sysroot for this rustc
     sysroot: PathBuf,
     /// All arguments passed to rustc
@@ -266,9 +272,18 @@ impl Rust {
     pub fn new<T>(mut creator: T,
                   executable: PathBuf,
                   env_vars: &[(OsString, OsString)],
+                  rustc_verbose_version: &str,
                   pool: CpuPool) -> SFuture<Rust>
         where T: CommandCreatorSync,
     {
+        // Taken from Cargo
+        let host = ftry!(rustc_verbose_version
+            .lines()
+            .find(|l| l.starts_with("host: "))
+            .map(|l| &l[6..])
+            .ok_or_else(|| Error::from("rustc verbose version didn't have a line for `host:`")))
+            .to_string();
+
         let mut cmd = creator.new_command_sync(&executable);
         cmd.stdout(Stdio::piped())
             .stderr(Stdio::null())
@@ -296,10 +311,12 @@ impl Rust {
             libs.sort();
             Ok((sysroot, libs))
         });
+
         Box::new(sysroot_and_libs.and_then(move |(sysroot, libs)| {
             hash_all(&libs, &pool).map(move |digests| {
                 Rust {
                     executable: executable,
+                    host,
                     sysroot,
                     compiler_shlibs_digests: digests,
                 }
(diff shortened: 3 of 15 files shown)


src/commands.rs
```diff
@@ -618,6 +618,36 @@ pub fn run_command(cmd: Command) -> Result<i32> {
...
+            let out_file = File::create(out)?;
+
+            let compiler = compiler::get_compiler_info(&creator, &executable, &env, &pool);
+            let packager = compiler.map(|c| c.get_toolchain_packager());
+            let res = packager.and_then(|p| p.write_pkg(out_file));
+            core.run(res)?
+        }
+        #[cfg(not(feature = "dist-client"))]
+        Command::PackageToolchain(_executable, _out) => {
+            bail!("Toolchain packaging not compiled in")
```
luser: This might be slightly better if it mentioned that you need to build the `dist-client` feature.

src/compiler/rust.rs
```diff
@@ -595,6 +686,7 @@ static ARGS: [ArgInfo<ArgData>; 33] = [
     take_arg!("--out-dir", PathBuf, CanBeSeparated('='), OutDir),
     take_arg!("--pretty", OsString, CanBeSeparated('='), NotCompilation),
     take_arg!("--print", OsString, CanBeSeparated('='), NotCompilation),
+    take_arg!("--remap-path-prefix", OsString, CanBeSeparated('='), TooHard),
```
aidanhs: Remapping path prefixes was already a failure case (since it would read incorrect things from the dep file), this addition just makes it graceful.

aidanhs:
Fixed the dep file generation - essentially, remapping path prefixes only applies to inputs, but some of the paths in a .d file are outputs so aren't mapped.

I hadn't noticed it before because nothing actually *uses* the created .d files. The chosen solution does mean that the .d files retrieved from remote compilation use forward slashes (i.e. `C:/a/b` rather than `C:\a\b`) but the Windows APIs accept either so most reasonable apps should be fine.

src/commands.rs
```diff
@@ -618,6 +618,36 @@ pub fn run_command(cmd: Command) -> Result<i32> {
...
+            let creator = ProcessCommandCreator::new(&core.handle(), &jobserver);
+            let env: Vec<_> = env::vars_os().collect();
+            let pool = CpuPool::new(1);
+            let out_file = File::create(out)?;
+
+            let compiler = compiler::get_compiler_info(&creator, &executable, &env, &pool);
+            let packager = compiler.map(|c| c.get_toolchain_packager());
+            let res = packager.and_then(|p| p.write_pkg(out_file));
+            core.run(res)?
+        }
```
luser: As a followup improvement, I think it would be nice if this command printed a status message when it started (since it could presumably take a little bit of time to write the package) and also printed the toolchain hash afterwards, for ease of configuring cross-compiling clients.

We also might want to audit potential error messages in the methods in use here, since most of them were only used internally in the server previously and this exposes them directly to command-line usage.

src/compiler/gcc.rs
```diff
@@ -402,6 +402,9 @@ pub fn generate_compile_commands(path_transformer: &mut dist::PathTransformer,
                                 env_vars: &[(OsString, OsString)])
                                 -> Result<(CompileCommand, Option<dist::CompileCommand>, Cacheable)>
 {
+    #[cfg(not(feature = "dist-client"))]
+    let _ = path_transformer;
```
luser: I don't love statement-level `#[cfg]` usage like this, but I don't know that there's a better pattern. Maybe using [`cfg-if`](https://github.com/alexcrichton/cfg-if) would make it nicer?

src/compiler/rust.rs
```diff
@@ -34,15 +46,29 @@ use std::io;
...
+use std::sync::{Arc, Mutex};
+use std::time;
 use tempdir::TempDir;
 use util::{fmt_duration_as_secs, run_input_output, Digest};
 use util::{HashToDigest, OsStrExt, ref_env};
 
 use errors::*;
 
+/// Can dylibs (like proc macros) be distributed on this platform?
+#[cfg(all(feature = "dist-client", target_os = "linux", target_arch = "x86_64"))]
```
luser: Are proc macro shared libraries portable enough that you can run them on different Linux systems? I guess we'll find out in practice once more people start using this code.

src/compiler/rust.rs
```diff
@@ -34,15 +46,29 @@ use std::io;
...
+const CAN_DIST_DYLIBS: bool = true;
+#[cfg(all(feature = "dist-client", not(all(target_os = "linux", target_arch = "x86_64"))))]
+const CAN_DIST_DYLIBS: bool = false;
+
+#[cfg(feature = "dist-client")]
+const RLIB_PREFIX: &str = "lib";
+#[cfg(feature = "dist-client")]
```
luser: consts like these would probably be fine with just `#[allow(unused)]`, but it's not a big deal.

src/compiler/rust.rs
```diff
@@ -296,22 +349,56 @@ impl Rust {
...
+                    host,
                     sysroot,
                     compiler_shlibs_digests: digests,
+                    rlib_dep_reader,
                 }
             })
-        }))
+        }));
+
+        #[cfg(not(feature = "dist-client"))]
```
luser: I wonder if this would be more readable if you created a do-nothing future like `f_ok(None)` for the `#[cfg(not(feature = "dist-client"))]` branch, and then joined into the same code from either version?

src/compiler/rust.rs
luser: We should probably figure out how to deal with this longer-term but it's probably fine for now. Was it harmful in the non-distributed case, or only in the distributed case?

src/compiler/rust.rs
```diff
@@ -927,10 +1021,13 @@ impl<T> CompilerHasher<T> for RustHasher
                         (o, p)
                     })
                     .collect::<HashMap<_, _>>();
-                if let Some(dep_info) = dep_info {
+                let dep_info = if let Some(dep_info) = dep_info {
```
luser: This feels like it'd be better written as `dep_info.map(|dep_info| ...)`.

src/compiler/rust.rs
```diff
@@ -1003,18 +1113,72 @@ impl Compilation for RustCompilation {
...
+                    }
+                    "CARGO" |
+                    "CARGO_MANIFEST_DIR" => {
+                        *v = path_transformer.to_dist(Path::new(v))?
+                    },
+                    _ => (),
+                }
+            }
+            // OUT_DIR was changed during transformation, check if this compilation is relying on anything
+            // inside it - if so, disallow distributed compilation (there are sometimes hardcoded paths present)
```
luser: This would happen for things like "my build script compiled some C code" or "my build script generated Rust code", right? Is it too hard to ship those additional files across the wire in those cases, or are there other cases I'm missing?

src/compiler/rust.rs
```diff
@@ -1023,18 +1187,88 @@ impl Compilation for RustCompilation {
...
+    }
+
+    fn outputs<'a>(&'a self) -> Box<Iterator<Item=(&'a str, &'a Path)> + 'a> {
+        Box::new(self.outputs.iter().map(|(k, v)| (k.as_str(), &**v)))
+    }
+}
+
+// TODO: we do end up with slashes facing the wrong way, but Windows is agnostic so it's
+// mostly ok. We currently don't get mappings for every single path because it means we need to
+// figure out all prefixes and send them over the wire.
```
luser: I think this is OK, sccache really only strives to handle the common cases of rustc invoked by cargo, and it seems unlikely that files will wind up on different drives there.

src/compiler/rust.rs
```diff
@@ -1023,18 +1187,88 @@ impl Compilation for RustCompilation {
...
+        use tar;
+
+        debug!("Packaging compile inputs for compile");
+        let RustInputsPackager { crate_link_paths, crate_types, inputs, mut path_transformer, rlib_dep_reader, env_vars } = *{self};
+
+        // If this is a cargo build, we can assume all immediate `extern crate` dependencies
+        // have been passed on the command line, allowing us to scan them all and find the
+        // complete list of crates we might need.
+        // If it's not a cargo build, we can't to extract the `extern crate` statements and
+        // so have no way to build a list of necessary crates - send all rlibs.
```
luser: It's nice that you have fallback support here, but it would have been fine to just not support distributed compilation for the non-cargo case, honestly.

src/compiler/rust.rs
```diff
@@ -1184,6 +1427,239 @@ impl pkg::ToolchainPackager for RustToolchainPackager {
...
+        use std::io::{Seek, Write};
+
+        // Outputs in dep files (the files at the beginning of lines) are untransformed - remap-path-prefix is documented
+        // to only apply to 'inputs'.
+        trace!("Pondering on rewriting dep file {:?}", self.dep_info);
+        if let Some(dep_info) = self.dep_info {
+            for (_dep_info_dist_path, dep_info_local_path) in output_paths {
+                trace!("Comparing with {}", dep_info_local_path.display());
+                if dep_info == dep_info_local_path {
+                    error!("Replacing using the transformer {:?}", path_transformer);
```
luser: This doesn't seem like an `error`.

src/compiler/rust.rs
```diff
@@ -1184,6 +1427,239 @@ impl pkg::ToolchainPackager for RustToolchainPackager {
...
+#[cfg(feature = "dist-client")]
+impl lru_cache::Meter<PathBuf, RlibDepsDetail> for DepsSize {
+    type Measure = usize;
+    fn measure<Q: ?Sized>(&self, _k: &Q, v: &RlibDepsDetail) -> usize where PathBuf: Borrow<Q>
+    {
+        use std::mem;
+
+        // TODO: unfortunately there is exactly nothing you can do with the k given the
+        // current trait bounds. Just use some kind of sane value;
+        //let k_size = mem::size_of::<PathBuf>() + k.capacity();
```
luser: I think this was mostly to account for the memory usage case where you would want to measure both the keys and values. If you only care about the values this is fine.

src/compiler/rust.rs
```diff
@@ -1184,6 +1427,239 @@ impl pkg::ToolchainPackager for RustToolchainPackager {
...
+        // approximates to `path_size + path + vec_size + num_deps * (systemtime_size + string_size + crate_name_len)`
+        //                 `   3*8    +  100 +   3*8    +    50    * (      8         +     3*8     +       20      )`
+        //                 `2748` bytes per crate
+        // Allowing for possible overhead of up to double (for unused space in allocated memory), this means we
+        // can cache information from about 570 rlibs - easily enough for a single project.
+        const CACHE_SIZE: u64 = 3*1024*1024;
+        let cache = lru_cache::LruCache::with_meter(CACHE_SIZE, DepsSize);
+
+        let rlib_dep_reader = RlibDepReader { cache: Mutex::new(cache), executable };
+        if let Err(e) = rlib_dep_reader.discover_rlib_deps(env_vars, &temp_rlib) {
```
luser: Is this just a sanity check to ensure that `-Z ls` works at all with the current rustc?

src/compiler/rust.rs
```diff
@@ -1184,6 +1427,239 @@ impl pkg::ToolchainPackager for RustToolchainPackager {
...
+            }
+        }
+
+        trace!("Discovering dependencies of {}", rlib.display());
+
+        let mut cmd = process::Command::new(&self.executable);
+        cmd.args(&["-Z", "ls"]).arg(&rlib)
+            .env_clear()
+            .envs(ref_env(env_vars))
+            .env("RUSTC_BOOTSTRAP", "1"); // TODO: this is fairly naughty
```
luser: Is this something we expect will ever become stable? If not it's probably a necessary evil, but it's a pretty minor thing.

src/compiler/rust.rs
```diff
@@ -1184,6 +1427,239 @@ impl pkg::ToolchainPackager for RustToolchainPackager {
...
+
+        let process::Output { status, stdout, stderr } = cmd.output()?;
+
+        if !status.success() {
```
luser: We do have `run_input_output` which handles a few of these things for you, but it returns a `Future` so you'd have to refactor this a bit:
https://github.com/mozilla/sccache/blob/dda88a658c0f48994527c201e15f736ad193da4f/src/util.rs#L142

src/compiler/rust.rs
```diff
@@ -1184,6 +1427,239 @@ impl pkg::ToolchainPackager for RustToolchainPackager {
...
+        let num: usize = line_splits.next().expect("Zero strings from line split").parse()
+            .chain_err(|| "Could not parse number from rustc -Z ls")?;
+        let libstring = line_splits.next().ok_or_else(|| "No lib string on line from rustc -Z ls")?;
+        if num != dep_names.len() + 1 {
+            bail!("Unexpected numbering of {} in rustc -Z ls output", libstring)
+        }
+        assert!(line_splits.next().is_none());
+
+        let mut libstring_splits = libstring.rsplitn(2, '-');
+        // Rustc prints strict hash value (rather than extra filename as it likely should be)
```
luser: You filed this upstream, right? If so it might be good to link the issue in this comment.

luser:
Obviously a bunch of minor comments there but nothing that felt serious enough to block merging.

PR #298, opened 2018-09-19
Use the counted_array macro for the ARGS arrays

glandium:
One of the most annoying thing from using static arrays in rust is that
there is no size inference, and you end up having to give the proper
size. Which makes updates to the arrays cumbersome.

I was reading "This Week in Rust 252", which linked to (RFC: Elide array
size)[https://github.com/rust-lang/rfcs/pull/2545], set to address this
very problem, and @durka linked to his counted-arrays crate, which
already existed and essentially implements the idea behind the RFC.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,6 +254,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "counted-array"
+version = "0.1.2"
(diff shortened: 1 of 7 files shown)


luser:
Thanks! I was thinking this same thing when I last touched one of the args arrays but I didn't get around to writing the patch. Let me just make sure this isn't going to conflict with #294.

PR #302, opened 2018-10-05
feat: Cache cargo check compilations

### LINKED ISSUE #301: Cache `rustc --emit=metadata`?
We currently cache Rust compiles with `--emit` values containing only `link` or `dep-info` since most of the other possible values have niche uses:
https://github.com/mozilla/sccache/blob/dda88a658c0f48994527c201e15f736ad193da4f/src/compiler/rust.rs#L135-L138

However, running `cargo check` nowadays will use `--emit=metadata` which we could reasonably cache, and is something that has widespread usage. I've seen recommendations for setting up CI for Rust projects to do `cargo check && cargo test`. It might be worth caching this type of compile if it provides a speedup. @alexcrichton: what do you think?

Marwes:
This is a big hack just to check that it works. Tests pass at least.

Closes #301

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -263,7 +263,7 @@ pub trait CompilerHasher<T>: fmt::Debug + Send + 'static
                         let mut entry = CacheWrite::new();
                         for (key, path) in &outputs {
                             let mut f = File::open(&path)?;
-                            let mode = get_file_mode(&path)?;
+                            let mode = get_file_mode(&f)?;
                             entry.put_object(key, &mut f, mode).chain_err(|| {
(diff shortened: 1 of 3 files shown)


luser:
The failure in Travis is real, you moved the body of `test_rust_cargo` to a separate function, but it uses the macros from `log`, and those don't get included on mac:
https://github.com/mozilla/sccache/blob/04f51467d4b0b9e567cc13795889c89cb0fea3a2/tests/sccache_cargo.rs#L9-L11

You can either conditionally not compile `test_rust_cargo_cmd` on mac, or remove the conditions on that `extern crate log`.

luser:
Thanks for the patch!

PR #306, opened 2018-10-11
Fix building 'gcs' feature on Windows and add OpenSSL static linking instructions.

fxb:


luser:
Thanks! I'd love to get https://github.com/mozilla/sccache/pull/272 sorted out so we could drop the openssl dependency here.

PR #307, opened 2018-10-12
Fix deprecation warning for env::home_dir by using directories

flxo:
This is simple small patch to get rid of `env::home_dir`which got deprecated with 1.29.0 and use `directories::UserDirs::home_dir` instead.

To be honest, I couldn't test this with S3 but since it just affects the location of the credentials it should work.

### DIFF
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -18,6 +18,7 @@ use cache::{
     CacheWrite,
     Storage,
 };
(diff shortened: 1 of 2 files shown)


luser:
Thanks! I saw this go by in the CI logs at some point but didn't have time to look at it. This is trivial enough that I suspect it'll be fine.

PR #309, opened 2018-10-12
Oauth for sccache

aidanhs:
Implicit flow, should be flexible enough to permit usage of https://sso.mozilla.com/info.

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -13,7 +13,9 @@ extern crate libmount;
 extern crate log;
 extern crate lru_disk_cache;
 extern crate nix;
+extern crate openssl;
 extern crate rand;
+extern crate reqwest;
 extern crate sccache;
 #[macro_use]
 extern crate serde_derive;
@@ -37,7 +39,7 @@ use std::net::SocketAddr;
 use std::path::Path;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use errors::*;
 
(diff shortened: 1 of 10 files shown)


src/bin/sccache-dist/main.rs
```diff
@@ -332,6 +371,62 @@ fn run(command: Command) -> Result<i32> {
...
+                            return false
+                        }
+                        // Cache the token
+                        if let Some(ref auth_cache) = maybe_auth_cache {
+                            let mut auth_cache = auth_cache.lock().unwrap();
+                            let (ref mut auth_cache, _) = *auth_cache;
+                            auth_cache.insert(s.to_owned(), Instant::now());
+                        }
+                        true
+                    })
```
luser: These two new match blocks are rather large. Maybe they'd be better off factored out into separate functions?

src/commands.rs
```diff
@@ -627,6 +627,32 @@ pub fn run_command(cmd: Command) -> Result<i32> {
...
         }
         #[cfg(feature = "dist-client")]
+        Command::DistAuth => {
+            use config::{self, CONFIG};
+            use dist;
+            use url::Url;
+
+            match &CONFIG.dist.auth {
+                config::DistAuth::Token { .. } => {
+                    info!("No authentication needed for type 'token'")
```
luser: This should probably just be `println!`.

src/config.rs
```diff
@@ -16,16 +16,23 @@ use directories::ProjectDirs;
...
+use std::sync::Mutex;
 use toml;
 
+use errors::*;
+
 lazy_static! {
-    pub static ref CONFIG: Config = { Config::create() };
+    pub static ref CONFIG: Config = Config::create();
+
+    static ref CACHED_CONFIG_PATH: PathBuf = CachedConfig::file_config_path();
```
luser: I'm not wild about "cached" as the naming here, especially since we have a lot of other usage of "cache" in the codebase. We could just directly name it "OauthTokens" or something like that.

src/dist/client_auth.rs
```diff
@@ -0,0 +1,225 @@
...
+use std::sync::mpsc;
+use std::time::{Duration, Instant};
+use url::Url;
+use uuid::Uuid;
+
+use errors::*;
+
+type BoxFut = Box<Future<Item = Response<Body>, Error = hyper::Error> + Send>;
+
+// These ports need to be registered as valid callback urls in the oauth provider you're using
```
luser: This would be good to mention in the docs somewhere.

PR #323, opened 2018-10-23
Oauth for mozilla

aidanhs:
If you test with the dist-tests feature on Linux, it will do the selenium tests in docker.

I also rolled in the suggested extraction of token checking into different functions.

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -39,7 +39,7 @@ use std::net::SocketAddr;
 use std::path::Path;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::time::{Duration, Instant};
+use std::time::{UNIX_EPOCH, Duration, Instant, SystemTime};
 
 use errors::*;
 
@@ -82,6 +82,8 @@ mod scheduler_config {
         Token { token: String },
         #[serde(rename = "jwt_validate")]
         JwtValidate { audience: String, issuer: String, jwks_url: String },
+        #[serde(rename = "mozilla")]
+        Mozilla,
         #[serde(rename = "proxy_token")]
         ProxyToken { url: String, cache_secs: Option<u64> }
     }
@@ -199,6 +201,7 @@ impl Jwk {
         Ok(der)
     }
 }
+
 enum Command {
     Auth(AuthSubcommand),
     Scheduler(scheduler_config::Config),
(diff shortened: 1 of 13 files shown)


luser:
[The Travis build failure](https://travis-ci.org/mozilla/sccache/jobs/445129289) looks like a missing crate declaration in the dist code:
```
error[E0432]: unresolved import `crypto`
  --> src/dist/client_auth.rs:70:9
   |
70 |     use crypto;
   |         ^^^^^^ no `crypto` in the root
error[E0433]: failed to resolve. Maybe a missing `extern crate crypto;`?
```

aidanhs:
Never tested with `--no-default-features` before!

src/compiler/compiler.rs
```diff
@@ -665,7 +665,7 @@ fn detect_compiler<T>(creator: &T,
                       -> SFuture<Option<Box<Compiler<T>>>>
     where T: CommandCreatorSync
 {
-    trace!("detect_compiler");
+    trace!("detect_compiler: {}", executable.display());
```
luser: You know, I was just looking at a log file from our CI the other day and realizing that this log message was not very informative. :)

tests/sccache_cargo.rs
```diff
@@ -6,6 +6,7 @@
 extern crate assert_cmd;
 extern crate chrono;
 extern crate env_logger;
+extern crate escargot;
```
luser: Just curious--what does escargot buy us over the existing assert_cmd usage?

src/bin/sccache-dist/main.rs
```diff
@@ -318,6 +321,123 @@ fn parse() -> Result<Command> {
...
+    #[derive(Deserialize)]
+    struct CISProfileV2AccessInformation {
+        ldap: CISProfileV2AccessInformationLDAP,
+    }
+    #[derive(Deserialize)]
+    struct CISProfileV2 {
+        access_information: CISProfileV2AccessInformation,
+    }
+    let profile: CISProfileV2 = res.json().chain_err(|| "Cannot decode CIS V2 profile from person API")?;
+    let is_allowed = profile.access_information.ldap.value.into_iter().any(|item| item.name == "hris_dept_firefox");
```
luser: I'm not 100% confident that this is the check we want here, but maybe it's close enough for now? I don't know if "is a MoCo employee" would be considered too broad? We should probably ask the security folks for clarification.

luser:
I won't claim that I've understood the meat of the authentication code here, but I've given everything a once-over and don't see anything that worries me so I'm going to merge this. If we discover problems in testing we can land fixes in separate PRs.

PR #331, opened 2018-11-10
Update default cache dir on MacOS in the readme

juliangehring:
Details on how the default value is constructed:
- https://github.com/mozilla/sccache/blob/06b999976d71a44eec94863acd326aa03783c495/src/config.rs#L51
- https://docs.rs/directories/1.0.2/directories/struct.ProjectDirs.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Some notes about using `sccache` with [Jenkins](https://jenkins.io) are [here](d
 Storage Options
 ---------------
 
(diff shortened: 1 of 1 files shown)


luser:
That must have changed when we switched from `app-dirs` to `directories`:
https://github.com/mozilla/sccache/commit/c13e04d507a315f5732766a8bfcfb79fc9483168#diff-635d4aa96f32dd80a837e858b4672359L315

Good catch!

PR #332, opened 2018-11-12
Handle compiler plugins with -Xclang flags

mshal:


src/compiler/gcc.rs
```diff
@@ -303,6 +309,55 @@ where
         };
     }
 
+    let xclang_it = ExpandIncludeFile::new(cwd, &xclangs);
+    for arg in ArgsIter::new(xclang_it, (&ARGS[..], &clang::ARGS[..])) {
```
luser: It feels a little unfortunate to conflate the set of options that `clang` takes with the set of options that are valid for `-Xclang` since they're not exactly the same. However, there's enough overlap (and we're definitely relying on that overlap with some of the things we use in our clang-cl builds) that it would probably be a pain to split them out.

src/compiler/gcc.rs
```diff
@@ -303,6 +309,55 @@ where
...
+                Some(s) if s.len() == 2 => NormalizedDisposition::Concatenated,
+                _ => NormalizedDisposition::Separated,
+            };
+            for arg in arg.normalize(norm).iter_os_strings() {
+                args.push("-Xclang".into());
+                args.push(arg)
+            }
+        }
+    }
+
```
luser: This code is identical to the code in the MSVC implementation, right?
https://github.com/mozilla/sccache/blob/e722f83b3eb3e3f6377524fd6eb20e72d57ee1d6/src/compiler/msvc.rs#L325-L373

Should we factor this out into a separate function that we could share?

src/compiler/c.rs
```diff
@@ -257,30 +261,33 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
-                    #[cfg(feature = "dist-client")]
-                    preprocessed_input: preprocessor_result.stdout,
-                    executable: executable,
-                    compiler: compiler,
-                    cwd,
-                    env_vars,
-                }),
-                weak_toolchain_key,
-            })
+            Box::new(extra_hashes.and_then(move |extra_hashes| {
```
luser: You might not actually need the `Box::new` here (since this isn't the return value of the function, it's just the return value of this *closure*), but it's not really harmful either.

luser:
Oh nevermind, that's just github showing me a confusing list of commits, apparently.

PR #304, opened 2018-10-09
Update tokio-core uses to tokio-0.1

Marwes:
`reqwest` has received an update to drop its `tokio-core` dependency so it is now possible to move to using `tokio` here as well.

`hyperx` were pulled it to paper over `hyper`'s removal of typed headers but were otherwise avoided.

Since there are quite a few non-Send futures I had to use `current_thread::Runtime` instead of the standard multi-threaded runtime.

PR #333, opened 2018-11-19
Make frame size configureable, closes #329

### LINKED ISSUE #329: make frame size configurable
With a lot of tools that generate C/C++ code, one ends up with big files (several dozend megabites).

Currently the default frame size is too small and sccache will abort.

drahnr:


luser:
This seems reasonable as a workaround. Ideally we wouldn't have hard-coded length limitations but I don't know if the tokio crates will allow this as written.

drahnr:
I don't think unlimited length works without introducing the risk of running out of memory on the server by a single malicious client (intentional or not). So using a sane default (in most cases 8M should work just fine) with the optional overwrite via env var is imo preferable.

Is there anything else for me to do?

luser:
I think this has conflicts with https://github.com/mozilla/sccache/pull/304 that I just merged. If you rebase over that I can merge this.

drahnr:
Ready for review.

Issue #319, opened 2018-10-17
`Bearer error="invalid_jwt"` pushing jobs to Linux64 server from macOS client


ncalexan:
I'm seeing a very surprising 401 pushing jobs to my Linux64 box from my macOS client.  Surprising because I've copy-pasted the `auth` blocks from https://github.com/mozilla/sccache/blob/063e66dc1810319bb58b8e25a63fda1ba1f02c96/docs/DistributedQuickstart.md _and_ these didn't cause authentication failures on another Linux64 box.  Is it possible that file encodings matter in some way here?

```
TRACE 2018-10-17T19:13:11Z: sccache::dist::http::server: Req 3 (192.168.1.139:38374): Request { method: "POST", url: "/api/v1/distserver/assign_job/8", headers: [("Host", "192.168.1.139:10501"), ("User-Agent", "reqwest/0.8.8"), ("Accept", "*/*"), ("Authorization", "Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJqb2JfaWQiOjh9.BjC7uIViC3SKaTrrlyflOFsBHOsaoYlIFMPgYRPvsTU"), ("Content-Type", "application/octet-stream"), ("Content-Length", "136"), ("Accept-Encoding", "gzip")], https: false, remote_addr: V4(192.168.1.139:38374) }
TRACE 2018-10-17T19:13:11Z: sccache::dist::http::server: Res 3: Response { status_code: 401, headers: [("WWW-Authenticate", "Bearer error=\"invalid_jwt\"")] }
```

/cc @aidanhs

Issue #318, opened 2018-10-17
assertion failed: self.custom_toolchains.lock().unwrap().insert(tc.clone(), custom_tc.clone()).is_none()', src/dist/cache.rs:174:21


ncalexan:
This might be a doc request, but feels like an actual issue.  I'm tripping the assertion:
```
DEBUG 2018-10-17T19:02:44Z: sccache::compiler::compiler: [conftest.R6cKbe.o]: Identifying dist toolchain for "/Users/nalexander/.mozbuild/macosx64-clang-6/bin/clang++"
thread 'main' panicked at 'assertion failed: self.custom_toolchains.lock().unwrap().insert(tc.clone(),
                                              custom_tc.clone()).is_none()', src/dist/cache.rs:174:21
```

I think what might be happening is that my configuration is wrong in some way.  The relevant piece is:

```
[[dist.toolchains]]
type = "path_override"
compiler_executable = "/Users/nalexander/.mozbuild/macosx64-clang-6/bin/clang"
archive = "/Users/nalexander/.mozbuild/clang-6.0.0-x86_64.tar.gz"
archive_compiler_executable = "usr/bin/clang"

[[dist.toolchains]]
type = "path_override"
compiler_executable = "/Users/nalexander/.mozbuild/macosx64-clang-6/bin/clang++"
archive = "/Users/nalexander/.mozbuild/clang-6.0.0-x86_64.tar.gz"
archive_compiler_executable = "usr/bin/clang++"
```

And the problem is that we're keying the `custom_toolchains` map by `Toolchain { archive_id }`, which are the same in this situation because one archive is providing two compiler executables.  Is there a way to specify what I want here?

/cc @aidanhs

ncalexan:
I'm pretty confident in my diagnosis.  I copied my archive to another path, and still hit the issue: that's because `archive_id` is a hash of the file contents.  I appended a byte to the archive for the second executable, and it works \o/

Issue #315, opened 2018-10-16
dist server `public_addr` setting is confusing


luser:
cc @aidanhs 

This config setting needs to be set to the IP:port that the server listens on, but the port here doesn't change the port the server actually listens on, so in practice it just needs to be IP:10501. Either the server should listen on the port specified in this setting, or the setting should just be an IP address instead of IP:port.

PR #334, opened 2018-11-20
Add tests, remove unwraps, fix bugs, auto https

aidanhs:
- Vast majority of unwraps removed
 - Tests for the dist system added (uses docker)
 - Script added to check the major combinations of features (I've heavily used it for the past few months, thought it might be useful)
 - Servers will auto-configure themselves to use https, informing the scheduler of certificates that clients should then use
 - Due to a bug in tiny-http, sccache currently uses a patch version
 - `scheduler_addr` no longer exists - the setting is now `scheduler_url`, to more accurately reflect the acceptance of http and https urls (though `ip:port` socket addresses are accepted and interpreted as HTTP for backwards compatibility and simplicity)

Fixes:
 - Ports are actually now configured in config files - fixes https://github.com/mozilla/sccache/issues/315
 - Server restarts are correctly registered in scheduler (+test) - fixes https://github.com/mozilla/sccache/issues/319
 - Toolchains now support being used for multiple binaries (+test) - fixes https://github.com/mozilla/sccache/issues/318

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -685,7 +685,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "libflate"
-version = "0.1.16"
+version = "0.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "adler32 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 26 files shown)


aidanhs:
After doing benchmarking it seems slightly (2min, 10min slower than measurements on https://github.com/mozilla/sccache/pull/294 on 2 runs), but that's still a good speedup over 100+min and the 10min could just have been variation based on not having firefox source in cache so loading from a slow HDD.

luser:
I admit I did not give this the most thorough of reviews but I gave it a once-over and it looks good.

PR #336, opened 2018-11-24
Fix parsing dep info containing escaped spaces.

reitermarkus:
Previously, a `dep-info` output like

```text
some\ directory/file.rs
```

would be parsed as `["some\\", "directory/file.rs"]`.

This fixes the `parse_dep_info` function so it will correctly return `["some directory/file.rs"]`.

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -261,11 +261,38 @@ fn parse_dep_info<T>(dep_info: &str, cwd: T) -> Vec<PathBuf>
         None => return vec![],
         Some(p) => p,
     };
-    let mut deps = line[pos + 2..]
-        .split(' ')
-        .map(|s| s.trim()).filter(|s| !s.is_empty())
-        .map(|s| cwd.join(s))
-        .collect::<Vec<_>>();
+
+    let mut deps = Vec::new();
+    let mut current_dep = String::new();
+
+    let mut iter = line[pos + 2..].chars().peekable();
+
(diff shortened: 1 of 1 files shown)


reitermarkus:
I found this:

https://github.com/rust-lang/cargo/blob/e07088baceb70e570977cc6d01b53c492b398f0a/src/cargo/core/compiler/fingerprint.rs#L833-L857

reitermarkus:
As it is now, it's not really reusable since it takes a path instead of a string and it is private.

luser:
This gets kind of hairy to follow. 😕 We've already got a dependency on the `regex` crate, what if we rewrote this as a regex instead? Maybe that's just trading one kind of complexity for another, I don't know. I fiddled around a bit and came up with this:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e712155a55925fb7cf87d6a67d150d45

reitermarkus:
> Maybe that's just trading one kind of complexity for another

I'll leave that decision up to you, but your current regex solution still has an error. The end result should be `"foo bar"` instead of `"foo\\ bar"`. (Should be as simple as adding a `.replace("\\ ", " ")`.)

reitermarkus:
What do you want me to do here?

luser:
My apologies, I forgot to come back to this. I don't think my regex solution is really any better (especially given the issue you pointed out!) so let's just land your changes since they work.

PR #347, opened 2018-12-27
Sort reasons by count when printing non-cacheable reasons.

### LINKED ISSUE #324: Sort the "Non-cacheable reasons" lines in the stats output
When I'm monitoring the stats output of sccache with `watch sccache -s` the "Non-cacheable reasons" lines in the output are displayed in a random order, they jump around the place which is rather distracting. It would be nice if this list would be sorted for example by the count column.

chmanchester:
Tested in a local build of Firefox. Closes #324


luser:
Thanks!

PR #349, opened 2019-01-03
Add support for '-mllvm' flag

hferreiro:
A normal chromium build has the following parameters: ` -Xclang -mllvm -Xclang -instcombine-lower-dbg-declare=0`, which aren't currently supported by sccache.


luser:
Thanks!

PR #353, opened 2019-01-11
Record cache hits, misses, and errors per language kind, closes #327

### LINKED ISSUE #327: Split compilation stats out by language
Currently when compiling something like Firefox which is a mix of Rust and C/C++ the stats aren't as useful as I'd like them to be:
```
18:51:51     INFO -  Compile requests              4151
18:51:51     INFO -  Compile requests executed     4042
18:51:51     INFO -  Cache hits                     413
18:51:51     INFO -  Cache misses                  3629
```

There are 3000+ C/C++ compilations in there and some number of Rust compilations, and it's very easy for the Rust compilation to get lost in the noise. Similarly, for Firefox CI we calculate a cache hit rate by taking "cache hits / compile requests executed", and if we break something so that sccache refuses to cache any C++ compilation (which is not that hard to do) it can still report a 99% cache hit rate because it is only taking Rust compiles into account.

I think it might be more useful to break this out by language. Lumping all of C/C++/ObjC into one bucket should be fine, and all Rust into the other should be fine. We don't currently support anything else, but if we do someday presumably that language would also benefit from that.

chmanchester:
This splits out the `cache_hits`, `cache_misses`, and `cache_errors` metrics
to be recorded per language "kind" (Rust or C/C++ at this point), the
printed output will look like this:

Cache hits                              236
Cache hits (Rust)                       156
Cache hits (C/C++)                       80

Many of the other stats could be split by language kind if we find that useful,
but this will give us a better indication for situations where caching rust
fails altogether, for instance.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -76,6 +76,15 @@ pub enum CompilerKind {
     Rust,
 }
 
+impl CompilerKind {
+    pub fn lang_kind(&self) -> String {
+        match self {
+            CompilerKind::C(_) => "C/C++",
+            CompilerKind::Rust => "Rust",
+        }.to_string()
+    }
+}
+
 /// An interface to a compiler for argument parsing.
 pub trait Compiler<T>: Send + 'static
 where
--- a/src/server.rs
+++ b/src/server.rs
@@ -17,8 +17,8 @@
 
 use cache::{storage_from_config, Storage};
 use compiler::{
-    get_compiler_info, CacheControl, CompileResult, Compiler, CompilerArguments, CompilerHasher,
-    DistType, MissType,
+    get_compiler_info, CacheControl, CompileResult, Compiler, CompilerKind,
+    CompilerArguments, CompilerHasher, DistType, MissType,
 };
 #[cfg(feature = "dist-client")]
 use config;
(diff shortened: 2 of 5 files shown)


chmanchester:
I have an update to the mozharness code that consumes these here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=88f85adceb06ee982b324efab80204d3d42a120c

luser:
This looks pretty good, I just have one thought--if we want to calculate cache hit rates per language then we'd also need to split out `requests_executed` per-language, right? Having this implemented as-is should be enough to let us sanity check changes by looking at the stats, but it might be nice to also start recording cache hit metrics per-language so we'd have notifications if we drop one to zero.

After writing that I realize that that could be done as a follow-up so I'm going to land this as-is.

src/server.rs
```diff
@@ -963,6 +965,36 @@ where
...
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct PerLanguageCount {
+    counts: HashMap<String, u64>,
+}
+
+impl PerLanguageCount {
+    fn increment(&mut self, kind: &CompilerKind) {
+        let key = kind.lang_kind().clone();
+        let count = match self.counts.get(&key) {
```
luser: FYI, `HashMap` has a great API for exactly this use case:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry

src/compiler/compiler.rs
```diff
@@ -76,6 +76,15 @@ pub enum CompilerKind {
     Rust,
 }
 
+impl CompilerKind {
+    pub fn lang_kind(&self) -> String {
```
luser: I would probably just make this return `&'static str` and have the caller deal with turning it into a `String` (you're already calling `clone` on it there anyway).

luser:
Thanks for fixing this!

PR #361, opened 2019-01-23
Mark -fprofile-use=path as TooHard

mshal:
The '-fprofile-use' flag without an argument is already TooHard, but
there is also a '-fprofile-use=path' variant that should be treated the
same.


luser:
Thanks!

PR #366, opened 2019-02-01
Use trim_end instead of trim_right as the latter is being deprecated.

SingingTree:
Rust 1.33 is deprecating trim_right in favour of trim_end. This patch
replaces instances of the trim_right with trim_end to avoid warnings.
Note, trim_end was introduced in 1.30, so this would bump the required
stable version (1.27 as of writing).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ matrix:
     - os: osx
 
     # rustc version compat
-    - rust: 1.27.0 # oldest supported version, keep in sync with README.md
-    - rust: 1.27.0
+    - rust: 1.30.0 # oldest supported version, keep in sync with README.md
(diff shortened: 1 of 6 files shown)


SingingTree:
As an aside, because `Error::cause` is used in by the macro [here](https://github.com/mozilla/sccache/blob/05bd9c7944d8c7c6e11f994c8cb5246f7638d887/src/errors.rs#L45-L74) rather than `Error::source`, a warning is also issued. But my understanding is that would need to be fixed in error_chain.

luser:
This looks good, the CI failures are just the 1.27 builds failing. I've added a change to bump the minimum Rust version to 1.30 which should fix them and then I'll merge this.

PR #367, opened 2019-02-05
Remove openssl dependency for GCS

Jake-Shadle:
This is a slight reimplementation of #272 that I think works slightly better.

* It doesn't add a dependency on `pem` which seems like overkill for just parsing a key out, at least if it's only going to be used for GCS specifically. This just uses a simple string split since the key format is not complicated.
* I opened a PR here https://github.com/Keats/jsonwebtoken/pull/74 to enable the key to be passed "as-is" rather than converting it to DER. It looks like a https://github.com/Keats/jsonwebtoken/issues/76 of jsonwebtoken coming, which might remove this API ambiguity, but for now pointing at the fork allows GCS to function properly, without OpenSSL.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,8 +116,6 @@ tokio-reactor = "0.1"
 [features]
 default = ["s3"]
 all = ["redis", "s3", "memcached", "gcs", "azure"]
-# gcs requires openssl, which is a pain on Windows.
-all-windows = ["redis", "s3", "memcached", "azure"]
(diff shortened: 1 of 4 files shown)


Jake-Shadle:
Went ahead and removed the now superfluous `all-windows` feature flag.

luser:
Thanks for the patch! I'll try to take a look at this today.

luser:
OK, I'm going to land this as-is. I don't love having crates.io patches because they complicate pushing new releases, but hopefully the jsonwebtoken API situation will get sorted out and we can switch back to an upstream release soon.

PR #368, opened 2019-02-06
Take -fplugin into account as an extra hash file. Fixes #288

### LINKED ISSUE #288: Handle clang plugins somehow
We use a clang plugin to do static analysis in Firefox CI, and @glandium noted that sccache does not currently handle clang plugins at all. The simplest thing I could think to do here would be to include the contents of the plugin file as part of the hash. That would only work well if the plugin is built in a reproducible manner, of course.

Annoyingly, the traditional way of loading a clang plugin is to pass it down to `cc1` [with a bunch of `-Xclang` options](https://clang.llvm.org/docs/ClangPlugins.html#using-the-cc1-command-line) and of course [that's what we use in the Firefox build](https://dxr.mozilla.org/mozilla-central/rev/9c13dbdf4cc9baf98881b4e2374363587fb017b7/build/autoconf/clang-plugin.m4#160). [The clang docs do say](https://clang.llvm.org/docs/ClangPlugins.html#using-the-clang-command-line) that you can use `-fplugin=plugin` now, and supporting that sounds easier, so maybe we can fix the Firefox build to use that and just support that?

chmanchester:


luser:
Thanks!

PR #371, opened 2019-02-06
Add scoop install, enhance Storage Options

Jake-Shadle:
* I've added `sccache` to the [main scoop bucket](https://github.com/lukesampson/scoop/blob/master/bucket/sccache.json) so noted that as an additional install option for Windows
* Enhanced the storage options to be easily clickable and I hope read better for people skimming the README.


luser:
Thanks!

PR #372, opened 2019-02-11
Fix checksum files

Jake-Shadle:
All of the .sha256 files had trailing newlines, which are now removed. Note that the chksum files generated with shasum are just the raw bytes of the checksum, not the ASCII hex like for Windows, but getting the same behavior seems to require using sha256sum instead of shasum, which isn't available by default on OSX, so I figured it was ok to just use the raw bytes on Linux and OSX.


luser:
Works for me.

Issue #350, opened 2019-01-03
Add support for clang plugin parameters


hferreiro:
Support for clang plugins was handled in #332 but, support for `-plugin-arg-<plugin-name>` is missing. An example invocation from a chromium build is the following:

```
-Xclang -load -Xclang ../../third_party/llvm-build/Release+Asserts/lib/libFindBadConstructs.so -Xclang -add-plugin -Xclang find-bad-constructs -Xclang -plugin-arg-find-bad-constructs -Xclang check-ipc
```

luser:
This should be easy enough to support, we should just need to add it to the list of supported clang arguments:
https://github.com/mozilla/sccache/blob/12008f60bc8b5335e8e8fa878e3bdc2749e50756/src/compiler/clang.rs#L88

cc @chmanchester

barrbrain:
For chromium clang, there are currently just 2 plugins for which arguments are passed in the build configuration:
1. `-plugin-arg-find-bad-constructs` at [build/config/clang/BUILD.gn:23](https://chromium.googlesource.com/chromium/src/+/master/build/config/clang/BUILD.gn#23)
2. `-plugin-arg-blink-gc-plugin` at [third_party/blink/renderer/BUILD.gn:123](https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/BUILD.gn#123)

Adding rules for these is sufficient to have a fully cacheable build when setting `clang_use_chrome_plugins = true`.

PR #374, opened 2019-02-13
Add support for '-fdebug-compilation-dir'

hferreiro:


luser:
Hi, thanks for the patch! First of all, the CI failures are due to:
> thread '<unnamed>' panicked at '-fprofile-instr-generate can't precede -fdebug-compilation-dir', src/compiler/args.rs:519:13

The arguments in the list need to be sorted, and the new argument is not in the right spot.

However, I have a more fundamental question--given that you're not adding code to do anything with this argument, what does this buy us? We should already handle unknown arguments, they just get passed along to the compiler.  The list is for arguments that need special handling.

hferreiro:
Without this patch I'm getting the following in the logs:
```
parse_arguments: CannotCache(Can't handle UnknownFlag arguments with -Xclang)
```
The same happens with https://github.com/mozilla/sccache/pull/349 and https://github.com/mozilla/sccache/issues/350.

Both `icecc` and `ccache` have no extra code for handling `Xclang` parameters, besides `-load`. What is different in `sccache`?

hferreiro:
Hey @luser, can you provide some input on my last comment?

luser:
> Both icecc and ccache have no extra code for handling Xclang parameters, besides -load. What is different in sccache?

Hm, I'm not actually sure why things are implemented this way. I know we have two-pass parsing to handle the set of `-Xclang` parameters (most specially for clang-cl where they're totally different). I'm going to merge this PR, feel free to open an issue on handling this more gracefully in the general case.

Issue #189, opened 2017-10-12
Working with CMake (failed to store ... no such file or directory)


rrichardson:
I am attempting to build a Rust project while using sccache only as a replacement for ccache for CC and CXX. 
The project I am attempting to build is the Rust `ssh2` crate. 

My first approach was to set CC="sccache gcc" and CXX="sccache g++" . 

However, cmake split the CC string into 'sccache' as the compiler and 'gcc' as a compiler flag.  (Also CC="'sccache gcc'" didn't work as well)

So I created bash files : 
sccache-cc
```
#!/bin/bash
sccache gcc "$@"
``` 
and set CC="sccache-cc"
this came from this tip for making ccache work with CMake [here](https://stackoverflow.com/questions/1815688/how-to-use-ccache-with-cmake) 

This got much farther, however it appears that CMake expected gcc to do something different with the .o file than was expected by sccache, because the result is : 

```
sccache: encountered fatal error
sccache: error : failed to store `x25519-x86_64.o` to cache
sccache:  cause: failed to store `x25519-x86_64.o` to cache
sccache:  cause: failed to zip up compiler outputs
sccache:  cause: No such file or directory (os error 2)
thread '<unnamed>' panicked at 'execution failed', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.11.0/build.rs:658:8
```

Has anyone encountered an error like this?  Better question : Has anyone convinced ccache to work nicely with CMake? 

Thanks.

rrichardson:
btw, the offending compilation step outputted by sccache is : 

```
running "/usr/local/bin/sccache-cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-m64" "-I" "include" "-std=c1x" "-Wbad-function-cast" "-Wmissing-prototypes" "-Wnested-externs" "-Wstrict-prototypes" "-fdata-sections" "-ffunction-sections" "-pedantic" "-pedantic-errors" "-Wall" "-Wextra" "-Wcast-align" "-Wcast-q
[...]

PR #376, opened 2019-02-15
Fix `-o` flag separator type

Jake-Shadle:
When I tried to enable sccache for native compilation in addition to rust I ran into errors (the same kind as #189), I found that `ring` was causing sccache to fail because it's a bit more involved in the use of `cc` and when specifying the output object file is just directly concatenating the flag and the arg, so the command ends up being

```
$ "scccache" "gcc" blah blah blah "-c" "-o/somepathto.o" "somepathto.c"
```

Just changing the type of separator to support both `-ofile` or `-o file` allowed sccache to compile `ring` successfully.

Resolves: #189


luser:
Gosh, compiler options are weird, aren't they?

PR #378, opened 2019-02-20
Add basic support for response files

Jake-Shadle:
Currently, the gcc and clang path will attempt to use response files (@<path>) if there are no quotes inside the file, and the MSVC path will just bail immediately, this PR attempts to provide a best effort parsing of response files. I am fairly(ish) confident of the gcc/clang part, as I was able to compile the entirety of UE4, which uses response files almost exclusively, but I'm less confident of the MSVC one because...Windows gonna Windows.


Jake-Shadle:
I just realized Windows is completely broken because of `\` separators, fixing it now.

jasonwhite:
@Jake-Shadle You may want to take a look at #192 for correct MSVC response file handling. Unfortunately, that PR never had much attention from the maintainers and was left to rot. Hopefully that doesn't happen here as well.

Jake-Shadle:
Damn, I could have sworn I searched for issues/PRs related to this, but maybe I forgot. Will take a look when I can!

PR #387, opened 2019-03-01
Sort stats by language name when printing them out. Fixes #386

### LINKED ISSUE #386: Per-language stats are printed in random order
The per-language statistics are gathered from the stats HashMap using the standard iterator which returns items in arbitrary order. Because of this every invocation of `sccache --show-stats` will print out the per-language stats with lines in different order.

gabrielesvelto:


src/server.rs
```diff
@@ -1094,7 +1094,9 @@ impl ServerStats {
         macro_rules! set_lang_stat {
             ($vec:ident, $var:expr, $name:expr) => {{
                 $vec.push(($name.to_string(), $var.all().to_string(), 0));
-                for (lang, count) in $var.counts.iter() {
+                let mut sorted_stats: Vec<_> = $var.counts.iter().collect();
+                sorted_stats.sort_by(|a, b| b.1.cmp(a.1));
```
luser:
```suggestion
                sorted_stats.sort_by_key(|v| v.1);
```

gabrielesvelto: Ah, sweet! I'll update the PR.

luser:
Thanks!

PR #392, opened 2019-03-04
Daemonize the scheduler process unless SCCACHE_NO_DAEMON is set in the environment. Fixes #382

### LINKED ISSUE #382: Make sccache scheduler able to daemonize
Currently the sccache scheduler runs in a purely foreground mode. Mozilla Ops has requested that it be able to daemonize itself so they don't have to maintain some custom code to daemonize it. We have existing code in sccache for daemonizing the local server that runs when using the client, so it shouldn't be much work to use that same (or similar) code in the scheduler.

chmanchester:
Fixes #382

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -19,6 +19,7 @@ use futures_cpupool::CpuPool;
 use mock_command::{CommandChild, RunCommand};
 use ring::digest::{Context, SHA512};
 use serde::Serialize;
+use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs::File;
 use std::hash::Hasher;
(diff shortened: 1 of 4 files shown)


luser:
Looks like this needs a tweak to build on Windows.

src/util.rs
```diff
@@ -431,6 +432,96 @@ mod http_extension {
...
+                libc::SIGILL => libc::sigaction(signum, &*PREV_SIGILL, 0 as *mut _),
+                _ => libc::sigaction(signum, &*PREV_SIGSEGV, 0 as *mut _),
+            };
+        }
+    }
+}
+
+/// This is a no-op on Windows.
+#[cfg(windows)]
+fn daemonize() -> Result<()> {
```
luser:
```suggestion
pub fn daemonize() -> Result<()> {
```

luser:
Thanks!

PR #403, opened 2019-04-03
Build the distributed compilation client support by default. Fixes #383

### LINKED ISSUE #383: Enable dist-client in releases
We don't currently enable dist-client in releases, but we should. That would make using the distributed compilation feature easier for users.

jwatt:
I'm not sure exactly why @luser phrased #383 as enabling the client support in "releases", rather than "by default". It seems more consistent to me to simply built it in all types of builds, release or otherwise, since it seems to build on Linux, macOS and Windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,8 +115,8 @@ tokio-named-pipes = "0.1"
 tokio-reactor = "0.1"
 
 [features]
-default = ["s3"]
-all = ["redis", "s3", "memcached", "gcs", "azure"]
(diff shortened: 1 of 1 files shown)


chmanchester:
I agree we should do this by default, it will save a slightly confusing step for people as we roll out sccache more widely. I'm able to reproduce that CI failure locally, I'll look into that before merging.

chmanchester:
#404 should fix the test issue.

jwatt:
Thanks @chmanchester

PR #418, opened 2019-04-29
fix build warnings of rustc 1.36.0-nightly

matthiaskrgr:


src/dist/pkg.rs
```diff
@@ -312,7 +311,7 @@ pub fn make_tar_header(src: &Path, dest: &str) -> io::Result<tar::Header> {
 
     // tar-rs imposes that `set_path` takes a relative path
     assert!(dest.starts_with("/"));
-    let dest = dest.trim_left_matches("/");
+    let dest = dest.trim_start_matches("/");
```
froydnj: Just for completeness: `trim_start_matches`, according to the docs, exists since 1.30.0, and `sccache`'s minimum supported Rust version is 1.30.0, so we don't have to have a debate about whether to bump the minimum supported version here.

PR #419, opened 2019-04-30
Various fixes for sccache-dist

chmanchester:


src/bin/sccache-dist/main.rs
```diff
@@ -556,7 +565,14 @@ impl SchedulerIncoming for Scheduler {
                         "Non-terminated job {} was cleaned up in server pruning",
                         job_id
                     );
-                    assert!(jobs.remove(&job_id).is_some())
+                    // A job may be missing here if it failed to allocate
+                    // initially, so just warn it's not present.
```
froydnj: Nit: "...warn if it's not present"

src/bin/sccache-dist/main.rs
```diff
@@ -556,7 +565,14 @@ impl SchedulerIncoming for Scheduler {
...
+                    if !jobs.remove(&job_id).is_some() {
```
froydnj: Nit: Maybe this is better expressed as the positive `if jobs.remove(...).is_none() {`, which is a little more in line with the comment above?

src/dist/http.rs
```diff
@@ -670,10 +670,21 @@ mod server {
...
                 ($request:ident) => {{
                     match bearer_http_auth($request).and_then(&*check_server_auth) {
-                        Some(server_id) if server_id.addr().ip() == $request.remote_addr().ip() => {
-                            server_id
-                        }
-                        Some(_) => return make_401("invalid_bearer_token_mismatched_address"),
+                        Some(server_id) => {
+                            let origin_ip = if let Some(header_val) = $request.header("X-Real-IP") {
+                                trace!("X-Real-IP: {:?}", header_val);
+                                header_val.parse().unwrap()
```
froydnj: I think it'd be more robust to handle the error case here rather than blindly calling `unwrap()`.  I'm not sure whether you can `header_val.parse()?` or something more verbose is required.

PR #427, opened 2019-05-16
Bug 1544873: Implement SCCACHE_GCS_CREDENTIALS_URL

walac:
This environment variable contains a URL used to fetch the
GCS access token. Its main use is for fetching temporary
credentials from the gcsCredentials [1] endpoint in the
taskcluster-auth service.

[1] https://github.com/taskcluster/taskcluster/pull/640


froydnj:
@chmanchester, are you able to review this?

src/config.rs
```diff
@@ -419,6 +420,7 @@ fn config_from_env() -> EnvConfig {
         .map(|bucket| {
             let cred_path = env::var_os("SCCACHE_GCS_KEY_PATH")
                 .map(|p| PathBuf::from(p));
+            let url = env::var("SCCACHE_GCS_CREDENTIALS_URL").ok();
```
chmanchester: This seems a little weird, because we're honoring the path ahead of the URL if both are set. Can we prevent both being set here or at least warn if they both are? The code to consume things would be a bit neater if `GCSCacheConfig` had just one of a url or path in it as well.

walac: iiuc, the role of config is to grab all available configs set and let the consumer handle what to do with them, that's why I didn't made `cred_path` and `url` mutually exclusive. I added the warning.

src/config.rs
```diff
@@ -417,8 +418,15 @@ fn config_from_env() -> EnvConfig {
 
     let gcs = env::var("SCCACHE_GCS_BUCKET").ok()
         .map(|bucket| {
+            let url = env::var("SCCACHE_GCS_CREDENTIALS_URL").ok();
             let cred_path = env::var_os("SCCACHE_GCS_KEY_PATH")
                 .map(|p| PathBuf::from(p));
+
+            if url.is_some() && cred_path.is_some() {
+                warn!("Both SCCACHE_GCS_CREDENTIALS_URL and SCCACHE_GCS_KEY_PATH are set");
+                warn!("You should set only one of them!");
```
chmanchester: Let's mention that the path will take precedence over the url here.

walac: Done

walac:
@chmanchester ok to merge?

PR #431, opened 2019-05-22
Add issuer and subject fields to certs generated by the server.

### LINKED ISSUE #406: sccache client on macOS failing to distribute compilation
I'm having problems getting sccache on my Mac to distribute compilation to my Linux workstation.

Given both machines are on my home network, I have the scheduler set up to use simple HTTP (not HTTPS), and the client and scheduler appear to be communicating fine. Specifically my client config file basically contains:

```
[dist]
scheduler_url = "http://my-workstation.localdomain:10600"
toolchain_cache_size = 1073741824

[[dist.toolchains]]
type = "path_override"
compiler_executable = "/usr/local/opt/llvm/bin/clang++"
archive = "/Users/me/toolchains/clang-8.0.0-x86_64.tar.gz"
#archive = "/Users/me/toolchains/clang.tar.xz"
archive_compiler_executable = "/clang-8.0.0-x86_64/bin/clang++"

[dist.auth]
type = "token"
# This should match the `client_auth` section of the scheduler config.
token = "my client token"
```

Running the client sccache server on my Mac via:

```
RUST_LOG=trace SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache
```

I see very little of use in the log output from sccache/reqwest/hyper etc. to help debug why toolchain upload to the workstation machine isn't working. The only bit of useful output is:

WARN 2019-04-04T04:37:29Z: sccache::compiler::compiler: [conftest.E2EDOm.o]: Could not perform distributed compile, falling back to local: Could not submit toolchain: Could not submit toolchain: `https://<build-server-ip-address>:10501/api/v1/distserver/submit_toolchain/5`: connection closed

I tried digging into the code using a debugger but there are a lot of layers of async indirection through the various modules which made that non-trivial and time consuming. It appears that the connection is already closed on the tx object before the body is even dispatched though, which is weird.

Another thing I discovered is that if I changed the sccache client code to dispatch a string instead of the toolchain file by changing [this line](https://github.com/mozilla/sccache/blob/9b04c5857bc9/src/dist/http.rs#L1185), then the log message changes from "connection closed" to "The certificate was not trusted":

WARN 2019-04-04T04:04:55Z: sccache::compiler::compiler: [conftest.Ius_Ek.o]: Could not perform distributed compile, falling back to local: Could not submit toolchain: Could not submit toolchain: `https://<build-server-ip-address>/api/v1/distserver/submit_toolchain/65`: an error occurred trying to connect: The certificate was not trusted.: The certificate was not trusted.

I haven't looked into the code for passing the certificate generated by the build server from the scheduler to the client, but it looks like that isn't working for me.

chmanchester:
Fixes #406.

I'm fairly sure this is the right thing to do here, and doesn't change any significant security properties of this part of the system, but I'm looking for a sanity check from someone with a bit more expertise in this area.

### DIFF
--- a/src/dist/http.rs
+++ b/src/dist/http.rs
@@ -344,6 +344,16 @@ mod server {
             .set_pubkey(privkey.as_ref())
             .chain_err(|| "failed to set pubkey for x509")?;
 
+        let mut name = openssl::x509::X509Name::builder()?;
+        name.append_entry_by_nid(openssl::nid::Nid::COMMONNAME,
+                                 "sccache-internal")?;
+        let name = name.build();
+
+        builder.set_subject_name(&name)
+            .chain_err(|| "failed to set subject name")?;
+        builder.set_issuer_name(&name)
+            .chain_err(|| "failed to set issuer name")?;
+
         // Add the SubjectAlternativeName
         let extension = openssl::x509::extension::SubjectAlternativeName::new()
             .ip(&addr.ip().to_string())


chmanchester:
That macOS timeout looks random, this builds and passes tests locally.

src/dist/http.rs
```diff
@@ -344,6 +344,16 @@ mod server {
             .set_pubkey(privkey.as_ref())
             .chain_err(|| "failed to set pubkey for x509")?;
 
+        let mut name = openssl::x509::X509Name::builder().unwrap();
```
froydnj: Can we do something better than `unwrap()` here?

froydnj:
The code as such looks fine to me; we are still waiting for input from a certificate expert on whether this is the right way to do things.

chmanchester:
> The code as such looks fine to me; we are still waiting for input from a certificate expert on whether this is the right way to do things.

We got the go ahead on this.

PR #449, opened 2019-06-12
Fix authorize url to correctly include "profile" scope.

chmanchester:


chmanchester:
@gdestuynder, can you sanity check this is the right thing for a consumer trying to check groups after https://github.com/mozilla-iam/auth0-deploy/commit/9b4b081c55db7a5602284d989964040f0ec8f681 ?

gdestuynder:
@chmanchester yes that's correct. Sorry, I didn't have you in my list to notify of this change but I'm glad you found out ;-)

Some additional info:

We basically respect the spec correctly now while we didn't before. The `profile` scope gives access to the group data and all other fields which can be had either in the id token or in the user profile endpoint (the later is recommended, we pass group data in the id token and it follows spec but it's potentially dangerous as the id token has a maximum size which depends on the web server and web browser configuration. in practice we don't currently go over the "well known limits" but someone who would have all groups set definitely would)

chmanchester:
Thanks!

Issue #390, opened 2019-03-04
Add a command to the sccache client to query distributed status


luser:
With an eye towards making the Firefox build system able to use distributed builds in a seamless way it would be nice if sccache exposed a command to query the status of its distributed compilation configuration. I'm imagining something like `sccache dist-status` which could return information about possible states:

1. Distributed compilation is disabled, either because sccache wasn't built with the `dist-client` feature or because it's not configured in the client's config file. (It might be worthwhile to report these as separate states to help users troubleshoot.)
2. Distributed compilation is enabled but we don't have a working connection to the scheduler. We should report the reason as well. This could be because the scheduler is not reachable or because authentication isn't configured properly.
3. Distributed compilation is enabled, configured, and working. Ideally the output here would also include the number of CPU cores available to the scheduler from all build servers. The scheduler doesn't currently expose that info, although [it does have a reported CPU count for each server in its list of servers internally](https://github.com/mozilla/sccache/blob/931cefa86987d538caf503839dd5da6c540368a9/src/bin/sccache-dist/main.rs#L467). This would be useful so the build could set the right amount of parallelism automatically without requiring developers to manually pass `-j100` or other arbitrary values. We could change the scheduler's `/status` API to include a `num_cores` alongside [the `num_servers` value](https://github.com/mozilla/sccache/blob/931cefa86987d538caf503839dd5da6c540368a9/src/bin/sccache-dist/main.rs#L683) it currently reports. The [`SchedulerStatusResult` struct is defined here](https://github.com/mozilla/sccache/blob/931cefa86987d538caf503839dd5da6c540368a9/src/dist/mod.rs#L520-L524).

[...]

PR #450, opened 2019-06-13
Add a field to the scheduler status result for the total number of cpus available across servers

chmanchester:


chmanchester:
This is a quick and dirty way to help people figure out how much concurrency the cluster supports. I'd like to incorporate this into a `dist-status` that the local daemon can query going forward.

luser:
Related: #390

PR #461, opened 2019-06-24
Add extra note about optional features during install

JeanMertz:
It took me an hour or two to debug what was going on here.

I was installing sccache on Travis, and could see the following in the logs:

```
DEBUG Trying GCS bucket(my-bucket, Some("/tmp/sccache-auth.json"), ReadWrite)
 INFO No configured caches successful, falling back to default
```

I couldn't figure out why it didn't work, and I spent some time re-creating access tokens, changing the access rights, etc. This of course took quite some time, as I had to wait for a CI run every time (I should've tested it locally, but I didn't).

It was extra confusing because I could see it was trying to use GCS, but it never printed any warnings about failing, even though the code does emit a warning if it fails.

Eventually, I found out that the `Trying GCS bucket` line is _before_ the feature check attribute, and so even it said it was trying, it never actually did:

https://github.com/mozilla/sccache/blob/6e3295a22283d6143859c8838f583cb37c176e03/src/cache/cache.rs#L185-L189

The README only mentions the feature flags in the build section, not in the install section. I kind of skipped over the build section, as I wasn't interested in building it, I just wanted to install the binary. In hindsight, it's also a bit of RTFM, but I hope this extra line saves others some time in the future 😄.

I guess another solution (or added solution) is to move that log line to after the cfg attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ set OPENSSL_LIBS=libcrypto64MT:libssl64MT
 
 `$ cargo install sccache`
 
+<sub>_use `--features=all` for non-local caching, see [build requirements](#build-requirements)_</sub>
+
 ### Windows
 
 sccache can also be installed via [scoop](https://scoop.sh/)


README.md
```diff
@@ -107,6 +107,8 @@ set OPENSSL_LIBS=libcrypto64MT:libssl64MT
 
 `$ cargo install sccache`
 
+<sub>_use `--features=all` for non-local caching, see [build requirements](#build-requirements)_</sub>
```
flxo: This is actually not completely correct since `S3` is in the list of [default features](https://github.com/mozilla/sccache/blob/5112041e0b1f95467e1a4da758eb2533bfb00d58/Cargo.toml#L119) and `non-local`.
The list of optional features in the `Build` section is outdated - think `memcached` is missing here - and this PR cries for adding that ;-)

luser:
Maybe we should just add all storage backends to the default features list? The only downside is that they drag in a bunch more dependencies so build times will go up, but people that are concerned about that can use `--no-default-features --features=...`. That's probably the right tradeoff vs. people building sccache and not having the storage backend they actually want.

flxo:
I just diffed the output of `cargo tree` with and without `--features all` and think enabling all backends by default do no harm. The only suspicious thing is `ring`.

```diff
> ├── jsonwebtoken v5.0.1 (https://github.com/Jake-Shadle/jsonwebtoken.git?rev=2f469a61#2f469a61)
> │   ├── base64 v0.9.3 (*)
> │   ├── chrono v0.4.6 (*)
> │   ├── ring v0.13.5
> │   │   └── untrusted v0.6.2
> │   │   [build-dependencies]
> │   │   └── cc v1.0.25 (*)
> │   ├── serde v1.0.80 (*)
> │   ├── serde_derive v1.0.80 (*)
> │   ├── serde_json v1.0.33
> │   │   ├── itoa v0.4.3 (*)
> │   │   ├── ryu v0.2.7
> │   │   └── serde v1.0.80 (*)
> │   └── untrusted v0.6.2 (*)
304a319,330
> ├── memcached-rs v0.3.0
> │   ├── bufstream v0.1.4
> │   ├── byteorder v1.2.7 (*)
> │   ├── conhash v0.4.0
> │   │   ├── log v0.4.6 (*)
> │   │   └── md5 v0.3.8
> │   ├── log v0.4.6 (*)
> │   ├── rand v0.5.5 (*)
> │   ├── semver v0.9.0 (*)
> │   └── unix_socket v0.5.0
> │       ├── cfg-if v0.1.6 (*)
> │       └── libc v0.2.44 (*)
308a335,356
> ├── redis v0.9.1
> │   ├── bytes v0.4.11 (*)
> │   ├── combine v3.6.3
> │   │   ├── ascii v0.9.1
> │   │   ├── byteorder v1.2.7 (*)
> │   │   ├── either v1.5.0
> │   │   ├── memchr v2.1.1 (*)
> │   │   └── unreachable v1.0.0 (*)
> │   ├── futures v0.1.25 (*)
> │   ├── sha1 v0.6.0
> │   ├── tokio-codec v0.1.1 (*)
> │   ├── tokio-executor v0.1.5 (*)
> │   ├── tokio-io v0.1.10 (*)
> │   ├── tokio-tcp v0.1.2 (*)
> │   └── url v1.7.2
> │       ├── idna v0.1.5
> │       │   ├── matches v0.1.8
> │       │   ├── unicode-bidi v0.3.4
> │       │   │   └── matches v0.1.8 (*)
> │       │   └── unicode-normalization v0.1.7
> │       ├── matches v0.1.8 (*)
> │       └── percent-encoding v1.0.1 (*)
366,369c414
< │   ├── serde_json v1.0.33
< │   │   ├── itoa v0.4.3 (*)
< │   │   ├── ryu v0.2.7
< │   │   └── serde v1.0.80 (*)
---
> │   ├── serde_json v1.0.33 (*)
374,381c419
< │   │   └── url v1.7.2
< │   │       ├── idna v0.1.5
< │   │       │   ├── matches v0.1.8
< │   │       │   ├── unicode-bidi v0.3.4
< │   │       │   │   └── matches v0.1.8 (*)
< │   │       │   └── unicode-normalization v0.1.7
< │   │       ├── matches v0.1.8 (*)
< │   │       └── percent-encoding v1.0.1 (*)
---
> │   │   └── url v1.7.2 (*)
391,394c429
< ├── ring v0.13.5
< │   └── untrusted v0.6.2
< │   [build-dependencies]
< │   └── cc v1.0.25 (*)
---
> ├── ring v0.13.5 (*)
507c542
< │   └── either v1.5.0
---
> │   └── either v1.5.0 (*)
```

glandium:
This was addressed in 194a25f344fc71d781974a715a7a1f51fcd7a099, although we may still want to do https://github.com/mozilla/sccache/pull/461#issuecomment-505058063

PR #465, opened 2019-06-25
Expose LruCache and Meter from lru-disk-cache

chmanchester:


chmanchester:
@luser, it looks like you'll need to publish a new version of `lru-disk-cache` with this once this is merged. If you prefer you can make me an owner on crates.io. Thanks!

luser:
> @luser, it looks like you'll need to publish a new version of `lru-disk-cache` with this once this is merged. If you prefer you can make me an owner on crates.io. Thanks!

I've added you as an owner, have at it.

PR #337, opened 2018-11-26
 Add authentication to S3 GET requests 

Marwes:
This is configurable through the `SCCACHE_S3_GET_AUTH` environment variable or the `get_auth` in the S3 config. When `"true"`, authentication will be performed in the same manner as for PUT requests, allowing completely private buckets to be used.

Based on and closes #135

PR #470, opened 2019-07-04
feat: Update all dependencies to the latest version

Marwes:
Most importantly this fixes an issue where sccache would not build if openssl 1.1.1 are installed (it was too recent).

PR #471, opened 2019-07-05
 fix: Correct the GET authentication for S3 using rusoto

Marwes:
For some unknown reason authenticated GET accesses fails under certain
configurations. So far I have been unable to determine the exact cause
but after adding `rusoto_s3` to deal with all communication to S3 it
fixed itself so at least for my case this would be a good fix.

I wasn't sure exactly what to do about the config file format as that probably needs to change to better fit what rusoto accepts. At least it should be expanded to support reading in the region which currently only seems to work through an environment variable.

Based on #470 since that fixes some build errors for me.

Closes #337


Marwes:
Rebased and moved up contained all accidental formatting into the first commit (see #502 ). https://github.com/mozilla/sccache/pull/471/files/dfe5f403dde8f56c24dd2a03acafa6cfaa378dcc..6cf2ac66baba83ebc28a97ba0e24820bf91e634d gives the actual changes which should be much simpler to review.

scooter-dangle:
@Marwes, have you tried to reproduce the issue after #490 was merged? After that change, I no longer see the S3 authentication issue that this PR solved.

Marwes:
Seems to work with #490 . Either way this has bit rotted to the point that another, fresher PR has taken its place https://github.com/mozilla/sccache/pull/522

PR #475, opened 2019-07-09
Try removing the cargo cache to make CI pass

chmanchester:


chmanchester:
Well, this seems to have sped things up as well as make CI pass.

PR #481, opened 2019-07-18
Add a version check for bubblewrap in the server. Fixes #316

### LINKED ISSUE #316: Add a bubblewrap version check in dist server startup
cc @aidanhs 

We need bubblewrap 0.3.0 or newer for its commandline handling. Ubuntu 18.04 ships 0.2.1, and using that produces a confusing error about `--` being an unsupported option. We should just version-check the supplied binary on startup and error if it's not new enough. `bwrap --version` works:
```
$ bwrap --version
bubblewrap 0.2.1
$ /build/bubblewrap-0.3.1/bwrap --version
bubblewrap 0.3.1
```

chmanchester:


froydnj:
WDYT about using `version-compare` for doing version checks, rather than all this manual parsing?

chmanchester:
> WDYT about using `version-compare` for doing version checks, rather than all this manual parsing?

I'm fine with it, but this seemed simpler at the time. I am having trouble finding a crate with that name...

chmanchester:
> > WDYT about using `version-compare` for doing version checks, rather than all this manual parsing?
> 
> I'm fine with it, but this seemed simpler at the time. I am having trouble finding a crate with that name...

Ok, I found it. Might be overkill but I'll take a look.

chmanchester:
@froydnj does this version seem preferable?

froydnj:
> @froydnj does this version seem preferable?

Huh, not that much shorter, but the logic seems easier to follow.  r+ from me either way.

Issue #477, opened 2019-07-11
Scheduler tracks capacity inaccurately in some cases


chmanchester:
I found after many builds that the cluster we have stopped taking jobs due to insufficient capacity. I was the only one using it at the time. According to the logs there were some cases we failed to assign a job. As far as I can tell in this case the scheduler loses track of the capacity of servers -- jobs are only removed from `jobs_assigned` when a job is registered as complete, not if assignment fails, which apparently can happen for servers intermittently.

Servers seem to have a similar issue, they don't register a job as complete if it fails, but I haven't seen evidence of that in practice.

PR #482, opened 2019-07-18
Notify the scheduler when a failed job completes. See #477

chmanchester:


src/bin/sccache-dist/main.rs
```diff
@@ -453,6 +453,7 @@ fn init_logging() {
 }
 
 const MAX_PER_CORE_LOAD: f64 = 10f64;
+const SERVER_ERROR_RESET: Duration = Duration::from_secs(300);
```
froydnj: This name makes me think of how often we reset (restart) servers, which is not the intended meaning.  Could we make this `SERVER_ERROR_REMEMBER_TIMEOUT` or something?  I'm not really sure what a good name is here that's not ridiculously long.

PR #486, opened 2019-07-31
Provide the failing process' stderr when failing to detect the compiler.

### LINKED ISSUE #485: Return a `Result` and error message from `detect_compiler`
Today if we complain about an unsupported compiler it sometimes mean we weren't able to run the rustc provided. This can maybe happen for a few reasons but https://github.com/mozilla/sccache/blob/6e3295a22283d6143859c8838f583cb37c176e03/src/compiler/compiler.rs#L854 means we don't provide the output of the compiler if it fails to run. That's fatal, so providing that output would be helpful.

chmanchester:
Fixes #485

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -827,7 +827,7 @@ fn detect_compiler<T>(
     executable: &Path,
     env: &[(OsString, OsString)],
     pool: &CpuPool,
-) -> SFuture<Option<Box<dyn Compiler<T>>>>
+) -> SFuture<Box<dyn Compiler<T>>>
 where
     T: CommandCreatorSync,
 {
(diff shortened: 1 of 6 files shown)


src/compiler/compiler.rs
```diff
@@ -847,11 +847,11 @@ where
             if output.status.success() {
                 if let Ok(stdout) = String::from_utf8(output.stdout) {
                     if stdout.starts_with("rustc ") {
-                        return Some(stdout);
+                        return Some(Ok(stdout));
                     }
                 }
             }
-            None
+            Some(Err(String::from_utf8_lossy(&output.stderr).into_owned()))
```
luser: FYI there's an existing [`run_input_output`](https://github.com/mozilla/sccache/blob/6ba6f6c15c106768b914a7697a763e2232fa253a/src/util.rs#L170) function that might be useful for you here, it returns a specific error when the process exits with a failure status that contains stdout/stderr.

chmanchester: Good to know! It doesn't quite map to what we're doing here, though.

froydnj: It seems better to adopt `run_input_output` to our needs here, perhaps by setting up the core `command` object with `env_clear().envs(...).args(...)` and calling into `run_input_output` rather than reproducing the code in a separate place.  Dealing with external commands is tricky enough that we want to centralize that logic in a single place.

chmanchester: We can use `run_input_output`, but the main advantage seems to be providing the process' output if it fails in a relevant error type that we hopefully wouldn't have to mess with ourselves. In this case though we're trying to fail when the process fails, but also if we're given something that looks like rustc but doesn't produce output like it's rustc, so we'd have to inspect and convert the output of `run_input_output` anyway, which is why I mentioned above it doesn't quite map to what we're doing here. The duplication was already around and doesn't seem egregious, but at this point I guess I'll just write that version up and see how it actually looks.

PR #490, opened 2019-08-05
Authenticate S3 GET requests if credentials are available

gferon:
An attempt to do the same as #337 and #471, with less changes to make sure it's easier to review & merge.


froydnj:
@chmanchester are you able to review this?  The change looks reasonable to me, but I'm not very knowledgeable about AWS credentialing.

chmanchester:
Sorry for the delay here, I'm also not very knowledgeable in this area. Looking over this now the change looks reasonable, thanks for the PR!

PR #491, opened 2019-08-05
Don't add --color when --json is passed to rustc. fixes #488

### LINKED ISSUE #488: sccache no longer works with rustc nightly
`rustc 1.38.0-nightly (435236b88 2019-08-01)`

It looks like https://github.com/rust-lang/rust/commit/17312337a9eb16389d11eca8472d1d083aa473bd changes how rustc emits diagnostics, if you try and use a nightly with this commit, you get an error when using sccache

```
error: Could not compile `rand_core`.

Caused by:
  process didn't exit successfully: `sccache rustc --crate-name rand_core /home/jake/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_core-0.4.0/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C debuginfo=2 --cfg 'feature="alloc"' --cfg 'feature="std"' -C metadata=073618b59550eeb8 -C extra-filename=-073618b59550eeb8 --out-dir /home/jake/code/ark/target/debug/deps -L dependency=/home/jake/code/ark/target/debug/deps --cap-lints allow` (exit code: 1)
warning: build failed, waiting for other jobs to finish...
error: cannot specify the `--color` option with `--json`
```

This occurs even if you specify `--color=never` due to how it checks the cmd line options, but regardless if it's a bug, it looks like sccache might have to change how it handles output from rustc due to the way the diagnostic messages are embedded in the JSON output.

chmanchester:


froydnj:
It looks like CI is very unhappy about some missed field initializations (?).

chmanchester:
> Just to make sure I understand: this change is _not_ problematic for non-nightly compilers because we've always been providing `--color`, and this patch changes nothing in that respect?

Right, this change is a no-op for non-nightly compilers because they wont pass be passing the `--json` flag at all, so they will get `--color=always` as they do today.

TriplEight:
Can we please get this change in the crate?

PR #500, opened 2019-08-22
Fix misleading error message

djmitche:
"Didn't get a parseable response body from instance role details"
suggests that something was parsed -- that's not the case.  Rather,
something failed in fetching the value, and that failure is `e`, so
should be included in the message.

---

I don't have the right version of OpenSSL installed to build this, so I'm hoping CI will tell me if this small change is correct.  If not, I'll figure out how to get OpenSSL built.


froydnj:
CI appears to be happy with your change. :)

PR #502, opened 2019-08-23
Run cargo fmt on the repo

Marwes:
Would be nice to have so that automatica formatting needn't be disabled when working on sccache (which I frequently forget, #471)


Marwes:
This could easily be done by any maintainer as well, just beware of the tiny fix in https://github.com/mozilla/sccache/pull/502/commits/20f889f812bc8c429c6286d3122b77daf723332b .

Would help reduce the clutter of #471

luser:
Subsumed by https://github.com/mozilla/sccache/pull/542

PR #505, opened 2019-09-03
Rewrap long line in sample config (to < 80 chars)

dholbert:
The sample config file has one long line (which is wider than a standard Mozilla-developer editor width, 80 characters).

This patch just rewraps one word (the word "provided") so that this sample config file is all less than 80 characters.

(Also, github shows the final line changing as well, even though I didn't touch it. It's just adding a trailing newline at the end of the file, where one was not previously present. This seems to have happened automagically by virtue of the fact that I was using the inline github editor (the github "pencil icon" tool), and it has some whitespace cleanup tooling built-in.)


froydnj:
Thanks!

PR #509, opened 2019-09-06
Fix accidental serialization on job_toolchains lock introduced in #482

chmanchester:


chmanchester:
> Doh. We should probably consider making a new release once this fix goes in? Version numbers are cheap.

Yes, will do.

Issue #453, opened 2019-06-15
sccache-dist corrupt deflate toolchain 


drahnr:
My wifi died while the toolchain was packaged and uploaded.

On the server side there then was a corrupted deflate which caused all subsequent compiles to fail.

Suggestion:

Delete toolchain fails if the fail to unpack.

drahnr:
I suspect the root cause here is that during the upload the toolchain is already marked as available, and as such the deflate stream is still incomplete and thus considered corrupted.

PR #511, opened 2019-09-08
[wip] track toolchains which are in progress

drahnr:
DO NOT MERGE 
This is work in progress to handle the case of #453 where the upload is proceeding while another request already states the toolchain is present.


froydnj:
Does #520 solve this problem for you?

drahnr:
I am not quite sure I understand that patch entirely: it will still cause the toolchain to appear avail during it's upload which is the root caues for #453 
I'll dig deeper and give #520 a shot and see if the issue still reproduces.

PR #515, opened 2019-09-16
Remove jobs from job list when replacing a server's identity. Fixes #513

### LINKED ISSUE #513: Builders may retain jobs and expend capacity indefinitely
We were noticing after #510 that `in_progress` seems to get stuck and not return to zero. I'm not sure exactly what's causing this, but reading through everything I found a server can retain jobs if the client goes away between the time a job is allocated and the job is run, which could be the culprit.

This wouldn't happen too often -- in the case someone interrupts a build the local daemon would still run pending jobs, but I was able to reproduce it on my laptop by stopping the local daemon after a build, and the case someone leaves the corporate network after interrupting a build would cause this, so it would certainly start wasting capacity eventually. #204 may also exacerbate the situation if people are hitting that.

chmanchester:


src/bin/sccache-dist/main.rs
```diff
@@ -668,9 +675,77 @@ impl SchedulerIncoming for Scheduler {
...
+                    }
+
+                    if stale_jobs.len() > 0 {
+                        warn!("The following stale jobs will be de-allocated: {:?}",
+                              stale_jobs);
+
+                        let mut jobs = self.jobs.lock().unwrap();
+
+                        for job_id in stale_jobs {
+                            if !details.jobs_assigned.remove(&job_id) {
```
froydnj: Is it possible to define a local enum or something so we can push (job_id, enumstate) triples into `stale_jobs`, and then we know exactly where to remove the jobs from in this loop?  I guess we have to worry about jobs appearing in `stale_jobs` from above and then getting removed before we get here?

src/bin/sccache-dist/main.rs
```diff
@@ -668,9 +675,77 @@ impl SchedulerIncoming for Scheduler {
...
+                                warn!(
+                                    "Stale job for server {} not assigned: {}",
+                                    server_id.addr(), job_id
+                                );
+                            }
+                            if details.jobs_unclaimed.remove(&job_id).is_none() {
+                                warn!(
+                                    "Stale job for server {} not marked as unclaimed: {}",
```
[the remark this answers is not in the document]
chmanchester: Yeah this is weird. It would mean it's not in the unclaimed list, which is how we got here in the first place. It "can't happen", this should probably become an assertion.

src/bin/sccache-dist/main.rs
chmanchester: I don't think I quite understand this suggestion. We want to remove the job from all these places whenever we find one that's expired. All this conditional checking is just to log something in case we think there should be a job somewhere and it isn't. What would the enumstate tell us?

froydnj: OK, I see now I was misunderstanding what was going on here.

chmanchester:
> Are the conversations you resolved addressed by changes in your local copy that you're going to push before merging?

Yes, I'll update and merge later today. Thanks!

PR #518, opened 2019-09-19
remove output files after preprocessing failures

### LINKED ISSUE #517: sccache should remove the object file when the preprocessor fails
See https://bugzilla.mozilla.org/show_bug.cgi?id=1540877#c13

froydnj:
Fixes #517.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -26,8 +26,6 @@ use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 use std::fmt;
-#[cfg(feature = "dist-client")]
-#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 use std::fs;
 use std::hash::Hash;
 #[cfg(feature = "dist-client")]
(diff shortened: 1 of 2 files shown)


froydnj:
It's not totally clear to me whether to remove everything, or just the output file, but removing everything is likely the safest course.

src/compiler/c.rs
```diff
@@ -245,8 +245,26 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
         let out_pretty = parsed_args.output_pretty().into_owned();
         let extra_hashes = hash_all(&parsed_args.extra_hash_files, &pool.clone());
+        let outputs = parsed_args.outputs.clone();
+        let args_cwd = cwd.clone();
 
         Box::new(result.or_else(move |err| {
+            // Errors remove all traces of potential output.
+            debug!("removing files {:?}", &outputs);
+
+            let v: std::result::Result<Vec<()>, std::io::Error> = outputs.values().map(|f| {
```
chmanchester: The Vec of empty values we don't consult seems unnecessary, maybe this can be something like a `fold`?

src/compiler/c.rs
```diff
@@ -245,8 +245,26 @@ impl<T, I> CompilerHasher<T> for CCompilerHasher<I>
...
+                let mut path = (&args_cwd).clone();
+                path.push(&f);
```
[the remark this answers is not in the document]
froydnj: The documentation for `PathBuf::Push` says that if the argument provided is an absolute path, it replaces anything in the `PathBuf`.  So I think we are OK here.

src/compiler/c.rs
froydnj: The motivation was not worrying too much about the performance of error cases, but I like the `fold` idea better.  I'll use that, thanks!

Issue #28, opened 2016-11-27
Rusoto now has a standalone credentials module


indiv0:
It looks like rusoto support was dropped in 657cb895f12bb0e51f950b6d62b1ff24a59ad9a8 for various reasons. However, you retained the rusoto credentials code in the form of an extracted file ([src/simples3/credential.rs](https://github.com/mozilla/sccache/blob/6eadccc9d6752747766d550388590199d87f23a5/src/simples3/credential.rs)).

It might interest you to know that rusoto now has a [standalone credentials module](https://crates.io/crates/rusoto_credential), which is published and functions independently of the main rusoto module. If you are interested in using it in your project, I can look into porting the modifications you've made to your version of the module into ours.

luser:
Oh, that would be super useful! FWIW, I liked Rusoto, I was just hitting some hard-to-diagnose issues and my S3 needs were fairly simple. I've only made a few small changes to the credentials module, I'd be happy to submit them as PRs.

indiv0:
That would be great, actually. As for rusoto itself, we're close to landing an important PR (https://github.com/rusoto/rusoto/pull/321) which will bring S3 in line with the rest of the codebase (currently it's mostly hand-crafted, legacy code from before our major re-write). Once that's done, I can help you migrate back to it as it's going to be much less problematic.

Xuanwo:
This issue has been resolved by https://github.com/mozilla/sccache/issues/1404

PR #522, opened 2019-09-25
Use Rusoto to handle S3 communication

bossmc:
This moves sccache over to using Rusoto instead of the inline S3 client.  The code is so similar that I can only assume one was based on the other but Rusoto has been developed further and supports some extra features (and probably bug-fixes).  In particular it supports AuthenticationV4 (as well as V2), which is very handy for interaction with S3 implementations that either don't support V2 or can only support one of V2 or V4 at a time (e.g. minio).

In the process, I ended up bumping quite a few dependencies and have updated the patch entries appropriately to match.

### DIFF
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -60,7 +60,7 @@ impl Storage for S3Cache {
     fn get(&self, key: &str) -> SFuture<Cache> {
         let key = normalize_key(key);
         let credentials = self.provider.credentials().chain_err(|| {
-            "failed to get AWS credentials"
+            "failed to get S3 credentials"
         });
 
         let bucket = self.bucket.clone();
@@ -73,7 +73,7 @@ impl Storage for S3Cache {
                         Ok(Cache::Hit(hit))
                     }
                     Err(e) => {
-                        warn!("Got AWS error: {:?}", e);
+                        warn!("Got S3 error: {:?}", e);
                         Ok(Cache::Miss)
                     }
                 }
@@ -89,7 +89,7 @@ impl Storage for S3Cache {
             Err(e) => return f_err(e),
         };
         let credentials = self.provider.credentials().chain_err(|| {
-            "failed to get AWS credentials"
+            "failed to get S3 credentials"
         });
 
         let bucket = self.bucket.clone();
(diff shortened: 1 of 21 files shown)


src/cache/s3.rs
```diff
@@ -46,19 +43,8 @@ pub struct S3Cache {
...
-        let user_dirs = UserDirs::new().ok_or("Couldn't get user directories")?;
-        let home = user_dirs.home_dir();
-
-        let profile_providers = vec![
-            ProfileProvider::with_configuration(home.join(".aws").join("credentials"), "default"),
-            //TODO: this is hacky, this is where our mac builders store their
-            // credentials. We should either match what boto does more directly
-            // or make those builders put their credentials in ~/.aws/credentials
-            ProfileProvider::with_configuration(home.join(".boto"), "Credentials"),
-        ];
```
bossmc: This change is the one bit that I can see being a problem.  The Rusoto client no longer supports configuring multiple `ProfileProvider`s for a single `ChainProvider`, I suspect it's not too hard to implement a local version, but if you're planning to move your credentials anyway, maybe this is your chance?

luser: I'm no longer at Mozilla but I suspect this is not a problem anymore because we had switched away from using native mac build machines while I was still there.

luser:
> The code is so similar that I can only assume one was based on the other

I did vendor a copy of Rusoto's S3 code a while back for some reason that I can no longer recall, but happy to see that get undone!

bossmc:
Well, that's promising @luser, thanks!

OTOH it seems I've bumped the minimum rustc version (due to dependencies)...

bossmc:
Re-enabled anonymous access to the cache (if no credentials are provided, try an anonymous read/write, if credentials are provided - use them).

luser:
This would subsume https://github.com/mozilla/sccache/issues/28

drahnr:
#28 already got solved :)

Hugal31:
Any news on this ?

bossmc:
From version 0.2.13 the original problem (that of inter-operation with minio) was resolved without this MR, which unfortunately has undermined my motivation for driving this change through.  The project still uses a local S3 client (in https://github.com/mozilla/sccache/tree/master/src/simples3) and I think it would still be valuable for maintenance to move over to rusoto, but it's probably easier now to start from scratch rather than trying to catch this MR up.

glandium:
Closing in favor of #826

Issue #514, opened 2019-09-11
Windows compiler invocation opens multiple console windows


Systemcluster:
This was previously reported under #16 and supposedly fixed in [this commit](https://github.com/mozilla/sccache/commit/c70dcf258fd71dd2f114930cc308d4d409314847) and again in #472 (which I presume was included in the `0.2.10` release), but the issue still persists - upon the start of a compilation process (in particular, a `cargo` invocation), multiple console windows pop up for a split second.

I'm encountering this with the following environment:
* `sccache` version `0.2.11` installed through `cargo` (it also occured with `0.2.10` and previous versions)
* Windows 10 (x64) version `10.0.18362.0`
* Running the `cargo` subcommands `run`, `build` or `install`, with the environment variable `RUSTC_WRAPPER=sccache` set and using the Rust `msvc` toolchain.

Please let me know if you need more information.

tangmi:
@Systemcluster Are you using PowerShell? Do you mind giving your builds a shot through Command Prompt (ensuring that sccache is not already running)? I remember reading somewhere that PowerShell does not like `CreateProcessW`'s `DETACHED_PROCESS` flag...

Systemcluster:
@tangmi I was using PowerShell, but I'm seeing the same behavior with CMD. Same with `sccache 0.2.12-alpha.0` compiled with the changes from your branch as well.

Systemcluster:
@tangmi I tried around a bit with some more start parameters, and after some experimentation, I decided to perform some logging with some `thread::sleep` calls to make it easier to pinpoint the source. As it turns out, the console popups are happening much later than the call to `connect_or_start_server`. (The initical sccache server start seems to create a very short lasting element in the taskbar, but no focus-stealing barrage of console windows as happening later.)

As far as I can tell (the sccache source is not the easiest to permeate) the popups appear inside `run_input_output` in util.rs. Which is very unfortunate, because at that point the process creation seems to be outsourced [to the Rust libstd](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/process.rs#L151).

gilescope:
Happens on a vanilla win7 cmd console also.

Shadlock0133:
Happens also on windows 8 in msys2's mintty.

gilescope:
And win10

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

galich:
it's a huge problem when running in Visual Studio Code + Rust Analyzer.

[...]

PR #525, opened 2019-10-02
fix(windows): remove `DETACHED_PROCESS` flag when starting the server process

tangmi:
The `DETACHED_PROCESS` flag causes the `CREATE_NO_WINDOW` to be ignored. See MS docs: https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags.

I believe `DETACHED_PROCESS` should avoid creating a window (unless one calls `AllocConsole`), but we're definitely seeing these calls create console windows. I've observed that sccache 0.2.11 works fine and does not create console windows in command prompt, but creates many console windows in powershell. The result is that this PR is a bit of a hack fix (unless we can justify not actually needing `DETACHED_PROCESS`...?).

Note: I'm planning on leaving the discussion at https://github.com/mozilla/sccache/issues/514 (and off this PR, at least until some decision is made).

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -157,7 +157,7 @@ fn run_server_process() -> Result<ServerStartup> {
     use winapi::um::handleapi::CloseHandle;
     use winapi::um::processthreadsapi::{CreateProcessW, PROCESS_INFORMATION, STARTUPINFOW};
     use winapi::um::winbase::{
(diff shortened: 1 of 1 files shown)


glandium:
Closing per https://github.com/mozilla/sccache/issues/514#issuecomment-741557322

glandium:
Reopening per https://github.com/mozilla/sccache/issues/514#issuecomment-742840236

glandium:
For good measure, I'll test this on Mozilla automation before merging.

PR #526, opened 2019-10-02
Update the docs with some minor notes.

emilio:
I had to figure these out while setting sccache up so it may be helpful to others too.


chmanchester:
Thanks for the PR, looks good to me!

PR #527, opened 2019-10-03
Add per-builder counts to dist stats. Fixes #478

### LINKED ISSUE #478: Provide more details to a client about distributed compilation
A common request when using `sccache-dist` is something like the icecc monitor. We may not be in a position to implement a gui right away, but it should be pretty straightforward to report where jobs are assigned and how long they're taking from the local daemon.

chmanchester:


chmanchester:
> LGTM, though I don't believe it addresses how long jobs are taking in #478, right? Or is that handled elsewhere?

I'm going to leave that for now -- I think it could end up being a little misleading because we don't have a metric for how long a compile is supposed to take, they might be doing more or less work.

PR #537, opened 2019-10-10
Update redis dep 0.9.0 -> 0.12.0

tiras-j:
It turns out that redis-rs 0.9.0 doesn't properly handle percent-decoding
of passwords extracted from the provided redis URL. This is particularly
problematic for Azure-hosted Redis infrastructure, as the access keys that
are generated are all base64 encoded (and thus have padding `=` chars) which
are percent-encoded when parsed into URL format. Bumping the redis dep handles
this use-case.

Unfortunately I discovered shortly thereafter that Azure disables the CONFIG
command (along with others) for their hosted Redis Cache... Oh well. The update
might see some help for someone using non-urlsafe passwords for Redis somewhere.

Also, if I'm missing some PR guidelines or whatnot sorry about that - just cutting this in the hopes that my frustration might benefit someone else, but it's such a small change I'm sure you could just inline it locally without any fuss.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ name = "aho-corasick"
 version = "0.6.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


froydnj:
Thanks for this change; it looks like there are some conflicts.  Can you rebase your changes and force-push your branch again, please?

glandium:
redis was updated to 0.15 in 34f391fe4f1c6b56db5faf7c34b8e32d6e85e2c4 and 0.18 in 9189ac6bca240ae7973e2e626e3ef364c73bbe17

PR #542, opened 2019-10-15
Format the entire repo.

chmanchester:


luser:
Nice, I kept meaning to do this but not getting around to it. :) You might want to add a rustfmt check to CI to ensure that things don't drift, something like `cargo fmt --all -- --check`. (I don't think that comes installed by default, although I don't know if that's the case on Travis CI, you might also need `rustup component add rustfmt-preview` in the setup steps, [ag_dubs has a good article on this](https://medium.com/@ag_dubs/enforcing-style-in-ci-for-rust-projects-18f6b09ec69d).)

emilio:
+1 for this, I tried to submit a minor cleanup patch and ran `cargo fmt` expecting my code to be formatted, and instead I got a massive diff.

chmanchester:
Let's handle the ci bits in a follow up (although they look straightforward enough).

PR #545, opened 2019-10-17
Improve preprocessor handling such that macros are not expanded.

### LINKED ISSUE #521: Preprocessor handling
Compared to what I know from the use of icecream/icecc, I find the handling of preprocessing by sccache less optimal.

With icecream/icecc, this worked transparently. The compiler output was the same, regardless of whether the compilation was done locally or remotely.

When using sccache, the compiler output is hardly useful in case of warnings or errors, because the compilation is done on the preprocessed source code, and all macro usages have been rolled out. If the code having warnings/errors uses macros heavily, this is virtually unreadable. In addition, this yields a lot additional meaningless warnings that only occur because the compiler only sees the result of the preprocessor (such as loop executed only once, e.g.). Also, the console coloring is not applied (but this wouldn't be helpful anyway in the cases mentioned before).

I am not sure if this is true also for compilation that goes through sccache but it actually done locally, or if this only applies to compilation done remotely.

In the former case, this issue might be mitigated to some extent, by allowing to disable remote compilation for certain paths (i.e. code that the user is actively working on), or rerunning the compilation locally when remote compilation had errors.

Why does this happen? I took not of #178, which seems to imply that the behaviour was already changed to match my expectation quite some time ago, but with the current version of sccache, this still happens to me. Can this be configured?

sigiesec:
Fixes #521

This surely deserves a test that checks that there will be no warnings emitted from macro expansions that should be suppressed by the additional flags. If someone provides me some guidance on how to add a test for that, I will be happy to do that.

- [x] Make enabling this behaviour explicit via a sccache configuration option.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -137,6 +137,7 @@ where
         env_vars: Vec<(OsString, OsString)>,
         may_dist: bool,
         pool: &CpuPool,
+        rewrite_includes_only: bool,
     ) -> SFuture<HashResult>;
 
     /// Return the state of any `--color` option passed to the compiler.
@@ -162,7 +163,18 @@ where
             Ok(Some(_)) => true,
             _ => false,
         };
-        let result = self.generate_hash_key(&creator, cwd.clone(), env_vars, may_dist, &pool);
+        let rewrite_includes_only = match dist_client {
+            Ok(Some(ref client)) => client.rewrite_includes_only(),
+            _ => false,
+        };
+        let result = self.generate_hash_key(
+            &creator,
+            cwd.clone(),
+            env_vars,
+            may_dist,
+            &pool,
+            rewrite_includes_only,
+        );
         Box::new(result.then(move |res| -> SFuture<_> {
             debug!(
                 "[{}]: generate_hash_key took {}",
@@ -390,7 +402,7 @@ where
 {
     let mut path_transformer = dist::PathTransformer::new();
     let compile_commands = compilation
-        .generate_compile_commands(&mut path_transformer)
+        .generate_compile_commands(&mut path_transformer, true)
         .chain_err(|| "Failed to generate compile commands");
     let (compile_cmd, _dist_compile_cmd, cacheable) = match compile_commands {
         Ok(cmds) => cmds,
@@ -420,9 +432,13 @@ where
     use futures::future;
     use std::io;
 
+    let rewrite_includes_only = match dist_client {
+        Ok(Some(ref client)) => client.rewrite_includes_only(),
+        _ => false,
+    };
     let mut path_transformer = dist::PathTransformer::new();
     let compile_commands = compilation
-        .generate_compile_commands(&mut path_transformer)
+        .generate_compile_commands(&mut path_transformer, rewrite_includes_only)
         .chain_err(|| "Failed to generate compile commands");
     let (compile_cmd, dist_compile_cmd, cacheable) = match compile_commands {
         Ok(cmds) => cmds,
@@ -585,6 +601,7 @@ pub trait Compilation {
     fn generate_compile_commands(
         &self,
         path_transformer: &mut dist::PathTransformer,
+        rewrite_includes_only: bool,
     ) -> Result<(CompileCommand, Option<dist::CompileCommand>, Cacheable)>;
 
     /// Create a function that will create the inputs used to perform a distributed compilation
@@ -1748,6 +1765,9 @@ mod test_dist {
         ) -> SFuture<(Toolchain, Option<String>)> {
             f_err("put toolchain failure")
         }
+        fn rewrite_includes_only(&self) -> bool {
+            false
+        }
     }
 
     pub struct ErrorAllocJobClient {
@@ -1789,6 +1809,9 @@ mod test_dist {
             _: Box<dyn pkg::ToolchainPackager>,
         ) -> SFuture<(Toolchain, Option<String>)> {
             f_ok((self.tc.clone(), None))
+        }
+        fn rewrite_includes_only(&self) -> bool {
+            false
         }
     }
 
@@ -1847,6 +1870,9 @@ mod test_dist {
             _: Box<dyn pkg::ToolchainPackager>,
         ) -> SFuture<(Toolchain, Option<String>)> {
             f_ok((self.tc.clone(), None))
+        }
+        fn rewrite_includes_only(&self) -> bool {
+            false
         }
     }
 
@@ -1907,6 +1933,9 @@ mod test_dist {
             _: Box<dyn pkg::ToolchainPackager>,
         ) -> SFuture<(Toolchain, Option<String>)> {
             f_ok((self.tc.clone(), Some("/overridden/compiler".to_owned())))
+        }
+        fn rewrite_includes_only(&self) -> bool {
+            false
         }
     }
 
@@ -1988,6 +2017,9 @@ mod test_dist {
             _: Box<dyn pkg::ToolchainPackager>,
         ) -> SFuture<(Toolchain, Option<String>)> {
             f_ok((self.tc.clone(), Some("/overridden/compiler".to_owned())))
+        }
+        fn rewrite_includes_only(&self) -> bool {
+            false
         }
     }
 }
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -52,6 +52,7 @@ impl CCompilerImpl for GCC {
         cwd: &Path,
         env_vars: &[(OsString, OsString)],
         may_dist: bool,
+        rewrite_includes_only: bool,
     ) -> SFuture<process::Output>
     where
         T: CommandCreatorSync,
@@ -64,6 +65,7 @@ impl CCompilerImpl for GCC {
             env_vars,
             may_dist,
             self.kind(),
+            rewrite_includes_only,
         )
     }
 
@@ -74,8 +76,16 @@ impl CCompilerImpl for GCC {
         parsed_args: &ParsedArguments,
         cwd: &Path,
         env_vars: &[(OsString, OsString)],
+        rewrite_includes_only: bool,
     ) -> Result<(CompileCommand, Option<dist::CompileCommand>, Cacheable)> {
-        generate_compile_commands(path_transformer, executable, parsed_args, cwd, env_vars)
+        generate_compile_commands(
+            path_transformer,
+            executable,
+            parsed_args,
+            cwd,
+            env_vars,
+            rewrite_includes_only,
+        )
     }
 }
 
@@ -445,6 +455,7 @@ pub fn preprocess<T>(
     env_vars: &[(OsString, OsString)],
     may_dist: bool,
     kind: CCompilerKind,
+    rewrite_includes_only: bool,
 ) -> SFuture<process::Output>
 where
     T: CommandCreatorSync,
@@ -465,14 +476,16 @@ where
     if !may_dist && !parsed_args.profile_generate {
         cmd.arg("-P");
     }
-    match kind {
-        CCompilerKind::Clang => {
-            cmd.arg("-frewrite-includes");
+    if rewrite_includes_only {
+        match kind {
+            CCompilerKind::Clang => {
+                cmd.arg("-frewrite-includes");
+            }
+            CCompilerKind::GCC => {
+                cmd.arg("-fdirectives-only");
+            }
+            _ => (),
         }
-        CCompilerKind::GCC => {
-            cmd.arg("-fdirectives-only");
-        }
-        _ => (),
     }
     cmd.arg(&parsed_args.input)
         .args(&parsed_args.preprocessor_args)
@@ -493,6 +506,7 @@ pub fn generate_compile_commands(
     parsed_args: &ParsedArguments,
     cwd: &Path,
     env_vars: &[(OsString, OsString)],
+    rewrite_includes_only: bool,
 ) -> Result<(CompileCommand, Option<dist::CompileCommand>, Cacheable)> {
     #[cfg(not(feature = "dist-client"))]
     let _ = path_transformer;
@@ -534,12 +548,16 @@ pub fn generate_compile_commands(
     #[cfg(feature = "dist-client")]
     let dist_command = (|| {
         // https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Overall-Options.html
-        let language = match parsed_args.language {
+        let mut language: String = match parsed_args.language {
             Language::C => "c",
             Language::Cxx => "c++",
             Language::ObjectiveC => "objective-c",
             Language::ObjectiveCxx => "objective-c++",
-        };
+        }
+        .into();
+        if !rewrite_includes_only {
+            language.push_str("-cpp-output");
+        }
         let mut arguments: Vec<String> = vec![
             "-x".into(),
             language.into(),
@@ -1150,6 +1168,7 @@ mod test {
             &parsed_args,
             f.tempdir.path(),
             &[],
+            false,
         )
         .unwrap();
         #[cfg(feature = "dist-client")]
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -56,7 +56,15 @@ impl CCompilerImpl for GCC {
     where
         T: CommandCreatorSync,
     {
-        preprocess(creator, executable, parsed_args, cwd, env_vars, may_dist)
+        preprocess(
+            creator,
+            executable,
+            parsed_args,
+            cwd,
+            env_vars,
+            may_dist,
+            self.kind(),
+        )
     }
 
     fn generate_compile_commands(
@@ -113,7 +121,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("--sysroot", PathBuf, Separated, PassThroughPath),
     take_arg!("-A", OsString, Separated, PassThrough),
     take_arg!("-B", PathBuf, CanBeSeparated, PassThroughPath),
-    take_arg!("-D", OsString, CanBeSeparated, PreprocessorArgument),
+    take_arg!("-D", OsString, CanBeSeparated, PassThrough),
     flag!("-E", TooHardFlag),
     take_arg!("-F", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-G", OsString, Separated, PassThrough),
@@ -128,7 +136,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("-MQ", OsString, Separated, PreprocessorArgument),
     take_arg!("-MT", OsString, Separated, DepTarget),
     flag!("-P", TooHardFlag),
-    take_arg!("-U", OsString, CanBeSeparated, PreprocessorArgument),
+    take_arg!("-U", OsString, CanBeSeparated, PassThrough),
     take_arg!("-V", OsString, Separated, PassThrough),
     take_arg!("-Xassembler", OsString, Separated, PassThrough),
     take_arg!("-Xlinker", OsString, Separated, PassThrough),
@@ -436,6 +444,7 @@ pub fn preprocess<T>(
     cwd: &Path,
     env_vars: &[(OsString, OsString)],
     may_dist: bool,
+    kind: CCompilerKind,
 ) -> SFuture<process::Output>
 where
     T: CommandCreatorSync,
@@ -455,6 +464,15 @@ where
     // With -fprofile-generate line number information is important, so don't use -P.
     if !may_dist && !parsed_args.profile_generate {
         cmd.arg("-P");
+    }
+    match kind {
+        CCompilerKind::Clang => {
+            cmd.arg("-frewrite-includes");
+        }
+        CCompilerKind::GCC => {
+            cmd.arg("-fdirectives-only");
+        }
+        _ => (),
     }
     cmd.arg(&parsed_args.input)
         .args(&parsed_args.preprocessor_args)
@@ -517,10 +535,10 @@ pub fn generate_compile_commands(
     let dist_command = (|| {
         // https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Overall-Options.html
         let language = match parsed_args.language {
-            Language::C => "cpp-output",
-            Language::Cxx => "c++-cpp-output",
-            Language::ObjectiveC => "objective-c-cpp-output",
-            Language::ObjectiveCxx => "objective-c++-cpp-output",
+            Language::C => "c",
+            Language::Cxx => "c++",
+            Language::ObjectiveC => "objective-c",
+            Language::ObjectiveCxx => "objective-c++",
         };
         let mut arguments: Vec<String> = vec![
             "-x".into(),
@@ -530,7 +548,6 @@ pub fn generate_compile_commands(
             "-o".into(),
             path_transformer.to_dist(out_file)?,
         ];
-        // We could do preprocessor_args here, but skip for consistency with msvc
         arguments.extend(dist::osstrings_to_strings(&parsed_args.common_args)?);
         Some(dist::CompileCommand {
             executable: path_transformer.to_dist(&executable)?,
--- a/src/config.rs
+++ b/src/config.rs
@@ -383,6 +383,7 @@ pub struct DistConfig {
     pub cache_dir: PathBuf,
     pub toolchains: Vec<DistToolchainConfig>,
     pub toolchain_cache_size: u64,
+    pub rewrite_includes_only: bool,
 }
 
 impl Default for DistConfig {
@@ -393,6 +394,7 @@ impl Default for DistConfig {
             cache_dir: default_dist_cache_dir(),
             toolchains: Default::default(),
             toolchain_cache_size: default_toolchain_cache_size(),
+            rewrite_includes_only: true,
         }
     }
 }
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -64,6 +64,8 @@ macro_rules! vec_from {
     };
 }
 
+// TODO: This will fail if gcc/clang is actually a ccache wrapper, as it is the
+// default case on Fedora, e.g.
 fn compile_cmdline<T: AsRef<OsStr>>(
     compiler: &str,
     exe: T,
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -81,6 +81,16 @@ const INPUT: &'static str = "test.c";
 const INPUT_ERR: &'static str = "test_err.c";
 const OUTPUT: &'static str = "test.o";
 
+// Copy the source files into the tempdir so we can compile with relative paths, since the commandline winds up in the hash key.
+fn copy_to_tempdir(inputs: &[&str], tempdir: &Path) {
+    for f in inputs {
+        let original_source_file = Path::new(file!()).parent().unwrap().join(f.clone());
+        let source_file = tempdir.join(f);
+        trace!("fs::copy({:?}, {:?})", original_source_file, source_file);
+        fs::copy(&original_source_file, &source_file).unwrap();
+    }
+}
+
 fn test_basic_compile(compiler: Compiler, tempdir: &Path) {
     let Compiler {
         name,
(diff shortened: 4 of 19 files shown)


luser:
There are some system tests that test interop with compilers in `tests/system.rs`. You can add a new one and make it compiler-specific by adding it to the right place in `run_sccache_command_tests`:
https://github.com/mozilla/sccache/blob/e5f8856901688c163baceb352bf766f72731dfed/tests/system.rs#L272

chmanchester:
@sigiesec are you planning to add a test for this? If not I'm happy to take a stab at it.

sigiesec:
I started to add a test. Will hopefully get to work on this in the next days.

emilio:
This breaks sccache-dist (I tried this out and had a pretty hard time figuring out why everything broke), because we don't pass `-D` arguments to the distributed builder, looks like.

sigiesec:
> This breaks sccache-dist (I tried this out and had a pretty hard time figuring out why everything broke), because we don't pass `-D` arguments to the distributed builder, looks like.

Mh, I finally got a test case that tests this, at least it compiles right now (puh, Rust :D ). I am not sure what exactly breaks here. The additional argument is only added when the preprocessor is run locally. On the distributed side, nothing changes. What do you mean by `-D` arguments?

sigiesec:
So, I finally added a test case. Most probably, the Rust syntax I used isn't the best, I am happy to incorporate any hints on how to do this better!

emilio:
So the test-case where I saw it failing was something like the following situation:

* Compile command: `clang -DTEST t.cpp`
* t.cpp:

```cpp
#ifndef TEST
#error "Something something"
#endif
```

Before this change, the preprocessor expands the whole file, and the distributed builder only needs to run `clang path/to/preprocessed/input.cpp`

After this change though, you need to pass the `-DTEST` flag to the remote builder, and this is not done at the moment, so your file will error.

In the situation I found it, the define was some configure stuff and what not that was passed to our clang plugin for feature detection.

sigiesec:
Ah I understand now. Sure, that needs to be taken care of.

sigiesec:
Hi @emilio: I added a test now that requires a define (https://github.com/mozilla/sccache/pull/545/commits/8c6240cddaf7431958a97c2c186f800afddb8496), but it passes the tests: https://travis-ci.org/mozilla/sccache/builds/609128687?utm_source=github_status&utm_medium=notification

Could you please check that the test case does what you meant, and maybe tell me what needs to be done differently?

emilio:
That test doesn't seem to exercise the `sccache-dist` command, does it? This was only a problem when the file is sent over to a remote builder.

The main issue is that [this line](https://github.com/mozilla/sccache/blob/84bac031a28a9dc63a231dfcfa2d893bab3ffbf6/src/compiler/gcc.rs#L516) only copies `common_args`, but not `preprocessor_args`.

sigiesec:
@emilio I thought that the CI job specified by https://github.com/mozilla/sccache/blob/master/.travis.yml#L11 is meant to run all the tests via sccache-dist, but maybe I am mistaken about that, or it doesn't work as it should?

I am not sure what args are copied by that line, but since you said it doesn't work for you, you are probably right this needs to be changed. I just would want to have a test case that fails with the current implementation to be sure that a fix is effective.

emilio:
Seems like the dist-tests do not run on CI per [this line](https://github.com/mozilla/sccache/blob/84bac031a28a9dc63a231dfcfa2d893bab3ffbf6/.travis.yml#L40)?

Also it looks like that only enables building with the `dist-*` features, but not running all the normal tests with a distributed compiler (I could be wrong though).

sigiesec:
> Seems like the dist-tests do not run on CI per [this line](https://github.com/mozilla/sccache/blob/84bac031a28a9dc63a231dfcfa2d893bab3ffbf6/.travis.yml#L40)?
> 
> Also it looks like that only enables building with the `dist-*` features, but not running all the normal tests with a distributed compiler (I could be wrong though).

That's possible. But wouldn't it make sense to run them also with a distributed compiler? For this particular issue, I can probably add a test that does that (if I find the right place), but it seems like this this should better be done in a more generic way.

emilio:
Oh to be clear I totally agree! I couldn't run the dist tests at all locally because some docker stuff.

sigiesec:
Added a check for the preprocessor define handling in the distributed test case test_dist_basic, and fixed issues as described by @emilio

src/compiler/gcc.rs
```diff
@@ -438,6 +439,11 @@ where
     if !may_dist && !parsed_args.profile_generate {
         cmd.arg("-P");
     }
+    match kind {
+       CCompilerKind::Clang => { cmd.arg("-frewrite-includes"); },
```
emilio: nit: braces are not necessary, so probably just: `=> cmd.arg(".."),`

src/compiler/gcc.rs
```diff
@@ -438,6 +439,11 @@ where
...
+       CCompilerKind::GCC => { cmd.arg("-fdirectives-only"); },
+        _ => (),
```
emilio: Indentation seems off here, maybe run rustfmt? Though maybe it's just github.

src/compiler/gcc.rs
```diff
@@ -438,6 +439,11 @@ where
...
+    };
```
emilio: Not 100% sure this semi-color is useful.

tests/system.rs
```diff
@@ -79,8 +81,21 @@ fn compile_cmdline<T: AsRef<OsStr>>(
 
 const INPUT: &'static str = "test.c";
 const INPUT_ERR: &'static str = "test_err.c";
+const INPUT_MACRO_EXPANSION: &'static str = "test_macro_expansion.c";
+const INPUT_WITH_DEFINE: &'static str = "test_with_define.c";
 const OUTPUT: &'static str = "test.o";
 
+// Copy the source files into the tempdir so we can compile with relative paths, since the commandline winds up in the hash key.
+fn copy_to_tempdir(inputs: Vec<String>, tempdir: &Path)
+{
```
emilio: This is not common style, please run rustfmt. Also since you don't need to resize `inputs` it may as well be an `&[String]`.


Furthermore, given these are static, you can just do `&[&str]`.

tests/system.rs
```diff
@@ -89,13 +104,7 @@ fn test_basic_compile(compiler: Compiler, tempdir: &Path) {
...
     trace!("run_sccache_command_test: {}", name);
     // Compile a source file.
-    // Copy the source files into the tempdir so we can compile with relative paths, since the commandline winds up in the hash key.
-    for f in &[INPUT, INPUT_ERR] {
-        let original_source_file = Path::new(file!()).parent().unwrap().join(f);
-        let source_file = tempdir.join(f);
-        trace!("fs::copy({:?}, {:?})", original_source_file, source_file);
-        fs::copy(&original_source_file, &source_file).unwrap();
-    }
+    copy_to_tempdir(vec![INPUT.to_string(), INPUT_ERR.to_string()], tempdir);
```
emilio: That way this becomes: `copy_to_tempdir(&[INPUT, INPUT_ERR], tempdir)` and so on.

src/compiler/gcc.rs
sigiesec: That yields me:
```
error[E0308]: match arms have incompatible types
   --> src/compiler/gcc.rs:453:14
    |
450 | /     match kind {
451 | |         CCompilerKind::Clang => cmd.arg("-frewrite-includes"),
    | |                                 ----------------------------- this is found to be of type `&mut <T as mock_command::CommandCreatorSync>::Cmd`
452 | |         CCompilerKind::GCC => cmd.arg("-fdirectives-only"),
    | |                               ---------------------------- this is found to be of type `&mut <T as mock_command::CommandCreatorSync>::Cmd`
453 | |         _ => (),
    | |              ^^ expected mutable reference, found ()
454 | |     }
    | |_____- `match` arms have incompatible types
    |
    = note: expected type `&mut <T as mock_command::CommandCreatorSync>::Cmd`
               found type `()`
```

src/compiler/gcc.rs
sigiesec: You're right, I reformatted this.

src/compiler/gcc.rs
sigiesec: Removed it.

src/compiler/gcc.rs
emilio: Ah, indeed. Nevermind then :)

tests/system.rs
sigiesec: Ok, thanks for the hint, I changed that!

tests/system.rs
sigiesec: Adapted.

sigiesec:
> Maybe remove the [WiP] label from the one commit in your branch that has it?

Sure, I oversaw that, thanks for pointing that out!

src/compiler/gcc.rs
```diff
@@ -512,8 +530,11 @@ pub fn generate_compile_commands(
             "-o".into(),
             path_transformer.to_dist(out_file)?,
         ];
-        // We could do preprocessor_args here, but skip for consistency with msvc
         arguments.extend(dist::osstrings_to_strings(&parsed_args.common_args)?);
+        // We need to include the preprocessor_args here, since the preprocessor run is
+        // configured to only resolve includes, so the preprocessor needs to run on the
+        // dist side again.
```
chmanchester: Would it be possible to do a quick benchmark of the impact of this? It may be worth it even if it adds some overhead, but it would be good to know.

sigiesec: I am not sure how to do a meaningful benchmark. It would need to use a containerized scheduler & server to be somewhat controllable. It would certainly be interesting to know, so with some guidance on how to perform such a benchmark, I can surely do that!

However, IMHO the modifications by this PR are essential for regular use cases, both locally and in CI. Distributing a build is usually expected to be transparent, and without these changes, transparency is very much broken: Any error and warning messages involving macros become hardly comprehensible, and it is essentially impossible to run a build with -Werror.

emilio: I tend to agree with @sigiesec. I was going to take some measurements by doing mozilla-central clobber builds, but it seems this needs more work as this breaks mozilla-central builds completely...

In particular, mozilla-central uses `-MD -MP -MF file.pp`, and with this patch those flags are passed to the remote builder (and it isn't there of course).

So I think rather than this, we should just make the `-D` flag a `PassThrough` flag rather than a `PreprocessorArg` so it goes to common_args and gets passed both to the preprocessor and the compiler, but without passing the other preprocessor flags like `-i*` and `-M*` flags.

sigiesec: So we should probably add another test case (or add it to the existing test cases) using these options, so we catch these issues in the sccache tests already :)

emilio: Hmm, doing that I get further along, but my build fails in a somewhat mysterious way in a system header...

emilio: So the command that's failing is `/home/emilio/.cargo/bin/sccache /home/emilio/.mozbuild/clang/bin/clang++ -o nsLocalFileUnix.o -c  -I/home/emilio/src/moz/gecko-2/obj-debug/dist/stl_wrappers -I/home/emilio/src/moz/gecko-2/obj-debug/dist/system_wrappers -include /home/emilio/src/moz/gecko-2/config/gcc_hidden.h -DDEBUG=1 -DOS_POSIX=1 -DOS_LINUX=1 -DSTATIC_EXPORTABLE_JS_API -DMOZ_HAS_MOZGLUE -DMOZILLA_INTERNAL_API -DIMPL_LIBXUL -I/home/emilio/src/moz/gecko-2/xpcom/io -I/home/emilio/src/moz/gecko-2/obj-debug/xpcom/io -I/home/emilio/src/moz/gecko-2/obj-debug/ipc/ipdl/_ipdlheaders -I/home/emilio/src/moz/gecko-2/ipc/chromium/src -I/home/emilio/src/moz/gecko-2/ipc/glue -I/home/emilio/src/moz/gecko-2/obj-debug/xpcom -I/home/emilio/src/moz/gecko-2/xpcom/build -I/home/emilio/src/moz/gecko-2/obj-debug/dist/include -I/home/emilio/src/moz/gecko-2/obj-debug/dist/include/nspr -I/home/emilio/src/moz/gecko-2/obj-debug/dist/include/nss -fPIC -DMOZILLA_CLIENT -include /home/emilio/src/moz/gecko-2/obj-debug/mozilla-config.h -Qunused-arguments -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fstack-protector-strong -ftrivial-auto-var-init=pattern -Qunused-arguments -Wall -Wbitfield-enum-conversion -Wempty-body -Wignored-qualifiers -Woverloaded-virtual -Wpointer-arith -Wshadow-field-in-constructor-modified -Wsign-compare -Wtype-limits -Wunreachable-code -Wunreachable-code-return -Wwrite-strings -Wno-invalid-offsetof -Wclass-varargs -Wfloat-overflow-conversion -Wfloat-zero-conversion -Wloop-analysis -Wc++1z-compat -Wc++2a-compat -Wcomma -Wimplicit-fallthrough -Wstring-conversion -Wtautological-overlap-compare -Wtautological-unsigned-enum-zero-compare -Wtautological-unsigned-zero-compare -Wno-error=tautological-type-limit-compare -Wno-inline-new-delete -Wno-error=deprecated-declarations -Wno-error=array-bounds -Wno-error=backend-plugin -Wno-error=return-std-move -Wno-error=atomic-alignment -Wformat -Wformat-security -Wno-gnu-zero-variadic-macro-arguments -Wno-unknown-warning-option -Wno-return-type-c-linkage -fno-sized-deallocation -fno-aligned-new -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fstack-protector-strong -ftrivial-auto-var-init=pattern -fno-exceptions -fno-strict-aliasing -fno-rtti -fno-exceptions -fno-math-errno -pthread -pipe -g -Os -fno-omit-frame-pointer -funwind-tables  -MD -MP -MF .deps/nsLocalFileUnix.o.pp  -fcolor-diagnostics  /home/emilio/src/moz/gecko-2/xpcom/io/nsLocalFileUnix.cpp`

And it does indeed compile if I remove the sccache wrapper. Not sure yet why.

emilio: Yeah I'm trying to reduce the issue with that suggestion too.

emilio: So the reduced test-case is a pile of sadness like this:

```c
 #define a __has_include                                                                                                                                                                                                                     
extern                                                                           
  #if a("linux/stat.h")                                                            
      ;                                                                            
  #endif
```

I think this is a bug in `clang`'s `-frewrite-include`. The tldr is that my glibc has something like:

```cpp
#ifdef __has_include
#define __glibc_has_include(header) __has_include(header)
#else 
#define __glibc_has_include(header) 0
#endif 

#if __glibc_has_include("linux/stat.h")
#include <linux/stat.h>
#endif
```

And then there's some code in gecko that includes that file unconditionally.

But clang rewrites the first include and not the second, but _doesn't_ remove the `__glibc_has_include` condition, so the builder, which doesn't have that header, doesn't process the included header and everything falls apart... GCC works fine, but GCC + sccache give me a `fatal error: cannot execute 'as': execvp: No such file or directory`.

emilio: So I think we should change this patch to do something like:

```diff
diff --git a/src/compiler/gcc.rs b/src/compiler/gcc.rs
index 8ef663b..e7a1354 100644
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -121,7 +121,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("--sysroot", PathBuf, Separated, PassThroughPath),
     take_arg!("-A", OsString, Separated, PassThrough),
     take_arg!("-B", PathBuf, CanBeSeparated, PassThroughPath),
-    take_arg!("-D", OsString, CanBeSeparated, PreprocessorArgument),
+    take_arg!("-D", OsString, CanBeSeparated, PassThrough),
     flag!("-E", TooHardFlag),
     take_arg!("-F", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-G", OsString, Separated, PassThrough),
@@ -136,7 +136,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("-MQ", OsString, Separated, PreprocessorArgument),
     take_arg!("-MT", OsString, Separated, DepTarget),
     flag!("-P", TooHardFlag),
-    take_arg!("-U", OsString, CanBeSeparated, PreprocessorArgument),
+    take_arg!("-U", OsString, CanBeSeparated, PassThrough),
     take_arg!("-V", OsString, Separated, PassThrough),
     take_arg!("-Xassembler", OsString, Separated, PassThrough),
     take_arg!("-Xlinker", OsString, Separated, PassThrough),
@@ -549,10 +549,6 @@ pub fn generate_compile_commands(
             path_transformer.to_dist(out_file)?,
         ];
         arguments.extend(dist::osstrings_to_strings(&parsed_args.common_args)?);
-        // We need to include the preprocessor_args here, since the preprocessor run is
-        // configured to only resolve includes, so the preprocessor needs to run on the
-        // dist side again.
-        arguments.extend(dist::osstrings_to_strings(&parsed_args.preprocessor_args)?);
         Some(dist::CompileCommand {
             executable: path_transformer.to_dist(&executable)?,
             arguments: arguments,
```

And then we should go:

 * Fix g++ so that it works right with sccache-dist.
 * Fix clang++'s `-frewrite-includes` to handle the above situation correctly.

emilio: So to elaborate, a more comprehensive test-case would be:

```cpp
#define has_include(header) __has_include(header)                                                                                                                                                                                           
                                                                                  
#if has_include("t.h")                                                           
#include <iostream>                                                              
#endif 

#include <iostream>                                                              
                                                                                   
int main() {                                                                     
  std::cout << "Hello world" << std::endl;                                       
  return 0;                                                                      
}
```

Where `t.h` is an empty file locally.

`clang -frewrite-includes` will leave the `has_include` check, but it will _not_ expand the second `#include` (this behavior seems to be limited to system headers somehow).

emilio: So I filed a clang bug in:

 * https://bugs.llvm.org/show_bug.cgi?id=43982

But right after doing so I realized of a similar bug:

 * https://bugs.llvm.org/show_bug.cgi?id=37990

Where Richard Smith explains that this is invalid per http://eel.is/c++draft/cpp.cond#7.sentence-2.

So I ended up filing a libc bug:

 * https://sourceware.org/bugzilla/show_bug.cgi?id=25189

And sending a patch because why not:

 * https://sourceware.org/ml/libc-alpha/2019-11/msg00431.html

emilio: Ok, with that glibc patch, and the diff I posted above applied to your tree, I can build Firefox without issues \o/

emilio: And I took some numbers, and these patches + mine make the build considerably faster (maybe due to the reduced output from the compiler).

I got a clobber debug build in Berlin in:

> 4:30.65 Your build was successful!

:)

glandium: Sadly, you can't rely on people having a new enough glibc once the patch lands.

chmanchester: > And I took some numbers, and these patches + mine make the build considerably faster (maybe due to the reduced output from the compiler).
> 
> I got a clobber debug build in Berlin in:
> 
> > 4:30.65 Your build was successful!
> 
> :)

That's impressive, was the cache empty for that build?

chmanchester: > I am not sure how to do a meaningful benchmark. It would need to use a containerized scheduler & server to be somewhat controllable. It would certainly be interesting to know, so with some guidance on how to perform such a benchmark, I can surely do that!
> 
> However, IMHO the modifications by this PR are essential for regular use cases, both locally and in CI. Distributing a build is usually expected to be transparent, and without these changes, transparency is very much broken: Any error and warning messages involving macros become hardly comprehensible, and it is essentially impossible to run a build with -Werror.

Yes, I agree, I just meant to try running a build with and without to see if it adds considerable overhead. But looking at this patch again I don't think it's going to result in too much extra work.

emilio: > Sadly, you can't rely on people having a new enough glibc once the patch lands.

I think the glibc version on my machine is pre-release. So assuming that's the case (not 100% sure) I think it's ok.

emilio: Gah, I lie, the problematic code is is in glibc 2.30 :(

sigiesec: We can certainly not land the patch relying on some very-recent/future glibc. What confuses me here is that this has been working with icecream (which uses `-frewrite-includes`/`-fdirectives-only`, so I think we are still missing something.

Hearing that this actually speeds up compilation is great! This might also explain why I had the impression that sccache has been slower than icecream despite the support for rust.

sigiesec: FWIW, these commits implemented this for icecream, for clang resp. gcc:
* clang: https://github.com/icecc/icecream/commit/b78ca31a18cbd24efe5a015da4734dc72afeb374 (note that the clang option was `-Wp,-rewrite-includes` originally, but was later changed to `-frewrite-includes`)
* gcc: https://github.com/icecc/icecream/commit/7ab429b5f3405ec8451c31f80573658c8818b972

This is on by default, but suppressed when -Wunused-macros is used, via https://github.com/icecc/icecream/commit/4f46699b945ed26fb9924609fdf68e606b0b53b7

There is also code to force local recompilation when there was an error remotely with gcc only due to some limitations of `-fdirectives-only`, added via https://github.com/icecc/icecream/commit/861030d617086560dacce2f000df5db0f7f19b9a

But I assume that @emilio's problem with glibc happened with clang?

emilio: Yes, my issue happened with clang. But it's a pretty recent issue, I suspect it'd happen with icecc as well.

sigiesec: By "recent" you mean the problematic `__has_include` usage was only introduced in a recent version of glibc?

chmanchester: Can we put this behind an option, as icecc did, and document the issue with glibc?

sigiesec: @chmanchester Makes sense. I will add that. Maybe at sometime in the future we can default it to on then.

sigiesec: So this means a (boolean) field needs to be added to DistConfig, right? Does this automagically generate parser code for the config file? I didn't find code that parser it explicitly.

luser: > So this means a (boolean) field needs to be added to DistConfig, right? Does this automagically generate parser code for the config file? I didn't find code that parser it explicitly.

It's using serde to `derive(Serialize, Deserialize)` so you get that for free:
https://github.com/mozilla/sccache/blob/a2cab6b35fd9cec47d09275a2448a272d771c82c/src/config.rs#L374

sigiesec:
I added a configuration option `rewrite_includes_only` now. As of now, it defaults to `true`. I am not sure how likely it is that the glibc issue occurs for a user. I can also default it to `false` if that seems better.

Please let me know if  any other changes are required before merging this.

src/compiler/gcc.rs
```diff
@@ -530,8 +566,11 @@ pub fn generate_compile_commands(
...
+        arguments.extend(dist::osstrings_to_strings(&parsed_args.preprocessor_args)?);
```
emilio: As per the thread above, I think this is not right. This passes too many flags to the dist builder, and fails when building Gecko. I posted a fix for this in the thread above, mainly:

```diff
diff --git a/src/compiler/gcc.rs b/src/compiler/gcc.rs
index 8ef663b..e7a1354 100644
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -121,7 +121,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("--sysroot", PathBuf, Separated, PassThroughPath),
     take_arg!("-A", OsString, Separated, PassThrough),
     take_arg!("-B", PathBuf, CanBeSeparated, PassThroughPath),
-    take_arg!("-D", OsString, CanBeSeparated, PreprocessorArgument),
+    take_arg!("-D", OsString, CanBeSeparated, PassThrough),
     flag!("-E", TooHardFlag),
     take_arg!("-F", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-G", OsString, Separated, PassThrough),
@@ -136,7 +136,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("-MQ", OsString, Separated, PreprocessorArgument),
     take_arg!("-MT", OsString, Separated, DepTarget),
     flag!("-P", TooHardFlag),
-    take_arg!("-U", OsString, CanBeSeparated, PreprocessorArgument),
+    take_arg!("-U", OsString, CanBeSeparated, PassThrough),
     take_arg!("-V", OsString, Separated, PassThrough),
     take_arg!("-Xassembler", OsString, Separated, PassThrough),
     take_arg!("-Xlinker", OsString, Separated, PassThrough),
@@ -549,10 +549,6 @@ pub fn generate_compile_commands(
             path_transformer.to_dist(out_file)?,
         ];
         arguments.extend(dist::osstrings_to_strings(&parsed_args.common_args)?);
-        // We need to include the preprocessor_args here, since the preprocessor run is
-        // configured to only resolve includes, so the preprocessor needs to run on the
-        // dist side again.
-        arguments.extend(dist::osstrings_to_strings(&parsed_args.preprocessor_args)?);
         Some(dist::CompileCommand {
             executable: path_transformer.to_dist(&executable)?,
             arguments: arguments,
```

Is this intentionally not addressed? If so, why?

src/compiler/compiler.rs
```diff
@@ -162,7 +163,23 @@ where
             Ok(Some(_)) => true,
             _ => false,
         };
-        let result = self.generate_hash_key(&creator, cwd.clone(), env_vars, may_dist, &pool);
+        let rewrite_includes_only = match dist_client {
+            Ok(Some(_)) => dist_client
```
emilio: Can you just do `Ok(Some(ref client)) => client.rewrite_includes_only(),`?

src/compiler/compiler.rs
```diff
@@ -420,9 +437,18 @@ where
     use futures::future;
     use std::io;
 
+    let rewrite_includes_only = match dist_client {
+        Ok(Some(_)) => dist_client
```
emilio: Same here.

src/compiler/gcc.rs
sigiesec: Not intentionally. Sorry, I think I applied these changes, but somehow I lost them accidentally during rebasing.

src/compiler/compiler.rs
sigiesec: Changed that.

src/compiler/compiler.rs
sigiesec: Dito

emilio:
FWIW, my glibc fix was uplifted to the current release, which was the only that contained the bug I think, so maybe we don't need the switch after all?

glandium:
Not everybody is going to build with the newest glibc.

emilio:
Sure, my point is that this was glibc regression, and the release that introduced the bug already has the fix. But anyhow not a big deal I guess.

src/compiler/compiler.rs
```diff
@@ -390,7 +402,7 @@ where
 {
     let mut path_transformer = dist::PathTransformer::new();
     let compile_commands = compilation
-        .generate_compile_commands(&mut path_transformer)
+        .generate_compile_commands(&mut path_transformer, true)
```
chmanchester: In the local case we want this to be `false`, no? Does it matter?

src/config.rs
```diff
@@ -393,6 +394,7 @@ impl Default for DistConfig {
             cache_dir: default_dist_cache_dir(),
             toolchains: Default::default(),
             toolchain_cache_size: default_toolchain_cache_size(),
+            rewrite_includes_only: true,
```
chmanchester: I'm inclined to say we should default this to `false` so we can get this merged until we know exactly how many of our users are going to have issues.

sigiesec: Well, I am not sure if we are going to find that out when defaulting to false. We can default to true initially and when it turns out that too many users need to disable and cannot update their glibc, we can still change the default. A note should be added to the docs and release notes though.

src/compiler/compiler.rs
sigiesec: In that case, is a separate preprocessing step done as well? If yes, we should definitely enable the behaviour. This won't run into the glibc issue then IIUC. Otherwise, it doesn't matter.

chmanchester: It is. Good point.

src/config.rs
chmanchester: Works for me.

PR #549, opened 2019-10-19
Fix default output directory

### LINKED ISSUE #548: Object files being output to source file path
This is surprising behavior that breaks FreeBSD's build scripts. Both clang and gcc output automatically named object files to the present working directory.

Current behavior:
```
sccache gcc -c ./foo/bar.c; find . -type f
./foo/bar.o
./foo/bar.c
```
Expected behavior:
```
sccache gcc -c ./foo/bar.c; find . -type f
./bar.o
./foo/bar.c
```

sterlingjensen:
Drop the source file path prefix when generating the object file name.

Fixes #548


froydnj:
Thank you!

PR #556, opened 2019-10-23
Add docs for setting up a server that spawns at boot time.

emilio:


luser:
You'll probably want to mention setting `SCCACHE_IDLE_TIMEOUT=0` in the environment so the server doesn't automatically shut down while idle.

emilio:
Hmm, does the server actually shutdown like that? It's not mentioned anywhere, and my service is running still since this morning (and I'm pretty sure it has been idle for a bit)

chmanchester:
> Hmm, does the server actually shutdown like that? It's not mentioned anywhere, and my service is running still since this morning (and I'm pretty sure it has been idle for a bit)

`SCCACHE_IDLE_TIMEOUT` changes the behavior of the "local" server (I've been trying to refer to it as the "local daemon" to avoid confusion). I don't think it's relevant in this case.

SimonSapin:
When following the instructions on Ubuntu 18.04 LTS, the last step prints:

```
$ sudo systemctl enable sccache-server
The unit files have no installation config (WantedBy, RequiredBy, Also, Alias
settings in the [Install] section, and DefaultInstance for template units).
This means they are not meant to be enabled using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another unit's
   .wants/ or .requires/ directory.
2) A unit's purpose may be to act as a helper for some other unit which has
   a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path, timer,
   D-Bus, udev, scripted systemctl call, ...).
4) In case of template units, the unit is meant to be enabled with some
   instance name specified.
```

SimonSapin:
After adding `WantedBy=multi-user.target` in the `[Install]` section I could run `enable` successfully. However on reboot the server failed to start:

```
oct. 24 16:34:42 usagi systemd[1]: Started sccache-dist server.
oct. 24 16:34:43 usagi sccache-dist[869]: error: Failed to start http server for sccache server
oct. 24 16:34:43 usagi sccache-dist[869]: caused by: Cannot assign requested address (os error 99)
oct. 24 16:34:43 usagi systemd[1]: sccache-server.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
oct. 24 16:34:43 usagi systemd[1]: sccache-server.service: Failed with result 'exit-code'.
```

emilio:
Hmm, probably needs a `WantedBy=network-online.target` instead.

emilio:
Huh, that also doesn't seem to work...

emilio:
```
$  journalctl -u network.service -u network.target -u sccache-server.service -b                                                                                                                                                            
-- Logs begin at Mon 2019-10-21 14:32:23 CEST, end at Thu 2019-10-24 20:42:38 CEST. --
Oct 24 20:37:48 localhost.localdomain systemd[1]: Started sccache-dist server.
Oct 24 20:37:48 localhost.localdomain bash[1250]: error: Failed to start http server for sccache server
Oct 24 20:37:48 localhost.localdomain bash[1250]: caused by: Cannot assign requested address (os error 99)
Oct 24 20:37:48 localhost.localdomain systemd[1]: sccache-server.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Oct 24 20:37:48 localhost.localdomain systemd[1]: sccache-server.service: Failed with result 'exit-code'.
Oct 24 20:37:49 localhost.localdomain systemd[1]: Reached target Network.
```

emilio:
> Oct 24 20:37:45 localhost.localdomain systemd[1]: /etc/systemd/system/sccache-server.service:9: Unknown key name 'Wants' in section 'Install', ignoring.
> Oct 24 20:37:45 localhost.localdomain systemd[1]: /etc/systemd/system/sccache-server.service:10: Unknown key name 'After' in section 'Install', ignoring.

That's not great... But I think I know how to fix it.

emilio:
Ok, what I have put in the PR now works :)

SimonSapin:
Works for me as well, thanks!

emilio:
I cannot merge it myself though :)

Issue #562, opened 2019-10-24
sccache-dist doesn't handle clang plugins really well


emilio:
Trying to run an sccache-dist build of Firefox with `--enable-clang-plugin` fails catastrophically because of:

```
[...]

Issue #558, opened 2019-10-23
Support file-level content-addressable storage for compile inputs in the distributed build server


luser:
I was looking at the source for the Rust `InputsPackager`, which is responsible for packaging up all the inputs necessary for a single compile to ship to the build server:
https://github.com/mozilla/sccache/blob/fc256ff3552274e210eb456f4a2f54ee9038307e/src/compiler/rust.rs#L1316-L1462

And I realized that distributed Rust compiles are probably doing a lot of redundant work. Every time a Rust compilation is distributed sccache has to package up any files referenced on the commandline which includes all the `rlib` files for crates the current crate uses, shared libraries for proc macros, etc. For a single `cargo` invocation this likely means that the same files will get packaged up over and over again.

A nice optimization here would be to give the build server an API for a content-addressable store, where clients could query the server for a file hash and find out if the server already has it and upload files that the server doesn't have. The server could simply store them on disk similar to how the existing `DiskCache` works. Then instead of packaging up all compile inputs the client would hash them all (this happens as part of cache lookup anyway), ask the server whether it already has them, and upload any that are not already present. Presumably to optimize the process the server should provide an API that allows querying a list of hashes and returns two lists of hashes: those that are already present and those that are missing. When the build server goes to execute a compilation then instead of providing a tarball of all the inputs it would provide a list of mappings from path -> hash, and the server would take care of retrieving all of those files from its local cache and placing them at the desired paths in the build filesystem (this could likely be done with hardlinks to avoid copying). A further optimization would be to preemptively store any outputs of Rust compilation in the cache since they are likely to be used as inputs to another Rust compile.

aidanhs:
Yes, I have a side interest in content-addressible-stores (actually in rolling checksums, but they're mainly interesting in that context) and sometimes think about it since there's a multitude of use cases, including:

 - this issue (build inputs)
 - allowing build machines to collaborate on receiving toolchains from slow clients
 - permitting clients to retrieve just changed parts of outputs

Might be worth also looking at the "Remote Execution API" as it uses CAS, with the caveat that it may be optimised for different use-cases to sccache - https://github.com/mozilla/sccache/issues/358#issuecomment-469032196

PR #563, opened 2019-10-24
Don't do distributed compilation if we're building with plugins.

emilio:
Partially addresses #562 by not building remotely.

That is not amazing, and we should probably look into how to address this use
case...

We could try to address it sending the plugins to the server, but that doesn't
seem to work when cross-compiling, and I don't have great ideas for that...

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -511,6 +511,13 @@ pub fn generate_compile_commands(path_transformer: &mut dist::PathTransformer,
     let dist_command = None;
     #[cfg(feature = "dist-client")]
     let dist_command = (|| {
+        if !parsed_args.extra_hash_files.is_empty() {
+            // If we're using a plugin of sorts, we have no guarantee that it'll
+            // be in the distributed server, so compile locally.
+            warn!("Cannot perform distributed compile due to gcc or clang plugin");
+            return None;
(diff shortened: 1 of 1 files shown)


luser:
The Rust distributed compiler handles this correctly, FYI:
https://github.com/mozilla/sccache/blob/fc256ff3552274e210eb456f4a2f54ee9038307e/src/compiler/rust.rs#L1336

emilio:
Hmm, I could give that a shot... But does that mean that, if I do it the obvious way at least, I'd be uploading the clang-plugin along with any translation unit?

emilio:
To be clear, maybe that is ok, I don't know the magnitude of the sizes involved here. For example in a local clang-plugin debug build:

```
$ du -h ./build/clang-plugin/libclang-plugin.so
1.7M	./build/clang-plugin/libclang-plugin.so
```

luser:
> Hmm, I could give that a shot... But does that mean that, if I do it the obvious way at least, I'd be uploading the clang-plugin along with any translation unit?

Yes, that's correct. Note that for distributed Rust compiles the client sends over every rlib (and .so for proc macro crates) referenced on the commandline. I filed https://github.com/mozilla/sccache/issues/558 with a proposal to optimize that.

PR #570, opened 2019-10-25
Replace rust-crypto with RustCrypto/hashes

chmanchester:


chmanchester:
> If I'm reading things correctly, the azure bits have tests that ensure these changes are a no-op, but the s3 bits do not. Can you add a small test there along the lines of the azure tests to ensure we don't break anything?

Does my addition here agree with what you had in mind for a test?

PR #577, opened 2019-10-31
Fix/silence a ton of clippy warnings

Jake-Shadle:
This fixes a majority of the clippy issues, or at least silences them. I think it may bump the minimum Rust version though, but I figured this might be ok as async/await is coming soon and presumably sccache will want to take advantage of that ASAP.


Jake-Shadle:
Yah, fair enough, I'll just go ahead and close this then, as this was just a drive by change while doing another PR.

Issue #489, opened 2019-08-04
GCS storage failing with "invalid RSA key"


brndnmtthws:
sccache recently stopped working with GCS. I tried rolling back to a previous version, and that didn't help.

I did some digging, and found the following being printed in the logs (perhaps something changed in the Google API?):

```
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
 WARN 2019-08-04T15:12:41Z: sccache::cache::gcs: Got GCS error: Error(Msg("invalid RSA key"), State { next_error: None, backtrace: InternalBacktrace })
[...]

PR #578, opened 2019-10-31
Replace usage of jsonwebtoken for GCS

Jake-Shadle:
This removes the use of the patched jsonwebtoken

`jsonwebtoken = { git = "https://github.com/Jake-Shadle/jsonwebtoken.git", rev = "2f469a61" }`

which is still needed due to no release from jsonwebtoken with the relevant changes (last release was May). Using patches are supposed to be temporary, because what ends up happening is that people that do `cargo install sccache --features=gcs` end up with a broken sccache because the patched version of jsonwebtoken that "works" is not used when installing from crates.io, and for some reason, the prebuilt binaries (at least the musl one) in github releases are also incorrect.

Rather than wait for a release, this change removes the need for jsonwebtoken altogether for `gcs` caching, by just doing the relevant pieces directly using base64/ring.

* Serializes  (JSON + base64) the JWT header (signing algorithm)
* Serializes (JSON + base64) the JWT claim (scope, timestamps, uri etc)
* Signs the JWT payload with the private key from the service account
* Concatenates the JWT payload and the base64 encoded signature

jsonwebtoken is still used in the dist stuff, only required 2 changes due to a field being removed from JWT.

Resolves: #489

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,6 +122,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bincode"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


froydnj:
Can you rebase this now that we're resolved the weird failures with `jobserver`?  Thanks!

froydnj:
Hm, it looks like development is happening on `jsonwebtoken`...but it doesn't seem to be happening fast enough.  We can merge this for now, but I'd love it if/when `jsonwebtoken` gets updated if this change were reverted.

Can you force-push to squash your commits and preserve bisectability, please?

Jake-Shadle:
Done!

froydnj:
Thank you!

PR #579, opened 2019-10-31
Prune stale servers more often in the scheduler

### LINKED ISSUE #538: Prune dead servers more often than on alloc_job
People rely on the server count / in progress stat when debugging. The fact we only prune dead servers when jobs are allocated has caused confusion more than a few times. I think we can additionally check on the status call.

chmanchester:
This updates to pruning dead servers when handling heartbeat or status rather than only when allocating jobs. Fixes #538

The idea is that this will happen more regularly so people attempting to debug a cluster will have an up-to-date view of what's happening, but less frequently and out of the critical path when allocating lots of jobs.

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -45,7 +45,7 @@ use std::io::{self, Write};
 use std::net::SocketAddr;
 use std::path::Path;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
 use std::time::{Duration, Instant};
 use syslog::Facility;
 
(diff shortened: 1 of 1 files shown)


froydnj:
> The idea is that this will happen more regularly

What is "this" here?  The heartbeat or status handling?  Or the pruning?  Or both?

I guess if you're trying to debug, you're asking for the status of the cluster a lot more often than you're allocating jobs?

chmanchester:
> > The idea is that this will happen more regularly
> 
> What is "this" here? The heartbeat or status handling? Or the pruning? Or both?
> 
> I guess if you're trying to debug, you're asking for the status of the cluster a lot more often than you're allocating jobs?

"this" is pruning dead servers. Job allocation happens a lot when people are running builds, but during a period of inactivity or when trying to debug something we can end up with a confusingly elevated in_progress/cpu_count. This is a little more "regular" in the sense it will get an update any time a server pings the scheduler, and in the case someone's trying to set up the first server on a cluster whenever they ask for status.

PR #587, opened 2019-11-09
cmdline: Allow running sccache under rr without crashing the remote compiler.

emilio:
Otherwise we remote LD_PRELOAD to the builder, and crash...

This is a huge hack and probably should only be done in the dist path. But it
worked for me so I wanted to post it somewhere. Happy to do the work to make
something like this landable or what not. It's clearly not perfect.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -201,11 +201,22 @@ pub fn parse() -> Result<Command> {
     } else if let Some(mut args) = cmd {
         if let Some(exe) = args.next() {
             let cmdline = args.map(|s| s.to_owned()).collect::<Vec<_>>();
+            let mut env_vars = env::vars_os().collect::<Vec<_>>();
+
+            // If we're running under rr, avoid the `LD_PRELOAD` bits, as it will
+            // almost surely do the wrong thing, as the compiler gets executed
+            // in a different process tree.
+            //
+            // FIXME: Maybe we should strip out `LD_PRELOAD` always?
+            if env::var_os("RUNNING_UNDER_RR").is_some() {
+                env_vars.retain(|(k, _v)| k != "LD_PRELOAD" && k != "RUNNING_UNDER_RR");
+            }
+
             Ok(Command::Compile {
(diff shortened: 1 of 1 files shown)


froydnj:
It's a little gross, but a lot less gross than it could be.  It needs at least a comment describing why we're doing this.

Actually, we probably shouldn't remote `LD_PRELOAD` under any circumstances...

emilio:
Yeah, I realized we need this even in the non-`dist` case, as the compile command gets executed elsewhere as well. So I think this is the right place to do this. I added a comment for that.

src/cmdline.rs
```diff
@@ -201,11 +201,22 @@ pub fn parse() -> Result<Command> {
     } else if let Some(mut args) = cmd {
         if let Some(exe) = args.next() {
             let cmdline = args.map(|s| s.to_owned()).collect::<Vec<_>>();
+            let mut env_vars = env::vars_os().collect::<Vec<_>>();
+
+            // If we're running under rr, avoid the `LD_PRELOAD` bits, as it will
+            // almost surely do the wrong thing, as the compiler gets executed
+            // in a different process tree.
```
froydnj: Wait, this happens even in the local case, not just the dist case?  So where you'd expect `rr -> sccache -> compiler`, you actually get `rr -> sccache` and `rr -> compiler` ?  Who's separating things out into a different process tree?

emilio: rr -> sccache -> sccache daemon -> compiler?

emilio: All the builds go through the daemon, afaict, which is what decides whether to remote again to a remote builder or not... My understanding here could be wrong, but I can trivially reproduce it even off-line with something like: `$ rr sccache /usr/bin/gcc -c t.c -o t.o`, where `t.c` is just `int main() { return 0; }`

froydnj: Ah, right.  OK.

froydnj:
Thanks!

emilio:
NP, thanks for merging :)

PR #589, opened 2019-11-09
compiler: Implement color diagnostics for gcc and clang. 

emilio:


emilio:
One day GitHub will learn to show the commits in topological order... `args:` goes before `compiler:`... :)

PR #591, opened 2019-11-11
tests: Install openssl 1.1 in the docker image.

emilio:
We copy the sccache-dist binary from the host to the container. If the host is
using openssl 1.1 (like Fedora by default) then the sccache-dist binary can't
find the right shared libs and fails.

This allows me to run dist-tests in Fedora 31, with some hackarounds like:

 * https://github.com/mozilla/kitsune/issues/3991#issuecomment-552154956


emilio:
(If travis fails, we may need to install both openssl versions or something)

sigiesec:
I also needed to modify the Dockerfile to work with my Fedora 30 host.

However, it should be kept in mind that the Dockerfile is not independent from the host environment. To ensure that the tests run in _any_ host environment, probably the build of the sccache-server would also need to be built in the container.

PR #595, opened 2019-11-12
compiler: support '.C' for C++ code

razeh:
Assume a file ending in '.C' is a C++ file.


froydnj:
Thanks!  (For the record, this is what gcc assumes by default as well.)

PR #597, opened 2019-11-12
Remove corrupt toolchains from cache when unpacking fails. Fixes #451

### LINKED ISSUE #451: Occassional "entity already exists" failure from build server
I set up sccache for distributed compilation on `Fedora 30` (client and dist server) following the guides under `docs/`

Normal sccache operation with redis works fine, but remote compilation always fails with this.
```
 WARN 2019-06-14T20:03:30Z: sccache::compiler::compiler: [serde_json]: Could not perform distributed compile, falling back to local: could not run distributed compilation job: could not run distributed compilation job: Error 500 (Headers={"server": "tiny-http (Rust)", "date": "Fri, 14 Jun 2019 20:03:30 GMT", "content-type": "application/json", "content-length": "152"}): {"description":"run build failed","cause":{"description":"failed to prepare overlay dirs","cause":{"description":"entity already exists","cause":null}}}
```
https://github.com/mozilla/sccache/blob/954897f9c933c7328f5289a38926ab7a6bbff3c2/src/bin/sccache-dist/build.rs#L133 fails since the directory seems to already exist. Is there some rationale behind it or should/could that particular error be ignored?

I applied that change with no side effects (tested without a custom toolchain on Linux x86_64 only).

chmanchester:
Additionally, don't allow unpacked toolchains to accumulate indefinitely.
This makes it so the unpacked toolchain directory will not be allowed
to grow to more elements than the backing toolchain cache.

### DIFF
--- a/src/bin/sccache-dist/build.rs
+++ b/src/bin/sccache-dist/build.rs
@@ -190,7 +190,18 @@ impl OverlayBuilder {
                         )
                     }
                 };
-                tar::Archive::new(GzDecoder::new(toolchain_rdr)).unpack(&toolchain_dir)?;
+
+                tar::Archive::new(GzDecoder::new(toolchain_rdr))
+                    .unpack(&toolchain_dir)
+                    .or_else(|e| {
+                        warn!("Failed to unpack toolchain: {:?}", e);
+                        fs::remove_dir_all(&toolchain_dir)
+                            .chain_err(|| "Failed to remove unpacked toolchain")?;
+                        tccache
+                            .remove(tc)
+                            .chain_err(|| "Failed to remove corrupt toolchain")?;
+                        Err(Error::from(e))
+                    })?;
 
                 let entry = (toolchain_dir, 1);
                 assert!(toolchain_dir_map
--- a/lru-disk-cache/src/lib.rs
+++ b/lru-disk-cache/src/lib.rs
@@ -302,6 +302,20 @@ impl LruDiskCache {
     pub fn get<K: AsRef<OsStr>>(&mut self, key: K) -> Result<Box<dyn ReadSeek>> {
         self.get_file(key).map(|f| Box::new(f) as Box<dyn ReadSeek>)
     }
+
+    /// Remove the given key from the cache.
+    pub fn remove<K: AsRef<OsStr>>(&mut self, key: K) -> Result<()> {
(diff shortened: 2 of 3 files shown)


src/bin/sccache-dist/build.rs
```diff
@@ -190,12 +202,36 @@ impl OverlayBuilder {
...
+                    let dir_map = toolchain_dir_map.clone();
+                    let mut entries: Vec<_> = dir_map.iter().collect();
+                    entries.sort_by(|a, b| (a.1).ctime.cmp(&(b.1).ctime));
+                    if let Some((tc, _)) = entries.first() {
+                        warn!("Removing old un-compressed toolchain: {:?}", tc);
+                        assert!(toolchain_dir_map.remove(tc).is_some());
+                        fs::remove_dir_all(&self.dir.join("toolchains").join(&tc.archive_id))
+                            .chain_err(|| "Failed to remove old toolchain directory")?;
+                    }
+                }
```
froydnj: I think I would prefer if this were split out into its own pull request and not ride along with the corruption fixes.

I'm also not sure if this is the best way to do things.  IIUC, we have `toolchain_dir_map`, which maps from `Toolchain` to (effectively) some path.  And then we have the `tccache`, which is mapping from paths to stuff.  And this block is compensating for having this double set of mappings, in some sense?  Should we really have a unified mapping that also manages the LRU-ness of the underlying disk cache?  Or would that confuse other parts of the system that don't really expect to have toolchains disappear from underneath them?

Maybe all these questions are easily answerable, maybe not.  In any event, I think this should be a separate thing.

chmanchester: I'm happy to split this out into more commits, but I'm not entirely keen on a deeper rewrite of these wonky caching layers. The current code is going to lead to servers running out of disk, and I think this will be a reasonably reliable way to prevent that in the short term. Do you have a specific approach in mind? The implementation is still quite weird, because we're allowing people to specify the max size of the compressed toolchain cache but leaving corresponding unpacked toolchains around on disk for quite a while. Perhaps a better approach would just be to cache unpacked toolchains rather than having a two tier system.

chmanchester:
Whoops, github ate my comment. I split this into separate commits. I'm fine with continuing discussion of the top two in a different forum, but the first issue is actively causing problems.

PR #599, opened 2019-11-14
feat: Support enabling TLS in custom S3 endpoints

bearcage:
Since the `simples3` module already handles this (and it seems the only logic flipped here is the url prefix), plumbing through a config is a pretty quick change. There doesn't appear to be a github issue for this yet, but there was a `TODO` in the source which this resolves.

The tests appear to be passing, and testing against a local minio instance shows this working:

image


bearcage:
ptal @aidanhs @chmanchester — picked y'all since you're the most recently-active contributors to this repo afaict

bearcage:
force-pushed to fixup the old-rust-version CI failure

PR #601, opened 2019-11-16
Add Windows client config path to DistributedQuickstart doc

nabijaczleweli:
Might save someone some swearing and patching `src/config.rs` to print the config path?


froydnj:
Thank you!

PR #603, opened 2019-11-17
Increase buffer size for file reading and other small IO improvements.

orium:


froydnj:
Can you comment on what kind of improvements this provides?  (I believe that it should theoretically make things faster, but I'd like to know if that improvement was measurable.)

orium:
I didn't measure it.  Probably the time spent on io is small enough that it doesn't matter that much.  I will benchmark it tomorrow with something big like servo.

froydnj:
> I will benchmark it tomorrow with something big like servo.

I will go ahead and merge this, because I think larger chunks of IO are better regardless.  I will be interested to hear what the results of your experiment are!

orium:
These are the results I got from compiling servo.

Before this PR:

```text
orium@laptop-orion servo % hyperfine -w 0 -r 3 -p './mach clean' './mach check'                                                                                                                            @ master
Benchmark #1: ./mach check
  Time (mean ± σ):     524.464 s ±  6.825 s    [User: 4699.042 s, System: 296.998 s]
  Range (min … max):   517.084 s … 530.547 s    3 runs

hyperfine -w 0 -r 3 -p './mach clean' './mach check'  14099.95s user 893.98s system 950% cpu 26:17.82 total
```

After this PR:

```text
orium@laptop-orion servo % hyperfine -w 0 -r 3 -p './mach clean' './mach check'                                                                                                                            @ master
Benchmark #1: ./mach check
  Time (mean ± σ):     521.629 s ± 11.372 s    [User: 4721.858 s, System: 300.042 s]
  Range (min … max):   508.578 s … 529.409 s    3 runs

hyperfine -w 0 -r 3 -p './mach clean' './mach check'  14168.68s user 903.28s system 960% cpu 26:09.68 total
```

It was slightly faster after the PR, however the variance increased in such a way that doesn't allow any conclusion.  Also this was a sample with 3 runs, which isn't great.  In any case, it is probably making things slightly faster...  Maybe.

PR #604, opened 2019-11-19
Prevent unpacked toolchains from accumulating indefinitely

chmanchester:


chmanchester:
I'll merge this for now, but I was thinking about it some more and there's a pathological case where least recently added and least recently used oppose one another and we end up with lots of sorting/deleting/recreation. I'll work on a better fix as a follow-up.

Issue #566, opened 2019-10-24
Use async / await when possible.


emilio:
They should be stable in the next rust release, and seems like sccache could get a fair cleanup...

I'd be interested in helping with this.

luser:
I think step one here is to switch things over to use `std::future::Future` which is a bit of an undertaking. It looks like the `futures` crate has [a `compat` layer](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.19/futures/compat/index.html) coming in its 0.3 release, and `tokio` has [an as-yet-unlanded change to add a compat runtime](https://github.com/tokio-rs/tokio/pull/1663). Using those would probably make it easier to to an incremental conversion.

The other option would be to wholesale switch to `tokio` 0.2 [which is still in alpha](https://tokio.rs/blog/2019-08-alphas/) but uses `std::future::Future`. This will involve a considerable amount of rework to match new APIs and changes to upstream tokio but is the right long-term direction.

luser:
`tokio` 0.2 was just released so it ought to be feasible for someone to start on this:
https://tokio.rs/blog/2019-11-tokio-0-2/

emilio:
Took a quick look, and hyper and reqwest still depend on older versions of tokio, so probably still need to wait for that.

zbraniecki:
`hyper 0.13` uses tokio 0.2, and `reqwest 0.10 alpha2` as well!

saneyuki:
I opened https://github.com/mozilla/sccache/pull/627 as the first step to migrate to use async/.await.

IMO, this migration would be a bit tough work and I feel we should migrate our code incrementally.

ncalexan:
FWIW not using `std::future` just surprised me reading the code.

FooBarWidget:
As far as I know, the async-await migration is now complete, and so I believe this issue can be closed.

PR #605, opened 2019-11-20
use a fixed thread pool for schedulers and servers

froydnj:
Rather than start a new thread for every request, which is moderately
expensive, let's just have a fixed thread pool to serve requests.

### DIFF
--- a/src/dist/http.rs
+++ b/src/dist/http.rs
@@ -836,6 +836,10 @@ mod server {
                 trace!("Res {}: {:?}", req_id, response);
                 response
             }).map_err(|e| Error::with_boxed_chain(e, ErrorKind::Msg("Failed to start http server for sccache scheduler".to_owned())))?;
+            // This limit is rouille's default for `start_server_with_pool`, which
+            // we would use, except that interface doesn't permit any sort of
+            // error handling to be done.
+            let server = server.pool_size(num_cpus::get() * 8);
             server.run();
 
             panic!("Rouille server terminated")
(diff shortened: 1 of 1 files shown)


froydnj:
`rouille`'s interfaces are really weird, and there's some evidence that other things might be faster than `threadpool`: https://gitlab.nebulanet.cc/xacrimon/uvth

I don't know if we want to enable the number of backing threads to be tuned, but I do know that all the "fork" overhead for spawning threads was visible in `perf` prior to this patch and is non-existent after this patch (`reqwest` seems to be consuming most of the CPU at this point?).  So I think this is valuable regardless of the exact default that we choose to go with.

luser:
If someone wanted to tackle https://github.com/mozilla/sccache/issues/566 I bet this code would benefit from using `async` (assuming there's a reasonable crate to use for the web framework part).

froydnj:
> If someone wanted to tackle #566 I bet this code would benefit from using `async` (assuming there's a reasonable crate to use for the web framework part).

Just to be clear, because I'm not totally sure what this entails, this means whatever's driving `async` under the hood would be responsible for using non-blocking I/O or threads, depending on what we directed the underlying driver to do?  (icecc manages to make everything work with a single-threaded scheduler, so clearly our threaded thing might be a little bit of overkill...)

luser:
> Just to be clear, because I'm not totally sure what this entails, this means whatever's driving `async` under the hood would be responsible for using non-blocking I/O or threads, depending on what we directed the underlying driver to do? (icecc manages to make everything work with a single-threaded scheduler, so clearly our threaded thing might be a little bit of overkill...)

Yep! This already happens all over the place in the existing background server where everything is async but it passes around a `CpuPool` for CPU-intensive work like hashing:
https://github.com/mozilla/sccache/blob/2a89cd6922469416d297f3fd9029d4d0a2686e3c/src/compiler/rust.rs#L1051

PR #612, opened 2019-11-25
Make the stats persistant by saving them to cache

### LINKED ISSUE #33: Save stats to disk

elichai:
Closes #33 
I tried making it as boilerplate in the Storage trait, but too much lifetime issues (upgrading to latest futures (if possible in mozilla?) should solve a lot of those)
 https://play.rust-lang.org/?gist=917b00ea0c752d6fa50f35965d969648

### DIFF
--- a/src/cache/azure.rs
+++ b/src/cache/azure.rs
@@ -16,12 +16,12 @@
 use crate::azure::BlobContainer;
 use crate::azure::*;
 use crate::cache::{Cache, CacheRead, CacheWrite, Storage};
+use crate::errors::*;
+use crate::server::ServerStats;
 use futures::future::Future;
 use std::io;
 use std::rc::Rc;
 use std::time::{Duration, Instant};
-
-use crate::errors::*;
 
 pub struct AzureBlobCache {
     container: Rc<BlobContainer>,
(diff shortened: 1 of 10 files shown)


froydnj:
I think that this saving would be better off being managed by something higher-level than sccache itself, since multiple instances of the `SccacheService` (e.g. different build processes on different machines) are going to be racing against each other to save their stats when e.g. saving them in S3 or a common disk cache on the network.

luser:
I had thought about this a while ago (as evidenced by the age of #33) and figured the local sccache server would just persist it to the local config location, not the cache (which I understand is what `ccache` does). It could write out stats on a background thread (or with async I/O) on a regular basis and during shutdown and read them on startup.

elichai:
So do you want it to save them into a file near where the config is saved or you don't want them to be saved by sccache at all?

sylvestre:
inactive, closing.

Danielmelody:
Saving to cache seems reasonable because it could be fully remote. Multi-users could share the real missing rate over time and machines.

PR #614, opened 2019-12-01
Add hits/misses ratio to stats

elichai:
Hi,
This adds a ratio to the stats. it looks like this:
```
$ ./target/debug/sccache -s
Compile requests                    862
Compile requests executed           648
Cache hits                          239
Cache hits (Rust)                   239
Cache misses                        409
Cache misses (Rust)                 409
Cache hits rate                   36.88 %
Cache hits rate (Rust)            36.88 %
Non-cacheable calls                 211
Unsupported compiler calls            3
Average cache read miss           1.158 s
Average cache read hit            0.002 s

Non-cacheable reasons:
crate-type                          137
incremental                          50
-                                    16
-o                                    8

Cache location                  Local disk: "/home/elichai2/.cache/sccache"
Cache size                            8 GiB
Max cache size                       20 GiB
```

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1221,6 +1221,14 @@ impl ServerStats {
             }};
         }
 
+        macro_rules! set_stat_precentage {
+            ($vec:ident, $count:expr, $sum:expr, $name:expr) => {
+                let ratio = $count as f64 / $sum as f64;
+                // name, value, suffix length
+                $vec.push(($name.to_string(), format!("{:.2} %", ratio*100.0), 2));
+            }
+        }
+
         macro_rules! set_lang_stat {
             ($vec:ident, $var:expr, $name:expr) => {{
                 $vec.push(($name.to_string(), $var.all().to_string(), 0));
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -1254,6 +1262,23 @@ impl ServerStats {
...
         set_lang_stat!(stats_vec, self.cache_misses, "Cache misses");
+
+        set_stat_precentage!(stats_vec, self.cache_hits.all(), self.cache_misses.all() + self.cache_hits.all(), "Cache hits rate");
+
+        let mut sorted_stats_hits: Vec<_> = self.cache_hits.counts.iter().collect();
+        sorted_stats_hits.sort_by_key(|v| v.0);
+        let mut sorted_stats_misses: Vec<_> = self.cache_misses.counts.iter().collect();
+        sorted_stats_misses.sort_by_key(|v| v.0);
+
+        for ((lang_hits, &count_hits), (lang_misses, &count_misses)) in sorted_stats_hits.into_iter().zip(sorted_stats_misses.into_iter()) {
```
glandium, line 1273: I don't think this would work in the case where there are only hits or misses for a given language. You'd need to enumerate the intersection of all languages in hits and misses.

src/server.rs
```diff
@@ -1221,6 +1221,14 @@ impl ServerStats {
             }};
         }
 
+        macro_rules! set_stat_precentage {
```
glandium, line 1224: Following the style of other macros, this should be set_percentage_stat (also, you have a typo on percentage).

glandium (changes requested):

sylvestre:
inactive, closing

PR #627, opened 2019-12-13
Prepare to migrate to async/.await

tetsuharuohzeki:
## Motivation

* This migration efforts would be tough work.
* I think we should migrate our code incrementally.

## Scope of this pull request

* This prepares async/.await infrastructures.
    * Bump minimum rustc version.
    * Add futures v3 and tokio v0.2 to dependency.
    * I think this migration would be a bit tough journey.
* This rename old tokio/future import to tokio01/future01.
    * This change aim to begin to use async/.await compatible code.

## Related Issues

* https://github.com/mozilla/sccache/issues/566


Marwes:
Having done conversions in other projects I would recommend waiting with tokio-0.2 as it is not usable until all dependencies are updated to use it (or the executor/timer/reactor won't match).

Updating incrementally while still using tokio-0.1 works perfectly well however, just needs a `.boxed().compat()` call before executing the top level future.

saneyuki:
> Updating incrementally while still using tokio-0.1 works perfectly well however, just needs a .boxed().compat() call before executing the top level future.

IMO, the migration of #566 would be a bit tough work with bumping up many dependencies once in the single pull request because this project uses a classic _future_ crate in wide area of code.

So I feel we should migrate our code incrementally even if we would not like to call `.boxed().compat()`...

Marwes:
I never said that this should be done all at once, it definitely should be done incrementally. I only said that adding tokio-0.2 only serves to add confusion since it isn't actually usable without all dependencies being updated from tokio-0.1 (so this could mean a lot of changes all at once, but it should still be fairly isolated).

It seems that https://crates.io/crates/tokio-compat was just released today though so it might actually be possible to mix 0.1 and 0.2 now.

saneyuki:
I see. 

> It seems that https://crates.io/crates/tokio-compat was just released today though so it might actually be possible to mix 0.1 and 0.2 now.

Thank you for your information. I'll try it.

glandium:
As Nathan mentioned in #733, it probably made this obsolete. I'm closing, but please feel free to reopen if you feel otherwise.

PR #628, opened 2019-12-19
Add instructions on how to use cargo's configuration file to setup sccache.

orium:


orium:
Note that `cargo` will issue an warning saying ```unused key `build.rustc-wrapper` in config file```.  This warning is incorrect and was fixed in rust-lang/cargo#7575, which will take effect in cargo 1.41.

README.md
```diff
@@ -120,11 +126,24 @@ Usage
...
+[build]
+rustc-wrapper = "[path to sccache]"
+```
+
+Alternatively you can use the environment variable `RUSTC_WRAPPER`:
 
-> $ RUSTC_WRAPPER=[path to sccache] cargo build
+```text
+RUSTC_WRAPPER=[path to sccache] cargo build
+```
```
froydnj: I think you are inconsistent about whether `text` or `bash` is used for commands (or not used at all, see the `scoop` instructions above).  I don't know what the right choice is, but we should pick one and stick with it.

README.md
```diff
@@ -120,11 +126,24 @@ Usage
...
+
+If you want to use sccache for your rust builds you can define `build.rustc-wrapper` in the
+[cargo configuration file](https://doc.rust-lang.org/cargo/reference/config.html).  For example, you can set it globally
+in `$HOME/.cargo/config` by adding:
 
-or use it with rust, like so:
+```toml
+[build]
+rustc-wrapper = "[path to sccache]"
```
froydnj: I think the cargo warning should be addressed here, so that people are not surprised.  (And if there's a minimum version requirement, that'd be good to mention too.)

README.md
orium: I did that because in these two places because it is actually not an executable bash command, since it includes the placeholder `[path to sccache]`.  In any case it doesn't matter much.  I will switch to `bash`.

orium:
@froydnj Can you take a look at the updated PR when you have some time?

froydnj:
Thanks!

PR #634, opened 2020-01-02
resolve relative paths against PATH during packaging

### LINKED ISSUE #389: Automatic toolchain packaging not working properly with gcc outputting relative paths
The code we landed in #321 doesn't seem to be working exactly right. I ran a simple test distributed compile on my Linux machine and got:
```
luser@eye7:/build$ /build/sccache/target/debug/sccache gcc -c hello.c -o hello.o
gcc: error trying to exec 'as': execvp: No such file or directory
Compiler killed by signal 1
```

Looking at the contents of the toolchain tar on disk shows that `as` is missing:
```
$ tar tzf /home/luser/sccache-dist/toolchains/tc/f/f/ff6b8de934e8d03f7bf720597e3f7818bf8e8f21288c6349273e480bd335caa564b7103a7ae2655e6b1f1702a440b4d2f8d212dd11c9c0e9cfcfc7e4be209d34
tmp
etc/ld.so.conf
lib/x86_64-linux-gnu/libc.so.6
lib/x86_64-linux-gnu/libdl.so.2
lib/x86_64-linux-gnu/libm.so.6
lib/x86_64-linux-gnu/libz.so.1
lib64/ld-linux-x86-64.so.2
usr/bin/gcc
usr/lib/gcc/x86_64-linux-gnu/7/cc1
usr/lib/gcc/x86_64-linux-gnu/7/cc1plus
usr/lib/gcc/x86_64-linux-gnu/7/liblto_plugin.so
usr/lib/x86_64-linux-gnu/libgmp.so.10
usr/lib/x86_64-linux-gnu/libisl.so.19
usr/lib/x86_64-linux-gnu/libmpc.so.3
usr/lib/x86_64-linux-gnu/libmpfr.so.6
```

If I look at an older toolchain tar I have that was created by `icecc-create-env` it does include `as`:
```
$ tar tzf /home/luser/sccache-dist/toolchains/tc/6/8/6805131d38b0c2b2f8cf967c15d96e2a5ce23d26cfbc4f60f2750b5f323a4cacffb1f96149e27be05ba927b7cec674e748e0efa1e06ce2eecda4673fd1060a5d
bin/true
etc/ld.so.cache
etc/ld.so.conf
lib/x86_64-linux-gnu/libc.so.6
lib/x86_64-linux-gnu/libdl.so.2
lib/x86_64-linux-gnu/libm.so.6
lib/x86_64-linux-gnu/libz.so.1
lib64/ld-linux-x86-64.so.2
usr/bin/as
usr/bin/cc1
usr/bin/cc1plus
usr/bin/g++
usr/bin/gcc
usr/bin/objcopy
usr/lib/gcc/x86_64-linux-gnu/7/liblto_plugin.so
usr/lib/x86_64-linux-gnu/libbfd-2.30-system.so
usr/lib/x86_64-linux-gnu/libgmp.so.10
usr/lib/x86_64-linux-gnu/libisl.so.19
usr/lib/x86_64-linux-gnu/libmpc.so.3
usr/lib/x86_64-linux-gnu/libmpfr.so.6
usr/lib/x86_64-linux-gnu/libopcodes-2.30-system.so
```

The sccache server's log output contains:
```
 INFO 2019-03-04T14:19:35Z: sccache::compiler::c: Generating toolchain /usr/bin/gcc
DEBUG 2019-03-04T14:19:35Z: sccache::compiler::c: find named prog as output:
as

===

DEBUG 2019-03-04T14:19:35Z: sccache::compiler::c: find named prog objcopy output:
objcopy

===

DEBUG 2019-03-04T14:19:35Z: sccache::compiler::c: find named prog cc1 output:
/usr/lib/gcc/x86_64-linux-gnu/7/cc1

===

DEBUG 2019-03-04T14:19:35Z: sccache::compiler::c: find named prog cc1plus output:
/usr/lib/gcc/x86_64-linux-gnu/7/cc1plus

===

DEBUG 2019-03-04T14:19:35Z: sccache::compiler::c: find named file specs output:
specs

===

DEBUG 2019-03-04T14:19:35Z: sccache::compiler::c: find named file liblto_plugin.so output:
/usr/lib/gcc/x86_64-linux-gnu/7/liblto_plugin.so

===
```

Looking at [the code involved](https://github.com/mozilla/sccache/blob/931cefa86987d538caf503839dd5da6c540368a9/src/compiler/c.rs#L392) it runs `gcc -print-prog-name=as`, which on my system just prints `as`:
```
$ gcc -print-prog-name=as
as
```

The packaging code [doesn't handle non-absolute paths here](https://github.com/mozilla/sccache/blob/931cefa86987d538caf503839dd5da6c540368a9/src/compiler/c.rs#L419), which seems to be the problem.

froydnj:
Fixes #389.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -36,6 +36,7 @@ use std::hash::Hash;
 use std::io;
 use std::path::{Path, PathBuf};
 use std::process;
+use which::which;
 
 use crate::errors::*;
 
@@ -544,12 +545,13 @@ impl pkg::ToolchainPackager for CToolchainPackager {
                 output.stdout.pop();
             }
 
-            // Create our PathBuf from the raw bytes, and return if absolute.
+            // Create our PathBuf from the raw bytes.  Assume that relative
+            // paths can be found via PATH.
             let path: PathBuf = OsString::from_vec(output.stdout).into();
             if path.is_absolute() {
                 Some(path)
             } else {
-                None
+                which(path).ok()
             }
         };
 


src/compiler/c.rs
```diff
@@ -544,12 +545,13 @@ impl pkg::ToolchainPackager for CToolchainPackager {
...
 
-            // Create our PathBuf from the raw bytes, and return if absolute.
+            // Create our PathBuf from the raw bytes.  Assume that relative
+            // paths can be found via PATH.
             let path: PathBuf = OsString::from_vec(output.stdout).into();
             if path.is_absolute() {
                 Some(path)
             } else {
-                None
+                which(path).ok()
```
chmanchester: gcc seems to just echo its `-print-prog-name` argument back to stdout if you don't pass a tool it recognizes, which `as` seems to not be. Are we worried about finding random/wrong assemblers by accident here? Maybe that's not an issue in practice. In any case, if this fixes something you care about go ahead and merge it.

froydnj: I'm not sure what GCC wants, exactly; the GCC 5 I have on my Ubuntu machine just prints `as`, while GCC 6 prints an absolute path to `as`.  I think the original issue was using another, newer GCC that *didn't* print absolute paths, and I don't know why the behavior changed (or if there are custom distribution patches installed).

There are obscure ways to change what GCC looks for here without modifying `PATH`, but I'm not worried about running into them.

PR #642, opened 2020-01-15
compatible with ldd 2.30 Fix #641

### LINKED ISSUE #641: package-toolchain not work on ldd version 2.30
ldd (verson 2.30) prints different entry for interpreter '/lib64/ld-linux-x86-64.so.2' from old versoin.

package-toolchain will package /usr/lib64/ld-linux-x86-64.so.2, but what we need
is /lib64/ld-linux-x86-64.so.2 

this results in a very confusing error message "No such file or directory" when sccache try to run
rustc with bwrap.

```bash
$ ldd --version
ldd (GNU libc) 2.30

$ ldd /usr/bin/ls
        linux-vdso.so.1 (0x00007ffedc39a000)
        libcap.so.2 => /usr/lib/libcap.so.2 (0x00007f7bc85e6000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f7bc841f000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f7bc8649000)
```

```
$ ldd --version
ldd (GNU libc) 2.17

$ ldd /usr/bin/ls
        linux-vdso.so.1 =>  (0x00007ffc8f436000)
        libselinux.so.1 => /lib64/libselinux.so.1 (0x00001490ae68e000)
        libcap.so.2 => /lib64/libcap.so.2 (0x00001490ae489000)
        libacl.so.1 => /lib64/libacl.so.1 (0x00001490ae280000)
        libc.so.6 => /lib64/libc.so.6 (0x00001490adeb3000)
        libpcre.so.1 => /lib64/libpcre.so.1 (0x00001490adc51000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00001490ada4d000)
        /lib64/ld-linux-x86-64.so.2 (0x00001490ae8b5000)
        libattr.so.1 => /lib64/libattr.so.1 (0x00001490ad848000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00001490ad62c000)
```

MoZhonghua:


src/dist/pkg.rs
```diff
@@ -262,7 +262,25 @@ mod toolchain_imp {
...
+                        let basename: &str = match libname.split('/').last() {
+                            Some(p) => p.into(),
+                            None => libname,
+                        };
+                        if basename.starts_with("ld-linux") {
+                            libs.push(PathBuf::from(libname));
+                        }
+                    }
+
+                    PathBuf::from(libpath)
```
froydnj: What extra case is being covered here?  I guess `ldd` is now explicitly showing that `/lib64/ld-linux-x86-64.so.2` is a symlink, and so it's not enough to package up `libpath`, because binaries also need the original symlink to exist as well?

If this is the case, it seems like it would be more correct to always `libs.push(PathBuf::from(libname)` if `libname` was absolute, rather than special-casing `ld-linux.*`.

MoZhonghua: I modify the code to add libname to deps if it's absolute and exists.

--------------
I use archlinux with ldd 2.30 as dev machine and centos 7 with ldd 2.17 as build server.

In archlinux, there are two lib64/ld-linux-x86-64.so.2:
- /lib64/lib64/ld-linux-x86-64.so.2 -> /lib64/ld-2.30.so
- /usr/lib64/ld-linux-x86-64.so.2  -> /usr/lib64/ld-linux-x86-64.so.2

ld-linux-x86-64.so.2 is not an ordinary dynlib but  dynamic linker, whose location
is hard coded in ELF binaries, like this:

```bash
$ readelf -a /usr/bin/ls | grep  interpreter
     [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
````

ldd(2.30) will use following command to get dynlibs:

```bash
LD_TRACE_LOADED_OBJECTS=1 LD_VERBOSE= /usr/lib64/ld-linux-x86-64.so.2 /usr/bin/ls
```
Note this command invokes dynamic linker explicitly, and there is a predefined list  of in ldd
source code. The output of  ld-linux-x86-64.so.2 is different on centos7 and archlinux.

Old output (centos 7)

```bash
        libpcre.so.0 => /usr/local/lib/libpcre.so.0 (0x00002b75eac85000)
        libc.so.6 => /lib64/libc.so.6 (0x000000378d200000)
        /lib64/ld-linux-x86-64.so.2 (0x000000378ce00000)
```

New output (archlinux)

```bash
        linux-vdso.so.1 (0x00007ffd5029e000)
        libcap.so.2 => /usr/lib/libcap.so.2 (0x00007f1abd8ba000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f1abd6f3000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f1abd91a000)
```

froydnj:
Thanks for the changes; would you please write a test featuring the output?  You can see examples of how to write such a test in the same file that you modified.

Issue #504, opened 2019-08-27
Reevaluate hash function used


apriori:
As seen [here](https://github.com/mozilla/sccache/blob/2448687217463b1d77b451ac7f63e5a4614693e7/src/util.rs#L41) SHA512 is used for file hashing. I wonder why, cryptographic quality of the hash function should be secondary, but instead the focus should be on performance.

I think the benchmark should be to offer similar performance to [CCache](https://github.com/ccache/ccache/) which [apparently](https://github.com/ccache/ccache/blob/9df29f16e2f64c1390da4bb77100c73fccc8416b/src/third_party/xxhash.h#L42) uses [xxhash](https://github.com/Cyan4973/xxHash).

I assume the choice for SHA512 was made due to implementation simplicity (its already there in the rust eco system) and robustness. Hashing on my laptop (i7-7700HQ CPU @ 2.80GHz) accounts for up to 60% of the same runtime as full compilation, which explains the somewhat bad speed up factor of ~2.3.

Apparently the rust eco system has an [implementation](https://github.com/shepmaster/twox-hash) for xxhash. Maybe that one could be used as a first step (should be easy to achieve), if one reluctantly wants to start researching the topic of "best hashing functions".

apriori:
I of course did not want to imply anything like that. I will see whether I can quickly hack up a fork using the mentioned and implemention and see how that performs for my project.

apriori:
Another candidate: https://github.com/eldruin/wyhash-rs

michaelwoerister:
We've used BLAKE2 in the Rust compiler for a while and it was pretty performant. We later switched to a 128 bit SipHash because it's even faster, 128 bits are enough for our purposes, and we don't need a full blown cryptographic hash function. The fastest function in our tests was [MetroHash](https://docs.rs/metrohash/1.0.6/metrohash/) but it's probably less tested than SipHash and supports 128 bit hashes at most.

apriori:
While a performant hash function is not irrelevant, testing shows runtime gets completely dominated by preprocessing overhead, to an extend that you would not even notice a super slow hash function.

luser:
I would be surprised to find that using a faster hash function made a noticeable difference in any performance testing but if someone wanted to write a patch and gather some data that would be a good start. Changing the hash function should be trivial in practice, everything goes through the methods linked in the original comment on this issue.

apriori:
[...]

PR #646, opened 2020-01-17
Switch from SHA-512 to BLAKE3

GeorgeHahn:
BLAKE3 is designed to be a very high performance cryptographic hash. The BLAKE3 team has shown significantly higher single-thread performance than SHA-512 on modern server hardware [1].

This change resulted in a minor improvement to observed local build times.

### Comparison

`cargo check` on sccache repo (at commit 32e40bba338020ae978cea5698aba35b4a710a88) with a warm cache for each hash. Ten samples were collected for each hash.

SHA512 mean: 60.06 seconds
BLAKE3 mean: 57.77 seconds

This data shows a mean improvement of 2.29 seconds (about 4%). My statistics skills are rough, but this appears to be a very statistically significant improvement.

This test was performed on an Intel i5-7600K and a fairly ancient SATA3 SSD. It would be interesting to see a test result from a higher performance machine. The BLAKE3 team's performance analysis shows 8.5x faster single threaded performance vs SHA-512 on an AWS `c5d.metal` instance.

<details>
<summary>Raw data</summary>
sha512 initial cold build: 1m 47s<br>
blake3 initial cold build: 1m 52s<br>
sha512 build times: 57.02s, 1m 03s, 1m 02s, 58.72s, 1m 00s, 59.48s, 1m 02s, 1m 01s, 58.54s, 58.85s<br>
blake3 build times: 54.87s, 59.82s, 56.72s, 56.91s, 1m 00s, 56.46s, 57.85s, 57.13s, 59.35s, 58.57s<br>
</details>

### Related issues & PRs:

- [#504: Reevaluate hash function used](https://github.com/mozilla/sccache/issues/504)
- [#310: Use RustCrypto crates](https://github.com/mozilla/sccache/pull/310)
- [#108: Faster hash algorithm?](https://github.com/mozilla/sccache/issues/108)
- [#109: Switch from SHA-1 to SHA-512](https://github.com/mozilla/sccache/pull/109)

### References:

1. [Blake3 repo](https://github.com/BLAKE3-team/BLAKE3)
2. [Blake3 paper](https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/blake3.pdf)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,7 +153,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "arrayref 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "arrayvec 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "constant_time_eq 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "constant_time_eq 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "blake3"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "arrayref 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "arrayvec 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cc 1.0.48 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "constant_time_eq 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -292,7 +304,7 @@ dependencies = [
 
 [[package]]
 name = "constant_time_eq"
-version = "0.1.4"
+version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


luser:
Per the other discussions I'm generally supportive of this. Two minor notes:
1. You should be able to make `ring` an optional dependency with this change since only the `gcs` feature will rely on it. That would fix some of the issues that motivated other people to look into replacing `ring` (as mentioned in #310).
2. For completeness' sake you probably ought to bump the instances of `CACHE_VERSION` to make it very clear that the keys are incompatible, although given that the digest length will change maybe this is overkill:
https://github.com/mozilla/sccache/blob/32e40bba338020ae978cea5698aba35b4a710a88/src/compiler/c.rs#L611
https://github.com/mozilla/sccache/blob/32e40bba338020ae978cea5698aba35b4a710a88/src/compiler/rust.rs#L190

GeorgeHahn:
Done & done, thanks for the comments.

froydnj:
An actual speedup from this switch sounds fantastic!

Just to play devil's advocate for a bit: my understanding of why SHA-512 was used is that we wanted something with actual collision resistance, and so we used a cryptographically secure hash instead of something like MurmurHash, xxhash, etc. etc.

But now we're going to replace SHA-512 with a hash that was published ~months ago and hasn't been subjected to analysis outside of the designers (AFAICT; I see the paper does include some results for BLAKE2 and why the designers felt comfortable with differing parameters for BLAKE3 based on the BLAKE2 results, but this is not quite the same thing as analysis of BLAKE3 itself)?  So we're replacing a known-good hash function with one that hasn't been subjected to nearly as much scrutiny?

I don't think this means that we'd reject this pull request, but I'd like us to be clear on the tradeoffs we're making.

jrmuizel:
For my curiosity, what's the threat model for a collision attack? A shared untrusted object cache? Do we include a hash of the entire toolchain in our hashes?

luser:
> Do we include a hash of the entire toolchain in our hashes?

For C compiles we only include a hash of the compiler binary but for Rust compiles we include a hash of all the shared libraries under `$SYSROOT/lib` (since the `rustc` binary is a shim in the rustup case, and a tiny wrapper around the implementation in the shared libraries in any case).

GeorgeHahn:
@froydnj That's a great outline of the hash change, thanks.

BLAKE3 aims to be secure against collision attacks and should offer excellent collision resistance for cached objects. However, there are currently no third-party assurances that BLAKE3's collision resistance will hold up to malicious actors.

Is a 'shared untrusted cache' scenario is within the planned threat model? If so, it may be desirable to wait for a cryptanalysis of BLAKE3.

luser:
> Is a 'shared untrusted cache' scenario is within the planned threat model? If so, it may be desirable to wait for a cryptanalysis of BLAKE3.

I really don't think so. sccache fully trusts the cache contents and does no verification of them. A malicious actor with write access to the cache can already easily poison the cache using known hashes without needing to find a hash collision.

drahnr:
is there any work item for this to be merged?

froydnj:
Thanks!

Issue #552, opened 2019-10-22
Use Zstd instead of gzip for compression


jrmuizel:
We're currently spending an appreciable amount of cpu time doing zlib compression:
```
  32.67%  sccache-dist     sccache-dist        [.] miniz_oxide::deflate::core::compress_inner
  11.67%  sccache-dist     sccache-dist        [.] miniz_oxide::deflate::core::compress_block
```
Using Zstd should give better compression and better performance.

jrmuizel:
FWIW, icecream uses lzo which is also a lot faster.

froydnj:
Isn't that just because `miniz_oxide` is written in pure Rust without the massive optimization work that's gone into zlib?

jrmuizel:
I'm not sure how the performance of miniz_oxide compares to zlib, but even zlib is appreciably slower than Zstd.

luser:
Currently cache entries are just zip files. It wouldn't be hard to change that, all the code for creating and unpacking cache entries is in this file:
https://github.com/mozilla/sccache/blob/d3093392af6f593361c229deb36dd14c964c0778/src/cache/cache.rs

You'd need a container format of some sort since there can be multiple output files + stdout/stderr, but it should be straightforward to use tar files.

luser:
Oh, it strikes me that since you're using `sccache-dist` there this is probably toolchain/inputs/outputs packaging.

Toolchain:
https://github.com/mozilla/sccache/blob/d3093392af6f593361c229deb36dd14c964c0778/src/dist/pkg.rs#L146-L159

Inputs packaging:
https://github.com/mozilla/sccache/blob/0cb2e77cd005ea027fcd102ade6b6158bbc765b0/src/dist/http.rs#L1246
Inputs unpackaging:
https://github.com/mozilla/sccache/blob/d3093392af6f593361c229deb36dd14c964c0778/src/dist/http.rs#L980

Outputs are handled here:
https://github.com/mozilla/sccache/blob/d3093392af6f593361c229deb36dd14c964c0778/src/dist/mod.rs#L424-L446

Unfortunately if you change this you'll either need to support both compression formats for interop or change the whole dist API to a v2.

luser:
Related, while looking at that code I realized that at least for Rust compilation packaging compile inputs is probably doing a lot of redundant work so I filed this issue on a potential optimization: https://github.com/mozilla/sccache/issues/558

froydnj:
[...]

PR #649, opened 2020-01-22
Add a feature enabling native zlib for cache entries.

chmanchester:
This is just one of a few things to try related to #552. As a first step it
would be interesting to see what kind of impact this has on Firefox automation
builds. Some initial measurements are promising for builds with lots of cache
misses, but this sort of thing is cumbersome to measure, so I'd like to try
enabling this for the sccache client built in Mozilla automation to see the
impact it has there.


chmanchester:
> Is it correct to assume that we're not enabling this for `flate2` (and therefore for `sccache-dist` requests) because we'd like to gather data on how this performs solely on cache entries first, and then make a decision informed by that?

Yes, that's my intention for now. Thanks for the review!

PR #657, opened 2020-01-28
Don't serialize inputs compression in the dist client

chmanchester:


chmanchester:
This isn't doing quite what I thought. It might be slightly helpful but will require further investigation.

PR #658, opened 2020-02-03
rust: workaround --print file-names emitting staticlibs / dylibs.

emilio:
This fixes cargo check in mozilla-central. The issue is that rustc --print
file-names emits a somewhat poor approximation of what's actually going to be
emitted.

So for a staticlib crate, it will also print the staticlib file, which is not
great.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1612855#c2 for a more
straight-forward explanation of the failure case.

Sccache would try to find the library and fail, erroring as a consequence.

Pile up on the existing workaround for rmeta files not showing up
(https://github.com/rust-lang/rust/issues/54852) by removing files that are not
metadata when we only request metadata.

https://github.com/rust-lang/rust/pull/68799 contains a rust-side fix that would
also fix this.

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -1165,6 +1165,26 @@ where
                         &env_vars,
                     )
                     .map(move |mut outputs| {
+                        // metadata / dep-info don't ever generate binaries, but
+                        // rustc still makes them appear in the --print
+                        // file-names output (see
+                        // https://github.com/rust-lang/rust/pull/68799).
+                        //
+                        // So if we see a binary in the rustc output and figure
+                        // out that we're not _actually_ generating it, then we
+                        // can avoid generating everything that isn't an rlib /
+                        // rmeta.
+                        //
+                        // This can go away once the above rustc PR makes it in.
+                        let emit_generates_only_metadata =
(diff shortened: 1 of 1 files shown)


froydnj:
I think we do have Rust version information at some point (maybe propagating it to here is hard), but maybe we should consider making this change conditional on the compiler version so that it's easy to detect regressions in the future?

emilio:
> I think we do have Rust version information at some point (maybe propagating it to here is hard), but maybe we should consider making this change conditional on the compiler version so that it's easy to detect regressions in the future?

This works regardless of the other fix being in rust or not... Not sure what a condition for this workaround would look like.

froydnj:
> > I think we do have Rust version information at some point (maybe propagating it to here is hard), but maybe we should consider making this change conditional on the compiler version so that it's easy to detect regressions in the future?
> 
> This works regardless of the other fix being in rust or not... Not sure what a condition for this workaround would look like.

I meant that `if rustc_version >= fixed_rustc_version`, we would not apply the workaround.  It probably doesn't matter that much, but it would help people detect if `rustc` suddenly broke again.

emilio:
Ah, that's fair... We'd need to fix rustc first though :P

PR #665, opened 2020-02-15
Update deps: remove tempdir and use log 0.4 for lru-disk-cache

Dushistov:
log 0.4 already used by main crate, so I do not see any reason to use previous version for lru-disk-cache. Plus tempdir deprecated and merge with tempfile and tempfile already used by sccache, so tempdir can be dropped and replaced by `tempfile::TempDir`


Dushistov:
Ping @chmanchester

PR #674, opened 2020-02-27
Support both -arg and /arg with msvc/clang-cl without altering paths

### LINKED ISSUE #669: sccache doesn't support clang-cl on non-Windows
This code: https://github.com/mozilla/sccache/blob/9d1346556378851fc4047387da2bcd9421cbee29/src/compiler/msvc.rs#L273-L282

unconditionally transforms all arguments starting with a `/` to start with a `-`, which is fine for both MSVC and clang-cl when they run on Windows, but breaks things when running clang-cl on e.g. Linux.

glandium:
34c6f9a added support for /arg flags by replacing the initial slash with
a -, and falling back to normal argument parsing. Unfortunately, when
running clang-cl on a non-Windows system, that means absolute paths get
garbled, and that can lead to bad argument parsing with separate
arguments (e.g. -I /absolute/path).

We change the arguments definition for MSVC to include both styles, by
hacking up a macro that generates two lists, one with - arguments, and
one with / arguments. Merging the two lists at build time doesn't seem
possible without going full procedural macro.

Fixes #669

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -227,29 +227,43 @@ ArgData! {
 
 use self::ArgData::*;
 
-counted_array!(static ARGS: [ArgInfo<ArgData>; _] = [
-    take_arg!("-D", OsString, Concatenated, PreprocessorArgument),
-    take_arg!("-FA", OsString, Concatenated, TooHard),
-    take_arg!("-FI", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
-    take_arg!("-FR", PathBuf, Concatenated, TooHardPath),
-    take_arg!("-Fa", PathBuf, Concatenated, TooHardPath),
-    take_arg!("-Fd", PathBuf, Concatenated, ProgramDatabase),
-    take_arg!("-Fe", PathBuf, Concatenated, TooHardPath),
-    take_arg!("-Fi", PathBuf, Concatenated, TooHardPath),
-    take_arg!("-Fm", PathBuf, Concatenated, TooHardPath),
-    take_arg!("-Fo", PathBuf, Concatenated, Output),
-    take_arg!("-Fp", PathBuf, Concatenated, TooHardPath),
-    take_arg!("-Fr", PathBuf, Concatenated, TooHardPath),
-    flag!("-Fx", TooHardFlag),
-    take_arg!("-I", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
-    take_arg!("-U", OsString, Concatenated, PreprocessorArgument),
-    take_arg!("-Xclang", OsString, Separated, XClang),
-    flag!("-Zi", DebugInfo),
-    flag!("-c", DoCompilation),
-    take_arg!("-deps", PathBuf, Concatenated, DepFile),
-    flag!("-fsyntax-only", TooHardFlag),
-    take_arg!("-o", PathBuf, Separated, Output), // Deprecated but valid
-    flag!("-showIncludes", ShowIncludes),
+macro_rules! msvc_args {
+    (static ARGS: [$t:ty; _] = [$($macro:ident ! ($($v:tt)*),)*]) => {
+        counted_array!(static ARGS: [$t; _] = [$(msvc_args!(@one "-", $macro!($($v)*)),)*]);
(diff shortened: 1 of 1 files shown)


emilio:
> Merging the two lists at build time doesn't seem
possible without going full procedural macro.

Doesn't this achieve what you want?

```diff
diff --git a/src/compiler/msvc.rs b/src/compiler/msvc.rs
index 3302856..7197b8c 100644
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -229,8 +229,10 @@ use self::ArgData::*;
 
 macro_rules! msvc_args {
     (static ARGS: [$t:ty; _] = [$($macro:ident ! ($($v:tt)*),)*]) => {
-        counted_array!(static ARGS: [$t; _] = [$(msvc_args!(@one "-", $macro!($($v)*)),)*]);
-        counted_array!(static SLASH_ARGS: [$t; _] = [$(msvc_args!(@one "/", $macro!($($v)*)),)*]);
+        counted_array!(static ARGS: [$t; _] = [
+            $(msvc_args!(@one "-", $macro!($($v)*)),)*
+            $(msvc_args!(@one "/", $macro!($($v)*)),)*
+        ]);
     };
     (@one $prefix:expr, msvc_take_arg!($s:expr, $($t:tt)*)) => {
         take_arg!(concat!($prefix, $s), $($t)+)
@@ -284,7 +286,7 @@ pub fn parse_arguments(
     let mut show_includes = false;
     let mut xclangs: Vec<OsString> = vec![];
 
-    for arg in ArgsIter::new(arguments.iter().cloned(), (&ARGS[..], &SLASH_ARGS[..])) {
+    for arg in ArgsIter::new(arguments.iter().cloned(), &ARGS[..]) {
         let arg = try_or_cannot_cache!(arg, "argument parse");
         match arg.get_data() {
             Some(TooHardFlag) | Some(TooHard(_)) | Some(TooHardPath(_)) => {
```

froydnj:
This doesn't work because of the requirement of `ARGS` to be sorted.  (I'm having trouble figuring out where we actually use this fact, but glandium pointed out the same thing to me when I made this argument last night and we do have tests for non-sorted arg arrays panic'ing, so I assume it must make a difference somewhere...)

glandium:
We're using a binary search in `args.rs`, which requires the list to be sorted. @emilio's diff would work if there wasn't the "@" argument. I tried something similar to that diff with a scheme to eliminate non-msvc_* arguments in prefixed passes and a final non-prefixed pass without the msvc_* arguments, but all I ended up with was this tweet: https://twitter.com/MikeHommey/status/1232932709409771522

peterjc123:
Caching is broken for commands like `cl /EHsc /c a.cpp` after this PR.
Related log:
```
DEBUG 2020-04-16T08:04:08Z: sccache::server: handle_client: compile
DEBUG 2020-04-16T08:04:08Z: sccache::server: check_compiler: Supported compiler
DEBUG 2020-04-16T08:04:08Z: sccache::server: parse_arguments: CannotCache(multiple input files): ["/c", "/EHsc", "a.cpp"]
```
But if I change to `cl -EHsc /c a.cpp`, it is working.
```
DEBUG 2020-04-16T08:02:45Z: sccache::server: handle_client: compile
DEBUG 2020-04-16T08:02:45Z: sccache::server: check_compiler: Supported compiler
DEBUG 2020-04-16T08:02:45Z: sccache::server: parse_arguments: Ok: ["/c", "-EHsc", "a.cpp"]
DEBUG 2020-04-16T08:02:45Z: sccache::compiler::compiler: [a.obj]: get_cached_or_compile: ["/c", "-EHsc", "a.cpp"]
DEBUG 2020-04-16T08:02:45Z: sccache::compiler::msvc: preprocess: Some("C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Enterprise\\VC\\Tools\\MSVC\\14.16.27023\\bin\\HostX64\\x64\\cl.EXE" "-E" "a.cpp" "-nologo" "-EHsc")
DEBUG 2020-04-16T08:02:45Z: sccache::compiler::compiler: [a.obj]: generate_hash_key took 0.232 s
```

glandium:
Please open a new issue.

peterjc123:
@glandium No problem.

PR #684, opened 2020-03-05
Added support for -z linker option.

edge90:
Without it we get `multiple input files` error.


edge90:
Added a test :)

froydnj:
Thank you!

PR #686, opened 2020-03-06
Fix many clippy warnings

ry:
This does not fix all of the warnings from clippy, only several dozen. 

Before committing more time to do the rest I want to make sure that this sort of PR will get merged : )


froydnj:
Thanks!

PR #691, opened 2020-03-09
Fix remaining clippy warnings

ry:


src/protocol.rs
```diff
@@ -23,11 +23,11 @@ pub enum Response {
     /// Response for `Request::Compile`.
     Compile(CompileResponse),
     /// Response for `Request::GetStats`, containing server statistics.
-    Stats(ServerInfo),
+    Stats(Box<ServerInfo>),
```
froydnj: Is this (and the `DistClientState` changes related to `Box`) due to `large_enum_variant`?

ry: yes

ry:
Ok I think this is ready for review now. It seems to be clippy clean for `--all-targets` and `--no-default-features`.

ry:
@froydnj This PR should be ready to land (the two commits can be squashed). I think it improves the code base and allows us to check clippy for new errors when making new changes. PTAL.

ry:
squashed and updated for Rust 1.42.0 ...

PR #692, opened 2020-03-09
Add github actions CI

ry:
All three operating systems build and test in about 10 minutes:
https://github.com/ry/sccache/runs/495956105

This does not include deployment. I figure we should try out GHA for a while before switching over completely to it. I can add this in a follow up PR - it would be modified from here: https://github.com/denoland/deno/blob/7f591c37835be55b8a426cca61e677fd62c23c93/.github/workflows/ci.yml#L175-L204

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,40 @@
+name: ci
+on: [push, pull_request]
+jobs:
+  build:
+    name: ${{ matrix.kind }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        os: [macOS-latest, windows-2019, ubuntu-16.04]
+        kind: ['test_debug']
+    steps:
+      - name: Clone repository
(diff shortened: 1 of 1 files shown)


ry:
@froydnj I think it would be prudent to keep travis/appveyor around until we're confident that GHA is working for everyone. So for now this would just be extra information for contributors, but in the future GHA can replace travis/appveyor.

luser:
Neat! If you get around to wanting to do deployment note that I created @sccachereleasebot for the existing release builds to avoid giving them full access to my GitHub account. I'd be happy to share access to that account with any of y'all that are currently maintaining sccache, or I can generate encrypted tokens or whatever you need on-demand to make it work. I don't think it required any changes after the first time I set it all up, FWIW.

froydnj:
> @froydnj I think it would be prudent to keep travis/appveyor around until we're confident that GHA is working for everyone. So yes, a future PR would remove travis/appveyor.

OK, thanks.  In that case, we can just iterate on fidelity to the existing testing matrix.  Thank you!

PR #695, opened 2020-03-11
azure: container_name shouldn't be optional

ry:
It's not clear to me if there is a purpose to having the container_name be optional for the Azure backend. If you don't have a container, how could it work? 

My ultimate goal is to enable anonymous reads for the Azure backend - I'm submitting this PR just to make sure I'm not misunderstanding something crucial. Any advice towards this ultimate goal would also be appreciated. In particular I'm wondering if we can default to `https://$ACCOUNT_NAME.blob.core.windows.net/$CONTAINER/` if the `$SCCACHE_AZURE_CONNECTION_STRING` is not set in the env (signifying an anonymous read).

cc @benjamin-bader

### DIFF
--- a/src/azure/blobstore.rs
+++ b/src/azure/blobstore.rs
@@ -61,14 +61,13 @@ impl fmt::Display for BlobContainer {
 }
 
 impl BlobContainer {
-    pub fn new(base_url: &str, container_name: &Option<String>) -> Result<BlobContainer> {
-        let container_url = match container_name {
-            &Some(ref name) => format!("{}{}/", base_url, name), // base_url is assumed to end in a trailing slash
-            &None => base_url.to_owned(),
-        };
-
+    pub fn new(base_url: &str, container_name: &str) -> Result<BlobContainer> {
+        assert!(
+            base_url.ends_with('/'),
+            "base_url is assumed to end in a trailing slash"
+        );
         Ok(BlobContainer {
-            url: container_url,
+            url: format!("{}{}/", base_url, container_name),
             client: Client::new(),
         })
     }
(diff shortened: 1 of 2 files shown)


benjamin-bader:
This makes sense.  It's been a while since I wrote this code so I don't remember just what I was thinking - probably I saw that _blob URLs_ have omit container names under some circumstances (see https://docs.microsoft.com/en-us/rest/api/storageservices/working-with-the-root-container), and got that confused with what I was actually working on.

You're right, it doesn't make sense for the container name to be optional, and probably doesn't work today with the container name missing.  Even if you wanted to use the root container, for writing you'd need `$root` as the container name.

I'm not sure I follow your goal here, so I'll paraphrase what I think I read - presumably you've got an environment that already has `ACCOUNT_NAME` and `CONTAINER` env vars defined, and you don't want to configure `SCCACHE_AZURE_CONNECTION_STRING`, but instead want sccache to infer from the two vars you already have.  Is that right?

Personally I've no objection to that.

ry:
Thanks for taking a look Ben.

>  Is that right?

Yes that's right. I want to use Azure Blob Storage to get free transfer bandwidth between Github Actions. In Github when people submit a PR, I don't want to allow them to write to the blob storage, this would be a security hole. But I do want them to be able to read from the blob storage and get fast builds. So I was thinking about allow sccache to accept a modified `SCCACHE_AZURE_CONNECTION_STRING` that doesn't include AccountKey ...  Or as you suggest have a new `AZURE_ACCOUNT_NAME` env var.

benjamin-bader:
> I want to use Azure Blob Storage to get free transfer bandwidth between Github Actions

Clever!  This seems totally doable - we'd just need to make writes a no-op in anonymous mode.

luser:
To get the behavior you want someone might have to fix the long-standing #30. Currently sccache only supports a single cache storage location and assumes it can both read from and write to it. We wanted to support something very similar to what you describe at Mozilla to allow "try" builds (very similar to builds from PRs) to get cache hits from the cache that CI builds of master write to, and also potentially let developers building on their local machines get cache hits from the CI cache (if we could resolve all of the path issues etc).

ry:
@luser Two tier caching would be extra great, but anonymous reads work fine for me with sccache s3 currently. I just want to reduce cross-datacenter bandwidth by doing the same on Azure. 

(Two-tier caching is something my colleague @piscisaureus and I have discussed quite a bit. Potentially this is something we can help with in the future.)

ry:
@froydnj This PR should already be good to land. I will follow up with a larger one for anonymous read on Azure.

froydnj:
Merging this based largely on @benjamin-bader's comments (thank you!), since I have zero Azure API knowledge.

Issue #680, opened 2020-03-05
Failing to compile with __rt missing in quote


drahnr:
```
error[E0433]: failed to resolve: could not find `__rt` in `quote`
   --> /home/bernhard/.cargo/registry/src/github.com-1ecc6299db9ec823/failure_derive-0.1.6/src/lib.rs:107:70
    |
107 | fn display_body(s: &synstructure::Structure) -> Result<Option<quote::__rt::TokenStream>, Error> {
    |                                                                      ^^^^ could not find `__rt` in `quote`

error: aborting due to previous error
```
is caused by

https://github.com/rust-lang-nursery/failure/issues/342

`cargo install --locked sccache` is a temporary workaround

PR #697, opened 2020-03-12
Don't lock quote to a version.

emilio:


froydnj:
So we locked quote because of #680.  Have things been fixed?

emilio:
Yes, I updated failure_derive and quote just to prove it ;)

PR #698, opened 2020-03-12
clang: Add support for -debug-info-kind.

emilio:
This is used by chromium for example.


froydnj:
Builds appear to be green; the beta failure is apparently because recent beta broke the `blake3` crate and/or we need to upgrade.

PR #700, opened 2020-03-12
CI: remove beta build

### LINKED ISSUE #699: CI red after 1.42.0 release
https://travis-ci.org/github/mozilla/sccache/jobs/661612775?utm_medium=notification&utm_source=github_status

This is because blake3 = "0.1.1" does not compile under rustc 1.43.0-beta.1.

As far as I'm concerned, removing the Rust beta build for now would be a reasonable solution:
https://github.com/mozilla/sccache/blob/86be547d0f8c3ba78d1eb7076c48d996dc708a73/.travis.yml#L13

ry:
The beta build is broken since Rust 1.42.0 was released because blake
0.1.1 is broken in 1.43.0-beta.1. Since sccache is an executable, not a
library, it's not very useful to test the build under many different
versions of Rust.

Fixes #699

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ matrix:
     - rust: 1.36.0 # oldest supported version, keep in sync with README.md
     - rust: 1.36.0
       env: DIST_SCCACHE=1
-    - rust: beta
     - rust: nightly
 
     # deployments


froydnj:
I understand the sentiment, but it sounds like there is a fixed version of `blake` ([0.1.5](https://github.com/BLAKE3-team/BLAKE3/issues/57#issuecomment-584356042)) and I think I would rather address issues as they come along, rather than stable suddenly turning builds red and having to figure out what's going on with several crates.

Can we just update the version of `blake` instead?

ry:
Sure, but I'd be interested in knowing what the benefit to supporting nightly, beta, and 1.36 are when sccache itself has no dependents. It's just more to manage - for example we cannot upgrade to the latest blake3 0.2.2 because we must also be backwards compatible with Rust 1.36. Likely the travis config for sccache was just copied from another Rust library which has actual concerns about dependents.

https://github.com/mozilla/sccache/pull/701

PR #701, opened 2020-03-12
upgrade: blake3 0.1.5

ry:
can't upgrade to latest version of blake3 because it doesn't support 1.36.0 (which sscache must)


froydnj:
Thank you!

PR #703, opened 2020-03-13
Remove github actions because Mozilla has a legacy repo

ry:


froydnj:
I think we still want these so people can get CI on their project branches, which might inform them of (at least) Windows errors sooner rather than later.

luser:
> I think we still want these so people can get CI on their project branches, which might inform them of (at least) Windows errors sooner rather than later.

I'm guessing that the reporting from the existing AppVeyor builds got broken when I was removed from the Mozilla GitHub organization? You might ask if Mozilla has an existing AppVeyor organization and if so enable sccache CI through that. [It's definitely still running builds via my account](https://ci.appveyor.com/project/luser/sccache2), it just doesn't seem to be reporting their status on PRs, probably because my account no longer has permission to do so. The releases still work because I explicitly added  @sccachereleasebot as a contributor to this project.

PR #706, opened 2020-03-13
azure: allow anonymous reads when account key is omitted

ry:
cc @benjamin-bader

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -272,7 +272,14 @@ mod test {
             "-debug-info-kind=constructor"
         );
         assert_eq!(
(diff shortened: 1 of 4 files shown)


src/compiler/clang.rs
```diff
@@ -272,7 +272,14 @@ mod test {
             "-debug-info-kind=constructor"
         );
         assert_eq!(
-            ovec!["-Xclang", "-mllvm", "-Xclang", "-instcombine-lower-dbg-declare=0", "-Xclang", "-debug-info-kind=constructor"],
```
benjamin-bader: Did you intend to include this change?  It looks fine but seems unrelated to the rest of the PR.

benjamin-bader:
This looks good to me!

src/compiler/clang.rs
ry: It's caused by "cargo fmt" - I will revert it

froydnj:
Thanks!

PR #707, opened 2020-03-14
Revert "use tokio_reactor::Handle::default"

piscisaureus:
The reverted commit introduced the following problem on Windows:

  Starting sccache server...
  error: Timed out waiting for server startup

This reverts commit b43d51fa2e536019c113cc79cd98dcc2861afad4.


froydnj:
This got landed in #734 without breaking Windows CI.

Issue #696, opened 2020-03-11
launching cargo from `build.rs` fails


drahnr:
Extraction from #676 :

I have yet to debug through this, the assumption currently is as you say: the build.rs is compiled via sccache-dist, the Cargo.toml it tries launch cargo for does not exist, hence it fails.
The environment is passed (as in not modified, besides the target dir IIRC) to the subprocess. So that cargo instance launched from the build.rs is also aware of the env var RUSTC_WRAPPER, but it never gets that far.
Hence my idea to exclude the build.rs by default, since it can contain all kind of mockery depending on the project setup and specifics (read git sha, inject things based on the current machine into the binary - while I do not recommend any of these, they are pretty common unfortunately).

So the issue that causes the failure is a different than the one I spotted:
https://github.com/paritytech/substrate/blob/master/bin/node/runtime/build.rs
https://github.com/paritytech/substrate/blob/master/utils/wasm-builder/src/prerequisites.rs#L63-L65
https://github.com/paritytech/substrate/blob/master/utils/wasm-builder/src/lib.rs#L188-L194
which would fail as I described before, but that is worth a separate ticket I guess

drahnr:
```
error: failed to run custom build command for `node-runtime v2.0.0-alpha.3 (/home/bernhard/Projects/substrate/bin/node/runtime)`

Caused by:
  process didn't exit successfully: `/home/bernhard/Projects/substrate/target/debug/build/node-runtime-76b70fc97179952e/build-script-build` (exit code: 1)
--- stdout
Executing build command: "rustup" "run" "nightly" "cargo" "rustc" "--target=wasm32-unknown-unknown" "--manifest-path=/home/bernhard/Projects/substrate/target/debug/wbuild/node-runtime/Cargo.toml" "--color=always" "--release"
Compiler killed by signal 1
Compiler killed by signal 1
```

Issue #675, opened 2020-02-28
add ability to skip caching/distributing for certain files


froydnj:
Google's `goma` has a configuration option to force local compilation for certain filenames (e.g. `conftest.c`).  I have definitely noticed slower configure times in Firefox when `sccache` is enabled because of the extra preprocessing + distributed compilation; it would be nice if we could avoid that.

(This is a little complicated because parts of Firefox's configure [don't use fixed filenames](https://searchfox.org/mozilla-central/rev/b2ccce862ef38d0d150fcac2b597f7f20091a0c7/build/moz.configure/util.configure#214), but maybe we could figure out a reasonable way of accommodating that as well.)

PR #710, opened 2020-03-17
[dist] feat/rust: blacklist support for rust builds

drahnr:
Tracking PR handling issues of

sense right now, the negative impact is pretty big for a config key change
- [x] issue #696 - black list for `Cargo.toml` provided `build = x.rs`
- [x] add environment and config file based configuration

~Prerequisistes: PR #666~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2836,9 +2836,9 @@ dependencies = [
 
 [[package]]
 name = "toml"
-version = "0.4.10"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "758664fc71a3a69038656bee8b6be6477d2a6c315a6b81f7081f591bffa4111f"
+checksum = "ffc92d160b1eef40665be3a05630d003936a3bc7da7421277846c2613e92c71a"
 dependencies = [
  "serde",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2103,7 +2103,6 @@ dependencies = [
  "serde_json",
  "sha-1",
  "sha2",
- "spin",
  "strip-ansi-escapes",
  "syslog",
  "tar",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,6 +403,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "384f8c53175c890920b6e0127b730709d2a173ca6c4dfdc81618ac9b46f648fe"
 
 [[package]]
+name = "coz-temporary"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1268a072653d044a6049f820e417af5778b75fe4922184dc9e4d08837575905c"
+dependencies = [
+ "libc",
+ "once_cell",
+]
+
+[[package]]
 name = "crc32fast"
 version = "1.2.0"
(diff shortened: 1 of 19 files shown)


froydnj:
FWIW, this is what I did for testing a patch for #675 solely for tests made during configure:

https://github.com/froydnj/sccache/commit/0325ae73276d2d17c9f41213525d941c99aeeef0

It is in some ways a hack, but I think it's a more elegant hack than trying to thread something through `CompilerArguments`.

drahnr:
After a brief look, I am not sure this solves the issue - there is still the injection within `parse_arguments` which one would need for rust too, so I'd be still in favour of fiddling an `RustBlacklist` which implements `trait Blacklist` as an `Option<dyn Blacklist>` through those two `fn` layers.

froydnj:
Sure, you’d have to do something similar for Rust as well.    I was just trying to point out what I thought might be a better way forward.

I tried to do something at the `CompilerArguments`/`generate_hash_key` layer and found myself getting very frustrated with what you have to do to make types line up and essentially having to lie about even having a hash key.  This was a much easier solution.

drahnr:
@froydnj I'll take another look probably tomorrow how to do this with the least friction :) I really appreciate you giving code pointers and samples :+1:

drahnr:
I don't think there is a good way of making this generic.

After all, the `BlackList` must depend on the type of the `Box<dyn Compilation>` for the compiler it tries to cache, since that contains also valid meta data to filter on.
The current approach is to make the `Blacklist` inherent to the compiler - and share it with the `Hasher` via an `Arc<_>` to avoid pointless copies.

But this also means that the common shared of blacklists will be the configuration, which frankly is also divided into the Rust/CXX(CompilerVarian) strains plus a few special case booleans.

drahnr:
```
DEBUG 2020-04-03T07:22:30Z: sccache::compiler::rust: blacklist(rust)> blacklist matched with rule: crate name is enlisted: syn
DEBUG 2020-04-03T07:22:30Z: sccache::compiler::compiler: [syn]: generate_hash_key took 1.078 s
TRACE 2020-04-03T07:22:30Z: sccache::compiler::compiler: [syn]: Hash key: 74d4a50c23f9ec2c21dd9045c680e4451edada2fdb21b849a3af321c09b38cb5
DEBUG 2020-04-03T07:22:30Z: sccache::compiler::compiler: [syn]: Cache blacklisted in 0.000 s
TRACE 2020-04-03T07:22:30Z: sccache::compiler::rust: [syn]: compile
DEBUG 2020-04-03T07:22:30Z: sccache::compiler::compiler: [syn]: Compiling locally
```
for

```
SCCACHE_BLACKLIST_RUST_BUILD_SCRIPT=1
SCCACHE_BLACKLIST_RUST_FILES=src/lib.rs
SCCACHE_BLACKLIST_RUST_CRATES=syn
SCCACHE_BLACKLIST_RUST=crate:regex,file:src/main.rs
```

I'd appreciate a review, thanks!

drahnr:
TODOs:
* [ ] performance measurments with `coz`/`flamegraph`
* [ ] documentation

drahnr:
I'd like some feedback on this in what direction this should progress or not

glandium:
I think this is superseded by #798.

drahnr:
Not sure I quite follow, #798 mingles the logic into the hash generation which I find rather unfortunate.

PR #712, opened 2020-03-17
chore/cargo/features: make `all` the default feature set

drahnr:
Enable all features by default, avoids missing features issues.


drahnr:
@chmanchester I could swear there was an issue suggesting this, but I can't find it anymore...

glandium:
Fixed in 5e447094d0039d7f41a887749c5fa48de92480a7

PR #713, opened 2020-03-18
bump minimum rust version to 1.41.1

drahnr:
Required for #666

Given the Cargo.lock fmt changed in 1.41, it might make sense to put the min version to `1.41.1` (the patched one) right away.


froydnj:
If we're going to update to 1.40 for a single function, we might as well just update to 1.41 so we can get the lockfile changes.  Please make that change.

drahnr:
Done

froydnj:
Thanks for also converting the `Cargo.lock` file!

PR #721, opened 2020-04-06
Correct minor linting issues detected by rust-clippy

robertmaynard:


robertmaynard:
Updated with cargo-fmt applied on the relevant sections I modified.

glandium:
It seems 3b340b3 and 669f99f cover this.

PR #724, opened 2020-04-15
support /external:I in msvc

froydnj:
MSVC 2017 15.6 came with a new feature to support system include paths:

https://devblogs.microsoft.com/cppblog/broken-warnings-theory/

Unfortunately, said new feature is unusable with sccache because sccache
thinks that for `/external:I $PATH`, `$PATH` is an input argument, and
bails out.  Let's teach sccache about this syntax.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -233,6 +233,7 @@ ArgData! {
     ExtraHashFile(PathBuf),
     XClang(OsString), // -Xclang ...
     Clang(OsString), // -clang:...
+    ExternalIncludePath(PathBuf),
 }
 
 use self::ArgData::*;
@@ -393,6 +394,7 @@ msvc_args!(static ARGS: [ArgInfo<ArgData>; _] = [
     msvc_flag!("experimental:module", TooHardFlag),
     msvc_flag!("experimental:module-", PassThrough), // Explicitly disabled modules.
     msvc_take_arg!("experimental:preprocessor", OsString, Concatenated, PassThroughWithSuffix),
+    msvc_take_arg!("external:I", PathBuf, CanBeSeparated, ExternalIncludePath),
     msvc_take_arg!("favor:", OsString, Separated, PassThroughWithSuffix),
     msvc_take_arg!("fp:", OsString, Separated, PassThroughWithSuffix),
     msvc_take_arg!("fsanitize-blacklist", PathBuf, Concatenated('='), ExtraHashFile),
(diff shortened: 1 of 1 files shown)


src/compiler/msvc.rs
```diff
@@ -261,6 +262,7 @@ msvc_args!(static ARGS: [ArgInfo<ArgData>; _] = [
     msvc_flag!("Zi", DebugInfo),
     msvc_flag!("c", DoCompilation),
     msvc_take_arg!("deps", PathBuf, Concatenated, DepFile),
+    msvc_take_arg!("external:I", PathBuf, Separated, ExternalIncludePath),
```
glandium, line 265: I wonder if it /has/ to be separated or if it's actually something that can also be concatenated, like normal `-I`.

src/compiler/msvc.rs
```diff
@@ -836,6 +844,44 @@ mod test {
         assert!(!msvc_show_includes);
     }
 
+    #[test]
+    fn test_parse_arguments_external_include() {
+        // Parsing -external:I relies on -experimental:external being parsed
+        // and placed into common_args.
```
glandium, line 850: shouldn't `-external:I` be in preprocessor_args, like `-I`?

src/compiler/msvc.rs
froydnj: I think it has to be separated -- at least the blogpost points that direction, and I would think that requiring it to be separated also eliminates parsing ambiguities if there was some future `/external:Ifoo` flag.  But I have not actually experimented with this to find out.

src/compiler/msvc.rs
froydnj: I was assuming, perhaps wrongly, that the actual analysis phase needed information on external directories to issue warnings properly.

glandium: You'd need both in the same set of args, presumably, but if `-external:I` is there, surely it will have an impact on preprocessing... and without it preprocessing might fail.

froydnj: But `common_args` are passed as part of preprocessing...?

glandium: Mmmm I was still under the model where compilation was using the preprocessed output, which it doesn't anymore.... except when doing distributed builds. And common_args is passed in that case. Do we want `-external:I` to be passed there?

src/compiler/msvc.rs
glandium: Trying out with MSVC 2019, it seems `-external:Ifoo` is valid.

PR #728, opened 2020-04-23
add support for clippy-driver to sccache

yaahc:
fixes https://github.com/rust-lang/rust-clippy/issues/5507

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -899,7 +899,9 @@ where
         None => return f_err("could not determine compiler kind"),
         Some(f) => f,
     };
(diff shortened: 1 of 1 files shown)


yaahc:
This is admittedly not a great solution, with this approach every single binary that wants to work as a `RUSTC_WORKSPACE_WRAPPER` and `sccache` will need to be added to this list. Long term I expect this should be changed to detect `rustc` without ever checking the name of the binary, I wouldn't mind doing that as part of this PR but I didn't want to spend the time exploring sccache enough to be confident in my changes beyond this minimal fix.

froydnj:
Thank you!

PR #736, opened 2020-04-29
test: Don't force the entire crate to be recompiled by escargot

Marwes:
Every time one of these commands were run in tests they forced the entire crate to be recompiled which is quite a pain. Removing the `current_target` fixes it though, at least for me.

### DIFF
--- a/tests/harness/mod.rs
+++ b/tests/harness/mod.rs
@@ -124,7 +124,6 @@ pub fn sccache_command() -> Command {
         CargoBuild::new()
             .bin("sccache")
             .current_release()
-            .current_target()
             .run()
             .unwrap()
             .command(),
@@ -148,7 +147,6 @@ pub fn sccache_command() -> Command {
             .arg("--features")
             .arg(features)
             .current_release()
-            .current_target()
             .run()
             .unwrap()
             .command(),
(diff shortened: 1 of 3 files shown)


froydnj:
I agree that this would be more convenient, but it would not be correct in general: e.g. if you were cross-compiling and attempted to test, this change would compile tests for your local machine, not whatever you had tried to cross-compile to.  (Blame Cargo and its consistent but weird compilation model.)  Maybe that's too much of a corner-case to care about?

Marwes:
Well the built sccache is immediately started on the local machine so if we were cross compiling it wouldn't work anyway?

froydnj:
Consider the case of building 32-bit Windows binaries on a 64-bit system or cross-compiling on Linux and using binfmt support in the kernel to automatically [run the cross binaries with QEMU](https://wiki.debian.org/QemuUserEmulation).

luser:
I think @Marwes' point is more that you can't really usefully run these integration tests in cross-compile scenarios anyway. It'd probably be fine to just disable them entirely using a `#[cfg(...)]` to check for that.

froydnj:
This was fixed by #775 in a slightly more general way.

PR #743, opened 2020-05-08
Convert a few things to newer Futures

glandium:
First batch of things. I've been struggling to move more from server.rs....

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -30,8 +30,8 @@ use crate::protocol::{Compile, CompileFinished, CompileResponse, Request, Respon
 use crate::util;
 use filetime::FileTime;
 use futures::sync::mpsc;
-use futures::task::{self, Task};
 use futures::{future, stream, Async, AsyncSink, Future, Poll, Sink, StartSend, Stream};
+use futures_03::compat::Compat;
 use futures_cpupool::CpuPool;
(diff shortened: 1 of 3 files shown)


glandium:
Snipped the commit that broke various tests.

src/server.rs
```diff
@@ -1708,24 +1707,23 @@ impl Drop for ActiveInfo {
...
 }
 
-impl Future for WaitUntilZero {
-    type Item = ();
-    type Error = io::Error;
+impl std::future::Future for WaitUntilZero {
+    type Output = io::Result<()>;
 
-    fn poll(&mut self) -> Poll<(), io::Error> {
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> std::task::Poll<Self::Output> {
```
froydnj, line 1720: I guess at some future point we should make this return `std::task::Poll<()>`, because we never care about the error case of `io::Result`?  Or am I missing something about futures-based code?

glandium: It's because of the compat stuff. To get an impl Future on Compat, you need an impl TryFuture, which you get if you have an impl Future with a Result Output.

froydnj: Fair enough!

PR #752, opened 2020-05-21
fix some workarounds now that we have non-lexical lifetimes

froydnj:
The TODO in server.rs that we didn't fix was somewhat cryptic and I
don't think it would have made the code any clearer even if it could
have been fixed.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -324,8 +324,8 @@ impl DistClientContainer {
             }};
         }
         // TODO: NLL would avoid this clone
-        match config.scheduler_url.clone() {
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -324,8 +324,8 @@ impl DistClientContainer {
             }};
         }
         // TODO: NLL would avoid this clone
```
emilio, line 326: Probably this comment should've been removed?

froydnj: Doh.  Yes, will fix.

Issue #411, opened 2019-04-11
MSVC: SCCACHE not printing included headers with /showIncludes and ninja can't detect changes


dpar39:
If I call `cl.exe -showIncludes ... /Fo$out /c $in` it correctly prints the list of include files to `stdout`. If I call `sccache cl.exe -showIncludes ... -c /Fo$out /c $in` only the name of the output object is printed to stdout.
Ninja relies on the list of include headers to figure out dependency builds.
I can reproduce the problem on 0.2.8 and master local build. I've tried both /showIncludes and -showIncludes, it doesn't make a difference.
If the compiler command has some error (e.g. a path to file doesn't exists), then the list of includes is correctly printed on the console.

piscisaureus:
I'm running into this as well. It appears that when sccache invokes the compiler after a cache miss, the `/showIncludes` flag is dropped.

I did the following test:

### First run (expecting a sccache cache miss):

I run:
  1. `sccache cl /showIncludes /c bar.cc`

sccache invokes:
  1. `"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe"" -E C:\Users\BERTBE~1\AppData\Local\Temp\sccache.ddMykFjaAGPF\testfile.c`
  2. `"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe"" -nologo -showIncludes -c -Fonul -I. C:\Users\BERTBE~1\AppData\Local\Temp\sccache.tfHJK22noDJy\test.c`
  3. `"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe"" -E bar.cc -nologo -showIncludes`
  4. `"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe"" -c bar.cc -Fobar.obj`

Note how step **4**, which is where it actually compiles the file, the `/showIncludes` flag is missing.

### Second run (expecting a sccache cache hit):

I run:
  1. `sccache cl /showIncludes /c bar.cc`

sccache invokes:
[...]

PR #753, opened 2020-05-21
msvc: handle -showIncludes as a dependency argument 

mathstuf:
Due to the way that MSVC works, these flags also need passed to the main
compilation as well.

Fixes: #411

---
~~Untested locally; I'll work on getting a binary onto a Windows machine to test it out live tomorrow or early next week.~~

Tested locally in these configurations:

  - clean build, clean cache
  - clean build, reused cache
  - reuse build, clean cache
  - reuse build, reused cache

Cc: @dpar39 @piscisaureus

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -468,7 +468,10 @@ pub fn parse_arguments(
                 compilation_flag =
                     OsString::from(arg.flag_str().expect("Compilation flag expected"));
             }
-            Some(ShowIncludes) => show_includes = true,
+            Some(ShowIncludes) => {
+                show_includes = true;
+                dependency_args.push(arg.to_os_string());
+            }
             Some(Output(out)) => {
                 output_arg = Some(out.clone());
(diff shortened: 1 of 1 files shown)


mathstuf:
The test case I'm using is specifically testing "do rebuilds happen when discovered dependencies change" (CMake's `BuildDepends` tests).

src/compiler/msvc.rs
```diff
@@ -505,9 +508,6 @@ where
         .env_clear()
         .envs(env_vars.iter().map(|&(ref k, ref v)| (k, v)))
         .current_dir(&cwd);
-    if parsed_args.depfile.is_some() || parsed_args.msvc_show_includes {
```
robertmaynard, line 508: I think you need to keep the `parsed_args.depfile.is_some()` conditional around. If somebody has passed `dep` and not/instead of `showIncludes`

mathstuf:
Ping?

autoantwort:
@froydnj Can you review?

mathstuf:
Thanks!

PR #761, opened 2020-05-30
docs: document all config options, includes a unit test for verification

drahnr:
An initial stab at documenting all configuration options, mostly to ease adoption pain for non-rust-natives.

CC @TriplEight


drahnr:
This is a non invasive, docs-only change not affecting behaviour, could we merge this?

Issue #757, opened 2020-05-27
write outputs from a cache hit in parallel


froydnj:
This bit of code:

https://github.com/mozilla/sccache/blob/35ea08f19ae23424a6c2d20705c2f6f970c91ae6/src/compiler/compiler.rs#L258-L275

requires every output to be written serially...when we have an entire pool of threads that are happy to handle multiple units of work.

This doesn't matter too much for C/C++ compilation (since there's a single output), but it might make a small amount of difference for Rust compilation, where there are ~three outputs (`.rlib`, `.rmeta`, and `.d`) and the first two tend to be large-ish.

We'd have to modify `CacheRead` to provide some sort of API for returning an iterator of `ZipFile` objects so that we're not capturing a mutable reference to the cache entry itself inside the closure spawned on each thread.

luser:
> We'd have to modify `CacheRead` to provide some sort of API for returning an iterator of `ZipFile` objects

It might be a nicer API to have something like `fn handle_objects<F: FnMut(&str, Option<u32>) -> Result<()>>(handler: F) -> Result<()>` so you could do something like:
```rust
// Probably `write_on_a_thread` would need to return something you could join on.
entry.handle_objects(|name, mode| write_on_a_thread(&pool, name, mode));
```

rickystewart:
I have a [commit](https://github.com/rickystewart/sccache/commit/2751ce4542d01c6179fd141dcfb7e9579fc59a92) demonstrating what an implementation of this might look like. Some takeaways:

1. The method on `ZipArchive`s to open up a `ZipFile` takes a mutable reference, so you can't just parallelize that across multiple cores.
2. `ZipFile`s themselves aren't `Send`, so you can't open a bunch of them at once then distribute them across cores to do the reading and the writing.
3. ... even if you could, `ZipFiles` have a lifetime dependency on their parent `ZipArchive`, so you would have to do lifetime finagling too to work around that.

In this commit I've done what we're allowed to do by the compiler, which is 1) in one thread, read all the `ZipFile` contents into memory, and then 2) distribute the work of writing the contents across multiple threads. I see no reason to believe this would be any more performant in anyway way than the current state of things, and having to read the entirety of every file into memory is a big downside. (Caveat: I'm not a Rust expert so I might be missing some other solution here.)

With all that in mind I would suggest we close this, since I don't see a feasible way to actually fix the issue in any way that has the expected perf improvement.

PR #774, opened 2020-06-04
move some functionality into the cache object layer

froydnj:
None of this is strictly necessary, but I do think it makes `compiler.rs` slightly easier to follow.  I think this makes the implementation of #757 slightly easier as well (and keeps the implementation details hidden from external clients, which is good).


froydnj:
Rebased and reformatted.

PR #781, opened 2020-06-05
[DRAFT] add an OS X + Rust 1.43.1 job

froydnj:
Trying to see if the OS + Rust stable crash looks like a Rust regression or not.  The TravisCI configuration is a little weird, so this might require some iterations to get it right.


froydnj:
This PR has served its purpose for debugging and for a reference in the Rust upstream bug report.

Issue #759, opened 2020-05-27
new compression format for cached artifacts


froydnj:
Zip files are super-convenient, but maybe using something that would provide a better compression/performance ratio would be better.  At least in the common case for C/C++ compiles where you have a single output and no compiler messages.  Maybe you could invent a small custom format for the typical Rust compile (three outputs) as well.

luser:
Related: https://github.com/mozilla/sccache/issues/552 .

IIRC the zip file cache entry format dates back to the original Python version of sccache. I don't know that inventing a custom format is worthwhile, but just switching to zstd-compressed tarballs ought to be a pretty simple change.

mostynb:
The bazel remote execution api's ActionResult protobuf message might be worth considering over tarballs. It provides a bit more structure, and is flexible enough to handle multiple output files (which would be good to support for C/C++ compiles too). It would also make it easier to integrate with REAPI cache servers.

https://github.com/bazelbuild/remote-apis/blob/7802003e00901b4e740fe0ebec1243c221e02ae2/build/bazel/remote/execution/v2/remote_execution.proto#L854

Issue #160, opened 2017-07-25
Local disk cache 3-4.5x slower than ccache


indygreg:
I frequently perform a number of fully cached clobber builds of Firefox as part of testing the Firefox build system. My workflow is build, clobber, build. This workflow essentially benchmarks non-cached aspects of the Firefox build system and the caching layer. I've previously been using ccache but switched to sccache because Rust compilation was being a long pole and I didn't want to wait dozens of seconds for Rust to finish compiling.

To my surprise, sccache is significantly slower than ccache for comparable operations. When you factor in cache hits from Rust, sccache is about the same speed (135-145s wall for a clobber build). But if you limit the comparison to just C/C++, then sccache is ~3x slower overall and ~4.5x slower to obtain object files:

```
ccache:   35s wall; 131s CPU; 20s wall to libxul link
sccache: 107s wall; 131s CPU; 93s wall to libxul link
```

My mozconfig has `export CC=clang` and `ac_add_options --disable-debug --disable-optimize`. Clang version is `4.0.0-1ubuntu1` (from Ubuntu 17.04). To get the numbers above, I did a full build then a `find objdir -name '*.o' | xargs rm` to nuke C/C++ object files. Then I did `mach build binaries` to measure C/C++ compile time from cache. This is with current master (`6267caf603d5a06e8a4b9391ffa02f6125e5ba9b`) built with `cargo build --release --features=all`

`perf` recording of all processes during an sccache build shows the following hot frames:

```
   4.79%  clang          libLLVM-4.0.so.1                   [.] llvm::StringMapImpl::LookupBucketFor
   4.29%  clang          clang                              [.] clang::Lexer::LexTokenInternal
   2.81%  clang          clang                              [.] clang::Lexer::LexIdentifier
   2.58%  clang          clang                              [.] clang::SourceManager::getFileIDLocal
   2.54%  clang          clang                              [.] clang::DoPrintPreprocessedInput
   2.01%  sccache        sccache                            [.] GFp_sha512_block_data_order_avx2
   1.82%  clang          libc-2.24.so                       [.] __memmove_avx_unaligned_erms
   1.40%  clang          libLLVM-4.0.so.1                   [.] llvm::raw_svector_ostream::write_impl
   1.22%  clang          clang                              [.] clang::Preprocessor::Lex
   1.20%  clang          clang                              [.] clang::SourceManager::getCharacterData
   1.11%  sccache        sccache                            [.] tinfl_decompress
[...]

PR #784, opened 2020-06-05
use zstd to compress cache entries

froydnj:
zstd is faster and gives slightly smaller (~5%) compressed blobs than
deflate does, as measured on a Firefox build.  Rather than inventing our
own compressed archive format, we piggyback on top of zip's "stored"
files to stuff zstd-compressed blobs in the zip archive.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -36,6 +36,7 @@ use std::time::Duration;
 use tempfile::NamedTempFile;
 use zip::write::FileOptions;
 use zip::{CompressionMethod, ZipArchive, ZipWriter};
+use zstd;
 
 use crate::errors::*;
 
@@ -111,12 +112,16 @@ impl CacheRead {
     where
         T: Write,
     {
-        let mut file = self
+        let file = self
             .zip
             .by_name(name)
             .context("Failed to read object from cache entry")?;
-        io::copy(&mut file, to)?;
-        Ok(file.unix_mode())
+        if file.compression() != CompressionMethod::Stored {
+            bail!("Cache entry is not stored as a zstd blob");
+        }
+        let mode = file.unix_mode();
+        zstd::stream::copy_decode(file, to)?;
+        Ok(mode)
     }
 
     /// Get the stdout from this cache entry, if it exists.
@@ -197,7 +202,9 @@ impl CacheWrite {
     where
         T: Read,
     {
-        let opts = FileOptions::default().compression_method(CompressionMethod::Deflated);
+        // We're going to declare the compression method as "stored",
+        // but we're actually going to store zstd-compressed blobs.
+        let opts = FileOptions::default().compression_method(CompressionMethod::Stored);
         let opts = if let Some(mode) = mode {
             opts.unix_permissions(mode)
         } else {
@@ -206,7 +213,7 @@ impl CacheWrite {
         self.zip
             .start_file(name, opts)
             .context("Failed to start cache entry object")?;
-        io::copy(from, &mut self.zip)?;
+        zstd::stream::copy_encode(from, &mut self.zip, 3)?;
         Ok(())
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,6 +285,9 @@ name = "cc"
 version = "1.0.54"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7bbb73db36c1246e9034e307d0fba23f9a2e251faa47ade70c1bd252220c8311"
+dependencies = [
+ "jobserver",
+]
 
 [[package]]
 name = "cfg-if"
@@ -883,6 +886,12 @@ name = "gimli"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bcc8e0c9bce37868955864dbecd2b1ab2bdf967e6f28066d65aaac620444b65c"
+
+[[package]]
+name = "glob"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
 
 [[package]]
 name = "h2"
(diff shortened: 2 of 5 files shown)


luser:
I feel like this has been discussed before, but you might look at using .tar.zstd if you're going to make this change, since that would allow compression across entries where zip compresses one entry at a time.

froydnj:
> I feel like this has been discussed before, but you might look at using .tar.zstd if you're going to make this change, since that would allow compression across entries where zip compresses one entry at a time.

I thought about doing this, but this change was much simpler; a quick run over my sccache cache corpus (a bunch of Firefox builds, ~8GB of zip files) says that `.tar.zstd` gains about half a percent more than just stuffing zstd blobs into zip files.  I'm not sure that's worth it.

jwatt:
> I feel like this has been discussed before

In both:
https://github.com/mozilla/sccache/issues/160
https://github.com/mozilla/sccache/issues/552

glandium:
I don't think it would be worth for compression size, but it could be worth for latency: with the zip format, you _have_ to have downloaded the entire thing before you can start decompressing. With tar, you can stream the decompression. With things ordered in the right way, you wouldn't even need to keep everything in memory.

glandium:
(OTOH, the overhead of decompressing is so low...)

froydnj:
> I don't think it would be worth for compression size, but it could be worth for latency: with the zip format, you _have_ to have downloaded the entire thing before you can start decompressing. With tar, you can stream the decompression. With things ordered in the right way, you wouldn't even need to keep everything in memory.

I can totally get behind this.  At the same time, making cache entries tar files would be somewhat more complicated than this, and then to take advantage of being able to stream decompress the cache entries would require significant rewriting of the storage layer (though I see from poking around that our GCS/S3/Azure support have basically copy-paste'd code for converting network requests into in-memory `Vec<u8>`s, so maybe it wouldn't be that bad?)

Experiments have also suggested that hiding the network latency here is not worth it...yet.  I do like the idea of `.tar.zstd`; I just don't think its time has come.  (And per the suggestion from @mostynb in #759, maybe we should be adopting Bazel's cache format -- though I'm not totally sure what that buys us at the current point beyond, I think, somewhat more code than even `.tar.zstd`.)

mostynb:
The potential benefits of using bazel's cache format (`ActionResult` protobuf) are orthogonal to whatever compression is applied to the serialized blobs:
* Adding an REAPI cache backend would be easy, and give us access to several existing implementations with some nice qualities.
* Supporting compile jobs that output additional files would be easy- they're simply stored by file name in the output_files field, which can't be confused with the stdout and stderr fields.

src/cache/cache.rs
```diff
@@ -197,7 +200,10 @@ impl CacheWrite {
     where
         T: Read,
     {
-        let opts = FileOptions::default().compression_method(CompressionMethod::Deflated);
+        // We are going to cheat a little bit here; we're going to declare the
```
glandium, line 203: That's not really cheating. Cheating would be using another compression method than stored.

src/cache/cache.rs
```diff
@@ -206,7 +212,9 @@ impl CacheWrite {
         self.zip
             .start_file(name, opts)
             .chain_err(|| "Failed to start cache entry object")?;
-        io::copy(from, &mut self.zip)?;
+        let mut encoder = zstd::Encoder::new(&mut self.zip, 1)?;
+        io::copy(from, &mut encoder)?;
```
glandium, line 216: `ztd::stream::copy_encode`?

src/cache/cache.rs
```diff
@@ -111,12 +112,14 @@ impl CacheRead {
...
-        let mut file = self
+        let file = self
             .zip
             .by_name(name)
             .chain_err(|| "Failed to read object from cache entry")?;
-        io::copy(&mut file, to)?;
-        Ok(file.unix_mode())
+        let mode = file.unix_mode();
+        let mut decoder = zstd::Decoder::new(file)?;
+        io::copy(&mut decoder, to)?;
```
glandium, line 121: `zstd::stream::copy_decode`?

Maybe check the compression method first?

src/cache/cache.rs
```diff
@@ -206,7 +212,9 @@ impl CacheWrite {
         self.zip
             .start_file(name, opts)
             .chain_err(|| "Failed to start cache entry object")?;
-        io::copy(from, &mut self.zip)?;
+        let mut encoder = zstd::Encoder::new(&mut self.zip, 1)?;
```
glandium, line 215: how much faster is 1 compared to 2 or 3 (the default)? how much space is left off by not using 2 or 3? 3 is still significantly faster than zlib, but can yield even smaller files, which means faster download times, which are arguably more important than upload time.

src/cache/cache.rs
froydnj: Ah, didn't think about `copy_{en,de}code`, those are reasonable things to use.

I guess we could check the compression method, but what would you want to do?  Throw an error if `!= Stored`?  I think we'll basically do that anyway.  Go ahead and use deflate otherwise?  That seems like it risks silently doing the wrong, slower thing.

src/cache/cache.rs
froydnj: I will reword this comment.

src/cache/cache.rs
froydnj: I went and checked this and I think I was doing my numbers wrong in the first place -- I thought I was using 1, but I was apparently using 3, so I'll change this to 3.

src/cache/cache.rs
glandium: You didn't.

src/cache/cache.rs
```diff
@@ -111,12 +112,13 @@ impl CacheRead {
...
     {
-        let mut file = self
+        let file = self
             .zip
             .by_name(name)
             .context("Failed to read object from cache entry")?;
-        io::copy(&mut file, to)?;
-        Ok(file.unix_mode())
+        let mode = file.unix_mode();
+        zstd::stream::copy_decode(file, to)?;
```
glandium, line 120: How about returning an error if it's not the STORED compression method? After all, that means the file is not in the right format....

PR #735, opened 2020-04-29
feat: Update redis to 0.15

Marwes:
With tokio-compat being used it is now possible to use dependencies that
use tokio-0.2

PR #786, opened 2020-06-08
feat: Update redis to 0.15

Marwes:
With tokio-compat being used it is now possible to use dependencies that
use tokio-0.2

Update of https://github.com/mozilla/sccache/pull/735, hopefully the timeouts were not by the redis update itself.


froydnj:
Thank you!

PR #787, opened 2020-06-08
refactor: Start translating server.rs to async await

Marwes:


froydnj:
This is really interesting, because we kind of thought there was no good way to incrementally transition to new futures.  I assume you basically sprinkle `.compat()` everywhere and then at some point you can go through and remove all of them?

@luser is this basically what y'all did internally?

luser:
We don't use tokio except for some testing tools so it wasn't as much of a hassle for us. Looks like that's basically what we did for those, though.

Marwes:
Yes. My preferred way is to start from the top, replacing things with async/await/std::fuuture::Future and calling `compat` on any old  futures. Then this can repeated, pushing the `compat` calls deeper into the stack until finally they are completely gone. (By starting from the top it is possible to replace cloning by borrowing in many cases).

src/server.rs
```diff
@@ -1582,6 +1596,33 @@ impl<R, B> Message<R, B> {
     }
 }
 
+struct BincodeCodec;
```
glandium, line 1579: github doesn't allow to comment on unpatches sections but there's a compile failure in `impl<R> futures_03::Stream for Body<R>` because the `receiver` changed type and doesn't have `poll` anymore. It does have `poll_next`, though, which simplifies the implementation.

src/server.rs
```diff
@@ -1582,6 +1596,33 @@ impl<R, B> Message<R, B> {
...
+impl<T> tokio_serde::Serializer<T> for BincodeCodec
+where
+    T: serde::Serialize,
+{
+    type Error = Error;
+
+    fn serialize(self: Pin<&mut Self>, item: &T) -> std::result::Result<Bytes, Self::Error> {
+        let mut bytes = BytesMut::new();
+        use bytes_05::buf::BufMutExt;
```
glandium, line 1588: This can go at the top of the file.

src/server.rs
```diff
@@ -1597,51 +1638,54 @@ impl<R, B> Message<R, B> {
...
 
-    fn poll(&mut self) -> Poll<Option<Self::Item>, io::Error> {
-        let msg = try_ready!(self.inner.poll().map_err(|e| {
-            error!("SccacheTransport::poll failed: {}", e);
-            io::Error::new(io::ErrorKind::Other, e)
-        }));
-        Ok(msg.map(Message::WithoutBody).into())
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Option<Self::Item>> {
+        Poll::Ready(
+            ready!(Stream::poll_next(Pin::new(&mut self.inner), cx))
```
glandium, line 1642: Is it more idiomatic than `Pin::new(&mut self.inner).poll_next(cx)`?

src/server.rs
```diff
@@ -1597,51 +1638,54 @@ impl<R, B> Message<R, B> {
...
+                .map(|r| r.map(Message::WithoutBody)),
```
glandium, line 1643: while it adds nesting, `.map(|r| r.map(|s| s.map(Message::WithoutBody)))` avoids the `Poll::Ready(ready!())` wrapping.

src/server.rs
```diff
@@ -1597,51 +1638,54 @@ impl<R, B> Message<R, B> {
...
-                }
-            },
-            Frame::Body { chunk: Some(chunk) } => match self.inner.start_send(chunk)? {
-                AsyncSink::Ready => Ok(AsyncSink::Ready),
-                AsyncSink::NotReady(chunk) => {
-                    Ok(AsyncSink::NotReady(Frame::Body { chunk: Some(chunk) }))
-                }
-            },
-            Frame::Body { chunk: None } => Ok(AsyncSink::Ready),
+            Frame::Message { message, body: _ } => Pin::new(&mut self.inner).start_send(message),
```
glandium: There's a warning about `body` here that you remove in a subsequent patch, but that should just be removed here.

src/server.rs
```diff
@@ -1597,51 +1638,54 @@ impl<R, B> Message<R, B> {
...
+impl<I: AsyncRead + AsyncWrite + Unpin> Stream for SccacheTransport<I> {
+    type Item = Result<Message<Request, Body<()>>>;
 
-    fn poll(&mut self) -> Poll<Option<Self::Item>, io::Error> {
-        let msg = try_ready!(self.inner.poll().map_err(|e| {
-            error!("SccacheTransport::poll failed: {}", e);
-            io::Error::new(io::ErrorKind::Other, e)
-        }));
-        Ok(msg.map(Message::WithoutBody).into())
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Option<Self::Item>> {
```
glandium, line 1640: `task::Context` is a reexport of `std::task::Context`, which is imported already, please just use `Context`.

src/server.rs
```diff
@@ -28,10 +28,17 @@ use crate::jobserver::Client;
...
-use futures::{future, stream, Async, AsyncSink, Future, Poll, Sink, StartSend, Stream};
-use futures_03::compat::Compat;
+use futures::Future as _;
+use futures_03::{
+    channel::mpsc,
+    compat::*,
+    future,
+    prelude::*,
+    ready, stream,
+    task::{self, Poll},
```
glandium, line 40: `Poll` is a reexport of `std::task::Poll`, please import `Poll` from there instead. As a follow-up, we can replace `std::task::Poll` with `Poll` in this file.

src/server.rs
```diff
@@ -1652,29 +1696,26 @@ struct ShutdownOrInactive {
...
+    fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<()> {
         loop {
-            match self.rx.poll().unwrap() {
-                Async::NotReady => break,
+            match Pin::new(&mut self.rx).poll_next(cx) {
+                Poll::Pending => break,
                 // Shutdown received!
-                Async::Ready(Some(ServerMessage::Shutdown)) => return Ok(().into()),
-                Async::Ready(Some(ServerMessage::Request)) => {
+                Poll::Ready(Some(ServerMessage::Shutdown)) => return ().into(),
```
glandium, line 1686: `return Poll::Ready(())`?

src/server.rs
```diff
@@ -1652,29 +1696,26 @@ struct ShutdownOrInactive {
...
+                Poll::Ready(Some(ServerMessage::Request)) => {
                     if self.timeout_dur != Duration::new(0, 0) {
-                        self.timeout = Some(Delay::new(Instant::now() + self.timeout_dur));
+                        self.timeout = Some(delay_for(self.timeout_dur));
                     }
                 }
                 // All services have shut down, in theory this isn't possible...
-                Async::Ready(None) => return Ok(().into()),
+                Poll::Ready(None) => return ().into(),
```
glandium, line 1693: Likewise.

src/server.rs
```diff
@@ -1652,29 +1696,26 @@ struct ShutdownOrInactive {
...
             }
         }
         match self.timeout {
-            None => Ok(Async::NotReady),
-            Some(ref mut timeout) => timeout
-                .poll()
-                .map_err(|err| io::Error::new(io::ErrorKind::Other, err)),
+            None => Poll::Pending,
+            Some(ref mut timeout) => ready!(Pin::new(timeout).poll(cx)).into(),
```
glandium, line 1698: Is the `ready!().into()` necessary? I now wonder if that's what was missing in my failed attempt at converting this impl.

src/server.rs
```diff
@@ -1102,25 +1090,28 @@ where
...
+                self.dist_client.get_client(),
+                self.creator.clone(),
+                self.storage.clone(),
+                arguments,
+                cwd,
+                env_vars,
+                cache_control,
+                self.pool.clone(),
+            )
+            .compat();
```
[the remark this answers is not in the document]
Marwes: `compat` calls are just temporary :shrug:

src/server.rs
```diff
@@ -1102,25 +1090,28 @@ where
...
-        let task = result.then(move |result| {
+        let task = async move {
+            let result = result.await;
             let mut cache_write = None;
-            let mut stats = me.stats.borrow_mut();
             let mut res = CompileFinished::default();
             res.color_mode = color_mode;
             match result {
                 Ok((compiled, out)) => {
+                    let mut stats = me.stats.borrow_mut();
```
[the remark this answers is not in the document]
Marwes: I had to move it to prevent it from living across an `await`

src/server.rs
```diff
@@ -702,45 +702,59 @@ where
...
-                    let _ = tx.send(ServerMessage::Shutdown).await;
-                    Ok(())
-                })
-                .compat();
-                let info_future = self.get_info();
-                return Box::new(future.join(info_future).map(move |(_, info)| {
-                    Message::WithoutBody(Response::ShuttingDown(Box::new(info)))
-                }));
+        let self_ = self.clone();
+        Box::pin(async move {
```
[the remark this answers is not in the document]
Marwes: Why? This allows await to be used, making it more flexible.

src/server.rs
Marwes: I guess, or the PR could just be squashed before merging.

Marwes:
Should have rebased and squashed these commits before making the PR, don't want to do it now since it messes up reviews.

src/server.rs
glandium: (It wasn't)

src/server.rs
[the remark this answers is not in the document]
glandium: It moves some code into a future, and as of right now, I'm not sure this is desired.

Cargo.lock
```diff
@@ -2103,7 +2103,6 @@ dependencies = [
  "bincode",
  "blake3",
  "byteorder",
- "bytes 0.4.12",
```
glandium, line 2129: Why does this not remove the corresponding `[[package]]` section?

Marwes: There are still other dependencies that require bytes 0.4

PR #794, opened 2020-06-11
Upgrade dependencies that don't require changes

glandium:
Except redis, which pulls in async-std, which needs some further
consideration.


glandium:
(turns out env_logger needs changes in tests ; I hadn't built the tests locally)

PR #795, opened 2020-06-12
Switch from futures_cpupool::CpuPool to futures_03::executor::ThreadPool

glandium:


src/util.rs
```diff
@@ -29,6 +31,18 @@ use std::time::Duration;
 
 use crate::errors::*;
 
+pub trait SpawnExt: task::SpawnExt {
+    fn spawn_fn<F, T>(&self, f: F) -> SFuture<T>
+    where
+        F: FnOnce() -> Result<T> + std::marker::Send + 'static,
+        T: std::marker::Send + 'static,
+    {
+        Box::new(self.spawn_with_handle(async move { f() }).unwrap().compat())
```
glandium, line 40: Gah, I forgot to remove the unwrap.

glandium:
The latter.

PR #796, opened 2020-06-12
update appveyor to MSVC 2017

froydnj:
This update should fix the BLAKE3 avx512 compilation issues.


froydnj:
Well, the nightly build is busted, but that might be some kind of transient download problem...or `cargo update` didn't correctly update packages on Windows?  We've made progress, at least.

PR #797, opened 2020-06-12
use a proper release for selenium-rs

froydnj:
0.1.2 was released earlier this year and includes the fix we need.


froydnj:
Travis apparently had a network outage during the checking process. :(

PR #798, opened 2020-06-15
Add ability to skip caching and distributing for certain files. Fixes #675

### LINKED ISSUE #675: add ability to skip caching/distributing for certain files

rickystewart:
I've implemented this as a regex which you can set via an environment variable; this seems to be a "sweet spot" that is better than a hard-coded list (which might not be flexible enough, especially for dynamic filenames such as you find in the Firefox configure process) but not extremely complicated to maintain or implement on our end either. If `SCCACHE_FORCE_LOCAL_REGEX` matches the basename of a file to be compiled, then we unconditionally perform a local compilation. For local `sccache` this doesn't seem to help a great deal, but this is expected to be a noticeable improvement both for `sccache-dist` as well as when cloud storage is used as a build cache.

In a test of the Firefox build in CI with an appropriately configured `SCCACHE_FORCE_LOCAL_REGEX`, this sped up `configure` by 6s (about 1/3rd of the total runtime of `configure`).


luser:
Out of curiosity, what regex did you use for building Firefox? (My guess would be that you're skipping autoconf test compiles.)

Given your implementation this feature is misnamed. It asks sccache to skip handling these compiles entirely, so they won't be cached either. The `DIST` in the env var leads one to believe that the compilation won't be distributed elsewhere.

rickystewart:
> Out of curiosity, what regex did you use for building Firefox? (My guess would be that you're skipping autoconf test compiles.)

Yeah: `^conftest\..*\..*$`

> Given your implementation this feature is misnamed. It asks sccache to skip handling these compiles entirely, so they won't be cached either. The `DIST` in the env var leads one to believe that the compilation won't be distributed elsewhere.

I'm aware of the naming issue. Do you have a concrete, better suggestion? The only other thing I could come up with was `SCCACHE_SKIP_DIST_AND_CACHE_REGEX` which is technically accurate but long. I guess `SCCACHE_FORCE_LOCAL_REGEX` would work?

rickystewart:
The "force local" nomenclature seemed fine to me so I swapped to that.

froydnj:
Comments:

1. (minor) We should probably fix the commit messages to refer to the new environment variable and squash the series.
2. I think we really want something separate for Rust and C/C++.  I guess that implies a separate environment variable(s?), but I don't think you can easily shoehorn ignoring Rust stuff into a single variable.  I would be OK just doing this for C/C++ and ignoring the Rust problems until a later point.
3. I think it's kind of gross that we have yet more arguments to propagate around, and especially into `generate_hash_key`, which shouldn't really have any business caring about where the compilation is ultimately going to be performed (cf. the comment in `rust.rs`).

drahnr:
Isn't this overlapping/complementary C/C++ coverage with #710 ?

rickystewart:
I pushed an update that attempts to address all 3 points made. (RE: 1, there's an option to [auto-squash a pull request when you close it](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/merging-a-pull-request) which I strongly recommend, unless preserving history is necessary for some reason. To make your job easier I've just squashed and force-pushed on my end.)

I think I'm rubbing against the [wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction) problem here. I agree I don't want to shunt this logic into the "hash" generation, but it's hard to separate it out without doing a more substantive refactor than I feel like it makes sense to commit to at this point in time. So I have a new abstraction which is like a two-phase thing (you try to get the necessary data to perform a forced-local compilation, unless you can't, in which case it gives you a data structure you can use to perform the whole hash generation). The type system cooperates with you here to make sure we don't do the second step until we've done the first. Naming these two phases is hard, especially not wanting to collide with existing data structures already in the codebase. I've opted with "light" and "full" to express these two phases for lack of better names, but if you see a better option I'll happily do a find-and-replace.

I was already planning on working on it, but having tried to wade through this I especially support additional refactoring to make further feature improvements even easier to work on.

froydnj:
@glandium, can you please review this?  I feel like I can't fairly review it because I already made a stab at this area.

luser:
> Naming these two phases is hard, especially not wanting to collide with existing data structures already in the codebase. I've opted with "light" and "full" to express these two phases for lack of better names but still don't really think they work awfully well, but if you see a better option I'll happily do a find-and-replace.

For the record, I don't think the existing names I settled on (`Compiler`, `CompilerHasher` et. al.) are all that amazing. I still like the *design* of having a separate trait per step, but I wouldn't stress too much about the names as long as they're not ridiculously obscure.

rickystewart:
Ping I guess?

mitchhentges:
Hey Ricky, long time no "👋" :)
I'm not sure that you still want to push this patch, but if you're available: can you rebase this against `main`, then I'll take a look and we can land this shortly after?

drahnr:
Gentle ping - I'd be interested in seeing this land

drahnr:
I'll tackle it, I think it has the potential to greatly ease the pain short term for all things similar to (but not limited to) #1426

sylvestre:

sylvestre:

PR #800, opened 2020-06-23
fix parsing for -dependency-file

froydnj:
This option was getting passed as a preprocessor argument and the
additional arguments that it requires (e.g. `-MT`) were getting passed
as dependency arguments.  So when we went to call the preprocessor, the
compiler complained about the missing arguments.  Instead, we should be
classifying it as specifying a dependency target, which will direct it
to the correct set of arguments.

We also take this opportunity to move it to be a clang-exclusive option;
the same option exists in GCC, but it's Darwin-specific and unlikely to
be actually used there.

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -157,7 +157,6 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("-aux-info", OsString, Separated, PassThrough),
     take_arg!("-b", OsString, Separated, PassThrough),
     flag!("-c", DoCompilation),
-    take_arg!("-dependency-file", PathBuf, Separated, PreprocessorArgumentPath),
     take_arg!("-fdiagnostics-color", OsString, Concatenated('='), DiagnosticsColor),
     flag!("-fno-diagnostics-color", NoDiagnosticsColorFlag),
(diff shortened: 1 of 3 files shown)


src/compiler/gcc.rs
```diff
@@ -157,7 +157,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     take_arg!("-aux-info", OsString, Separated, PassThrough),
     take_arg!("-b", OsString, Separated, PassThrough),
     flag!("-c", DoCompilation),
-    take_arg!("-dependency-file", PathBuf, Separated, PreprocessorArgumentPath),
+    take_arg!("-dependency-file", PathBuf, Separated, DepArgumentPath),
```
glandium, line 160: Why is it not in clang.rs in the first place?

froydnj: I don't know; on the GCC side, it is apparently an actual option, though it appears to have a completely different meaning:

https://gcc.gnu.org/onlinedocs/gcc-9.3.0/gcc/Darwin-Options.html#index-dependency-file

froydnj: I'll move it as part of this PR; I need to tweak the title anyway.

PR #764, opened 2020-06-01
allow caching with more msvc slash style arguments

mostynb:
#674 broke caching for msvc /foo style arguments that weren't handled in our arg list. So let's expand that list.

Fixes #725.

PR #801, opened 2020-06-24
fix msvc passthrough arguments

froydnj:
#764 and its followon patch, #768, failed to take into account that
passthrough arguments actually need to be passed through to the compiler
invocation: they don't prevent caching, but they do need to be
considered when computing the cache key...and when actually performing
the compilation.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -496,7 +496,11 @@ pub fn parse_arguments(
                     arg.normalize(NormalizedDisposition::Concatenated)
                         .iter_os_strings(),
                 ),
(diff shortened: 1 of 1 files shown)


mathstuf:
This appears to fix #802 for me.

src/compiler/msvc.rs
```diff
@@ -500,6 +500,12 @@ pub fn parse_arguments(
                 arg.normalize(NormalizedDisposition::Concatenated)
                     .iter_os_strings(),
             ),
+            Some(PassThrough) | Some(PassThroughWithPath(_)) | Some(PassThroughWithSuffix(_)) => {
```
glandium, line 503: This can be folded with the case above, which does the exact same thing.

PR #805, opened 2020-06-30
Validate server started on the correct port

sum12:
The patch implements the TODO mentioned.

The server writes the port of on which it is listening to the socket
placed at SCCACHE_STARTUP_NOTIFY.  This port should match the port
expected by the client that requested the startup.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -271,8 +271,20 @@ fn connect_or_start_server(port: u16) -> Result<ServerConnection> {
         {
             // If the connection was refused we probably need to start
             // the server.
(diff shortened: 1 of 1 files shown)


froydnj:
Thanks!

sum12:
Thanks for merging. There are quite a few TODO's. I would like to work on them. 
Just practicing some rust and this is a cool project. And I have some experience with ccache.
What do you think ?

froydnj:
Fixing TODOs sounds like a great idea!

philn:
Seems like this is causing #824

PR #807, opened 2020-07-06
Add a note for how to use sccache with cmake.

anholt:
It feels like the obvious way would be to ask cmake to use "sccache
gcc" as CMAKE_C_COMPILER, but that option can't take arguments.  Leave
a note for the next person trying to figure it out.


froydnj:
Thanks!

PR #809, opened 2020-07-14
update deps of lru-disk-cache

Dushistov:
I use lru-disc-cache as separate crate. And it is a pity that I have several versions of the same crates
in my dependency tree.


froydnj:
Thanks!

PR #811, opened 2020-07-19
Add support for key prefixes when using S3 as the storage

### LINKED ISSUE #1: [feature-request] Support for S3 prefixes
Hi,

I was discussing TC support for sscache with @gregarndt recently, and we found that it would be extremely nice if sscache supported S3 prefixes.

Essentially, instead of giving sscache just a `bucket` we would like to give it `bucket` and a `prefix`.
Such that all keys are stored under the given prefix in the bucket.

In `taskcluster-auth` we have an end-point that given a scope `auth:aws-s3:read-write:<bucket>/<prefix>` will return temporary S3 credentials with read-write access to a specific subfolder/prefix on S3.
We already need a bucket for every region, not having to do separate buckets for try and production, would be nice. Instead we could just use separate prefixes, which we can grant access to using scopes.

Notably, this will make it a lot easier if someone wanted to experiment with sscache and wanted isolated storage to avoid poisoning try or production. As giving such isolated storage would be a matter of issuing a scope, rather than creating a new bucket in every region.

http://docs.taskcluster.net/auth/api-docs/#awsS3Credentials

jakubadamw:
The corresponding environment variable is `SCCACHE_S3_KEY_PREFIX`.

Closes #1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,8 @@ You can use `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to set the S3 creden
 
 If you need to override the default endpoint you can set `SCCACHE_ENDPOINT`. To connect to a minio storage for example you can set `SCCACHE_ENDPOINT=<ip>:<port>`. If your endpoint requires TLS, set `SCCACHE_S3_USE_SSL=true`.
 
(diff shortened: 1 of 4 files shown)


jakubadamw:
> This looks reasonable, but please fix up the formatting errors in the Travis logs.

Done! 🙂

froydnj:
Thanks!

PR #815, opened 2020-07-22
Add known caveats section to readme

Timmmm:
These are pretty serious limitations that I would have liked to read about rather than discover later! I also added an example for `RUST_LOG` so you don't have to go digging in the docs to learn how to use it.

There might be more caveats but these are the ones I've found so far.

Ref #806

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ the container for you - you'll need to do that yourself.
 Debugging
 ---------
 
-You can run the server manually in foreground mode by running `SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache`, and send logging to stderr by setting the `RUST_LOG` environment variable, the format of which is described in more detail in the [env_logger](https://docs.rs/env_logger/0.5.3/env_logger/#enabling-logging) documentation.
+You can run the server manually in foreground mode by running `SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache`, and send logging to stderr by setting the [`RUST_LOG` environment variable](https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging) e.g. `RUST_LOG=debug`.
 
 Alternately, you can set the `SCCACHE_ERROR_LOG` environment variable to a path and set `RUST_LOG` to get the server process to redirect its logging there (including the output of unhandled panics, since the server sets `RUST_BACKTRACE=1` internally).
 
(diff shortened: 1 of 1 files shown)


luser:
I think most of this is captured in [the (very short) Rust-specific doc](https://github.com/mozilla/sccache/blob/master/docs/Rust.md). Also there used to be a "Known Caveats" section in the README but [I removed it recently](https://github.com/mozilla/sccache/commit/98c62b5404fac2149bcebaf1f05050d55133859d#diff-04c6e90faac2675aa89e2176d2eec7d8L223) because it didn't seem very useful, but perhaps a more thorough listing of the limitations of sccache for both C++ and Rust usage would be useful? It's certainly better to let people know up-front what the tool can and cannot do, I agree.

README.md
```diff
@@ -223,7 +223,7 @@ the container for you - you'll need to do that yourself.
 Debugging
 ---------
 
-You can run the server manually in foreground mode by running `SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache`, and send logging to stderr by setting the `RUST_LOG` environment variable, the format of which is described in more detail in the [env_logger](https://docs.rs/env_logger/0.5.3/env_logger/#enabling-logging) documentation.
+You can run the server manually in foreground mode by running `SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache`, and send logging to stderr by setting the [`RUST_LOG` environment variable](https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging) e.g. `RUST_LOG=debug`.
```
froydnj, line 226: I think we should have an example and a pointer to further documentation; could you make that change, please?

README.md
```diff
@@ -235,3 +235,13 @@ Interaction with GNU `make` jobserver
...
 This is most useful when using sccache for Rust compilation, as rustc supports using a jobserver for parallel codegen, so this ensures that rustc will not overwhelm the system with codegen tasks. Cargo implements its own jobserver ([see the information on `NUM_JOBS` in the cargo documentation](https://doc.rust-lang.org/stable/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)) for rustc to use, so using sccache for Rust compilation in cargo via `RUSTC_WRAPPER` should do the right thing automatically.
+
+---
+
+Known Caveats
+-------------
+
+* `proc-macro` crates like `serde_derive` cannot be cached.
+* `bin` crates cannot be cached.
+* Incrementally compiled crates cannot be cached. First party crates are compiled incrementally when building in debug mode. You may find that release builds are faster than debug builds because of this.
```
froydnj, line 246: It would be useful to mention setting the environment variable `CARGO_INCREMENTAL=0` to disable incremental compilation.

README.md
```diff
@@ -235,3 +235,13 @@ Interaction with GNU `make` jobserver
...
 
 This is most useful when using sccache for Rust compilation, as rustc supports using a jobserver for parallel codegen, so this ensures that rustc will not overwhelm the system with codegen tasks. Cargo implements its own jobserver ([see the information on `NUM_JOBS` in the cargo documentation](https://doc.rust-lang.org/stable/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)) for rustc to use, so using sccache for Rust compilation in cargo via `RUSTC_WRAPPER` should do the right thing automatically.
+
+---
+
+Known Caveats
+-------------
+
+* `proc-macro` crates like `serde_derive` cannot be cached.
+* `bin` crates cannot be cached.
```
froydnj, line 245: Like @luser said, I think this is covered in the Rust-specific documentation, but it's maybe less than obvious that `--emit` controls this; perhaps the Rust-specific documentation should be updated instead?

README.md
```diff
@@ -235,3 +235,13 @@ Interaction with GNU `make` jobserver
...
+* Incrementally compiled crates cannot be cached. First party crates are compiled incrementally when building in debug mode. You may find that release builds are faster than debug builds because of this.
+* Absolute paths to files must match to get a cache hit. This means that even if you are using a shared cache, everyone will have to build at the same absolute path (i.e. not in $HOME) in order to benefit each other.
```
froydnj, line 247: I think it'd be good to mention that this applies for both first-party crates and third-party crates.  So in most situations where one might want a shared cache, this would mean that you would have to relocate your cargo cache to someplace other than `$HOME/.cargo` to benefit.

README.md
luser: I wrote the Rust caveats while thinking about implementation details. It could stand to be rewritten to be understandable from the perspective of someone using sccache via cargo, perhaps something like "sccache cannot cache compiles of crates that invoke the system linker, such as `bin`, `dylib`, `cdylib`, or `proc-macro` crates".

README.md
luser: I'm not sure I'd want to directly recommend this here, TBH. It'd probably be best to just state the facts and link to [the cargo documentation](https://doc.rust-lang.org/cargo/reference/profiles.html#incremental). "sccache cannot cache incrementally compiled crates. By default, in the debug profile, cargo will use incremental compilation for workspace members and path dependencies."

README.md
Timmmm: What would be an example of `SCCACHE_ERROR_LOG`? Is this right?

    SCCACHE_ERROR_LOG=/tmp/sccache_log.txt RUST_LOG=debug sccache

README.md
Timmmm: I didn't actually know there was any documentation other than the Readme, because it isn't linked anywhere. I'll add a link.

README.md
froydnj: Yup!

PR #817, opened 2020-07-23
clang: Allow chrome plugin arguments

### LINKED ISSUE #350: Add support for clang plugin parameters

barrbrain:
For chromium clang, there are currently just 2 plugins for which arguments are passed in the build configuration:

- `-plugin-arg-find-bad-constructs` at [build/config/clang/BUILD.gn:23](https://chromium.googlesource.com/chromium/src/+/master/build/config/clang/BUILD.gn#23)
- `-plugin-arg-blink-gc-plugin` at [third_party/blink/renderer/BUILD.gn:123](https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/BUILD.gn#123)

When setting `clang_use_chrome_plugins = true`, the build is fully cacheable after adding rules for these 2 plugins.

Fixes #350. Supporting arbitrary plugins requires deeper changes.

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -113,8 +113,7 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-include-pch", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-load", PathBuf, Separated, ExtraHashFile),
     take_arg!("-mllvm", OsString, Separated, PassThrough),
-    take_arg!("-plugin-arg-blink-gc-plugin", OsString, Separated, PassThrough),
-    take_arg!("-plugin-arg-find-bad-constructs", OsString, Separated, PassThrough),
+    take_arg!("-plugin-arg", OsString, Concatenated('-'), PassThrough),
     take_arg!("-target", OsString, Separated, PassThrough),
     flag!("-verify", PreprocessorArgumentFlag),
 ]);
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -113,6 +113,8 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-include-pch", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-load", PathBuf, Separated, ExtraHashFile),
     take_arg!("-mllvm", OsString, Separated, PassThrough),
+    take_arg!("-plugin-arg-blink-gc-plugin", OsString, Separated, PassThrough),
+    take_arg!("-plugin-arg-find-bad-constructs", OsString, Separated, PassThrough),
     take_arg!("-target", OsString, Separated, PassThrough),
     flag!("-verify", PreprocessorArgumentFlag),
 ]);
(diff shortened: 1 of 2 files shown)


glandium:
Why not go with https://github.com/mozilla/sccache/issues/350#issuecomment-462807691 ?

barrbrain:
> Why not go with [#350 (comment)](https://github.com/mozilla/sccache/issues/350#issuecomment-462807691) ?

I think this change follows that suggestion, adding the specific plugin args to that list.
If there is a straightforward way to match the prefix, we could support arbitrary plugins rather than having an allow-list.

glandium:
That's exactly what the code in that comment was doing...

glandium:
Except it would need `Concatenated('-')` rather than `Separated`

src/compiler/clang.rs
```diff
@@ -113,6 +113,8 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-include-pch", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-load", PathBuf, Separated, ExtraHashFile),
     take_arg!("-mllvm", OsString, Separated, PassThrough),
+    take_arg!("-plugin-arg-blink-gc-plugin", OsString, Separated, PassThrough),
+    take_arg!("-plugin-arg-find-bad-constructs", OsString, Separated, PassThrough),
```
barrbrain, line 117: If we simply use `Concatenated` to capture the plugin name, additional logic is needed to consume the argument value.
```suggestion
    take_arg!("-plugin-arg", OsString, Concatenated('-'), PassThrough),
```

```
running 2 tests
test compiler::clang::test::test_parse_xclang_plugin_arg_blink_gc_plugin ... FAILED
test compiler::clang::test::test_parse_xclang_plugin_arg_find_bad_constructs ... FAILED

failures:

---- compiler::clang::test::test_parse_xclang_plugin_arg_blink_gc_plugin stdout ----
thread 'compiler::clang::test::test_parse_xclang_plugin_arg_blink_gc_plugin' panicked at 'Got unexpected parse result: CannotCache("Can\'t handle Raw arguments with -Xclang", None)', src/compiler/clang.rs:285:17

---- compiler::clang::test::test_parse_xclang_plugin_arg_find_bad_constructs stdout ----
thread 'compiler::clang::test::test_parse_xclang_plugin_arg_find_bad_constructs' panicked at 'Got unexpected parse result: CannotCache("Can\'t handle Raw arguments with -Xclang", None)', src/compiler/clang.rs:316:17
```

glandium:
I see what's going on. Yeah, it's more involved than just adding `-plugin-arg`. But your patch is also incomplete because it only supports the case where the arguments are used with `-Xclang`, which is not the only valid case for -plugin-arg-*.

Now, I'd understand if you don't want to go all the way to support the `-plugin-arg-*` arguments (plus, I'm not all that sure that plugins should be blindly trusted to be behaving in a cacheable way), so maybe what would be more helpful is a way to configure a set of arguments to pass through or something, although it's not clear how you'd tell "this argument takes another".

@froydnj wdyt?

froydnj:
Apologies for the delayed response here.

> I see what's going on. Yeah, it's more involved than just adding `-plugin-arg`. But your patch is also incomplete because it only supports the case where the arguments are used with `-Xclang`, which is not the only valid case for -plugin-arg-*.

Is this referring to how to can invoke `clang++ -cc1 -plugin-arg-* ...` or something else (`/clang:-plugin-arg-*` in `clang-cl`?)?

Even though [the "official" example](https://github.com/llvm/llvm-project/blob/30e45f339eb0841dc7fe27fad119cc5db0c052f3/clang/examples/PrintFunctionNames/README.txt) uses `-cc1 -plugin-arg-*`, I'm inclined to just require `-Xclang` here and call that good enough.  My sense has always been that `-cc1` is kind of magic and not really meant for external consumption, but maybe that's wrong!

> Now, I'd understand if you don't want to go all the way to support the `-plugin-arg-*` arguments (plus, I'm not all that sure that plugins should be blindly trusted to be behaving in a cacheable way), so maybe what would be more helpful is a way to configure a set of arguments to pass through or something, although it's not clear how you'd tell "this argument takes another".
> 
> @froydnj wdyt?

I'm not totally clear on where this extra configuration would take place -- if it's in the current argument parsing code, I think that's...mostly OK?  But it sounds like it would need to be more dynamic to handle cases like plugin args, and so it would be controlled through environment variables?  Which sounds terrible.  I don't think we want to make argument parsing any more complicated than it already is.

I think we should just go with the patch here, and if we wind up needing something more robust later, we can see if improvements present themselves.

glandium:
> Is this referring to how to can invoke clang++ -cc1 -plugin-arg-* ... or something else (/clang:-plugin-arg-* in clang-cl?)?

Yes. Surely, clang-cl is not the only use case for plugin arguments.

DAC324:
Sorry for reviving that old thread. Just a humble question: Will it be possible to use these Chrome clang plugins even when the system's clang is used, or do they require to use the buildtools bundled with Chrome, under all circumstances?

PR #818, opened 2020-07-27
Implement `g++` and `clang++` behavior when used on a `.c` input file. Fixes #803

### LINKED ISSUE #803: sccache's explicit language selection causes subtle behaviour changes
Based on https://github.com/servo/mozangle/issues/38, I tracked down the issue to the following:
* we were using cc-rs to build a collection of C++ and C files into a static library
* the C++ files required the -std=c++14 flag
* clang via cc-rs had no problem building the C files with the -std=c++14 flag present
* adding sccache to the build yielded an error that -std=c++14 couldn't be used with C files

A minimal testcase:
```
    cc::Build::new()
        .file("a.c")
        .cpp(true)
        .flag("-std=c++14")
        .compile("test");
```
When run with RUSTC_WRAPPER=sccache, this yields `cargo:warning=error: invalid argument '-std=c++14' not allowed with 'C'`. I can also reproduce without sccache by adding `.flag("-xc")`, which is what sccache does internally when invoking the compiler by guessing the language based on the filename.

rickystewart:
`g++` and `clang++` both compile `.c` files in C++ mode in the absence of any other directions (like the `-x` flag); add code to distinguish between the two compilers and their `++` variations, then fix up language selection and caching accordingly.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -166,6 +166,8 @@ pub enum CCompilerKind {
 pub trait CCompilerImpl: Clone + fmt::Debug + Send + 'static {
     /// Return the kind of compiler.
     fn kind(&self) -> CCompilerKind;
+    /// Return true iff this is g++ or clang++.
+    fn plusplus(&self) -> bool;
     /// Determine whether `arguments` are supported by this compiler.
     fn parse_arguments(
         &self,
@@ -344,6 +346,7 @@ where
                                 &extra_hashes,
                                 &env_vars,
                                 &preprocessor_result.stdout,
+                                compiler.plusplus(),
                             )
                         };
                         // A compiler binary may be a symlink to another and so has the same digest, but that means
@@ -621,7 +624,7 @@ impl pkg::ToolchainPackager for CToolchainPackager {
 }
 
 /// The cache is versioned by the inputs to `hash_key`.
-pub const CACHE_VERSION: &[u8] = b"9";
+pub const CACHE_VERSION: &[u8] = b"10";
 
 lazy_static! {
     /// Environment variables that are factored into the cache key.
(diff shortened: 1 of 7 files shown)


src/compiler/compiler.rs
```diff
@@ -1008,11 +1012,30 @@ diab
         for line in stdout.lines() {
             //TODO: do something smarter here.
             match line {
```
froydnj, line 1014: We should be able to reduce duplication here by:

```
match line {
  "clang" | "clang++" => {
    ... Clang { clangplusplus: line == "clang++" }
  // and similarly for the gcc/g++ case.
```

it does duplicate computation, but I am not super concerned about the hotness of this code.

src/compiler/c.rs
```diff
@@ -660,20 +664,43 @@ pub fn hash_key(
         }
     }
     m.update(preprocessor_output);
+    // clang and clang++ have different behavior despite being byte-for-byte identical binaries, so
+    // we have to incorporate that into the hash as well.
+    m.update(&[plusplus as u8]);
```
froydnj, line 669: Let's add this further up near the `compiler_digest` addition` so we don't have to worry about weird interactions with the preprocessor output.

rickystewart: Done, let me know if this is in line with what you meant.

rickystewart:
Sorry for the confusion here, I did something wrong as I was wrapping up this change on Friday evening (of course) and some of the code I wrote didn't make it to this PR. Here's the whole thing, which I tested and ensured works. :)

src/compiler/gcc.rs
```diff
@@ -420,9 +426,17 @@ where
         // We can't cache compilation without an input.
         None => cannot_cache!("no input file"),
     };
-    if language == None {
-        language = Language::from_file_name(Path::new(&input));
-    }
+    let language = match language {
+        None => {
+            let mut lang = Language::from_file_name(Path::new(&input));
+            // g++ and clang++ override the language to c++ if the -stdlib argument is passed.
```
froydnj, line 432: I think this should be "the -std argument" or something like that?

Actually, is that even right, because the behavior that the code is actually implementing is "if the compiler is g++ or clang++, parse it as C++ regardless of the file extension".  So perhaps we should just drop this comment entirely?

Issue #810, opened 2020-07-18
Consider moving log configuration to a dedicated env var


Nemo157:
Currently when running `RUST_LOG=debug cargo run` to get debug output from the target executable you also receive extra log info from `sccache`. Cargo and rustc avoid this by using custom `CARGO_LOG` and `RUSTC_LOG` variables respectively. It would be useful if `sccache` also moved to something like `SCCACHE_LOG` to avoid interference with projects using the `env_logger` default variable.

froydnj:
I'm not sure what the use case is here; are you saying that when you `RUSTC_WRAPPER=/path/to/sccache RUST_LOG=debug cargo run` and things need to be rebuilt, you're getting debug logging from `sccache` as well?

Is `RUST_LOG=${TARGET_EXE_NAME}=debug` not sufficient for this purpose?

Nemo157:
> [...] you're getting debug logging from sccache as well?
Exactly.

`RUST_LOG=${TARGET_EXE_NAME}=debug` will only print debug information from the module-tree `${TARGET_EXE_NAME}`. If for example I'm debugging something Tokio related in my application I would use `RUST_LOG=tokio=debug`, but since `sccache` also uses `tokio` that will still print debug output from `sccache` as well.

froydnj:
> `RUST_LOG=${TARGET_EXE_NAME}=debug` will only print debug information from the module-tree `${TARGET_EXE_NAME}`. If for example I'm debugging something Tokio related in my application I would use `RUST_LOG=tokio=debug`, but since `sccache` also uses `tokio` that will still print debug output from `sccache` as well.

Thanks for the explanation!  I agree this would be useful.

PR #822, opened 2020-08-04
Modify logging to use SCCACHE_LOG.

rthomas:
As per #810, modify env_logger to use the SCCACHE_LOG environment
variable instead of the RUST_LOG one. This is to enable control of
sccache logs whilst debugging other rust binaries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,13 +223,13 @@ the container for you - you'll need to do that yourself.
 Debugging
 ---------
 
-You can run the server manually in foreground mode by running `SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache`, and send logging to stderr by setting the [`RUST_LOG` environment variable](https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging) for example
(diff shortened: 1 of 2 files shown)


docs/Releasing.md
```diff
@@ -1,3 +1,15 @@
+# Building and Debuging
+
+Like `rustc` and `cargo`, `sccache` uses a separate `env_logger` environment
+variable to configure the logging. For sccache this is called `SCCACHE_LOG`.
```
froydnj, line 4: I don't think a section on "building and debugging" belongs in documentation about releasing `sccache`.  You probably want to update https://github.com/mozilla/sccache#debugging instead.

rthomas: Ah yes totally - will make this change.

rthomas: Done.

mstange:
This didn't update the sccache-dist binary, which still uses `RUST_LOG`:
https://github.com/mozilla/sccache/blob/b3f4727eda81c45b0632d036395af85429c4774b/src/bin/sccache-dist/main.rs#L433

rthomas:
Thanks @mstange I must have missed this. I'm unavailable for the next week and will get to this in about 10 days.

rthomas:
I've added this in #977.

PR #829, opened 2020-08-16
Allow disabling DETACHED_PROCESS per environment variable

Systemcluster:
This is a workaround for #514. It leaves the current behavior as default but allows setting an environment variable to disable the `DETACHED_PROCESS` flag.

This is an alternative to #525.


Systemcluster:
Any interest in merging this? The issue is affecting quite a few people and this should be fairly non-invasive.

glandium:
Closing per https://github.com/mozilla/sccache/issues/514#issuecomment-741557322

PR #836, opened 2020-08-25
In case we fail to read the compressed content from the cache, treat the failure as a forced recache.

rickystewart:
This prevents us from failing build jobs unconditionally if cache entries are corrupt (or in case of other sporadic failures), in which case we should continue on with force-recompiling and recaching.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -96,6 +96,18 @@ pub struct CacheRead {
     zip: ZipArchive<Box<dyn ReadSeek>>,
 }
 
+/// Represents a failure to decompress stored object data.
+#[derive(Debug)]
+pub struct DecompressionFailure;
+
+impl std::fmt::Display for DecompressionFailure {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "failed to decompress content")
+    }
+}
+
+impl std::error::Error for DecompressionFailure {}
+
 impl CacheRead {
     /// Create a cache entry from `reader`.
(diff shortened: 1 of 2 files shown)


rickystewart:
Sorry about the big diff -- I had to indent everything because futures.

The actual issue from https://bugzilla.mozilla.org/show_bug.cgi?id=1660614 was difficult for me to reproduce locally, but I screwed up my own cache by replacing every cache entry with a zip file containing a malformed garbage zstd blob and ran a new build of Firefox which succeeded and re-populated the cache with correctly formatted entries as expected.

src/cache/cache.rs
```diff
@@ -113,15 +125,12 @@ impl CacheRead {
...
-            .by_name(name)
-            .context("Failed to read object from cache entry")?;
+        let file = self.zip.by_name(name).or(Err(DecompressionFailure))?;
         if file.compression() != CompressionMethod::Stored {
-            bail!("Cache entry is not stored as a zstd blob");
+            bail!(DecompressionFailure);
         }
         let mode = file.unix_mode();
-        zstd::stream::copy_decode(file, to)?;
+        zstd::stream::copy_decode(file, to).or(Err(DecompressionFailure))?;
```
froydnj, line 133: It might be useful to differentiate between zipfile errors/corruption and zstd errors, probably for purposes of `std::fmt::Display`.  But that might be a distinction without a difference, dunno.

rickystewart: Yeah, for my money, there's not a meaningful difference. Either way decompression failed and the course of action is going to be the same (namely, do the compilation and re-cache).

PR #840, opened 2020-09-09
New entry in ServerStartup enum to indicate AddrInUse Error

sum12:
when cargo fires of parallel build jobs, we encounter a race to start
the server if the server is not already running to start the server

This is an AddrInUse error causing atleast one of the parallel build
jobs fail.

This patch adds a new entry to ServerStartup enum to clearly communicate
the AddrInUse error from the server to the client that requested the
server start. Using this information the client can retry connecting
instead of failing.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -79,6 +79,8 @@ const DIST_CLIENT_RECREATE_TIMEOUT: Duration = Duration::from_secs(30);
 pub enum ServerStartup {
     /// Server started successfully on `port`.
     Ok { port: u16 },
+    /// Server Addr already in suse
+    AddrInUse,
     /// Timed out waiting for server startup.
     TimedOut,
     /// Server encountered an error.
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -412,8 +414,14 @@ pub fn start_server(config: &Config, port: u16) -> Result<()> {
         }
         Err(e) => {
             error!("failed to start server: {}", e);
-            let reason = e.to_string();
-            notify_server_startup(&notify, ServerStartup::Err { reason })?;
+            if let Some(io_err) = e.downcast_ref::<io::Error>() {
+                if let io::ErrorKind::AddrInUse = io_err.kind() {
```
luser, line 418: This doesn't need to be `if let`, it could just be `if io_err.kind() == io::ErrorKind::AddrInUse {`. But furthermore, I believe you could write this as:
```rust
match e.downcast_ref::<io::Error>() {
  Some(io_err) if io_err.kind() == io::ErrorKind::AddrInUse => notify_server_startup(&notify, ServerStartup::AddrInUse)?;
  _ => {
    let reason = e.to_string();
    notify_server_startup(&notify, ServerStartup::Err { reason })?;
  }
}
```

sum12: thanks

mstange:
Can this be merged? @glandium?

rickystewart (approved):

PR #842, opened 2020-09-09
Consume sanitizer blacklist as extra hash in msvc

squarewave:
This just extends PR #664 to work with `clang-cl` / `--driver-mode=cl`, given that it still accepts the `-fsanitizer-blacklist` argument. Given that we use the `--driver-mode=cl` argument when building Firefox on windows, this fixes caching issues when building Firefox for Windows.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -228,6 +228,7 @@ ArgData! {
     PassThrough, // Miscellaneous flags that don't prevent caching.
     PassThroughWithPath(PathBuf), // As above, recognised by prefix.
     PassThroughWithSuffix(OsString), // As above, recognised by prefix.
+    ExtraHashFile(PathBuf),
     XClang(OsString), // -Xclang ...
     Clang(OsString), // -clang:...
 }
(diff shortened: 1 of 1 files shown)


squarewave:
@froydnj this appears to be the culprit WRT the issue we were chatting about a few days ago.

froydnj (approved):
Oof, nice find.  Does this mean that we also don't properly handle `-fplugin` (or `-load`) argument as an extra hash file?

froydnj:
Also, is it possible to add a test for this @squarewave ?

squarewave:
> 
> 
> Oof, nice find. Does this mean that we also don't properly handle `-fplugin` (or `-load`) argument as an extra hash file?

It doesn't look like `clang-cl` accepts the `-fplugin` argument, and I can't find documentation on the `-load` argument. Any insights here @froydnj?



> 
> 
> Also, is it possible to add a test for this @squarewave ?

Sure!

froydnj:
> > Oof, nice find. Does this mean that we also don't properly handle `-fplugin` (or `-load`) argument as an extra hash file?
> 
> It doesn't look like `clang-cl` accepts the `-fplugin` argument, and I can't find documentation on the `-load` argument. Any insights here @froydnj?

Ah, I see what this is.  The plugin-related arguments (at least in Firefox) are passed through behind `-Xclang` flags, so they get processed further down; the problem here only arises because `-fsanitize-blacklist` is handled as a "toplevel" option.  Whew!

PR #866, opened 2020-10-09
server: support a configurable timeout

mathstuf:
With large caches, the server may take a long time to respond.

---
Cc: @bradking


luser:
This feels like an awkward band-aid papering over the actual problem, honestly. It's obviously simpler than fixing the underlying issue but pushing this complexity onto users doesn't feel great. Making the cache creation code `async` so that `LRUDiskCache` could do the full scan on a background thread would be much nicer.

mathstuf:
I took a quick glance at it and it doesn't seem trivial. I think an async `WalkDir` would be needed for that to really be effective. I think just shoving the `LruDiskCache` into a background thread behind a `Mutex` and a worker thread is likely more doable today.

mathstuf:
#868 is likely a better solution overall. Closing in favor of it.

Issue #45, opened 2016-12-12
Fail compiles on misconfigured S3 bucket


alexcrichton:
We discussed this on IRC, but a nifty feature to help debug misconfiguration of S3 credentials would be to fail compilations (perhaps with an error message) if the S3 cache management fails.

I ran into two separate problems during integration:

* First I forgot to make the whole S3 bucket public-readable
* Next, I used the wrong credentials by accident

I think both cases have a fairly predictable error which can flag "misconfiguration" as opposed to transient error, and it'd definitely help working through the setup!

alexcrichton:
One thing I just remembered on https://github.com/rust-lang/rust/pull/38359 though was that we don't have secret env vars on PRs, so the configuration where the access key is specified but the secret key isn't would be nice to continue working.

nierob:
I had the same, a typo in credentials killed 15 min of my time :-)

Ten0:
same here, 30min

sylvestre:
don't hesitate to propose patches to improve this :)

PR #868, opened 2020-10-12
disk: initialize the LruDiskCache in a background thread

mathstuf:
---
Cc: @luser

### DIFF
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -15,18 +15,48 @@
 use crate::cache::{Cache, CacheRead, CacheWrite, Storage};
 use crate::lru_disk_cache::Error as LruError;
 use crate::lru_disk_cache::LruDiskCache;
-use std::ffi::OsStr;
+use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
 use crate::errors::*;
 
+enum LazyDiskCache {
+    Uninit { root: OsString, max_size: u64 },
+    Init(LruDiskCache),
+}
+
+impl LazyDiskCache {
+    fn get_or_init(&mut self) -> Result<&mut LruDiskCache> {
+        match self {
+            LazyDiskCache::Uninit { root, max_size } => {
+                *self = LazyDiskCache::Init(LruDiskCache::new(&root, *max_size)?);
+                self.get_or_init()
+            }
+            LazyDiskCache::Init(d) => Ok(d),
+        }
+    }
+
+    fn get(&mut self) -> Option<&mut LruDiskCache> {
+        match self {
+            LazyDiskCache::Uninit { .. } => None,
+            LazyDiskCache::Init(d) => Some(d),
+        }
+    }
+
+    fn path(&self) -> &Path {
+        match self {
+            LazyDiskCache::Uninit { root, .. } => root.as_ref(),
+            LazyDiskCache::Init(d) => d.path(),
+        }
+    }
+}
+
 /// A cache that stores entries at local disk paths.
-#[derive(Clone)]
 pub struct DiskCache {
     /// `LruDiskCache` does all the real work here.
-    lru: Arc<Mutex<LruDiskCache>>,
+    lru: Arc<Mutex<LazyDiskCache>>,
     /// Thread pool to execute disk I/O
     pool: tokio::runtime::Handle,
 }
@@ -34,15 +64,15 @@ pub struct DiskCache {
 impl DiskCache {
     /// Create a new `DiskCache` rooted at `root`, with `max_size` as the maximum cache size on-disk, in bytes.
     pub fn new<T: AsRef<OsStr>>(
-        root: &T,
+        root: T,
         max_size: u64,
         pool: &tokio::runtime::Handle,
     ) -> DiskCache {
         DiskCache {
-            //TODO: change this function to return a Result
-            lru: Arc::new(Mutex::new(
-                LruDiskCache::new(root, max_size).expect("Couldn't instantiate disk cache!"),
-            )),
+            lru: Arc::new(Mutex::new(LazyDiskCache::Uninit {
+                root: root.as_ref().to_os_string(),
+                max_size,
+            })),
             pool: pool.clone(),
         }
     }
@@ -64,7 +94,7 @@ impl Storage for DiskCache {
         self.pool
             .spawn_blocking(move || {
                 let mut lru = lru.lock().unwrap();
-                let io = match lru.get(&path) {
+                let io = match lru.get_or_init()?.get(&path) {
                     Ok(f) => f,
                     Err(LruError::FileNotInCache) => {
                         trace!("DiskCache::get({}): FileNotInCache", key);
@@ -93,7 +123,7 @@ impl Storage for DiskCache {
             .spawn_blocking(move || {
                 let start = Instant::now();
                 let v = entry.finish()?;
-                lru.lock().unwrap().insert_bytes(key, &v)?;
+                lru.lock().unwrap().get_or_init()?.insert_bytes(key, &v)?;
                 Ok(start.elapsed())
             })
             .await?
@@ -104,9 +134,9 @@ impl Storage for DiskCache {
     }
 
     async fn current_size(&self) -> Result<Option<u64>> {
-        Ok(Some(self.lru.lock().unwrap().size()))
+        Ok(self.lru.lock().unwrap().get().map(|l| l.size()))
     }
     async fn max_size(&self) -> Result<Option<u64>> {
-        Ok(Some(self.lru.lock().unwrap().capacity()))
+        Ok(self.lru.lock().unwrap().get().map(|l| l.capacity()))
     }
 }


src/cache/disk.rs
```diff
@@ -18,17 +18,59 @@ use futures_03::executor::ThreadPool;
...
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
+use std::thread::{self, JoinHandle};
 use std::time::{Duration, Instant};
 
 use crate::errors::*;
 
+enum DiskCacheInit {
+    Initializing(JoinHandle<LruDiskCache>),
+    Joining,
```
mathstuf, line 31: This is a dummy state while we're inspecting the background thread.

src/cache/disk.rs
```diff
@@ -18,17 +18,59 @@ use futures_03::executor::ThreadPool;
...
+        }
+    }
+}
+
 /// A cache that stores entries at local disk paths.
-#[derive(Clone)]
 pub struct DiskCache {
     /// `LruDiskCache` does all the real work here.
-    lru: Arc<Mutex<LruDiskCache>>,
+    lru_init: Mutex<DiskCacheInit>,
```
mathstuf, line 73: Something like `LazyInit` would be best, but it initializes on the same thread. Is there something that I'm not aware of like that in the crate ecosystem already?

src/cache/disk.rs
```diff
@@ -38,12 +80,20 @@ impl DiskCache {
...
-            )),
+            lru_init: Mutex::new(DiskCacheInit::new(root.as_ref(), max_size)),
             pool: pool.clone(),
         }
     }
+
+    fn lru(&self) -> Arc<Mutex<LruDiskCache>> {
+        let mut lru_init = self.lru_init.lock().unwrap();
+
+        if let Some(lru) = lru_init.opt_lru() {
```
mathstuf, line 91: This fast path is needed because of the `mem::replace` in the slow path. Might be able to elide it with a `BackgroundLazyInit` mechanism.

mathstuf:
FWIW, testing it in our deployment seems to resolve the startup timeout woes. We ask for stats right afterwards and it does stall (but without a timeout).

src/cache/disk.rs
samlh: Perhaps a dumb question, but why not simply add the fast path at the start of DiskCacheInit's lru function?
```rust
if let Self::Initialized(lru) = self {
   return lru;
}
```

mathstuf: Indeed, thanks. :)

mathstuf:
Ping?

src/cache/disk.rs
glandium, line 73: This makes it necessary to take 2 mutexes for any normal access to the disk cache, which is not a huge thing considering disk will be slow anyways, but sounded enough of a burden code-wise that I looked into it more closely, and it turns out it's not necessary to initialize in a background thread at all. All you really need is to not initialize in `DiskCache::new`. IOW, lazy initialization would work just as well.
My patch looks like this:
```diff
diff --git a/src/cache/disk.rs b/src/cache/disk.rs
index 6a49062..2a6285f 100644
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -17,30 +17,47 @@ use crate::util::SpawnExt;
 use futures_03::executor::ThreadPool;
 use lru_disk_cache::Error as LruError;
 use lru_disk_cache::LruDiskCache;
-use std::ffi::OsStr;
+use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
 use crate::errors::*;
 
+enum LazyDiskCache {
+    Uninit { root: OsString, max_size: u64 },
+    Init(LruDiskCache),
+}
+
+impl LazyDiskCache {
+    fn get_or_init(&mut self) -> &mut LruDiskCache {
+        match self {
+            LazyDiskCache::Uninit { root, max_size } => {
+                *self = LazyDiskCache::Init(
+                    LruDiskCache::new(&root, *max_size).expect("Couldn't instantiate disk cache!"),
+                );
+                self.get_or_init()
+            }
+            LazyDiskCache::Init(d) => d,
+        }
+    }
+}
+
 /// A cache that stores entries at local disk paths.
 #[derive(Clone)]
 pub struct DiskCache {
     /// `LruDiskCache` does all the real work here.
-    lru: Arc<Mutex<LruDiskCache>>,
+    lru: Arc<Mutex<LazyDiskCache>>,
     /// Thread pool to execute disk I/O
     pool: ThreadPool,
 }
 
 impl DiskCache {
     /// Create a new `DiskCache` rooted at `root`, with `max_size` as the maximum cache size on-disk, in bytes.
-    pub fn new<T: AsRef<OsStr>>(root: &T, max_size: u64, pool: &ThreadPool) -> DiskCache {
+    pub fn new<T: AsRef<OsStr>>(root: T, max_size: u64, pool: &ThreadPool) -> DiskCache {
         DiskCache {
             //TODO: change this function to return a Result
-            lru: Arc::new(Mutex::new(
-                LruDiskCache::new(root, max_size).expect("Couldn't instantiate disk cache!"),
-            )),
+            lru: Arc::new(Mutex::new(LazyDiskCache::Uninit { root: root.as_ref().to_os_string(), max_size })),
             pool: pool.clone(),
         }
     }
@@ -59,7 +76,7 @@ impl Storage for DiskCache {
         let key = key.to_owned();
         Box::new(self.pool.spawn_fn(move || {
             let mut lru = lru.lock().unwrap();
-            let f = match lru.get(&path) {
+            let f = match lru.get_or_init().get(&path) {
                 Ok(f) => f,
                 Err(LruError::FileNotInCache) => {
                     trace!("DiskCache::get({}): FileNotInCache", key);
@@ -85,19 +102,19 @@ impl Storage for DiskCache {
         Box::new(self.pool.spawn_fn(move || {
             let start = Instant::now();
             let v = entry.finish()?;
-            lru.lock().unwrap().insert_bytes(key, &v)?;
+            lru.lock().unwrap().get_or_init().insert_bytes(key, &v)?;
             Ok(start.elapsed())
         }))
     }
 
     fn location(&self) -> String {
-        format!("Local disk: {:?}", self.lru.lock().unwrap().path())
+        format!("Local disk: {:?}", self.lru.lock().unwrap().get_or_init().path())
     }
 
     fn current_size(&self) -> SFuture<Option<u64>> {
-        f_ok(Some(self.lru.lock().unwrap().size()))
+        f_ok(Some(self.lru.lock().unwrap().get_or_init().size()))
     }
     fn max_size(&self) -> SFuture<Option<u64>> {
-        f_ok(Some(self.lru.lock().unwrap().capacity()))
+        f_ok(Some(self.lru.lock().unwrap().get_or_init().capacity()))
     }
 }
```

However, both this patch and your patch suffer from the problem that when errors happen during the disk cache initialization (example: can't create the `SCCACHE_DIR`), then sccache fails to shutdown.

glandium (changes requested):

src/cache/disk.rs
glandium: This version works better:
```
diff --git a/src/cache/disk.rs b/src/cache/disk.rs
index 6a49062..d8707b6 100644
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -17,30 +17,62 @@ use crate::util::SpawnExt;
 use futures_03::executor::ThreadPool;
 use lru_disk_cache::Error as LruError;
 use lru_disk_cache::LruDiskCache;
-use std::ffi::OsStr;
+use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
 use crate::errors::*;
 
+enum LazyDiskCache {
+    Uninit { root: OsString, max_size: u64 },
+    Init(LruDiskCache),
+}
+
+impl LazyDiskCache {
+    fn get_or_init(&mut self) -> Result<&mut LruDiskCache> {
+        match self {
+            LazyDiskCache::Uninit { root, max_size } => {
+                *self = LazyDiskCache::Init(LruDiskCache::new(&root, *max_size)?);
+                self.get_or_init()
+            }
+            LazyDiskCache::Init(d) => Ok(d),
+        }
+    }
+
+    fn get(&mut self) -> Option<&mut LruDiskCache> {
+        match self {
+            LazyDiskCache::Uninit { .. } => None,
+            LazyDiskCache::Init(d) => Some(d),
+        }
+    }
+
+    fn path(&self) -> &Path {
+        match self {
+            LazyDiskCache::Uninit { root, .. } => root.as_ref(),
+            LazyDiskCache::Init(d) => d.path(),
+        }
+    }
+}
+
 /// A cache that stores entries at local disk paths.
 #[derive(Clone)]
 pub struct DiskCache {
     /// `LruDiskCache` does all the real work here.
-    lru: Arc<Mutex<LruDiskCache>>,
+    lru: Arc<Mutex<LazyDiskCache>>,
     /// Thread pool to execute disk I/O
     pool: ThreadPool,
 }
 
 impl DiskCache {
     /// Create a new `DiskCache` rooted at `root`, with `max_size` as the maximum cache size on-disk, in bytes.
-    pub fn new<T: AsRef<OsStr>>(root: &T, max_size: u64, pool: &ThreadPool) -> DiskCache {
+    pub fn new<T: AsRef<OsStr>>(root: T, max_size: u64, pool: &ThreadPool) -> DiskCache {
         DiskCache {
             //TODO: change this function to return a Result
-            lru: Arc::new(Mutex::new(
-                LruDiskCache::new(root, max_size).expect("Couldn't instantiate disk cache!"),
-            )),
+            lru: Arc::new(Mutex::new(LazyDiskCache::Uninit {
+                root: root.as_ref().to_os_string(),
+                max_size,
+            })),
             pool: pool.clone(),
         }
     }
@@ -59,7 +91,7 @@ impl Storage for DiskCache {
         let key = key.to_owned();
         Box::new(self.pool.spawn_fn(move || {
             let mut lru = lru.lock().unwrap();
-            let f = match lru.get(&path) {
+            let f = match lru.get_or_init()?.get(&path) {
                 Ok(f) => f,
                 Err(LruError::FileNotInCache) => {
                     trace!("DiskCache::get({}): FileNotInCache", key);
@@ -85,7 +117,7 @@ impl Storage for DiskCache {
         Box::new(self.pool.spawn_fn(move || {
             let start = Instant::now();
             let v = entry.finish()?;
-            lru.lock().unwrap().insert_bytes(key, &v)?;
+            lru.lock().unwrap().get_or_init()?.insert_bytes(key, &v)?;
             Ok(start.elapsed())
         }))
     }
@@ -95,9 +127,9 @@ impl Storage for DiskCache {
     }
 
     fn current_size(&self) -> SFuture<Option<u64>> {
-        f_ok(Some(self.lru.lock().unwrap().size()))
+        f_ok(self.lru.lock().unwrap().get().map(|l| l.size()))
     }
     fn max_size(&self) -> SFuture<Option<u64>> {
-        f_ok(Some(self.lru.lock().unwrap().capacity()))
+        f_ok(self.lru.lock().unwrap().get().map(|l| l.capacity()))
     }
 }
```

However, both this new patch and your patch suffer from the problem that the first compile is going to wait for the cache initialization, and if that takes more than 60s, will bail out.

The ideal solution would be to not have the cache initialization blocking reads and writes from the cache at all.

mathstuf:
No idea what's up with the macOS build, but there doesn't seem to be a "rerun failed", but just a "rerun all"?

That said, this takes @glandium's suggestion. Making the read/write not block on the init sounds fine to me, but out-of-scope right now. I suppose it would queue a log of added contents to the repo so that it can delete as needed when it finally catches up?

Is there a model for multiple `sccache` servers using the same on-disk cache at once?

luser:
Personally, if I had a do-over I would just make [`storage_from_config`](https://github.com/mozilla/sccache/blob/b3f4727eda81c45b0632d036395af85429c4774b/src/cache/cache.rs#L288) `async`, then make the server startup process return a response once that future was spawned, but not handle compile requests until it finishes. The simplest way to do that would be to make it return `SFuture<Arc<dyn Storage>>`, then block on that future in [`SccacheServer::new`](https://github.com/mozilla/sccache/blob/c447c904ce7da76f5e0ab61803564a1e969fca0a/src/server.rs#L511) before accepting incoming connections. That might not be ideal though, since that could make compile requests timeout if scanning the disk cache takes too long. Alternately, you could make `SccacheService` store something like:

```rust
enum MaybeStorage {
  Initializing(SFuture<Arc<dyn Storage>>),
  Ready(Arc<dyn Storage>),
}
```

...and `handle_compile` could match on that and in the `Initializing` case it could `await` the storage future before serving the compile request.

As a bonus, this would give the server a way to notify of errors on a misconfigured cache (#45, longstanding issue), as other cache backends could do more work during initialization without blocking server startup. (The s3 backend could attempt to write a test entry to the bucket and then read it back to validate that the provided credentials work, for example.)

luser:
> Is there a model for multiple `sccache` servers using the same on-disk cache at once?

That's explicitly not supported. As currently written, `sccache` wants exclusive access to its disk cache because it maintains the current LRU state in memory. Disk storage was added as a nice-to-have to make it useful for local development, the original purpose of `sccache` was to be useful in CI with cloud storage. It would certainly be possible to make disk storage in `sccache` work more like `ccache`, where metadata is stored on-disk as part of the cache and concurrent access is allowed, but honestly it'd probably be less work to just write a simple cache daemon that implements the `memcached` protocol and stores cache entries on disk if someone wanted a lightweight way to share local disk cache between several `sccache` instances. (You could take the existing `LruDiskCache` code and wrap it in a simple network daemon and be 95% of the way there.)

mathstuf:
Is there anything else I need to do here?

mitchhentges (approved):
This looks good to me - can you rebase this off of `main` and we'll get this landed?
Thanks for the patch mate :beers:

mathstuf:
Rebased.

mitchhentges:
> I get an email that the CI has failed, can find the job through those links

That's weird, I can't see it either. Can you share one of the links here so I can take a look?

mathstuf:
Oh! It was the CI run for the push. It was just a timeout, so nothing dire, just…confusing.

mitchhentges:
Sweet! Let's land this then :sunglasses:

PR #826, opened 2020-08-11
Use Rusoto to handle S3 communication

Hugal31:
Replacement for PR #522.

I am not familiar with futures (my god when this mayhem of mixing future::Future and core::Future will stop?), so feel free to add your remarks.

- [x] Use rusuto
- [x] Checks that behavior is the same as the old one (I am thinking of the HTTP parameters, such as cache duration, etc...)
- [x] Remove old simples3 and its dependencies.
- [x] Test other regions.
- [x] Add a region parameter.
- [x] Add support for the `endpoint` S3 parameter.
- [x] Add docs for the new configuration, and document SCCACHE_REGION.

BREAKING CHANGES:

- The credentials are not searched in `$HOME/.boto` anymore. It is not required if I understand correcly luser comment in #522, but I may re-add it if one ask for it.
- `use_ssl` is gone. You'll have to set the endpoint url beginning with `http://` to use S3 without SSL.

Closes #632.

Issue #632, opened 2019-12-27
403 when trying to write to S3


dsilvasc:
Hi, I've created a private S3 bucket and verified that I can write to it:

```bash
export AWS_ACCESS_KEY_ID=...
export AWS_SECRET_ACCESS_KEY=...
export BUCKET=...
echo hi > hello.txt
aws s3 cp hello.txt s3://$BUCKET/test.txt
aws s3 cp s3://$BUCKET/test.txt /dev/stdout
```

It prints `hi` as expected.

With the same environment variables, I start an sccache server:

```bash
export SCCACHE_BUCKET=$BUCKET
SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 RUST_LOG=sccache=trace sccache
```

Now when I try a `cargo build` somewhere, I get lots of failures like this:

```
 WARN 2019-12-27T23:25:54Z: sccache::cache::s3: Got AWS error: Error(BadHTTPStatus(403), State { next_error: None, backtrace: InternalBacktrace })
...
TRACE 2019-12-27T23:25:54Z: sccache::simples3::s3: PUT failed with HTTP status: 403 Forbidden
DEBUG 2019-12-27T23:25:54Z: sccache::compiler::compiler: [arrayvec]: Cache write error: Error(Msg("failed to put cache entry in s3"), State { next_error: Some(Error(BadHTTPStatus(403), State { next_error: None, backtrace: InternalBacktrace })), backtrace: InternalBacktrace })
DEBUG 2019-12-27T23:25:54Z: sccache::server: Error executing cache write: failed to put cache entry in s3
ERROR 2019-12-27T23:25:54Z: sccache::server: SccacheTransport::poll failed: io error.
ERROR 2019-12-27T23:25:54Z: sccache::server: io error.
...
```

Does sccache need anything other than AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY to write to s3?

chin33z:
I had exactly the same issue with you but can't figure it out. What is your s3 bucket region? My `ap-southeast-1` is fine but `ap-east-1` is not.

nickpascucci:
I'm having the same issue as well. It would be great if the `sccache` documentation would list exactly what permissions are needed for S3 bucket cache stores, as I don't think this is actually documented anywhere — I certainly have not been able to find it.

pcorliss:
[...]

PR #869, opened 2020-10-18
Rusoto for S3 communication (with support for public buckets).

rexhoffman:
Picking up on the work of @Hugal31 in https://github.com/mozilla/sccache/pull/826 with support for s3 buckets, but now with support for public buckets as well (read only mode).

Prior to this fix, the the https://rusoto.github.io/rusoto/rusoto_credential/struct.DefaultCredentialsProvider.html called the https://rusoto.github.io/rusoto/rusoto_credential/struct.ChainProvider.html and would error out if no credentials were found.

Based on this pr https://github.com/rusoto/rusoto/pull/1566 a it seems anonymous mode is a mechanism that is appearing in Rusoto, but is not yet a supported via DefaultCredentialsProvider, or perhaps ever, so using a different S3Client constructor.  

It would also be possible to build our own extension of DefaultCredentialsProvider, but an upstream builder might be a better solution.


rexhoffman:
By the way we are currently using this in libra, https://github.com/libra/libra/blob/master/x.toml#L11-L23, via our build extension, cargo xbuild.   Can confirm success with read/write and public read from modern s3 bucket.

Crunch09:
Using this branch as well in our build now with a recently created bucket and can confirm that it works great, thank you @rexhoffman ! 💯 

Getting this warning which might be an import you no longer need?

image

rexhoffman:
Rebased since the Cargo.lock file was conflicting, also fixed the warn @Crunch09 commented on.

rexhoffman:
@glandium any chance of this being reviewed, or possibly merged in?

rexhoffman:
@froydnj perhaps you've time to look at this pr?   Not sure who's engaged w/ this project?

froydnj:
> @froydnj perhaps you've time to look at this pr? Not sure who's engaged w/ this project?

I can look, but as I do not have write access, I doubt my looking will do much good. :grin:  It certainly looks a lot nicer than the old code, though!

glandium:
I'm going through all the PRs this week, I'll eventually reach this one.

glandium:
I haven't gone through the entire PR yet, but I'm noticing that it's removing support for AWS_IAM_CREDENTIALS_URL and rusoto doesn't seem to have a knob for something equivalent, and we use AWS_IAM_CREDENTIALS_URL at Mozilla: https://searchfox.org/mozilla-central/rev/ffdb6a4d934b3f5294f18cf0e1df618109ccb36b/build/mozconfig.cache#77
So adding some shim to keep that supported would be great.

rexhoffman:
@glandium has [rusoto going maintain mode](https://github.com/rusoto/rusoto/issues/1651) changed anything on your side?   I final have some time to dig in to sccache again.

I was about to start drilling in to the shim you'll need for AWS_IAM_CREDENTIALS_URL today.

Fuuzetsu:
I tried this branch today because due to https://github.com/mozilla/sccache/issues/632 we were unable to use a private bucket in `ap-northeast-1` region. It seems to be working.

Any plan to push this forward?

tomjohnburton:
We would love this as well. Getting a 400 error for a private bucket.

nicholastmosher:
Ran into this problem today as well, this fix would be greatly appreciated!

tobz:
Bumping this as without it, using a recently-created private (maybe public, too) S3 bucket is impossible.

The tests pass, it seems to work fine… is there a specific hold up? Does it need reviewer time that could be provided outside of the core maintainers?

tomelliff:
Given that Rusoto isn't being actively maintained any more and that [AWS have announced an official Rust SDK](https://aws.amazon.com/blogs/developer/a-new-aws-sdk-for-rust-alpha-launch/) that is generated from their API I'd suggest that maybe it would be better to attempt to switch to this new official AWS Rust SDK instead as Rusoto is likely a dead end.

tobz:
If someone wants to roll a PR to use it, that seems fine, although I'm not sure choosing an alpha library is demonstrably better than a mature (even if maintenance mode) library, given that this PR fixes problems _right now_.  As evidenced by the number of other issues and PRs linked to this one, the can has already been kicked down the road significantly.

Xanewok:
I just want to echo a statement that using a mature, albeit only maintained library seems like a good solution for now until the official SDK is properly released (at least the S3 parts of it, that is).

benhannel:
This worked great for us! We're now able to use a private bucket for our cache. Thanks!

sylvestre:
would it be possible to rebase this PR? it is full of conflicts, sorry :(

jschwe:
> So adding some shim to keep that supported would be great.

Is that requirement from mozilla still up to date? The v2 implementation uses Sha1 as a digest for signing and AFAIK sha1 is viewed as insecure for cryptographic purposes.

PR #879, opened 2020-11-12
remove openssl

drahnr:
replace `openssl` with `rustls`, `rsa`, `rsa-pem`, `rsa-export` and `picky`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2462,6 +2462,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "rsa-export"
+version = "0.1.1"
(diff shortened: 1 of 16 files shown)


TriplEight:
Please do not force-push. It shatters gitlab mirroring.

mitchhentges:
Ooh, nice, I think that this makes a lot of sense - and a pretty large patch too, dang.
Would you be able to update this patch, rebasing it off of `master`? This would be a sweet upgrade.
I see that `cachepot` is already `openssl`-less, so the best approach for this migration might be to graft their solution.

Anyways, thanks for the patch, and sorry for the delay here.

PR #896, opened 2020-12-03
Pass -clang:-frewrite-includes to clang-cl.

### LINKED ISSUE #894: Support rewrite_includes_only with clang-cl
To avoid large swathes of compiler warnings during distributed compilation, we try to limit client-side preprocessing to only resolve includes, and not substitute #defines etc. This mode is enabled when `rewrite_includes_only = true` is set in the dist config.

However, this mode is currently unimplemented on Windows: msvc.rs ignores the `rewrite_includes_only` parameter.

https://github.com/mozilla/sccache/blob/e66c9c15142a7e583d6ab80bd614bdffb2ebcc47/src/compiler/msvc.rs#L69

MSVC's cl.exe does not appear to support an argument to unlock this mode. However, clang-cl does, in an underhanded way, with the `-clang:-frewrite-includes` argument. So we can support `rewrite_includes_only` in msvc.rs if clang-cl is used.

With this change, I've been able to build Firefox on Windows with distributed sccache successfully. Without this change, the build was 1. failing because some directories had -Werror enabled and 2. extremely slow, because the console delayed everything by trying to keep up with the huge amounts of warning output.

mstange:
Fixes #894.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -66,7 +66,7 @@ impl CCompilerImpl for MSVC {
         cwd: &Path,
         env_vars: &[(OsString, OsString)],
         may_dist: bool,
-        _rewrite_includes_only: bool,
+        rewrite_includes_only: bool,
     ) -> SFuture<process::Output>
     where
         T: CommandCreatorSync,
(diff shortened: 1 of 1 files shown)


emilio (approved):
LGTM, for what is worth, but I can't merge this.

PR #906, opened 2020-12-16
Set the current directory of the server to the directory containing the executable

### LINKED ISSUE #722: File handle to root dir left open
For my sins I'm building rust in a conda environment. That environment's directory is renamed at the end of the build.

But fails to be renamed as sscache has an open handle to the dir in question: 
C:\conda-bld\my-proj_1586446013252\work

It would be really great if it didn't keep file handles open. You can see it in process explorer that it's kept a handle on the build dir open.

glandium:
On Windows, files and directories can't be modified if they are "locked" by some process that has them opened. When the server is spawned on the first compile, the directory it was launched from is kept open as its current directory, which prevents some actions on the directory later on (like removing it or renaming it), as long as the server keeps running.

One directory that is less likely to be renamed or removed during the lifetime of the server is the directory containing sccache itself, so we use that as sccache's current directory instead.

Fixes #722

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -84,7 +84,9 @@ fn run_server_process() -> Result<ServerStartup> {
     let mut runtime = Runtime::new()?;
     let listener = tokio_uds::UnixListener::bind(&socket_path)?;
     let exe_path = env::current_exe()?;
-    let _child = process::Command::new(exe_path)
+    let workdir = exe_path.parent().expect("executable path has no parent?!");
+    let _child = process::Command::new(&exe_path)
+        .current_dir(workdir)
         .env("SCCACHE_START_SERVER", "1")
         .env("SCCACHE_STARTUP_NOTIFY", &socket_path)
         .env("RUST_BACKTRACE", "1")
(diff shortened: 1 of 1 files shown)


src/commands.rs
```diff
@@ -209,6 +211,13 @@ fn run_server_process() -> Result<ServerStartup> {
         v.push(0);
         v
     };
+    let workdir = exe_path
+        .parent()
+        .expect("executable path has no parent?!")
+        .as_os_str()
+        .encode_wide()
+        .chain(Some(0u16))
```
mitchhentges, line 219: Hmm, I'm surprised that there isn't a better existing abstraction that we can use here that encapsulates these technical details. Something like:
```
let workdir: WindowsStringParameterThinger = exe_path
    .parent()
    .expect(...)
    .as_os_str()
    .into::<WindowsStringParameterThinger>();

// ...

CreateProcessW(
    ...,
    workdir.to_ptr()
    ...,
)
```

mitchhentges (approved):

cgsheeh (approved):
Looks good to me - guessing we can't avoid taking the lock on the directory altogether as the issue reporter suggests?

glandium:
I don't think it's possible to close the handle for the current directory, no.

src/commands.rs
glandium: There is https://crates.io/crates/widestring but we're not currently using it. I just used the same pattern as already in use in the same file.

PR #907, opened 2020-12-16
Use a full path relative to `cwd` for extra hash files

### LINKED ISSUE #843: -fsanitize-blacklist does not work correctly with relative paths
I'm trying to replace ccache with sccache in the AOSP build system but I'm facing a few errors
This is the first issue that stems from using the `-c` flag of clang enabling the use of the integrated clang assembler

```
cat foo.cpp              
#include <stdlib.h>
void bad_foo() {
  int *a = (int*)malloc(40);
  a[10] = 1;
}
int main() { bad_foo(); }
```

```
cat blacklist.txt 
# Ignore reports from bad_foo function.
fun:bad_foo
```

```
sccache clang++ -c -fsanitize=address -fsanitize-blacklist=blacklist.txt foo.cpp
sccache: error : Failed to open file for hashing: "blacklist.txt"
sccache:  cause: Failed to open file for hashing: "blacklist.txt"
sccache:  cause: No such file or directory (os error 2)
```

The same command without `-c` works properly
```
sccache clang++ -fsanitize=address -fsanitize-blacklist=blacklist.txt foo.cpp
```

glandium:
When the current directory of the sccache process is not the same as
the current directory for the compiler process, extra hash files are
currently not found, leading to errors.

Fixes #843

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -145,7 +145,7 @@ mod test {
         Clang {
             clangplusplus: false,
         }
-        .parse_arguments(&arguments, ".".as_ref())
+        .parse_arguments(&arguments, &std::env::current_dir().unwrap())
     }
 
     macro_rules! parses {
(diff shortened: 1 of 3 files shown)


mitchhentges (approved):
Looks good, but isn't the "full path relative to the `cwd`" equivalent to the "absolute path"? Perhaps the commit message should be simplified accordingly ("Use an absolute path for extra hash files"?)

glandium:
It's not necessarily absolute, although it is in practice (for example, if I hadn't changed the tests to use absolute paths, the paths would have been relative). But the important part is that the path is complete rather than absolute.

PR #908, opened 2020-12-16
Ignore both -FS and -MP on MSVC

### LINKED ISSUE #859: Only block /FS when building with PDBs
Currently /FS is marked as too hard to cache, see https://github.com/mozilla/sccache/blob/5197fbbca5b6a07bcec7e91df8b0ca8d0f32e8da/src/compiler/msvc.rs#L266
This is problematic since CMake enables this flag for all MSVC builds on windows: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-MSVC-CXX.cmake

It seems to me that caching with this flag is safe when building without /Zi.

glandium:
Fixes #859


mitchhentges (approved):
Makes sense to me

cgsheeh (approved):

PR #911, opened 2020-12-22
Switch CI to Github Actions

glandium:
Fixes: #910

PR #914, opened 2020-12-24
 Address clippy issues raised in #911

### LINKED ISSUE #913: Address clippy issues raised in #911

glandium:
Fixes #913

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1401,10 +1401,7 @@ LLVM version: 6.0",
             }))
             .unwrap();
         // Ensure that the object file was created.
-        assert_eq!(
-            true,
-            fs::metadata(&obj).and_then(|m| Ok(m.len() > 0)).unwrap()
-        );
+        assert_eq!(true, fs::metadata(&obj).map(|m| m.len() > 0).unwrap());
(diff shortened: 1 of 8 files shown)


src/compiler/compiler.rs
```diff
@@ -1401,10 +1401,7 @@ LLVM version: 6.0",
             }))
             .unwrap();
         // Ensure that the object file was created.
-        assert_eq!(
-            true,
-            fs::metadata(&obj).and_then(|m| Ok(m.len() > 0)).unwrap()
-        );
+        assert_eq!(true, fs::metadata(&obj).map(|m| m.len() > 0).unwrap());
```
mitchhentges, line 1404: Should this be `assert!(...)`, instead of asserting that `true` is equal to your computation?
This is old code, but hey, while we're fixing lint-worthy issues :)

mitchhentges (approved):
Clippy is fantastic!

cgsheeh (approved):
Same thought as Mitch, `s/assert_eq!(true, /assert!(/g` :)

cgsheeh (approved):

mitchhentges:
Looks good :+1:

PR #921, opened 2021-01-06
Still read config file from Library/Preferences if it exists

### LINKED ISSUE #919: Config file location on macOS changed in 0.2.14 and needs to be documented
185c6ba4b0d360e6be8daf15a9fb80899c030df5 updated the `directories` dependency from 2 to 3. This contained a breaking change for the location of the config dir on macOS: https://github.com/dirs-dev/directories-rs#3

As a result, the config file location on macOS changed from `~/Library/Preferences/Mozilla.sccache/config` to `~/Library/Application Support/Mozilla.sccache/config`.

The documentation needs to be updated to reflect that. Ideally, if there are sccache release notes somewhere, this should be mentioned there, too.

glandium:
Fixes #919

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -546,6 +546,29 @@ fn config_from_env() -> EnvConfig {
     EnvConfig { cache }
 }
 
+// The directories crate changed the location of `config_dir` on macos in version 3,
+// so we also check the config in `preference_dir` (new in that version), which
+// corresponds to the old location, for compatibility with older setups.
+fn config_file(env_var: &str, leaf: &str) -> PathBuf {
+    if let Some(env_value) = env::var_os(env_var) {
+        return env_value.into();
+    }
+    let dirs =
+        ProjectDirs::from("", ORGANIZATION, APP_NAME).expect("Unable to get config directory");
+    // If the new location exists, use that.
+    let path = dirs.config_dir().join(leaf);
+    if path.exists() {
+        return path;
+    }
+    // If the old location exists, use that.
+    let path = dirs.preference_dir().join(leaf);
+    if path.exists() {
+        return path;
+    }
+    // Otherwise, use the new location.
+    dirs.config_dir().join(leaf)
+}
+
 #[derive(Debug, Default, PartialEq, Eq)]
 pub struct Config {
     pub caches: Vec<CacheType>,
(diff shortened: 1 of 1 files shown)


glandium:
@mstange Can you double check that this uses a config from ~/Library/Preferences if you move it back there?

mstange:
I will check tomorrow.

Looking at the patch, I think that if both files are present, the patch prefers the one in Preferences. I think it would be better to prefer the one in the new location.

mitchhentges (approved):
Looks good, but I agree with :mstange's suggestion on prioritizing our new config path first.

mstange:
> @mstange Can you double check that this uses a config from ~/Library/Preferences if you move it back there?

Confirmed, the patch works.

```
% mv ~/Library/Application\ Support/Mozilla.sccache/config ~/Library/Preferences/Mozilla.sccache/
% SCCACHE_LOG=trace ./target/release/sccache --dist-status
[...]
[2021-01-06T17:23:37Z DEBUG sccache::config] Attempting to read config file at "/Users/mstange/Library/Preferences/Mozilla.sccache/config"
[...]
{"SchedulerStatus":["http://192.168.2.146:10600/",{"num_servers":1,"num_cpus":16,"in_progress":0}]}%
```

src/config.rs
```diff
@@ -546,6 +546,19 @@ fn config_from_env() -> EnvConfig {
...
+        return env_value.into();
+    }
+    let dirs =
+        ProjectDirs::from("", ORGANIZATION, APP_NAME).expect("Unable to get config directory");
+    let path = dirs.preference_dir().join(leaf);
+    if path.exists() {
+        return path;
+    }
+    dirs.config_dir().join(leaf)
+}
```
cgsheeh, line 560: Maybe we fast-path the new path if it already exists, if not we check the old path and return that if it exists, and return the new path if neither exist yet.

Another option is to migrate the config file and/or log a warning somewhere saying we're moving the config file and we're deprecating use of the old one.

src/config.rs
```diff
@@ -546,6 +546,19 @@ fn config_from_env() -> EnvConfig {
     EnvConfig { cache }
 }
 
+fn config_file(env_var: &str, leaf: &str) -> PathBuf {
```
cgsheeh, line 549: Regardless of what we do, we should have a comment describing why we need to do this config-file-search dance.

cgsheeh (changes requested):

cgsheeh (approved):

Issue #235, opened 2018-03-13
C compilers bottoming out in scripts may not cache correctly


alexcrichton:
This is sort of similar to https://github.com/mozilla/sccache/issues/87 except for C compilers. First discovered [here](https://github.com/rust-lang/rust/pull/48647#issuecomment-372795772) we had a situation that looked like:

* Two distinct toolchains (but built by the same entity) were installed in two different locations
* Two different compilers, `mips-openwrt-linux-gcc` and `mipsel-openwrt-linux-gcc`, were used
* Both compilers were a trail of symlinks bottoming out in a shell scripts (but shell scripts at two different paths on the filesystem)
* Both shell scripts, however, were exactly the same! They used dispatch based on `$0` to figure out where to go (and is why the shell script operated differently in the two locations)
* Inside of `sccache` the [`executable_digest` part of the hash key](https://github.com/mozilla/sccache/blob/25eba29e8e5fa2bc76bc2a34c3e1fc296d4d1d95/src/compiler/c.rs#L239-L245) was then the same for both compilers

As a consequence this meant that the mips object files (built first) were used for the mipsel compiler, causing our issues!

I'm not really sure what the best way to solve this would be :(

luser:
Oh, this sucks. :-( gcc has a `--print-prog-name=` option that sounds almost useful, but it doesn't seem to do anything useful if you pass it cc or gcc:
```
luser@eye7:/build$ gcc --print-prog-name=gcc
gcc
luser@eye7:/build$ gcc-5 --print-prog-name=gcc-5
gcc-5
luser@eye7:/build$ gcc --print-prog-name=cc
cc
luser@eye7:/build$ gcc-5 --print-prog-name=cc
cc
```

clang's implementation of this seems to be more useful:
```
luser@eye7:/build$ clang --print-prog-name=clang
/usr/bin/clang
luser@eye7:/build$ PATH=/build/clang+llvm-5.0.1-x86_64-linux-gnu-ubuntu-16.04/bin/:$PATH clang --print-prog-name=clang
/build/clang+llvm-5.0.1-x86_64-linux-gnu-ubuntu-16.04/bin/clang
```

gcc's `-print-search-dirs` option does seem to include an `install:` key which might be usable:
```
luser@eye7:/build$ gcc -print-search-dirsinstall: /usr/lib/gcc/x86_64-linux-gnu/7/
[...]

PR #922, opened 2021-01-07
Add C compiler versions to the hash calculation when available

glandium:
The clang compiler is a relatively thin wrapper around libclang-cpp and
libllvm, and in some cases, it's possible that two identical clang
executables are actually for different builds of clang. While ideally
we could find all the libraries it could be using, a work-around,
good enough for now, is to check the compiler version, which we get from
the __VERSION__ macro in the preprocessor we're already making.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1685034

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -205,11 +205,25 @@ impl<I> CCompiler<I>
 where
     I: CCompilerImpl,
 {
-    pub fn new(compiler: I, executable: PathBuf, pool: &ThreadPool) -> SFuture<CCompiler<I>> {
+    pub fn new(
+        compiler: I,
+        executable: PathBuf,
+        version: Option<String>,
+        pool: &ThreadPool,
+    ) -> SFuture<CCompiler<I>> {
         Box::new(
             Digest::file(executable.clone(), &pool).map(move |digest| CCompiler {
                 executable,
-                executable_digest: digest,
+                executable_digest: {
+                    if let Some(version) = version {
+                        let mut m = Digest::new();
+                        m.update(digest.as_bytes());
+                        m.update(version.as_bytes());
+                        m.finish()
+                    } else {
+                        digest
+                    }
+                },
                 compiler,
             }),
         )
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1012,7 +1012,10 @@ g++
 gcc
 #elif defined(__DCC__)
 diab
+#else
+unknown
 #endif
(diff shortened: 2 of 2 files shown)


src/compiler/compiler.rs
```diff
@@ -1040,17 +1043,30 @@ diab
...
+            } else {
+                Some(line)
+            }
+        });
+        if let Some(kind) = lines.next() {
+            let version = lines
+                .next()
+                // In case the compiler didn't expand the macro.
+                .filter(|&line| line != "__VERSION__")
+                .map(str::to_owned);
```
mitchhentges, line 1059: Nice

src/compiler/compiler.rs
```diff
@@ -1173,10 +1188,7 @@ mod test {
         let f = TestFixture::new();
         let creator = new_creator();
         let pool = ThreadPool::sized(1);
-        next_command(
-            &creator,
-            Ok(MockChild::new(exit_status(0), "clang\nfoo", "")),
-        );
+        next_command(&creator, Ok(MockChild::new(exit_status(0), "clang\n", "")));
         let c = detect_compiler(creator, &f.bins[0], f.tempdir.path(), &[], &pool, None)
```
mitchhentges, line 1192: Should we modify this test to check for `version`?

src/compiler/c.rs
```diff
@@ -205,11 +205,25 @@ impl<I> CCompiler<I>
...
-                executable_digest: digest,
+                executable_digest: {
+                    if let Some(version) = version {
+                        let mut m = Digest::new();
+                        m.update(digest.as_bytes());
+                        m.update(version.as_bytes());
+                        m.finish()
+                    } else {
+                        digest
+                    }
```
mitchhentges, line 225: :100:

mitchhentges (approved):
Looks good mate

froydnj (commented):
You probably want to bump `CACHE_VERSION` in this file, too?

glandium:
> You probably want to bump `CACHE_VERSION` in this file, too?

Having the compiler version added to the hash takes care of anything that would require a CACHE_VERSION bump.

src/compiler/compiler.rs
glandium: I guess I could add a test that checks that cache keys for 2 different versions are different...

luser:
Related discussion in #235

mitchhentges (approved):
Thanks!

Issue #912, opened 2020-12-22
Publish sccache-dist artifacts in releases


glandium:
#902 mostly covered this, but a) was missing a bit that made the artifact not actually published and b) forgot about the fact we remove the `[patch]` for tiny-http during the release process, which would have produced broken binaries.

luser:
> forgot about the fact we remove the `[patch]` for tiny-http during the release process

I guess if upstream is not going to fix that issue you ought to either fork it, vendor a copy, or find a replacement.

glandium:
The issue is fixed, and released, but we're not using tiny-http directly, rouille is, and rouille hasn't been updated in more than a year.

Issue #151, opened 2017-07-13
Need a better story wrt logging


glandium:
There are essentially two logging setups, and they don't work the same way.
- If you set `RUST_LOG`, you end up with the env_logger setup, which allows fine-grained log output (I've found `RUST_LOG=sccache::compiler=debug` to be incredibly useful, so much so that I'm thinking we should set it by default on Firefox builds, as well as `SCCACHE_ERROR_LOG` and upload the log as artifact. BUT, it has no timestamps.
- If you instead set `SCCACHE_LOG_LEVEL`, you can, AIUI, only set the global log level (and `debug` is already too verbose), BUT you get timestamps (although you don't get fractions of seconds, which makes the timestamps less useful for measures). OTOH, you also always get the log sent to either stdout or stderr, I'm not sure which, on the client side (which doesn't follow `SCCACHE_ERROR_LOG`), and that breaks most build systems (for Firefox, configure just plain fails)

luser:
Agreed. It's sort of a mess. `RUST_LOG` + `SCCACHE_ERROR_LOG` is currently the best you can do. Not having the fine-grained log level filtering makes the logging unusable due to other crates having verbose logging. @alexcrichton has a PR that I have not found time to look at that might help: https://github.com/mozilla/sccache/pull/115

If you want to review that that would be appreciated!

More generally: we currently pull in both `env_logger` and `fern`, and the latter doesn't support the fine-grained log level syntax that the former does. I thought about extracting those bits into a standalone crate to make things nicer, but never quite got motivated.

PR #925, opened 2021-01-08
Workaround tiny_http issue #151 by disabling HTTP pipelining

glandium:
Instead of relying on a crate patch.

Fixes: #912

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2903,7 +2903,8 @@ dependencies = [
 [[package]]
 name = "tiny_http"
 version = "0.6.2"
-source = "git+https://github.com/tiny-http/tiny-http.git?rev=619680de#619680dea2c4f1dfb31b552d5e7b472a9f25b8b8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1661fa0a44c95d01604bd05c66732a446c657efb62b5164a7a083a3b552b4951"
 dependencies = [
  "ascii",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,10 +95,6 @@ syslog = { version = "5", optional = true }
 void = { version = "1", optional = true }
 version-compare = { version = "0.0.11", optional = true }
 
-[patch.crates-io]
-# Waiting for #151 to make it into a release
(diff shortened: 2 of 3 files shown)


glandium:
@mstange @dholbert Can you check whether a) it works properly (tests pass, which suggests it might) b) it has an impact on performance?

mitchhentges (changes requested):
~`tiny-http` has the fix from [#151](https://github.com/tiny-http/tiny-http/pull/151) in release `v0.7.0`, why don't we use that release instead of working around the bug?~

Ah, that's where `rouille fits into this, my bad.

mitchhentges (approved):

cgsheeh (approved):

dholbert:
Seems to work fine on an Ubuntu client/server/scheduler setup.

Only caveat, it doesn't seem to work when I use my Mac client (which is currently still using the latest mach-bootstrap-provided sccache version, which doesn't have this patch, if that matters).  From that client, it looks like compile jobs just hang, when the scheduler/builder are using my local build of glandium's patched revision here.

glandium:
> Only caveat, it doesn't seem to work when I use my Mac client (which is currently still using the latest mach-bootstrap-provided sccache version, which doesn't have this patch, if that matters).

For good measure, can you check updating the mac client? (you can find a prebuild sccache in the checks tab, under artifacts)
Can you also check that build times are not much different?

dholbert:
Yeah, I don't hit the compile-jobs-hanging issue anymore on Mac if I use the [prebuilt sccache client binary from the artifacts here](https://pipelines.actions.githubusercontent.com/xTZDbRczhQWc6twoYOC6b79HzZEBlkEMmvBunZE5uwkUcnjFFY/_apis/pipelines/1/runs/36/signedartifactscontent?artifactName=sccache-952c8b9c2ed2d8f4881f3c5b3b5814f3351e6d3f-x86_64-apple-darwin&urlExpires=2021-01-08T21%3A06%3A50.7288246Z&urlSigningMethod=HMACV1&urlSignature=GMgs7rUqm5Ff7YMWFnGwyyee2Dfq4oUNCwJWLeHerOc%3D) (instead of the one from bootstrap).

Build times don't seem to be particularly affected by this fix, in my nonscientific investigations so far; and core saturation (via `sccache --dist-status`) seems similar to what they were before too.

PR #927, opened 2021-01-09
Upgrade reqwest

tetsuharuohzeki:
It's complex to upgrade all crates related to http, but it's easy relatively to upgrade only `reqwest`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ futures = "0.1.11"
 futures_03 = { package = "futures", version = "0.3", features = ["compat", "thread-pool"] }
 hmac = { version = "0.10", optional = true }
 http = "0.1"
+http_02 = { package = "http", version = "0.2" }
 hyper = { version = "0.12", optional = true }
 hyperx = { version = "0.12", optional = true }
 jobserver = "0.1"
@@ -58,7 +59,7 @@ percent-encoding = { version = "2", optional = true }
 rand = "0.7"
 redis = { version = "0.17", optional = true }
 regex = "1"
-reqwest = { version = "0.9.11", optional = true }
+reqwest = { version = "0.10", optional = true, features = ["blocking", "json"] }
 retry = "1"
 ring = { version = "0.16", optional = true, features = ["std"] }
 sha-1 = { version = "0.9", optional = true }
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -58,7 +59,7 @@ percent-encoding = { version = "2", optional = true }
 rand = "0.7"
 redis = { version = "0.17", optional = true }
 regex = "1"
-reqwest = { version = "0.9.11", optional = true }
+reqwest = { version = "0.10", optional = true, features = ["blocking", "json"] }
```
cgsheeh, line 62: Can we upgrade to `v0.11`? If not, why not go for `v0.10.10`?

glandium: Presumably because 0.11 is 5 days old. (v0.10.10 *is* what's being targeted here, see Cargo.lock)

Note, though, that the dist tests are busted.

tetsuharuohzeki: > Can we upgrade to v0.11? If not, why not go for v0.10.10?

Yes, but I hesitated to do it by reasons:

1. what @glandium said.
2. I'm not sure that it's nice to work, so I intend to avoid to include (at least) 3 versions Tokio in our dependency graph, v0.1 come from direct dependency, v0.2 come from tokio-compat, v1.0 comes from reqwest 0.11.....

-------

> Note, though, that the dist tests are busted.

Ugh, but I don't have any Linux environment to test failure cases...

-----

I'll address the conflict and push it once.

tetsuharuohzeki:
BTW, we can compile dish-server on linux only, and I also could not cross compile from macOS to Linux
It's a bit annoyed because I did not know dist-server could not run tests on non linux platform and I could not test basic compilations via `cargo check`...

Cargo.toml
cgsheeh: > Presumably because 0.11 is 5 days old. (v0.10.10 is what's being targeted here, see Cargo.lock)

Yes, indeed. My mistake. :)

mitchhentges:
Hey, I appreciate the patch, and I'm sorry about the delay here.
Since it's been a year (:flushed:) since this patch was up, we should probably embrace `reqwest==0.11` at this point.
I'm going to close this since this patch is specific to `0.10`, but I'll happily review a newer dependency-update patch.

Thanks again, and sorry we couldn't get this landed in time.

PR #940, opened 2021-01-28
Use timezone name in azure headers

johnor:
* Azure claims they follow rfc1123 here, which says implementations
  must accept both numeric and name for timezone:
  https://docs.microsoft.com/en-us/rest/api/storageservices/representation-of-date-time-values-in-headers
* Using a numeric timezone always returns a http status with 403 though,
  but calls work with a timezone name work.
* The examples in the above link use timezone names, so it should be
  safe to switch.


glandium:
Merged in c161213

yoavk:
@glandium Is there a scheduled upcoming release? I'm trying to use sccache with Azure Blob Storage and need this fix. 
I'm using it in CI, so ideally would like to download the pre-build binaries from the release (instead of building from source every time)

Thanks!

PR #942, opened 2021-02-03
[env] CARGO env can't be ignored.

rexhoffman:
Until it's last release trybuild embedded the CARGO env.   When we were switching our ci platform the location of CARGO changed.   This resulted in failed tests that could not find the the cargo executable.


glandium:
This feels like a band-aid, not addressing the core problem, which is that some of your compiled rust code uses `env!`. It happens that the environment variable associated with it was `CARGO`, but it would have been `FOO`, and the cache would have ignored it just as much. OTOH, the existing code is exactly the same kind of band-aid, so :shrug:.

glandium:
Merged in 33623e1.

PR #945, opened 2021-02-05
refactor: Convert server.rs to use async await

Marwes:
Reopening https://github.com/mozilla/sccache/pull/787 . I believe I fixed all nits in the review.

PR #946, opened 2021-02-05
futures migration to 0.3

drahnr:
This is mostly a PR to raise awareness that the migration to futures 0.3 is WIP, see https://github.com/paritytech/sccache/pull/31 for current status and remaining steps.

Now this is a rather big chunk of changes, partially non-atomic changes over multiple commits.

Ref #945


drahnr:
Superseded by #985

Issue #948, opened 2021-02-06
SCCACHE_START_SERVER=1 fails with "Too many commands specified"


Be-ing:
The documentation says for debugging to run:
```
SCCACHE_LOG=debug SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache
```
but this fails with:
```
[...]

PR #949, opened 2021-02-06
README: clarify how to debug on CI services

Be-ing:
I got confused by the current documentation because I tried to run cmake with `SCCACHE_START_SERVER=1`, which failed as described in #948.


Be-ing:
Thanks for merging!

Issue #941, opened 2021-02-03
Support caching cc invocations with -fprofile-use


michaelwoerister:
At the moment `sccache` does not try to cache C/C++ compiler invocations with the `-fprofile-use` flag. However, that would be rather useful for cases where profile data is cached by a build system (instead of regenerated for every build). 

It looks like implementing this shouldn't be too hard using the existing `ExtraHashFile` infrastructure. But one caveat there is that profile data files can become rather big (tens of megabytes are not uncommon). I suspect that re-hashing all that data for each compiler invocation is not very fast. 

Does `sccache` have some kind of mechanism to cache such file hashes (and only re-hash if the file in question has changed)?

cc @luser

luser:
> Does `sccache` have some kind of mechanism to cache such file hashes (and only re-hash if the file in question has changed)?

It doesn't at present, but #758 is open on the topic. Note that last year the hash algorithm used was switched to BLAKE3, so it should be pretty fast. sccache server logs contain timing information on how long it spends hashing files so it should be easy enough to run some before and after tests if you implement this functionality: https://github.com/mozilla/sccache/blob/b3f4727eda81c45b0632d036395af85429c4774b/src/util.rs#L141

michaelwoerister:
Yeah, BLAKE3 looks pretty fast. On the other hand compiling (for example) all of LLVM means invoking Clang 2000-3000 times, and for each invocation we would re-hash the same 20MB `.profdata` file. Although it might not actually take that long on a fast machine, each time `sccache` hashes ~40-60GB of instead of 20MB you'd not be wrong to expect a sad little tear drop sparkling in the corner of my eye. 

From the readme and #758 it sounds like sccache will always spawn a server process, right? That would be the perfect spot for keeping such a cache of file hashes.

Hopefully I'll have time to take a closer like at the whole setup next week.

PR #952, opened 2021-02-11
Support caching Clang invocations with -fprofile-use and -fprofile-instr-use.

michaelwoerister:
This PR adds support for caching Clang invocations with `-fprofile-use` and `-fprofile-instr-use`, something that will be required for supporting [PGOed Rust compiler builds](https://github.com/rust-lang/rust/issues/79562).

The PR *only* implements support for Clang. I have not had time to look into how GCC works (although from a superficial glance it looks like it might be harder to find out which files to hash exactly). Adding support for Rust should be straightforward though.

I added some unit tests and verified that things work locally, but found it hard to write end-to-end tests because of Clang's PGO setup which requires the external `llvm-profdata` tool to generate the `.profdata` file. Let me know if this is something that is required.

In https://github.com/mozilla/sccache/issues/941 I expressed the concern that hashing these rather big `.profdata` files would take a long time. In practice this does not seem to be true. In my local tests compiling LLVM with a full cache took about 50 seconds. Caching the file hashes (so that most of the hashing work could be skipped) only reduced the total build time by a couple of seconds or so.

r? @luser

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -122,6 +122,7 @@ ArgData! { pub
     Language(OsString),
     SplitDwarf,
     ProfileGenerate,
+    ClangProfileUse(PathBuf),
     TestCoverage,
     Coverage,
     ExtraHashFile(PathBuf),
@@ -168,6 +169,8 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     flag!("-c", DoCompilation),
     take_arg!("-fdiagnostics-color", OsString, Concatenated('='), DiagnosticsColor),
     flag!("-fno-diagnostics-color", NoDiagnosticsColorFlag),
+    flag!("-fno-profile-generate", TooHardFlag),
+    flag!("-fno-profile-use", TooHardFlag),
     flag!("-fno-working-directory", PreprocessorArgumentFlag),
     flag!("-fplugin=libcc1plugin", TooHardFlag),
     flag!("-fprofile-arcs", ProfileGenerate),
@@ -278,6 +281,9 @@ where
                     OsString::from(arg.flag_str().expect("Compilation flag expected"));
             }
             Some(ProfileGenerate) => profile_generate = true,
+            Some(ClangProfileUse(path)) => {
+                extra_hash_files.push(clang::resolve_profile_use_path(path, cwd));
+            }
             Some(TestCoverage) => outputs_gcno = true,
             Some(Coverage) => {
                 outputs_gcno = true;
@@ -337,6 +343,7 @@ where
         let args = match arg.get_data() {
             Some(SplitDwarf)
             | Some(ProfileGenerate)
+            | Some(ClangProfileUse(_))
             | Some(TestCoverage)
             | Some(Coverage)
             | Some(DiagnosticsColor(_))
@@ -380,6 +387,7 @@ where
         let args = match arg.get_data() {
             Some(SplitDwarf)
             | Some(ProfileGenerate)
+            | Some(ClangProfileUse(_))
             | Some(TestCoverage)
             | Some(Coverage)
             | Some(DoCompilation)
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -603,6 +603,12 @@ pub fn parse_arguments(
                     profile_generate = true;
                     &mut common_args
                 }
+
+                Some(ClangProfileUse(path)) => {
+                    extra_hash_files.push(clang::resolve_profile_use_path(path, cwd));
+                    &mut common_args
+                }
+
                 Some(ExtraHashFile(path)) => {
                     extra_hash_files.push(cwd.join(path));
                     &mut common_args
@@ -1061,6 +1067,7 @@ mod test {
             "-Xclang",
             "host_dictionary.obj",
             "-clang:-fprofile-generate",
+            "-clang:-fprofile-use=xyz.profdata",
             "dictionary.c"
         ];
         let ParsedArguments {
@@ -1068,6 +1075,7 @@ mod test {
             preprocessor_args,
             common_args,
             profile_generate,
+            extra_hash_files,
             ..
         } = match parse_arguments(args) {
             CompilerArguments::Ok(args) => args,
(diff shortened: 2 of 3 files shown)


src/compiler/gcc.rs
```diff
@@ -275,6 +283,11 @@ where
                     OsString::from(arg.flag_str().expect("Compilation flag expected"));
             }
             Some(ProfileGenerate) => profile_generate = true,
+            Some(ProfileUse(_)) => {
+                if compiler_kind != CCompilerKind::Clang {
+                    cannot_cache!(arg.flag_str().unwrap(), "only Clang supported".into())
```
luser, line 288: Just use a string literal here in `cannot_cache!`, since we use the strings as buckets to count non-cacheable compile requests and output them in the stats. You could use "-fprofile-use but not clang" or something like that.

src/compiler/gcc.rs
```diff
@@ -341,6 +354,20 @@ where
...
+
+                let mut path = cwd.join(path);
+
+                // Clang allows specifying a directory here, in which case it
+                // will look for the file `default.profdata` in that directory.
+                if path.is_dir() {
+                    path.push("default.profdata");
+                }
+
+                extra_hash_files.push(path);
```
luser, line 368: This feels like it belongs more in the `match` above? I'm not sure why `ExtraHashFile` has logic down here, this match exists just to split arguments up into ones that get passed to the preprocessor or not.

luser (approved):
This LGTM. Just two small things I would change.

michaelwoerister:
Thanks for the review, @luser! I think I addressed your comments.

@glandium, would you maybe be able to spare some time for merging/reviewing this PR? The new functionality is a necessary prerequisite for [distributing PGOed versions of rustc](https://github.com/rust-lang/rust/issues/79442).

src/compiler/gcc.rs
```diff
@@ -169,7 +176,7 @@ counted_array!(pub static ARGS: [ArgInfo<ArgData>; _] = [
     flag!("-fplugin=libcc1plugin", TooHardFlag),
     flag!("-fprofile-arcs", ProfileGenerate),
     flag!("-fprofile-generate", ProfileGenerate),
-    take_arg!("-fprofile-use", OsString, Concatenated, TooHard),
+    take_arg!("-fprofile-use", PathBuf, Concatenated('='), ProfileUse),
```
glandium, line 179: `-fprofile-use` can also be used without a value associated with it. It looks like that's true of `-fprofile-instr-use` too, and you don't handle those cases. There's also `-fno-profile-instr-use`, which you might as well add support for while you're here.

src/compiler/gcc.rs
```diff
@@ -275,6 +283,21 @@ where
                     OsString::from(arg.flag_str().expect("Compilation flag expected"));
             }
             Some(ProfileGenerate) => profile_generate = true,
+            Some(ProfileUse(path)) => {
+                if compiler_kind != CCompilerKind::Clang {
+                    cannot_cache!("-fprofile(-instr)-use caching only supported for Clang");
```
glandium, line 288: no need for the `(-instr)` part.

src/compiler/gcc.rs
```diff
@@ -275,6 +283,21 @@ where
                     OsString::from(arg.flag_str().expect("Compilation flag expected"));
             }
             Some(ProfileGenerate) => profile_generate = true,
+            Some(ProfileUse(path)) => {
+                if compiler_kind != CCompilerKind::Clang {
```
glandium, line 287: Another way you could do this, without adding the compiler kind, is to simply add -fprofile-use to the clang list of arguments (see `args::test_multi_search`), in which case it would probably be more appropriate to call this `ClangProfileUse` rather than `ProfileUse`.

src/compiler/msvc.rs
```diff
@@ -587,7 +587,7 @@ pub fn parse_arguments(
             let mut args = match arg.get_data() {
                 Some(SplitDwarf) | Some(TestCoverage) | Some(Coverage) | Some(DoCompilation)
                 | Some(Language(_)) | Some(Output(_)) | Some(TooHardFlag) | Some(XClang(_))
-                | Some(TooHard(_)) => cannot_cache!(arg
+                | Some(TooHard(_)) | Some(ProfileUse(_)) => cannot_cache!(arg
```
glandium, line 590: it shouldn't be too hard to add support for clang-cl here.

michaelwoerister:
Thanks for the review, @glandium!

> Can you also rebase on top of #971 after it's merged? Thank you. 

Will do.

> (BTW, did you intend your user email not to contain a TLD?)

That was a halfhearted (and probably ineffective) attempt to keep my email address somewhat protected from spammers. It set up git like that years ago and it seems to have stuck `:/`

src/compiler/gcc.rs
michaelwoerister: Ah, you can "override" arguments that way? That does sound like the better approach for this case.

src/compiler/gcc.rs
michaelwoerister: It looks like `-fprofile-use` without argument will try to read from `./default.profdata` -- and that the implementation here already handles this correctly. I'll add a unit test for that.

michaelwoerister:
OK, I think the version I just pushed addresses your comments, @glandium. And it does look cleaner than the previous version `:)`

Note: I added "support" for the `-fno-profile-*` flags by explicitly marking them as `TooHardFlag`. I can look into making them actually cacheable but I'll only do so if you think it's worth the trouble. 

I was a bit surprised to find that `GCC::parse_arguments()` seems to treat unknown arguments as always cacheable. Is that on purpose? It seems like a somewhat dangerous default.

michaelwoerister:
@glandium, do you have an estimate on when this can be merged? Is there anything I can do to help?

dancing-leaves:
Hi, would love to see these changes upstreamed, tracking this PR for a while. Thanks for the work, and lets rebase and merge it already. :slightly_smiling_face:

FilipAndersson245:
@glandium Any idea of when/if this pull request can be merged?

mitchhentges (approved):
Thanks Michael, and sorry about the review delay.
Let's get this landed :sunglasses:

michaelwoerister:
Well, that was quick :) Thanks for the review!

Issue #758, opened 2020-05-27
consider caching file hashes in the sccache server


froydnj:
...so that, particularly for Rust compilations, we don't waste a bunch of time re-hashing the same files over and over again.

I don't know how much this really helps, because it probably hurts a little bit (?) on single-shot builds where the server doesn't stay up very long (e.g. Firefox automation builds)...though maybe not having to touch the disk (or the kernel's file cache, or whatever) is a win overall.  Also unsure how easy it is to arrange things so you don't wind up badly serializing the whole process on your hash cache; maybe the locking overhead would really not be that large.

luser:
From a cursory search there appear to be several concurrent hashtable crates out there, one of those might be suitable. Given that this is a cache there might be some special-purpose data structure that would work better, I don't know. Would you need to cache by (filename, mtime) for this to be correct? I assume your concern is mostly the time spent hashing rlib / rmeta files, right? The existing cargo fingerprinting there probably makes that less of a concern.

A data structure that allowed lock-free reads ought to keep the fast path fast, and given that the file hashing code is already async, the write path could be something like:
1. Attempt to insert a "pending" entry (I'd probably use the read end of a channel).
2a. If a pending entry already exists (some other thread racing), `await` it.
2b. If not, insert the pending entry and kick off the hash calculation.
3. When hash calculation finishes, swap out the pending entry for an actual calculated hash entry, and resolve the pending entry so anyone `await`ing it is unblocked.

Honestly if you get updated to the latest `tokio` and whatnot you could likely define a better threading model for sccache where most of the server code always runs on the main thread (you've already got a thread pool for the CPU-intensive stuff) and this could just be a standard `HashMap` that the server owns.

the8472:
> Would you need to cache by (filename, mtime) for this to be correct?

It can't be entirely correct since there are ways to modify files without updating mtime. But yeah, on windows that's probably a decent heuristic. On unix systems `(dev, ino, mtime, size)` could be used instead as a more compact alternative.

PR #953, opened 2021-02-11
Cache file hashes

michaelwoerister:
This PR adds a cache for file hashes in the server, as suggested in https://github.com/mozilla/sccache/issues/758. 

The implementation is kind of simplistic (maybe even too much so):

- It does not try to avoid races between multiple threads hashing the same file. If one thread is already hashing file `x` and another thread wants to do the same, it will just go ahead an calculate the hash by itself. Both threads will write the hash to the cache (in the order they finish). This should be a benign race since hashes are always associated with their timestamps. In the worst case, a long-running hash computation overwrites a new entry with an outdated hash. In that case the next request for the file hash would re-do the hashing again.

- The cache is simply kept in a local static variable. Keeping it somewhere else requires lots of additional parameter passing in many places. I'm undecided on whether I think this ugly or an example of good encapsulation `:)`

In my local testing this kind of caching did not make a lot of difference for C/C++ builds. Those seem to be completely dominated by the C preprocessor. It did not get a chance to benchmark this with Rust code yet.

Feedback welcome!

Closes https://github.com/mozilla/sccache/issues/758.

cc @froydnj @luser

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -59,11 +59,20 @@ impl Digest {
 
     /// Calculate the BLAKE3 digest of the contents of `path`, running
     /// the actual hash computation on a background thread in `pool`.
-    pub fn file<T>(path: T, pool: &ThreadPool) -> SFuture<String>
+    /// The `Duration` in the result is the time it took to compute the hash.
+    pub fn file<T>(path: T, pool: &ThreadPool) -> SFuture<(String, Duration)>
     where
         T: AsRef<Path>,
     {
-        Self::reader(path.as_ref().to_owned(), pool)
+        let path = path.as_ref().to_owned();
+
+        Box::new(pool.spawn_fn(move || -> Result<_> {
+            let start = time::Instant::now();
+            let reader = File::open(&path)
+                .with_context(|| format!("Failed to open file for hashing: {}", path.display()))?;
+            let hash = Digest::reader_sync(reader)?;
+            Ok((hash, start.elapsed()))
+        }))
     }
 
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -107,6 +128,103 @@ impl Default for Digest {
...
+
+#[derive(Clone)]
+struct CachedFileHash {
+    timestamp: FileTimeStamp,
+    hash: Arc<String>,
+}
+
+#[derive(Default)]
+struct FileHashCache {
+    hashes: Arc<Mutex<HashMap<PathBuf, CachedFileHash>>>,
```
glandium, line 189: Without looking too deeply into this, I feel like this shouldn't be free to grow indefinitely, and should be a LRU cache of some sort.

glandium (changes requested):
I haven't looked deeply in the details yet. Intuitively, I'd guess the reduction in how much hashing happens compensates for the extra syscalls to get the file metadata, even on Windows where I/O tends to be awful, but I'd like to benchmark this somehow before considering for a merge. I'm also not sure whether I'd queue that for 0.2 or 0.3...

michaelwoerister:
Good point -- I did not benchmark this on Windows.

sylvestre:
it needs to be rebased too :)
(hello Michael ;)

michaelwoerister:
Hi Sylvestre :)

I probably won't have time to work on this in the next couple of months. If anybody else is interested in this, feel free to take over.

PR #959, opened 2021-02-18
Allow caching gcc calls if same -arch is passed twice

ahoppen:
https://github.com/mozilla/sccache/pull/898 prevents `sccache` from caching `gcc`/`clang` calls that have two `-arch` arguments. However, this also prevents caching of calls that just specify the same arch twice (like `-arch x86_64 -arch 86_64`). In particular this prevents `sccache` from caching any compilations when building the [Swift compiler](https://github.com/apple/swift). 

This PR changes the logic to allow caching of `gcc`/`clang` calls with multiple `-arch` arguments if their value is the same.

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -236,7 +236,7 @@ where
     let mut outputs_gcno = false;
     let mut xclangs: Vec<OsString> = vec![];
     let mut color_mode = ColorMode::Auto;
-    let mut seen_arch = false;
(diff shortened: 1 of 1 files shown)


ahoppen:
Is there a chance to get this reviewed? The CI failures seem unrelated to my changes.

CC @indygreg @glandium

glandium:
Merged a slightly simpler version in 7f8d643.

PR #963, opened 2021-02-26
README: document that sccache is incompatible with /Zi with MSVC

Be-ing:


README.md
```diff
@@ -108,6 +108,21 @@ To use sccache with cmake, provide the following command line arguments to cmake
 -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
 ```
 
+To generate PDB files for debugging with MSVC, only the [`/Z7` option](https://docs.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-160) works with sccache; the PDB files generated with the `/Zi` and `/ZI` options cannot be cached. Note that CMake sets `/Zi` by default, so if you use CMake, you will need something like this in your CMakeLists.txt:
```
glandium, line 111: That is not entirely true. `/Zi` is supported if you also use `/Fd` with a different pdb name for each object file you create.

Be-ing: Thank you for correcting that. I'll update the documentation.

PR #968, opened 2021-03-02
Add reqwest to azure feature

### LINKED ISSUE #967: sccache doesn't compile just with the azure feature
Simply the following command doesn't work:
`cargo build --no-default-features --features=azure`

It needs `reqwest` as a dependency.

omid:
Fixes #967


Xanewok:
@glandium this is relatively straightfoward patch and just adds a missing, required `reqwest` dep when compiling only with the `azure` feature. Would it be possible to merge that?

PR #971, opened 2021-03-10
Fix clippy warnings with new version of clippy

glandium:
https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_constants
https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps

Partially based upon patch from Michael Woerister <michaelwoerister@posteo> in #952.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -47,6 +47,7 @@ fn get_file_mode(file: &fs::File) -> Result<Option<u32>> {
 }
 
 #[cfg(windows)]
+#[allow(clippy::unnecessary_wraps)]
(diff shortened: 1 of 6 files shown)


src/cache/cache.rs
```diff
@@ -47,6 +47,7 @@ fn get_file_mode(file: &fs::File) -> Result<Option<u32>> {
 }
 
 #[cfg(windows)]
+#[allow(clippy::unnecessary_wraps)]
```
mitchhentges, line 50: Smells like an `impl Trait` is needed around here, but that's unrelated to this patch.

mitchhentges (approved):

PR #976, opened 2021-03-19
Enable TLS support for Redis

alnr:
This enables TLS for Redis backends. Tested against an AWS ElastiCache Redis with TLS. I'm a Rust newbie so I don't know if the dependency changes (in `Cargo.lock`) are OK.


Xanewok:
I don't think these lockfile changes are good since they pull Tokio 1.0 in addition to 0.1 used now. I think it's best to defer that until the futures migration is done (se #946); from there migrating to Tokio 0.3 or 1.0 should be somewhat easier. Also see https://github.com/mozilla/sccache/commit/10b072d57eaacf45e1bd84463460952ed290732b for more context

alnr:
@Xanewok Could you take another look? This time, I haven't upgraded the redis-rs version at all, just fiddled with the features. Turns out several default features are not used and were bloating `Cargo.lock`.

I've also added a small improvement (masking the password in the Redis URL if there is one).

src/cache/redis.rs
```diff
@@ -32,8 +33,12 @@ pub struct RedisCache {
 impl RedisCache {
     /// Create a new `RedisCache`.
     pub fn new(url: &str) -> Result<RedisCache> {
+        let mut parsed = Url::parse(url)?;
+        if parsed.password().is_some() {
```
sylvestre, line 37: Could you please add a comment to explain that it is for the TLS support? thanks

alnr: This change is unrelated to TLS. Still, I updated the diff to match https://github.com/paritytech/cachepot/pull/98.

alnr:
Hi @sylvestre 
I rebased; pipeline is green :)

PR #980, opened 2021-03-31
MSVC: add /Brepro to the list of known args

aguinet:
`/Brepro` is a hidden flag of `cl.exe` that can be used to generate reproducible binaries. More information can be found here: https://blog.conan.io/2019/09/02/Deterministic-builds-with-C-C++.html .

sccache currently does not understand this flag, and thus fails to cache builds that use them. One noticable project that "forces" the usage of `/Brepro` is the LLVM project: https://github.com/llvm/llvm-project/blob/9d474be11d71b5ae13490d3d8bd66150765560e3/llvm/cmake/modules/HandleLLVMOptions.cmake#L489-L516

It should be used only as a linker flag, but MS cl.exe does understand this flag, and pass it to the linker if calling it. I think sccache should just be able to support this, in case some projects use it like LLVM.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -258,6 +258,7 @@ macro_rules! msvc_args {
 // https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=vs-2019
 msvc_args!(static ARGS: [ArgInfo<ArgData>; _] = [
     msvc_flag!("?", SuppressCompilation),
+    msvc_flag!("Brepro", PassThrough),
     msvc_flag!("C", PassThrough), // Ignored unless a preprocess-only flag is specified.
     msvc_take_arg!("D", OsString, CanBeSeparated, PreprocessorArgument),
     msvc_flag!("E", SuppressCompilation),


jviotti:
Is there any reason why this can't be merged? Is sccache relying on the mtime in any other way?

zturnerbx:
It's definitely not true that it should be used only as a linker flag.  For example, /Brepro causes things like `__DATE__`, `__TIME__`, and a couple of other builtin pre-processor macros to evaluate to a constant.  I believe it also effects debug info generation in object files (especially when /Z7 is used, but perhaps other times as well).

Either way, I would also like to see this merged.

aguinet:
About testing: I'll add one. Should I put it in `tests/system.rs`?

yuanfang-chen:
There is a `test_parse_arguments_passthrough` unit test in the same file. I guess it should be good enough?

mitchhentges (approved):
Hey, thanks for this patch!
We already have some tests for `PassThrough` flags, so I'm ok with landing this as-is - I don't think the tradeoff for you having to update a year-old patch at _this_ point is worth it :)

aguinet:
Thanks a lot @mitchhentges !

PR #981, opened 2021-04-02
Fix warnings with latest Rust stable

Xanewok:
Out of the two warnings fixed, the latter is more interesting as it fixes a warning that's going to be phased out in Rust 2021 (calling `panic` with something else than a format string literal).


sylvestre:
Sorry but it is probably fixed by https://github.com/mozilla/sccache/pull/1058

PR #983, opened 2021-04-06
Make integration tests more environment-agnostic

Xanewok:
Discovered when we were running unit tests in our CI environment with `CARGO_INCREMENTAL` set to `0` (https://github.com/paritytech/sccache/issues/23).

It seems this relies on the incremental compilation being enabled by default, so this change explicitly disables it to make the test more robust/independent of the environment. The test crate compiled is effectively empty so incremental comp. should be negligible in cost (or even cheaper - less overhead to setup).

### DIFF
--- a/tests/harness/mod.rs
+++ b/tests/harness/mod.rs
@@ -2,7 +2,6 @@
 use sccache::config::HTTPUrl;
 use sccache::dist::{self, SchedulerStatusResult, ServerId};
 use sccache::server::ServerInfo;
-#[cfg(feature = "dist-server")]
 use std::env;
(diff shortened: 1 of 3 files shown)


Xanewok:
This also adds another commit (https://github.com/mozilla/sccache/pull/983/commits/06bb2baee4ce1b799c0bfa833444345157a0ec2c) which should help isolate the tests when running with other sccache-related environment set (e.g. `SCCACHE_REDIS` like we had in https://github.com/paritytech/sccache/pull/42)

drahnr (approved):

PR #987, opened 2021-04-13
Add SCCACHE_C_CUSTOM_CACHE_BUSTER to CACHED_ENV_VARS

t184256:
I needed a way to make hashes CWD-dependent, and it feels ugly to use other variables names for that purpose.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -638,6 +638,11 @@ pub const CACHE_VERSION: &[u8] = b"10";
 lazy_static! {
     /// Environment variables that are factored into the cache key.
     static ref CACHED_ENV_VARS: HashSet<&'static OsStr> = [
+        // SCCACHE_C_CUSTOM_CACHE_BUSTER has no particular meaning behind it,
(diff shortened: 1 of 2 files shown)


src/compiler/c.rs
```diff
@@ -643,6 +643,7 @@ pub const CACHE_VERSION: &[u8] = b"10";
 lazy_static! {
     /// Environment variables that are factored into the cache key.
     static ref CACHED_ENV_VARS: HashSet<&'static OsStr> = [
+        "SCCACHE_C_CUSTOM_CACHE_BUSTER",
```
sylvestre, line 646: Could you please add a comment in the code explaining the why

t184256:
Added a comment as requested + rebased.

sylvestre:
thinking about this a bit more, it should be also documented in https://github.com/mozilla/sccache/blob/master/README.md 
sorry

t184256:
No problem, I tried to add a section for it and the two already existing variables.

PR #990, opened 2021-04-26
Add support for NVCC ccbin argument

kerrysnyder:
The "[ccbin](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#file-and-path-specifications-compiler-bindir)" argument is used to control the "host" compiler that is used by NVCC, which is useful because NVCC only supports certain versions of the host compiler (collected [here](https://gist.github.com/ax3l/9489132#nvcc)). This argument is used by CMake when the `CMAKE_CUDA_HOST_COMPILER` [parameter](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_HOST_COMPILER.html) is set (or alternatively the `CUDAHOSTCXX` environment variable)

### DIFF
--- a/src/compiler/nvcc.rs
+++ b/src/compiler/nvcc.rs
@@ -184,6 +184,7 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-Xnvlink", OsString, CanBeSeparated('='), PassThrough),
     take_arg!("-Xptxas", OsString, CanBeSeparated('='), PassThrough),
     take_arg!("-arch", OsString, CanBeSeparated('='), PassThrough),
+    take_arg!("-ccbin", OsString, CanBeSeparated('='), PassThrough),
     take_arg!("-code", OsString, CanBeSeparated('='), PassThrough),
     flag!("-dc", DoCompilation),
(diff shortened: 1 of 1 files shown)


src/compiler/nvcc.rs
```diff
@@ -118,6 +118,7 @@ impl CCompilerImpl for NVCC {
         //NVCC only supports `-E` when it comes after preprocessor
         //and common flags.
         cmd.arg("-E")
+            .arg("-D__FILE__=\"empty\"")
```
mitchhentges, line 121: I don't quite follow, why do we need to "get rid of line info from nvcc"?
I'm _guessing_ that it's another potential improvement with its own merits - perhaps it should be split into a separate patch?

mitchhentges (changes requested):
The `ccbin` change makes sense to me, good catch.
I don't understand the other line change, but otherwise this looks good.
Thanks for the patch :)

mitchhentges:
Hmm, looks like this got closed due to the rename of `master` to `main` (unsure yet why some PRs were affected, and others weren't).
This still looks like a beneficial change, so if you're able to rebase off of `main` and re-push, I'll happily review and land this.

Issue #935, opened 2021-01-17
RUSTSEC-2020-0054: directories: directories is unmaintained, use directories-next instead


brightly-salty:
When `cargo-audit` is run on anything that depends directly or indirectly on sccache, the following is output (for more info see https://rustsec.org/advisories/RUSTSEC-2020-0054):

```
Version:       3.0.1
Warning:       unmaintained
Title:         directories is unmaintained, use directories-next instead
Date:          2020-10-16
ID:            RUSTSEC-2020-0054
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0054
Dependency tree:
directories 3.0.1
└── sccache 0.2.15
```

It can be easily remedied by replacing directories with directories-next.

I would be willing to attempt to implement this and submit a PR if it would be accepted.

glandium:
This RUSTSEC advisory seems a little disingenuous, considering a month earlier, the directories-next repository got a change that replaced its comment about the status of  the directories crate from "unmaintained" to "once-abandoned". https://github.com/xdg-rs/dirs/commit/0a73f5f6ce919a24623d37591dc196fbc0c81638

sfc-gh-anoyes:
It looks like there's even more `cargo-audit` output now:

```
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 263 security advisories (from /home/anoyes/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (382 crate dependencies)
Crate:         hyper
Version:       0.12.35
Title:         Multiple Transfer-Encoding headers misinterprets request payload
Date:          2021-02-05
ID:            RUSTSEC-2021-0020
URL:           https://rustsec.org/advisories/RUSTSEC-2021-0020
Solution:      Upgrade to >=0.14.3 OR >=0.13.10, <0.14.0
Dependency tree: 
hyper 0.12.35
├── sccache 0.2.16-alpha.0
├── reqwest 0.9.24
│   ├── selenium-rs 0.1.2
│   │   └── sccache 0.2.16-alpha.0
│   └── sccache 0.2.16-alpha.0
└── hyper-tls 0.3.2
    └── reqwest 0.9.24

Crate:         smallvec
Version:       0.6.13
Title:         Buffer overflow in SmallVec::insert_many
Date:          2021-01-08
ID:            RUSTSEC-2021-0003
URL:           https://rustsec.org/advisories/RUSTSEC-2021-0003
[...]

PR #991, opened 2021-04-27
Update Cargo.lock to address security advisories

sfc-gh-anoyes:
I'm not very familiar with rust so many I haven't done this correctly, but here's my attempt.

Related: #935

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1175,9 +1175,9 @@ checksum = "3c1ad908cc71012b7bea4d0c53ba96a8cba9962f048fa68d143376143d863b7a"
 
 [[package]]
 name = "hyper"
-version = "0.12.35"
+version = "0.12.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9dbe6ed1438e1f8ad955a4701e9a944938e9519f6888d12d8558b645e247d5f6"
+checksum = "5c843caf6296fc1f93444735205af9ed4e109a539005abb2564ae1d6fad34c52"
 dependencies = [
  "bytes 0.4.12",
  "futures 0.1.30",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -66,6 +66,7 @@ sha2 = { version = "0.9", optional = true }
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
+smallvec = ">=0.6.14"
```
Xanewok, line 69: I think that's fine because by default (without using explicit caret or tilde or similar semver requirements or ranges) this falls back to patch-compatible versions and does not pull a separate 1.x version.
```suggestion
smallvec = "0.6.14" 
```

Could you add a comment saying that this dependency here is only to force a version requirement across the entire dependency graph but is not directly used by sccache? In case all other transitive dependencies upgrade to 1.x, so we know that it's safe to remove this. Thanks!

sfc-gh-anoyes:
It might be the case that all transitive dependencies already updated (at least enough to fix the security advisories) . If I remove Cargo.lock on master, then run `cargo build` and `cargo audit`, it now only gives some "deprecated/unmaintained" warnings.

sfc-gh-anoyes:
Would you accept a PR that simply rebuilds Cargo.lock instead? If it makes more sense for you to make this change that's also fine :)

Xanewok:
Hm, updating the entire lockfile seems a bit aggressive. To only address the smallvec advisory, could you try resetting the change and doing only `cargo update -p smallvec:0.6.13` (or whichever version we currently have resolved)?

> Would you accept a PR that simply rebuilds Cargo.lock instead? If it makes more sense for you to make this change that's also fine :)

I'm not a maintainer, just doing a drive-by review =)

sfc-gh-anoyes:
@Xanewok I think I have the minimum set of updates that addresses all vulnerabilities reported by cargo audit

sfc-gh-anoyes:
@glandium I looked at the git history and it looks like you might be a maintainer? Would you mind taking a look please?

sylvestre:
thanks!

PR #998, opened 2021-05-05
Update blobstore.rs

veniqa:


Xanewok:
This doesn't look correct - it adds an extra leading `}` brace and removes the `x-ms-version` request header.

What's the reasoning behind this change?

sylvestre:
no update for a while, closing

PR #1002, opened 2021-05-12
Use deserializer to parse size in human readable format from fileconfig

sum12:
This patch allows the disk cache size to be specified in the similar
format as SCCACHE_CACHE_SIZE. Human readable format with suffix K/M/G/T.

To maintain backwards compatibility a visit_u64 is implemented as a
transparent fn.  The patch also addresses the requested TODO

Added tests for the corresponding part


sum12:
The Failure seems unreleated to the change.

Xanewok:
It seems like CI base images for dist tests need to be updated, see https://github.com/paritytech/sccache/pull/74

src/config.rs
```diff
@@ -145,12 +146,43 @@ impl HTTPUrl {
...
+where
+    D: de::Deserializer<'de>,
+{
+    struct Visitor;
+
+    impl<'de> de::Visitor<'de> for Visitor {
+        type Value = u64;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("string disk size with suffix K/M/G/T or integer for Bytes ")
```
mitchhentges, line 159: I'm not sufficiently familiar with `serde`, so just to confirm: is the extra space at the end of this string deliberate?

mitchhentges (approved):
Looks good, thanks!
Can you rebase this against `main` so that we can have modern CI run against this change?

src/config.rs
sum12: thanks, fixed it

mitchhentges:
Heads up that there's some CI failures that need to be resolved here first before I can land this.

drahnr:
Gentle ping.

drahnr (approved):

sylvestre:

PR #1010, opened 2021-06-10
Fix compiler warning

theidexisted:


sylvestre:
Sorry but it is probably fixed by https://github.com/mozilla/sccache/pull/1058

Issue #1017, opened 2021-07-07
CUDA source files cannot be cached


zhouwei25:
The `-P` option causes all of the pre-processed content to be written to the file (please refer to
https://docs.microsoft.com/en-gb/cpp/build/reference/p-preprocess-to-a-file?view=msvc-160). The stdout content is 0, and the hash key is generated according to the contents of the stdout. Therefore, the hash key of any CUDA source file is the same, which makes it can only cache one CUDA file. In order to make the contents of the CUDA source file preprocessed output to STDOUT, the preprocess option should  be can ``-P``, then different hash values will be generated, haven't you detected this bug？

https://github.com/mozilla/sccache/blob/master/src/compiler/nvcc.rs#L121

```
[0m[38;5;8m[[0m2021-07-07T08:59:45Z [0m[36mTRACE[0m sccache::compiler::c[0m[38;5;8m][0m [math_function.cu.obj]: Preprocessor output is 0 bytes
[0m[38;5;8m[[0m2021-07-07T08:59:45Z [0m[36mTRACE[0m sccache::util[0m[38;5;8m][0m Hashed 0 files in 8.913 s
[0m[38;5;8m[[0m2021-07-07T08:59:45Z [0m[34mDEBUG[0m sccache::compiler::compiler[0m[38;5;8m][0m [math_function.cu.obj]: generate_hash_key took 8.913 s
[0m[38;5;8m[[0m2021-07-07T08:59:45Z [0m[36mTRACE[0m sccache::compiler::compiler[0m[38;5;8m][0m [math_function.cu.obj]: Hash key: 7e5027a5bb3d395b95819336d7def508c0c4125d11fc8bc862c15daafd3a0cc1
[0m[38;5;8m[[0m2021-07-07T08:59:45Z [0m[36mTRACE[0m sccache::cache::disk[0m[38;5;8m][0m DiskCache::get(7e5027a5bb3d395b95819336d7def508c0c4125d11fc8bc862c15daafd3a0cc1)
[0m[38;5;8m[[0m2021-07-07T08:59:45Z [0m[36mTRACE[0m tokio_reactor[0m[38;5;8m][0m loop process - 0 events, 0.000s
```

sylvestre:
it has been implemented now

PR #1018, opened 2021-07-07
fix NVCC hash key generate error on windows

zhouwei25:
Resolve https://github.com/mozilla/sccache/issues/1017
===

The `-P` option causes all of the pre-processed content to be written to the file (please refer to
https://docs.microsoft.com/en-gb/cpp/build/reference/p-preprocess-to-a-file?view=msvc-160). The STDOUT content is 0, and the hash key is generated according to the contents of the STDOUT. Therefore, the hash key of any CUDA source file is the same, which makes it can only cache one CUDA file. In order to make the contents of the CUDA source file preprocessed output to STDOUT, the preprocess option should  be  ``-EP``, then different hash values will be generated, have you detected this bug？

https://github.com/mozilla/sccache/blob/master/src/compiler/nvcc.rs#L121

```
[0m [math_function.cu.obj]: Preprocessor output is 0 bytes
[0m Hashed 0 files in 8.913 s
[0m [math_function.cu.obj]: generate_hash_key took 8.913 s
[0m [math_function.cu.obj]: Hash key: 7e5027a5bb3d395b95819336d7def508c0c4125d11fc8bc862c15daafd3a0cc1
[0m DiskCache::get(7e5027a5bb3d395b95819336d7def508c0c4125d11fc8bc862c15daafd3a0cc1)
[0m loop process - 0 events, 0.000s
```

All hash key is ``7e5027a5bb3d395b95819336d7def508c0c4125d11fc8bc862c15daafd3a0cc1``, no matter which CUDA source files!!!


ZyhHelen (approved):

src/compiler/nvcc.rs
```diff
@@ -117,9 +117,13 @@ impl CCompilerImpl for NVCC {
 
         //NVCC only supports `-E` when it comes after preprocessor
         //and common flags.
-        cmd.arg("-E")
-            .arg("-Xcompiler=-P")
-            .env_clear()
+        cmd.arg("-E");
+        if cfg!(windows) {
+            cmd.arg("-Xcompiler=-EP");
```
sylvestre, line 122: Could you please add a comment to explain what you are doing?

zhouwei25: If windows, `-Xcompiler=-P` will send precompiled content to a file, but hash key will generatate from stdout, so that hash key is wrong and remain the same. `-Xcompiler=-E` or  `-Xcompiler=-EP` will send precompiled content to stdout, it's right.

sylvestre: Sorry, i meant in the code :)

mitchhentges: Hey, thanks for this patch, your explanation in #1017 makes a lot of sense, good investigation :+1: 
I have one question:

> To suppress the generation of #line directives, use [/EP (Preprocess to stdout Without #line Directives)](https://docs.microsoft.com/en-gb/cpp/build/reference/ep-preprocess-to-stdout-without-hash-line-directives?view=msvc-160) as well as /P.

Should we be providing both `-EP` and `-P`? Otherwise, will we get `stdout`, but `file.cpp.i` won't be created, right? Unsure if I'm understanding this correctly.

I agree with Sylvestre though: let's add a little comment in the code to explain what's going on:
```suggestion
            // The hash key is generated from stdout, so do "-EP" to print the output.
            cmd.arg("-Xcompiler=-EP");
```

mitchhentges (changes requested):

zhouwei25:
`-EP` and `-E` preprocess to stdout, `-P` preprocess to file, and no cache will hit. So, it can only use `-EP` or `-E`

sylvestre:

Issue #413, opened 2019-04-13
sccache cant start on windows10 20h1 build 18875


gettf:
sccache --start-server
**Starting sccache server...
error: Server startup failed: an attempt was made to access a socket in a way forbidden by its access permissions (os error 10013)**

sccache -s
**error: Connection to server timed out**

aminya:
I was having this issue. I had installed sccache from chocolatey. Updating sccache using Cargo fixes the issue

```
cargo install sccache --git https://github.com/mozilla/sccache 
```

PR #1027, opened 2021-08-18
Address Windows port binding errors

emabrey:
On Windows 10 running either WSL or Hyper-V, the operating system will acquire random and ephemeral binds of various port ranges. Once these port ranges are allocated for the usage of these two OS features, they are no longer available for usage by other processes, even though at a given point in time they may technically not be actively bound. It just so happens that on these machines, the default port of sccache (TCP 4226) conflicts with the default excluded port ranges for some configurations of winNAT and Hyper-V. The ideal way to resolve this would be to tell the user to modify the operating system to pick it's TCP ports from much higher in the TCP port space (such as the dynamic TCP port range). However, this is much more complicated than simply manually instructing sccache to utilize a different port than the default, not to mention the user is more likely to blame sccache for the failure than the OS regardless of who is actually in the wrong. For information on how the Docker project handled this same problem, see docker/for-win#3171.

Within sccache, to address this issue the proper method would be usage of the environmental variable `SCCACHE_SERVER_PORT`, however, the documentation does not go into detail on the availability of this configurable. To make users aware of how to fix this issue, I have added a warning message that instructs the user on how to change the default sccache port via the variable. Additionally, instead of hoping that the port we are using will be valid when sccache attempts to bind it, now startup of the sccache executable verifies that the port we will use, default or manually configured, is valid. This is much better than the previous behavior of letting the operating system give the user cryptic and essentially unhelpful error messages (for example OS Error 10013, which I experienced myself and which other people have also experienced e.g. #413).

This fix was implemented using the addition of a new function which runs between log initialization and the handoff to the main commands dispatch logic. This function verifies that either the `commands::DEFAULT_PORT` or the `SCCACHE_SERVER_PORT` values are valid and bindable ports, but only if we are running a `--start-server` command, since checking for the bind on other options would be incorrect, especially in the `--stop-server` scenario. To enable this I needed to make a few modifications to the `cmdline::StatsFormat` and `cmdline::Commands` enums, specifically I needed to add the `PartialEq` trait so that the port check code would be able to tell what the command line args were without resorting to a match statement. If the check fails the `sccache::main` function will quickly exit using `std::process::exit(1)` to indicate an error occurred. The check function can optionally emit the aforementioned error messages to stderr, but in the `sccache::main` usecase it **must** emit the error messages to comply with the unit tests' expectation that a failed `--start-server` will emit the text `Server startup failed:`.

The logic for evaluating if a port is "valid" utilizes a new dependency on `getport-4.0.0` to provide the ability to test the validity of the port binds without resorting to ugly command line parsing.

Resolves mozilla/sccache#413

#### `SCCACHE_SERVER_PORT` set but not valid
image

#### `SCCACHE_SERVER_PORT` not set (using invalid default port)
image

#### `SCCACHE_SERVER_PORT` set using valid value
image


src/lib.rs
```diff
@@ -66,6 +68,11 @@ const LOGGING_ENV: &str = "SCCACHE_LOG";
 
 pub fn main() {
     init_logging();
+
+    if !check_valid_port_available(true) {
+        std::process::exit(1);
+    }
```
mitchhentges, line 74: Instead of checking if the port is available at the very beginning of the application, can we instead catch error (probably the `10013` one) when we attempt to allocate the port? At that point, we can recommend setting/changing `SCCACHE_SERVER_PORT`.

The downsides with the current approach are:
* You need to infer from the args whether or not a port will be allocated here, which is unfortunate. We're parsing args twice now :(
* There's a potential race condition (what if the port is available here, but then something else steals it before `sccache` actually does its real port allocation?
* We can probably avoid the `PartialEq` implementations, perhaps.
* We'll definitely be able to remove `get-port`, I believe.

src/lib.rs
```diff
@@ -97,3 +104,67 @@ fn init_logging() {
...
+
+    // If neither are valid for binding a new server
+    if !env_port_valid && !default_port_valid {
+        let cmdline_command = cmdline::parse();
+
+        if cmdline_command.is_ok() {
+            let cmd = cmdline_command.unwrap();
+            
+            // and we are running a --start-server 
+            if cmd == cmdline::Command::StartServer {
```
mitchhentges, line 136: FWIW, it's possible to do this without `PartialEq` and also without a match by doing:
```suggestion
            if let cmdline::Command::StartServer = cmd { ... }
```
I believe :)

src/lib.rs
```diff
@@ -97,3 +104,67 @@ fn init_logging() {
...
+                if print_error_message {
+                    let mut chosen_port = DEFAULT_PORT;
+                    let mut user_output_port_term = "default port";
+                    let mut user_output_verb = "setting";
+
+                    if env_var_raw.is_some() {
+                        chosen_port = env_port;
+                        user_output_port_term = "user-provided port";
+                        user_output_verb = "resetting"
```
mitchhentges, line 145: Nit: Let's use the term "changing" instead here, because "resetting" is slightly ambiguous between the literal "re-set: set it again" meaning and the "reset: remove any custom changes" interpretation. We want the first meaning, and "changing" may be more clear.

mitchhentges (changes requested):
This is a great usability improvement, thanks for the patch!
I really think that tests will be valuable here, though it may require some fancy mocking/injection/abstraction work, unsure.
Let me know how viable you think testing this will be.

emabrey:
Planning on looking into unit tests/suggested changes today hopefully.

emabrey:
@mitchhentges 
@sylvestre 

I renamed my default branch and didn't realize that would auto-close this PR. I should prolly make a copy/paste PR using the new branch. I will at you both once the changes are ready and make a new PR.

mitchhentges:
No worries, and I'll be notified automatically when you create the new PR, so the `@` is redundant :)
(heads up that I'll be away next week, so a review will likely stroll by the week afterwards).

Issue #1009, opened 2021-05-27
(Again) no cache hits with cl.exe, cmake and ninja


bilke:
All misses are due to `multiple input files`. In the following is an example, can you help me find the wrong compilation flag?

```json
parse_arguments: CannotCache(multiple input files): [
    "/nologo",
    "-DNDEBUG",
    "-D_BIND_TO_CURRENT_VCLIBS_VERSION=1",
    "-D_CONSOLE",
    "-D_CRT_SECURE_NO_WARNINGS",
    "-IC:\\cpm\\hdf5\\b29b430dd2bdc4f489cfe8f50785e974d9afda4c\\src",
    "-I_deps\\hdf5-build",
    "/DWIN32",
    "/D_WINDOWS",
    "-openmp",
    "/MD",
    "/O2",
    "/Ob2",
    "/DNDEBUG",
    "/MP",
    "/wd4290",
    "/wd4267",
    "/wd4996",
    "/bigobj",
    "-D_CRT_SECURE_NO_WARNINGS",
    "-D_CRT_NONSTDC_NO_WARNINGS",
    "-D_CRT_XNONSTDC_NO_WARNINGS",
    "-D__restrict__=__restrict",
    "-DNOMINMAX",
    "-DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE",
    "/experimental:external",
    "/external:anglebrackets",
    "/external:W0",
    "/favor:blend",
    "/W0",
    "/showIncludes",
    "/Fo_deps\\hdf5-build\\src\\CMakeFiles\\hdf5-static.dir\\H5Dbtree.c.obj",
    "/Fdlib\\libhdf5.pdb",
    "/FS",
    "-c",
    "C:\\cpm\\hdf5\\b29b430dd2bdc4f489cfe8f50785e974d9afda4c\\src\\H5Dbtree.c"
]
```

Thanks!

AJIOB:
@bilke, hi.

Is the issue is still relevant?

PR #1050, opened 2021-10-10
Add some msvc flags

yuyoyuppe:
- */external:W** arg family added which mimics */W** family
- some msvc versions required /experimental:external to be able to use the */external:W** family, see e.g. https://github.com/mozilla/sccache/issues/1009
- fixed compile warnings on 1.55


src/compiler/msvc.rs
```diff
@@ -391,10 +391,19 @@ msvc_args!(static ARGS: [ArgInfo<ArgData>; _] = [
     msvc_take_arg!("doc", PathBuf, Concatenated, TooHardPath), // Creates an .xdc file.
     msvc_take_arg!("errorReport:", OsString, Concatenated, PassThroughWithSuffix), // Deprecated.
     msvc_take_arg!("execution-charset:", OsString, Concatenated, PassThroughWithSuffix),
+    msvc_flag!("experimental:external", PassThrough),
     msvc_flag!("experimental:module", TooHardFlag),
     msvc_flag!("experimental:module-", PassThrough), // Explicitly disabled modules.
     msvc_take_arg!("experimental:preprocessor", OsString, Concatenated, PassThroughWithSuffix),
     msvc_take_arg!("external:I", PathBuf, CanBeSeparated, ExternalIncludePath),
+    msvc_flag!("external:W0", PassThrough),
```
sylvestre, line 399: Could you please add comments explaining why we are adding these flags?
thanks

yuyoyuppe: @sylvestre Adding those flags are required because otherwise sccache treats them wrongly, resulting in false-positive cache misses.

yuyoyuppe:
Regarding the failed tests on macos, can anyone please help with that?

milahu:
> ```note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.``` ([line 571](https://github.com/mozilla/sccache/runs/4007582992?check_suite_focus=true#step:5:571))

meh ...

src/compiler/rust.rs
```diff
@@ -1131,7 +1131,7 @@ fn parse_arguments(arguments: &[OsString], cwd: &Path) -> CompilerArguments<Pars
                 cannot_cache!(concat!("missing ", stringify!($x)));
             };
         };
-    };
+    }
```
mitchhentges, line 1134: Let's resolve clippy/compiler warnings in a separate patch, as they're unrelated to the MSVC flag shenanigans we're up to here.

src/compiler/msvc.rs
mitchhentges: Can you elaborate on what it means for "sccache treats them wrongly"?
Is this an issue with an equivalent of `-Werror` where different warning levels may change whether or not a compile passes or succeeds?

If you can provide this information in the commit message, that'd be ideal :)

mitchhentges (changes requested):
Thanks for the patch - I'm pretty sure that I understand why this is beneficial, but I had to do some significant digging to do so.

I'd like to inline some of this information to be part of your commit here so that future readers will find it easier to understand what's going on.

Speaking of which, for future readers, the `external:` flags are documented [here](https://docs.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170)

yuanfang-chen:
MSVC flags handling needs some change. But not in a way to teach it about every flags.

I have a patch here https://github.com/yuanfang-chen/sccache/commit/ae7df77e3ea1235b9739b230c33d31b64bc37330 to address most of the MSVC tickets. I cannot upstream it for the moment, feel free to use it if it looks useful (it has some dependencies though).

mitchhentges:
> I have a patch here https://github.com/yuanfang-chen/sccache/commit/ae7df77e3ea1235b9739b230c33d31b64bc37330 to address most of the MSVC tickets. I cannot upstream it for the moment, feel free to use it if it looks useful (it has some dependencies though).

Nice, thanks! I've created a ticket over [here](https://github.com/mozilla/sccache/issues/1122) to track that work.

sylvestre:

yuyoyuppe:
nope, [this pr](https://github.com/mozilla/sccache/pull/1141) superseded it.

PR #1055, opened 2021-10-16
Add -fminimize-whitespace preprocessor flag when Clang >= 14 to increase cache hits

InnovativeInventor:
This proof of concept tries to take advantage of the new `-fminimize-whitespace` preprocessor option in Clang (14.0.0, not released yet). `-fminimize-whitespace` was added to Clang with the intention to help increase the chance of a cache hit in tools like `sccache` when only whitespace changes.

To illustrate this change, here are two sample C programs that previously wouldn't trigger a cache hit before this PR, but now do (when using Clang >= 14):
```c
#include <stdio.h>

int main() 


{


   printf("Hello World!");
return 0;
}
```
```c
#include <stdio.h>

int main() {printf("Hello World!");return 0;}
```

I also added some code to expose the compiler version to the preprocessor. This could be useful if version-dependent features are added to `sccache` in the future.

Feedback on this is definitely welcome! Note that Clang 14.0.0 is still in progress, so there is no rush to merge in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,16 @@ jobs:
         with:
           toolchain: ${{ matrix.rustc }}
 
+      - name: Install LLVM/Clang
+        if: contains(matrix.os, 'ubuntu')
+        run: 
+          sudo apt-get update || true
+          sudo apt-get install wget
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 13
+          sudo ./llvm.sh 14
+
       - name: Build tests
         run: cargo test --no-run --locked --all-targets --verbose ${{ matrix.extra_args }}
 
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -45,6 +45,7 @@ where
     executable: PathBuf,
     executable_digest: String,
     compiler: I,
+    version: Option<String>,
 }
 
 /// A generic implementation of the `CompilerHasher` trait for C/C++ compilers.
@@ -165,6 +166,8 @@ pub enum CCompilerKind {
 pub trait CCompilerImpl: Clone + fmt::Debug + Send + Sync + 'static {
     /// Return the kind of compiler.
     fn kind(&self) -> CCompilerKind;
+    /// Return the version string of the compiler, if it exists
+    fn version(&self) -> Option<String>;
     /// Return true iff this is g++ or clang++.
     fn plusplus(&self) -> bool;
     /// Determine whether `arguments` are supported by this compiler.
@@ -184,6 +187,7 @@ pub trait CCompilerImpl: Clone + fmt::Debug + Send + Sync + 'static {
         env_vars: &[(OsString, OsString)],
         may_dist: bool,
         rewrite_includes_only: bool,
+        version: Option<String>,
     ) -> Result<process::Output>
     where
         T: CommandCreatorSync;
@@ -215,7 +219,7 @@ where
         Ok(CCompiler {
             executable,
             executable_digest: {
-                if let Some(version) = version {
+                if let Some(ref version) = version {
                     let mut m = Digest::new();
                     m.update(digest.as_bytes());
                     m.update(version.as_bytes());
@@ -225,6 +229,7 @@ where
                 }
             },
             compiler,
+            version,
         })
     }
 }
@@ -295,6 +300,7 @@ where
                 &env_vars,
                 may_dist,
                 rewrite_includes_only,
+                compiler.version(),
             )
             .await;
         let out_pretty = parsed_args.output_pretty().into_owned();
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -35,12 +35,16 @@ use crate::errors::*;
 pub struct Clang {
     /// true iff this is clang++.
     pub clangplusplus: bool,
+    pub version: Option<String>,
 }
 
 #[async_trait]
 impl CCompilerImpl for Clang {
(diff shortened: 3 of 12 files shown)


Meinersbur:
Great addition! `-fminimize-whitespace` should allow reusing the cached artifact even after running clang-format over the source (at least without ` --sort-includes`).

sylvestre:
Could you please fix the conflict?
Also, would it be possible to add a test?

drahnr:
Does this change the resulting binary behavior? Will `__LINE__` and debug info point to the correct locations in the source code?

Meinersbur:
> Does this change the resulting binary behavior? Will `__LINE__` and debug info point to the correct locations in the source code?

`__LINE__` directives are already replaced in the preprocessor output. With just `-E`, line directives and newlines remain the same. With `-E -P`, no line directives are emitted anyway so there are no line numbers to preserve anyway. This patch only adds `-fminimize-whitespace` when `-P` is added, although there would be no problem to add it as well when just using `-E`.

InnovativeInventor:
> Could you please fix the conflict? 

Just did. I also addressed the style issues that clippy was warning about.

> Also, would it be possible to add a test?

I'm not quite sure how I should do this (since clang 14 isn't released yet, I was thinking about pinning a particular commit and building it in the CI environment). Do you have any suggestions?

sylvestre:
@InnovativeInventor maybe create a job CI task and use a clang-14 package from apt.llvm.org for Ubuntu  ?

sylvestre:
It seems that it need a clear rebase:
"This branch cannot be rebased due to conflicts"

InnovativeInventor:
> @InnovativeInventor maybe create a job CI task and use a clang-14 package from apt.llvm.org for Ubuntu ?

Ok, just pushed some tests (only for Ubuntu right now) that check that whitespace causes a cache miss on clang-13 (as expected) and a cache hit on clang-14. I included a check for clang-13 mostly to demonstrate that the tests would fail if the feature was not implemented properly (to show that the tests actually work).

I also cleaned up the commit history, so it should be ready now (CI appears to pass on my fork: https://github.com/InnovativeInventor/sccache/actions/runs/1473112165).

sylvestre:
Would it be possible to not hardcode the llvm version ?
or it is too much work?

glandium:
You can't rely on version number for clang, because Xcode's clang has a version that corresponds to the Xcode version, and that doesn't match the LLVM version.

Meinersbur:
What would be the alternative to use the llvm version? One could call the compiler at the beginning with the `-fminimize-whitespace` to see whether it it is accepted, but it would mean another compiler invocation every time. Would that be acceptable?

mitchhentges (changes requested):
Hey, this patch is really cool, good idea on using `-fminimize-whitespace`.

Unfortunately, as alluded by glandium, we can't lean on the version number because XCode's version isn't aligned with LLVM's - see [this table](https://en.wikipedia.org/wiki/Xcode#Xcode_11.x_-_13.x_(since_SwiftUI_framework)_2).

> What would be the alternative to use the llvm version? One could call the compiler at the beginning with the -fminimize-whitespace to see whether it it is accepted, but it would mean another compiler invocation every time. Would that be acceptable?

I think the right move would be to call `clang` once to query if it supports `-fminimize-whitespace`, caching the result in the `sccache` server (probably), then reusing that value for future usages of the same `clang`.

It does cost a compiler call to determine support, but IIRC we already incur a "call the compiler" cost to determine its version. We can call again (or, ideally, append to that existing "detect_c_compiler" call) to get other feature information.

Meinersbur:
Wouldn't it be sufficient to identify Apply clang with `__apple_build_version__ ` or `__APPLE_CC__`? [This is how CMake detects ApplyClang](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/AppleClang-DetermineCompiler.cmake#L2).

InnovativeInventor:
@Meinersbur is there a `__clang_minimize_whitespace__` thing I can check for? If not, I can call `--help` and parse the flags (might be more brittle, though).

mitchhentges:
> Wouldn't it be sufficient to identify Apply clang with __apple_build_version__ or __APPLE_CC__? [This is how CMake detects AppleClang](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/AppleClang-DetermineCompiler.cmake#L2).

Hmm, that could work, yeah.
We'll probably want to store both pieces of information (version + apple/llvm clang) on the sccache compiler object - or, front-load the decision-making, and just store "supports minimize whitespace" instead.
Related: can we remove `version` (or whatever we store) from the top level `CCompiler` struct, and just store the values needed for minimize-whitespace decision-making on the `Clang` struct instead?

Meinersbur:
> @Meinersbur is there a `__clang_minimize_whitespace__` thing I can check for? If not, I can call `--help` and parse the flags (might be more brittle, though).

There is not. Usually preprocessor macros are about language features, not which command line flags are accepted. Before [feature test macros](https://en.cppreference.com/w/cpp/feature_test), compiler version was the go-to mechanisms for language features. I could try to add it though, but clang-14 has already branched and I am not sure it would be included before release.

Before parsing `--help`, just invoke it with `-fminimize-whitespace` and see whether you get an error.

Meinersbur:
With clang 14 now released, any update on this PR?

Meinersbur:
This PR is superseded by #1162 and can be closed. Thanks @drahnr for merging!

InnovativeInventor:
> This PR is superseded by #1162 and can be closed. Thanks @drahnr for merging!

Nice!

PR #1030, opened 2021-08-20
Update CI Rust/Ubuntu versions and address build warnings

emabrey:
* Update Rust version from `1.50.0` to `1.54.0`
* Update Ubuntu CI version to `18.04` because Github no longer supports version `16.04`
* Fix incorrect usage of `panic!(&str)` instead of `panic!("{}", &str)`
* Replace usage of outdated clippy options
* Refactor codebase to bring it in line with rustfmt standards
* Refactor codebase to remove needless borrowing
* Refactor enum/struct names to be Camel Case instead of UPPER CASE
* Refactor code which creates a reference to a reference
* Refactor usages of redundant semicolons
* Refactor usages of outdated `panic!(format!(...))`

PR #1057, opened 2021-11-02
chore: apply clippy lint

lightsing:
apply clippy lint


sylvestre:
Seems that it is a dup of https://github.com/mozilla/sccache/pull/1030 no ?

PR #1062, opened 2021-11-05
update the blake3 dependency to v1

oconnor663:
The blake3 crate currently has a MSRV of 1.51 for const generics.


sylvestre:
The ci fails with
```

    Updating crates.io index
error: the lock file /home/runner/work/sccache/sccache/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, use the --offline flag.
Error: Process completed with exit code 101.
```

oconnor663:
Thanks, added the lockfile changes and force pushed.

oconnor663:
Ah it looks like there are some CI runs that test rustc 1.43. This PR will have to wait until it's ok to bump the MSRV.

mitchhentges:
Hey, thanks for the patch!
I'm going to close this because we're going to do [a larger update-bump pass](https://github.com/mozilla/sccache/issues/1110), which is also contingent on [us having a more concrete understanding of the restrictions on having the MSRV bumped](https://github.com/mozilla/sccache/issues/1110).

PR #1024, opened 2021-07-29
fix Date header

lightsing:
The `Date` header in s3 module doesn't obey the RFC7231 definition.
Causing s3-compatible server refuses sccache's request.

original request:
```
GET /0/1/2/012620f0a29db1fa159c8471bb82f9c25cad332cfffa5f680f7bb0cfd9c30e62 HTTP/1.1
date: Thu, 29 Jul 2021 06:56:44 +0000
authorization: AWS **RETRACTED**
user-agent: reqwest/0.9.24
accept: */*
accept-encoding: gzip
host: **RETRACTED**.oss-accelerate.aliyuncs.com
```
server response:
```
HTTP/1.1 403 Forbidden
Server: AliyunOSS
Date: Thu, 29 Jul 2021 06:56:44 GMT
Content-Type: application/xml
Content-Length: 252
Connection: keep-alive
x-amz-request-id: **RETRACTED**
x-oss-server-time: 0

<?xml version="1.0" encoding="UTF-8"?>
<Error>
  <Code>AccessDenied</Code>
  <Message>OSS authentication requires a valid Date.</Message>
  <RequestId>6102512CBB04C5A3079D59B5</RequestId>
  <HostId>sccache.oss-accelerate.aliyuncs.com</HostId>
</Error>
```

patched request:
```
GET /c/c/6/cc65e6de3e04ef4f9eaae65e71fbba04019c78792364467217248f1b06bc3e71 HTTP/1.1
date: Thu, 29 Jul 2021 07:28:02 GMT
authorization: AWS **RETRACTED**
user-agent: reqwest/0.9.24
accept: */*
accept-encoding: gzip
host: **RETRACTED**.oss-accelerate.aliyuncs.com
```

PR #1064, opened 2021-11-09
Fix compilation error when not using Azure support

Xanewok:
This also fixes CI errors. Introduced in #1024.

r? @sylvestre


sylvestre:
thanks!

PR #1068, opened 2021-11-18
Ignore profiling flag if no link is being emitted

montekki:
1. Give the error some context;
2. If no `--emit=link` is passed to `rustc` (as it is in case of `cargo check`) `-Zprofile` flag has to be ignored since no `*gcno` file will be generated.


montekki:
Also extend a test on `nightly` target.

sylvestre:
Looks great, thanks
Sorry for the latency, I missed this PR

PR #1072, opened 2021-11-27
Add SccacheServer::with_listener

cgrenz:
Currently a `SccacheServer` can only be constructed via `SccacheServer::new` with a port number. That prevents third-party crates from creating a server instance e.g. not bound to 127.0.0.1 or with an existing socket, e.g. from systemd socket activation.
    
This adds `SccacheServer::with_listener` to allow creating a server with an externally provided `TcpListener`.


mitchhentges (approved):
Makes sense, nice! Thanks for the patch :+1:

PR #1083, opened 2021-12-18
msvc: pass through -Brepro

yuanfang-chen:
Basically #980 with a unit test.


yuanfang-chen:
I'll send over an alternative patch.

PR #1085, opened 2021-12-20
msvc: -Xclang -ivfsoverlay support

yuanfang-chen:
https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-ivfsoverlay-arg
It is both driver and cc1 argument for clang, but only cc1 argument for clang-cl.

This is the discussion on the ccache side: https://github.com/ccache/ccache/pull/700

Since sccache always performs preprocess, add it to the preprocessor args.


glandium:
Considering https://ccache.dev/releasenotes.html#_new_features_4 I'm not sure this shouldn't be TooHardPath for now.

yuanfang-chen:
> Considering https://ccache.dev/releasenotes.html#_new_features_4 I'm not sure this shouldn't be TooHardPath for now.

Thanks for the reply. Do you mind elaborating a little bit? Like when this is not desirable for sccache? Ccache's implementation basically adds an environment variable to ignore the flag for certain scenarios (Xcode in this case). Ccache seems not able to assume "preprocessor mode" all the time. This seems not a restriction for sccache?

sylvestre:
Could you please make rustfmt happy? thanks

yuanfang-chen:
ping?

mitchhentges:
> Thanks for the reply. Do you mind elaborating a little bit?

I'm going to take a guess here: I don't think it's sufficient to reuse a cached resource if the `-ivfsoverlay` parameter is identical to that of the cache. The files referenced in the vfs may have different contents, which would mean that we're serving an invalid artifact.

I suppose that that's why the `ccache` chose to allow ignoring the parameter, but only if `SLOPPY_IVFSOVERLAY` was set.

So, I agree that this should probably be a `TooHardPath` for now, right?
Does that answer why our management of preprocessor flags isn't sufficient?

sylvestre:

Issue #281, opened 2018-08-27
Support AWS signature version 4 on S3


Marwes:
The version 2(?) signature that is currently used is deprecated and does not work on newer regions such as frankfurt. It might be worthwhile to modify the s3 implementation to use https://github.com/rusoto/rusoto instead of the custom one (which I assume supports the new signature).

https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html

luser:
I originally used rusoto for the s3 backend but switched off of it because rusoto didn't support something I needed to do. It has been a few years now so it's likely that things have changed to the point where that's no longer true. However, sccache uses hyper to do asynchronous HTTP requests currently, and given the unstable state of futures/async in the Rust ecosystem I don't know if switching would be straightforward.

If you or someone else would like to look into that I'd be happy to advise or look at patches. It might be the case that waiting for futures to stabilize in the Rust standard library and the crates ecosystem to settle around them is the best path forward.

luser:
An alternate option here would be to simply copy the rusoto v4 signature code into the sccache codebase. We did that with the authentication code originally (although it has since been heavily modified).

Marwes:
I will probably not implement this as just changing the s3 region of the bucket is an easy workaround. Still wanted it documented in case someone else gets bit by it!

zaidenwerg:
> The version 2(?) signature that is currently used is deprecated and does not work on newer regions such as frankfurt. It might be worthwhile to modify the s3 implementation to use https://github.com/rusoto/rusoto instead of the custom one (which I assume supports the new signature).
> 
> https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html

Hey, starting April 2020 new S3 buckets will not support SigV2 anymore. Maybe it's time to reconsider working on this ? 

https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/

andrewwebber:
Is anyone using SCCACHE outside of the US?
In Europe I would probably not want to have network hops over the Atlantic just to use caching on a CI server in Europe?

Xuanwo:
We have implemented this. @sylvestre Please close this issue.

Issue #217, opened 2018-01-13
Support using authentication on s3 get requests


johnklai1:
Our s3 bucket requires authorization for get requests as well, and I noticed that the s3 implementation only sets authorization for put requests.  I have a change that uses credentials for the get requests as well, but I'm not sure if this needs to be configurable (e.g. only use credentials for get requests if some environment variable is set).

https://github.com/johnklai1/sccache/commit/8660c006d965947b4eac0082a072f6e98d8d3a80

luser:
Yeah, I would like to be able to preserve the existing behavior, it's what we use in Firefox CI (our s3 buckets are world-readable). It's slightly faster to do unauthenticated GET requests if possible. I'd love to support this as well, though. We've been hung up for a while by sccache not supporting a config file, which would make it much more straightforward to have nontrivial configurations.

jechase:
#135 implements this feature in a configurable manner. I've been maintaining my fork which includes it for use in my company's build pipeline until it gets merged.

Xuanwo:
Has been implemented.

cc @sylvestre Please close this issue.

PR #1086, opened 2022-01-04
New S3 backend

alessandrod:
The backend uses the official AWS SDK for Rust. 

Supersedes https://github.com/mozilla/sccache/pull/869 (including readonly mode with `SCCACHE_S3_NO_CREDENTIALS`), and fixes https://github.com/mozilla/sccache/issues/632 https://github.com/mozilla/sccache/issues/281 and https://github.com/mozilla/sccache/issues/217

The new backend is largely compatible with the old one in terms of conf settings, except that it does not support `SCCACHE_S3_USE_SSL=0`. SSL is always required at the moment.


glandium:
This doesn't address https://github.com/mozilla/sccache/pull/869#issuecomment-742368633 AFAICT.

alessandrod:
> This doesn't address [#869 (comment)](https://github.com/mozilla/sccache/pull/869#issuecomment-742368633) AFAICT.

Oh oops my bad, I completely missed that. I'll look into it!

sylvestre:
@alessandrod do you have an update? thanks

mitchhentges (changes requested):
Hey, would you mind rebasing this off `main`?
Sorry for the delay in reviewing this so far, but I'll dig into it if you're able to update it :)

alessandrod:
Rebased, sorry for the delay 😊

mitchhentges:
Looks like we have some CI failures:
```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.3.7/Cargo.toml`

Caused by:
  feature `resolver` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["resolver"]` to enable this feature
```

I'm guessing that this will be blocked on #1121 anyways, which is itself blocked on [Milestone 3](https://github.com/mozilla/sccache/milestone/3).

alessandrod:
Yep, it looks like `resolver` was added in cargo 1.51

jrcichra:
Just wanted to say thanks @alessandrod for this PR. The `main` branch wasn't working with my Minio cluster (getting `404 Not Found` ) errors. Your PR solved those issues and works great for me 💯

GZGavinZhao:
I have resolved conflicts and made a PR to the author's branch. The struggle for a new S3 backend will not be abandoned :P

alessandrod:
I've rebased this, sorry for the delay! Someone kick tests pls.

GZGavinZhao:
Is there a particular reason why the aws dependencies are on version 0.3? I tested with 0.12 and it works fine on my machine.

src/cache/s3.rs
```diff
@@ -96,33 +67,99 @@ impl Storage for S3Cache {
...
+    config: Config,
+}
+
+impl S3Client {
+    async fn new(bucket: &str, region: Option<&str>, endpoint: Option<&str>) -> Result<S3Client> {
+        let region_provider =
+            RegionProviderChain::first_try(region.map(|r| Region::new(r.to_owned())))
+                .or_default_provider();
+
+        let shared_config = aws_config::from_env().region(region_provider).load().await;
```
drahnr, line 114: We should not pull in env vars here, but rather use the arguments and the arguments alone. Everything else should be part of the config file. There are two reasons, mainly maintainability, reproducibility of tests.

drahnr (approved):
LGTM, there is one nit that must be resolved, otherwise looks good to me. Going to test on a rather old minio cluster before merging. Could you also bump the dependencies to their latest versions? 

Does this require openssl under the hood or is there an option to use rustls instead, if so that should be the default.

Thank you for your work and endurance on this!

src/cache/s3.rs
woss: i'd like to chip in. why not pull envs in? the whole AWS compatible cli works the same way. and passing the AWS_ACCESS_SECRET_KEY as an argument sounds terribly risky

drahnr: I am not saying to remove the functionality, but with clap there is an easier way to provide the information in one place. If env is pulled in from allover the place, that's not maintainable.

ThomasK33: The clap approach of centrally pulling in AWS credentials from related arguments would lead to less maintainability in this case.

The aws_config crate pulls credentials from [different sources](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence) with their corresponding precedences.

If the access key id and secret access key method-usage is hard-coded, authentication options such as a [CLI credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html), [IAM roles for ECS tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) and [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) become unusable right from the get-go.

Also, future CLI authentication method integrations, e.g., IAM roles anywhere, would have to be added manually instead of just updating the aws_config crate.

ajschmidt8:
bump! This is so close to being merged.

@alessandrod, can you bump the dependencies as per the request in the previous comment above?

messense:
Opened a rebased version of this PR in #1403, hopefully it will speed things up.

sylvestre:
merged here:
https://github.com/mozilla/sccache/pull/1403

PR #1087, opened 2022-01-04
gcc/clang: -D/-U are preprocessor flags

yuanfang-chen:
Diab is correct in this regard.

Without this, the flags are passed to the compilation and may end up like:
`clang: warning: argument unused during compilation: '-U NDEBUG' [-Wunused-command-line-argument]`


sylvestre:
sorry, I merged it too quickly
thanks @glandium

PR #1088, opened 2022-01-04
Packaging clang: always add a clang-cl symlink to clang for convenience

yuanfang-chen:
I don't see existing tests for packaging and I'm not sure it is worthwhile to add them just for this. I've tried locally that this works well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 strip-ansi-escapes = "0.1"
-tar = "0.4"
+tar = "^0.4.38"
 tempfile = "3"
 tokio = { version = "1", features = ["rt-multi-thread", "io-util", "time", "net", "process", "macros"] }
 tokio-serde = "0.8"
--- a/docs/DistributedQuickstart.md
+++ b/docs/DistributedQuickstart.md
@@ -190,10 +190,9 @@ sufficient to run the compiler without relying on any external files. If you hav
 compatible with icecream (created with `icecc-create-env`, like
 [these ones](https://github.com/jyavenard/mozilla-icecream) for macOS), they should also work
 with sccache. To create a Windows toolchain, it is recommended that you download the [Clang
-binaries for Ubuntu 16.04](http://releases.llvm.org/download.html) and extract them,
-package up the toolchain using the extracted `bin/clang` file (requires
-[PR #321](https://github.com/mozilla/sccache/pull/321)) and then insert `bin/clang-cl` at
-the appropriate path as a symlink to the `bin/clang` binary.
+binaries for Ubuntu](https://github.com/llvm/llvm-project/releases) and extract them,
+package up the toolchain using the extracted `bin/clang` file (`bin/clang-cl` would be created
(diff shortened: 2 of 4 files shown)


sylvestre:
@yuanfang-chen ping?

Cargo.toml
```diff
@@ -69,7 +69,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 strip-ansi-escapes = "0.1"
-tar = "0.4"
+tar = "^0.4.38"
```
mitchhentges, line 72: Is this change necessary as part of this patch?

docs/DistributedQuickstart.md
```diff
@@ -190,10 +190,9 @@ sufficient to run the compiler without relying on any external files. If you hav
 compatible with icecream (created with `icecc-create-env`, like
 [these ones](https://github.com/jyavenard/mozilla-icecream) for macOS), they should also work
 with sccache. To create a Windows toolchain, it is recommended that you download the [Clang
-binaries for Ubuntu 16.04](http://releases.llvm.org/download.html) and extract them,
-package up the toolchain using the extracted `bin/clang` file (requires
-[PR #321](https://github.com/mozilla/sccache/pull/321)) and then insert `bin/clang-cl` at
-the appropriate path as a symlink to the `bin/clang` binary.
+binaries for Ubuntu](https://github.com/llvm/llvm-project/releases) and extract them,
+package up the toolchain using the extracted `bin/clang` file (`bin/clang-cl` would be created
```
mitchhentges, line 194:
```suggestion
package up the toolchain using the extracted `bin/clang` file (`bin/clang-cl` will be created
```

src/dist/pkg.rs
```diff
@@ -169,8 +169,20 @@ mod toolchain_imp {
                 builder.append_dir(tar_path, dir_path)?
             }
             for (tar_path, file_path) in file_set.into_iter() {
-                let file = &mut fs::File::open(file_path)?;
-                builder.append_file(tar_path, file)?
+                let file = &mut fs::File::open(&file_path)?;
+                builder.append_file(&tar_path, file)?;
+                // Should have no need to package clang-cl.exe since the dist
+                // server must a Linux environemnt.
```
mitchhentges, line 175:
```suggestion
                // Should have no need to package clang-cl.exe since the dist
                // server must a Linux environment.
```
I don't quite understand this comment: isn't `clang-cl` only used on Windows, therefore we want `clang-cl.exe`? Or, is `clang-cl` (without `.exe`) used on other platforms for cross-compiling? What is the "dist server" in this context?

mitchhentges (approved):
This looks pretty good, but before I land this:

> I don't see existing tests for packaging and I'm not sure it is worthwhile to add them just for this

How challenging will it be to add tests for this? I think it'd be worth it to avoid breakage here in the future.
If it requires nontrivial infra changes (I don't _think_ so?), then of course it might not be worth it :)

mitchhentges (changes requested):
I'm going to mark this as Requesting Changes to make it more clear that this is blocked at least on confirming why we only package `clang-cl.exe` on Linux environments.

PR #1090, opened 2022-01-06
add one option to continue building even there's any server errors.

### LINKED ISSUE #1089: There should be an option that it would run compilation command locally  if there's any server errors.
In handle_compile_response, only if the server returned `UnhandledCompile`, it will run compilation command locally.
As the result, the CI might be broken for some intermittent sccache server errors, like` unexpected response from server`.
In practice, slowness is better than exception in CI.
For any sccache server errors, it only means that cache isn't available at some time.
So, there should an option that build could be continued even there's any server errros, like `UnhandledCompile`

mszhanyi:
Fixes #1089

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -53,6 +53,16 @@ fn get_port() -> u16 {
         .unwrap_or(DEFAULT_PORT)
 }
 
+/// Check if ignoring all response errors
+fn ignore_all_server_errors() -> bool {
+    let ignore;
+    match env::var("SCCACHE_IGNORE_RESPONSE_ERROR") {
+        Ok(val) => ignore = val,
+        Err(_e) => ignore = "none".to_string(),
+    }
+    ignore == "1"
+}
+
 async fn read_server_startup_status<R: AsyncReadExt + Unpin>(
     mut server: R,
 ) -> Result<ServerStartup> {
@@ -464,7 +474,18 @@ where
                 Ok(Response::CompileFinished(result)) => {
                     return handle_compile_finished(result, stdout, stderr)
                 }
-                Ok(_) => bail!("unexpected response from server"),
+                Ok(_) => {
+                    if ignore_all_server_errors() {
+                        eprintln!(
+                            "sccache: warning: unexpected response from server \
+                             compiling locally instead"
+                        );
+                    }
+                    else {
+                        bail!("unexpected response from server")
+                    }
+
+                }
                 Err(e) => {
                     match e.downcast_ref::<io::Error>() {
                         Some(io_e) if io_e.kind() == io::ErrorKind::UnexpectedEof => {
@@ -475,7 +496,15 @@ where
                         }
                         _ => {
                             //TODO: something better here?
-                            return Err(e).context("error reading compile response from server");
+                            if ignore_all_server_errors() {
+                                eprintln!(
+                                    "sccache: warning: error reading compile response from server \
+                                     compiling locally instead"
+                                );
+                            }
+                            else {
+                                return Err(e).context("error reading compile response from server");
+                            }
                         }
                     }
                 }
@@ -483,7 +512,15 @@ where
         }
         CompileResponse::UnsupportedCompiler(s) => {
             debug!("Server sent UnsupportedCompiler: {:?}", s);
-            bail!("Compiler not supported: {:?}", s);
+            if ignore_all_server_errors() {
+                eprintln!(
+                    "sccache: warning: Compiler not supported: {:?} \
+                     compiling locally instead", s
+                );
+            }
+            else {
+                bail!("Compiler not supported: {:?}", s);
+            }
         }
         CompileResponse::UnhandledCompile => {
             debug!("Server sent UnhandledCompile");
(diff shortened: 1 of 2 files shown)


src/commands.rs
```diff
@@ -464,7 +474,18 @@ where
                 Ok(Response::CompileFinished(result)) => {
                     return handle_compile_finished(result, stdout, stderr)
                 }
-                Ok(_) => bail!("unexpected response from server"),
+                Ok(_) => {
```
yuanfang-chen, line 477: It is a programming error to see an unexpected response. I think this is not needed.

src/commands.rs
```diff
@@ -475,15 +496,31 @@ where
...
+                            }
                         }
                     }
                 }
             }
         }
         CompileResponse::UnsupportedCompiler(s) => {
             debug!("Server sent UnsupportedCompiler: {:?}", s);
-            bail!("Compiler not supported: {:?}", s);
+            if ignore_all_server_errors() {
```
yuanfang-chen, line 515: It should be a user config problem to supply an unsupported compiler. I think this is not needed.

src/commands.rs
```diff
@@ -53,6 +53,16 @@ fn get_port() -> u16 {
         .unwrap_or(DEFAULT_PORT)
 }
 
+/// Check if ignoring all response errors
+fn ignore_all_server_errors() -> bool {
+    let ignore;
+    match env::var("SCCACHE_IGNORE_RESPONSE_ERROR") {
```
yuanfang-chen, line 59: I'd prefer SCCACHE_IGNORE_SERVER_IO_ERROR and it would be nice to have a short description of this flag in the top-level readme.

yuanfang-chen (changes requested):

sylvestre:
rustfmt isn't happy ;)

mszhanyi:
@sylvestre  Thank your reply. updated and verified it locally.

sylvestre:
thanks
would it be possible to add a test for this or it is too complex?

src/commands.rs
```diff
@@ -475,7 +485,15 @@ where
                         }
                         _ => {
                             //TODO: something better here?
-                            return Err(e).context("error reading compile response from server");
+                            if ignore_all_server_io_errors() {
+                                eprintln!(
+                                    "sccache: warning: error reading compile response from server \
+                                     compiling locally instead"
```
sylvestre, line 491:
```suggestion
                                     compiling locally instead (option SCCACHE_IGNORE_SERVER_IO_ERROR passed)"
```

mszhanyi:
> thanks would it be possible to add a test for this or it is too complex?

So far, I haven't find how to mimic the exception in tests

mszhanyi:
@sylvestre 
I add one commit to print the unexpected exception.
Because it's a random error so I don't use the `debug!`
https://github.com/mszhanyi/sccache/blob/41b9e279a72b05d0869772d839f0aa4cae98f072/src/commands.rs#L488

src/commands.rs
```diff
@@ -53,6 +53,16 @@ fn get_port() -> u16 {
...
 }
 
+/// Check if ignoring all response errors
+fn ignore_all_server_io_errors() -> bool {
+    let ignore;
+    match env::var("SCCACHE_IGNORE_SERVER_IO_ERROR") {
+        Ok(val) => ignore = val,
+        Err(_e) => ignore = "none".to_string(),
+    }
+    ignore == "1"
```
mitchhentges, line 63:
```suggestion
    match env::var("SCCACHE_IGNORE_SERVER_IO_ERROR") {
        Ok(ignore_server_error) => ignore_server_error == "1"
        Err(_) => false
    }
```
This should be simplify-able a bit :)

src/commands.rs
```diff
@@ -475,7 +485,16 @@ where
                         }
                         _ => {
                             //TODO: something better here?
-                            return Err(e).context("error reading compile response from server");
+                            eprintln!("unexpected io error: {:?}", e.downcast_ref::<io::Error>());
```
mitchhentges, line 488: Since you're already `eprintln!()`-ing/returning `Err(e)`, this `eprintln!()` is probably redundant.

src/commands.rs
```diff
@@ -475,7 +485,16 @@ where
...
+                            if ignore_all_server_io_errors() {
+                                eprintln!(
+                                    "sccache: warning: error reading compile response from server \
+                                     compiling locally instead (option SCCACHE_IGNORE_SERVER_IO_ERROR passed)"
```
mitchhentges, line 492: IMO, the "option SCCACHE_IGNORE_SERVER_IO_ERROR passed" part of this message is redundant, let's remove it.
Worst comes to worst, if someone is _really_ curious why they're seeing this error message, they can check the source code.
This'll also make use consistent with the "server looks like it shut down" message.
```suggestion
                                    "sccache: warning: error reading compile response from server, \
                                     compiling locally instead"
```

mitchhentges (changes requested):

src/commands.rs
mszhanyi: Agree! revert the commit https://github.com/mozilla/sccache/pull/1090/commits/878988e4ebe8a7bc99393f5355da6518a5909edc

mitchhentges: Whoops, missed that Sylvestre requested that, my mistake. I think it's ok either way, though I have my preference ;)

mitchhentges (approved):
Looks good, thanks for the update!

PR #1093, opened 2022-01-13
add more input file extensions

milahu:
add more file extensions from https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html

note: these are case-sensitive, so gcc does not recognize `input.CpP`

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -803,4 +803,48 @@ mod test {
             hash_key(digest, Language::C, &args, &[], &[], PREPROCESSED, false)
         );
     }
+
+    #[test]
+    fn test_language_from_file_name() {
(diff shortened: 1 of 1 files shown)


sylvestre:
Could you please add a test ?

sylvestre:
and rustfmt isn't happy

milahu:
> Could you please add a test ?

no : )

sylvestre:
can i ask why ? :)

milahu:
too trivial

sylvestre:
Well, it is but the goal is to make sure we don't regress in the future (that extensions aren't removed by mistake at some point). 
Not about the complexity of the patch

milahu:
(push)

PR #1095, opened 2022-01-14
normalize: gcc -frandom-seed

### LINKED ISSUE #1092: normalize: gcc -frandom-seed=xxx
this is required for reproducible builds and cache hits

build tools like cmake will set `random-seed` to a random string, which prevents cache hits

when `random-seed` is not set, then the compiler will choose a random value

bazel solves this by always setting `-frandom-seed=%{output_file}`, see https://github.com/bazelbuild/bazel/issues/6540

milahu:
fix #1092 

works for me with gcc ... add test?

todo: fix all the test_parse_arguments_* tests

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -128,6 +128,7 @@ ArgData! { pub
     // Only valid for clang, but this needs to be here since clang shares gcc's arg parsing.
     XClang(OsString),
     Arch(OsString),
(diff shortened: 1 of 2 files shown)


sylvestre:
Could you please add a test to verify it works as expected?

sylvestre:
And many tests are failing

milahu:
nope, sorry ... \*abandon\*

PR #1104, opened 2022-01-31
Support Google Cloud Storage bucket key prefixes

FooBarWidget:
Adds support for Google Cloud Storage bucket key prefixes, similar to `SCCACHE_S3_KEY_PREFIX`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,8 @@ If you're using authentication, either set `SCCACHE_GCS_KEY_PATH` to the locatio
 a URL that returns the oauth token.
 By default, SCCACHE on GCS will be read-only. To change this, set `SCCACHE_GCS_RW_MODE` to either `READ_ONLY` or `READ_WRITE`.
 
+You can also define a prefix that will be prepended to the keys of all cache objects created and read within the GCS bucket, effectively creating a scope. To do that use the `SCCACHE_GCS_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
+
 ### Azure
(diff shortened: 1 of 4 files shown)


sylvestre:
i guess it is too hard to add a test for this?

FooBarWidget:
There were nearly no tests for that module in the first place. I tested by hand. If you want I can attempt to add a test, but that might involve changing various structures in order to make them testable (e.g. adding mocks for network calls). What do you think?

FooBarWidget:
I added some basic tests and fixed some bugs too.

sylvestre:
Excellent, thanks.
Do you know if we have the same feature in AWS?

FooBarWidget:
We do. The feature is already in sccache (`SCCACHE_S3_KEY_PREFIX`).

sylvestre:
it is possible to factor some code? thanks :)

FooBarWidget:
What do you mean by that?

sylvestre:
Sharing code between AWS & GCP support on this :)

FooBarWidget:
Looking at respective codebases, I don't see what can be shared. Perhaps the normalize_key function can be shared, but there is no guarantee that this function can be shared with other object stores (which may have different key schemes), which makes me question the usefulness of sharing that function.

If you think there are more sharing opportunities, would you please advice me on what concrete steps should be taken?

FooBarWidget:
Thanks for merging! Can I ask you to do a release soon? I need this feature for https://github.com/fullstaq-labs/fullstaq-ruby-server-edition/issues/86

PR #1106, opened 2022-02-03
Display error details when Azure connection string parsing fails

FooBarWidget:
When using the Azure backend, and there is a problem with the connection string, then sccache does not report what's wrong with the connection string:

~~~
[2022-02-03T16:23:34Z WARN  sccache::cache::cache] Failed to create Azure cache: Could not find Azure credentials in the environment
~~~

After this fix, sccache properly includes the reason:

~~~
[2022-02-03T16:27:07Z WARN  sccache::cache::cache] Failed to create Azure cache: Could not find Azure credentials in the environment: Can not infer blob endpoint; connection string is missing BlobEndpoint, AccountName, and/or EndpointSuffix.
~~~


FooBarWidget:
Done.

PR #1107, opened 2022-02-05
Parse env-deps from dep-info

### LINKED ISSUE #804: Handle `# env-dep` directives in dep-info files
[rustc nightly added support](https://github.com/rust-lang/rust/pull/71858) for adding extra annotations in the form of Makefile comments to dep-info files to track environment variables accessed by way of the `env!` macro. sccache should parse these directives and add referenced environment variables as hash inputs.

@alexcrichton has [a pull request open to add support to cargo](https://github.com/rust-lang/cargo/pull/8421). I'm sure it'd be straightforward to add support to sccache's parser.

### LINKED ISSUE #972: Better support for env! in rust code
`env!` in rust code includes the literal value of environment variables set at compilation time. sccache currently hashes environment variables that start with `CARGO_` except `CARGO_METAFLAGS`, or `CARGO` (per #89 and #942), but that doesn't cover all possible use cases and is thus not future-proof.

Rust has recently added support for outputing the variables used by the source in its dep-info (https://github.com/rust-lang/rust/issues/40364), so for appropriate versions of rustc, we could actually do the right thing.

jschwe:
Parse the environment variables (and their values) referenced in rust code via `env!` or `option_env!` from the `dep-info` file and add them to the hash. The `CARGO_` env variables apparently still have to be blanket added to the hash, since they don't appear in the `dep-info` file.

Closes #972 
Closes #804

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -197,8 +197,8 @@ lazy_static! {
 /// Version number for cache key.
 const CACHE_VERSION: &[u8] = b"6";
 
-/// Get absolute paths for all source files listed in rustc's dep-info output.
-async fn get_source_files<T>(
+/// Get absolute paths for all source files and env-deps listed in rustc's dep-info output.
+async fn get_source_files_and_env_deps<T>(
     creator: &T,
     crate_name: &str,
     executable: &Path,
@@ -206,7 +206,7 @@ async fn get_source_files<T>(
     cwd: &Path,
     env_vars: &[(OsString, OsString)],
     pool: &tokio::runtime::Handle,
-) -> Result<Vec<PathBuf>>
+) -> Result<(Vec<PathBuf>, Vec<(OsString, OsString)>)>
 where
     T: CommandCreatorSync,
 {
@@ -238,22 +238,23 @@ where
         })
         .await?;
 
-    parsed.map(move |files| {
+    parsed.map(move |(files, env_deps)| {
         trace!(
-            "[{}]: got {} source files from dep-info in {}",
+            "[{}]: got {} source files and {} env-deps from dep-info in {}",
             crate_name,
             files.len(),
+            env_deps.len(),
             fmt_duration_as_secs(&start.elapsed())
         );
         // Just to make sure we capture temp_dir.
         drop(temp_dir);
-        files
+        (files, env_deps)
     })
 }
 
 /// Parse dependency info from `file` and return a Vec of files mentioned.
 /// Treat paths as relative to `cwd`.
-fn parse_dep_file<T, U>(file: T, cwd: U) -> Result<Vec<PathBuf>>
+fn parse_dep_file<T, U>(file: T, cwd: U) -> Result<(Vec<PathBuf>, Vec<(OsString, OsString)>)>
 where
     T: AsRef<Path>,
     U: AsRef<Path>,
@@ -261,7 +262,7 @@ where
     let mut f = fs::File::open(file)?;
     let mut deps = String::new();
     f.read_to_string(&mut deps)?;
-    Ok(parse_dep_info(&deps, cwd))
+    Ok((parse_dep_info(&deps, cwd), parse_env_dep_info(&deps)))
 }
 
 fn parse_dep_info<T>(dep_info: &str, cwd: T) -> Vec<PathBuf>
@@ -313,6 +314,20 @@ where
     let mut deps = deps.iter().map(|s| cwd.join(s)).collect::<Vec<_>>();
     deps.sort();
     deps
+}
+
+fn parse_env_dep_info(dep_info: &str) -> Vec<(OsString, OsString)> {
+    let mut env_deps = Vec::new();
+    for line in dep_info.lines() {
+        if let Some(env_dep) = line.strip_prefix("# env-dep:") {
+            let mut split = env_dep.splitn(2, '=');
+            match (split.next(), split.next()) {
+                (Some(var), Some(val)) => env_deps.push((var.into(), val.into())),
+                _ => env_deps.push((env_dep.into(), "".into())),
+            }
+        }
+    }
+    env_deps
 }
 
 /// Run `rustc --print file-names` to get the outputs of compilation.
@@ -1299,8 +1314,8 @@ where
             .collect::<Vec<_>>();
         // Find all the source files and hash them
         let source_hashes_pool = pool.clone();
-        let source_files_and_hashes = async {
-            let source_files = get_source_files(
+        let source_files_and_hashes_and_env_deps = async {
+            let (source_files, env_deps) = get_source_files_and_env_deps(
                 creator,
                 &crate_name,
                 &executable,
@@ -1311,7 +1326,7 @@ where
             )
             .await?;
             let source_hashes = hash_all(&source_files, &source_hashes_pool).await?;
-            Ok((source_files, source_hashes))
+            Ok((source_files, source_hashes, env_deps))
         };
 
         // Hash the contents of the externs listed on the commandline.
@@ -1323,8 +1338,11 @@ where
         let abs_staticlibs = staticlibs.iter().map(|s| cwd.join(s)).collect::<Vec<_>>();
         let staticlib_hashes = hash_all(&abs_staticlibs, pool);
 
-        let ((source_files, source_hashes), extern_hashes, staticlib_hashes) =
-            futures::try_join!(source_files_and_hashes, extern_hashes, staticlib_hashes)?;
+        let ((source_files, source_hashes, mut env_deps), extern_hashes, staticlib_hashes) = futures::try_join!(
+            source_files_and_hashes_and_env_deps,
+            extern_hashes,
+            staticlib_hashes
+        )?;
         // If you change any of the inputs to the hash, you should change `CACHE_VERSION`.
         let mut m = Digest::new();
         // Hash inputs:
@@ -1374,12 +1392,15 @@ where
         {
             m.update(h.as_bytes());
         }
-        // 7. Environment variables. Ideally we'd use anything referenced
-        // via env! in the program, but we don't have a way to determine that
-        // currently, and hashing all environment variables is too much, so
-        // we'll just hash the CARGO_ env vars and hope that's sufficient.
-        // Upstream Rust issue tracking getting information about env! usage:
-        // https://github.com/rust-lang/rust/issues/40364
+        // 7. Environment variables: Hash all environment variables listed in the rustc dep-info
+        //    output. Additionally also has all environment variables starting with `CARGO_`,
+        //    since those are not listed in dep-info but affect cacheability.
+        env_deps.sort();
+        for &(ref var, ref val) in env_deps.iter() {
+            var.hash(&mut HashToDigest { digest: &mut m });
+            m.update(b"=");
+            val.hash(&mut HashToDigest { digest: &mut m });
+        }
         let mut env_vars: Vec<_> = env_vars
             .iter()
             // Filter out RUSTC_COLOR since we control color usage with command line flags.
--- a/tests/sccache_cargo.rs
+++ b/tests/sccache_cargo.rs
@@ -3,135 +3,198 @@
 //! Any copyright is dedicated to the Public Domain.
 //! http://creativecommons.org/publicdomain/zero/1.0/
 
-#![deny(rust_2018_idioms)]
+use anyhow::{Context, Result};
+use once_cell::sync::Lazy;
+
+use assert_cmd::prelude::*;
+use chrono::Local;
+use predicates::prelude::*;
+use serial_test::serial;
+use std::convert::Infallible;
+use std::ffi::OsString;
+use std::fs;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
 
 #[cfg(all(not(target_os = "windows"), not(target_os = "macos")))]
 #[macro_use]
 extern crate log;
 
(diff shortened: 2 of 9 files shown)


mitchhentges (changes requested):
Looks pretty good, though I'm mostly just marking this as requesting changes for the extra log lines.

Thanks for the patch :)

src/compiler/rust.rs
```diff
@@ -316,6 +317,27 @@ where
     deps
 }
 
+fn parse_env_dep_info(dep_info: &str) -> Vec<(OsString, OsString)> {
```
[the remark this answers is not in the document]
jschwe: Actually, the first pass only reads the first line of the `dep_info` file, so I think from the performance perspective there is no benefit in doing so. Keeping it separate helps readability in my opinion.

tests/sccache_cargo.rs
```diff
@@ -133,5 +142,34 @@ fn test_rust_cargo_cmd(cmd: &str, extra_envs: &[(&str, &std::ffi::OsStr)]) {
         .assert()
         .stdout(predicates::str::contains(r#""cache_hits":{"counts":{"Rust":2}}"#).from_utf8())
         .success();
+
+    if cmd == "run" {
```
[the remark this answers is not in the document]
jschwe: The test for the env vars is now separate. While doing so I refactored basically the whole test and I hope it is easier to add new tests now without duplicating a lot of code.

I removed the deny for rust 2018 idioms from the integration test file, since the test itself should always be built with a cargo version >= the sccache MSRV (which I guessed to be 1.48 by looking at CI). 

Depending on how important the `cfg` excluding windows and macos was, I may also just put everything in the test into a mod.

jschwe:
This might need another round of iterations, but if everything is okay, please ping me so I can rebase and squash before merging.

tests/sccache_cargo.rs
```diff
@@ -64,74 +112,123 @@ fn test_rust_cargo_cmd(cmd: &str, extra_envs: &[(&str, &std::ffi::OsStr)]) {
...
     // compilations.
     trace!("sccache --show-stats");
-    sccache_command()
+    Command::new(SCCACHE_BIN.as_os_str())
+        .args(&["--show-stats", "--stats-format=json"])
+        .assert()
+        .try_stdout(predicates::str::contains(r#""cache_hits":{"counts":{"Rust":2}}"#).from_utf8())?
+        .try_success()?;
+
+    drop(test_info);
```
mitchhentges, line 184: Isn't `test_info` dropped automatically at the end of this function?
If we need to explicitly stop the server to make it shut down between tests, perhaps we should do so explicitly without leaning on `drop()`?

tests/sccache_cargo.rs
```diff
@@ -3,53 +3,101 @@
...
+static SCCACHE_BIN: Lazy<PathBuf> = Lazy::new(|| assert_cmd::cargo::cargo_bin("sccache"));
+static CARGO: Lazy<OsString> = Lazy::new(|| std::env::var_os("CARGO").unwrap());
+static CRATE_DIR: Lazy<PathBuf> =
+    Lazy::new(|| Path::new(file!()).parent().unwrap().join("test-crate"));
 
-    #[cfg(feature = "unstable")]
-    test_rust_cargo_cmd("check", &[("RUSTFLAGS", std::ffi::OsStr::new("-Zprofile"))]);
-    #[cfg(feature = "unstable")]
-    test_rust_cargo_cmd("build", &[("RUSTFLAGS", std::ffi::OsStr::new("-Zprofile"))]);
+/// Used as a test setup fixture. The drop implementation cleans up after a _successfull_ test.
```
mitchhentges, line 26:
```suggestion
/// Used as a test setup fixture. The drop implementation cleans up after a _successful_ test.
```

tests/sccache_cargo.rs
```diff
@@ -3,53 +3,101 @@
...
+/// It is therefore important that the asserts use `Result` to return errors instead of panicking.
+/// Otherwise the temporary directory will not be cleaned up and the started sccache server process
+/// will not be stopped.
```
mitchhentges, line 29: If we need to do post-test cleanup, we can catch the unwind:
```
let test_result = panic::catch_unwind(|| {
    // do majority of the test
});
// do test cleanup
assert!(test_result.is_ok())
```

This will help avoid a footgun for consumers of `SccacheTest`.

tests/sccache_cargo.rs
```diff
@@ -3,53 +3,101 @@
...
+struct SccacheTest<'a> {
+    /// Tempdir used for Sccache cache and cargo output. It is kept in the struct only to have the
+    /// destructor run when SccacheTest goes out of scope, but is never used otherwise.
+    #[allow(dead_code)]
+    tempdir: tempfile::TempDir,
```
mitchhentges, line 34: Perhaps have a `run_cargo()` method on `SccacheTest` that builds the `env` dict so that the compiler can more easily understand the lifecycle of the `tempdir`. As it is now, we still have "references" to the `tempdir` in the form of raw string paths, but this is too abstract for the compiler to track the lifetime automatically.

tests/sccache_cargo.rs
```diff
@@ -64,74 +112,123 @@ fn test_rust_cargo_cmd(cmd: &str, extra_envs: &[(&str, &std::ffi::OsStr)]) {
...
-        .success();
+    stop_sccache()?;
+
+    test_rust_cargo_cmd("check", SccacheTest::new(None)?)
+        .context("Sccache failed for `cargo check`")?;
+    test_rust_cargo_cmd("build", SccacheTest::new(None)?)
+        .context("Sccache failed for `cargo build`")?;
+
+    test_rust_cargo_env_dep(SccacheTest::new(None)?)
+        .context("Sccache failed for `cargo build` with changed environment variables.")?;
```
mitchhentges, line 127: Let's split these into three test functions:
* One checking that a base cargo build has cache hits
* One checking that a cargo `-Zprofile` build has cache hits
* One checking that changing a needed env var has a cache miss.

mitchhentges (changes requested):
Looks pretty good, almost there! I want to avoid `drop()` footguns here, then we should be good :)

tests/sccache_cargo.rs
```diff
@@ -3,53 +3,101 @@
...
-/// when built a second time.
-#[test]
-#[cfg(all(not(target_os = "windows"), not(target_os = "macos")))]
-fn test_rust_cargo() {
-    test_rust_cargo_cmd("check", &[]);
-    test_rust_cargo_cmd("build", &[]);
+static SCCACHE_BIN: Lazy<PathBuf> = Lazy::new(|| assert_cmd::cargo::cargo_bin("sccache"));
+static CARGO: Lazy<OsString> = Lazy::new(|| std::env::var_os("CARGO").unwrap());
+static CRATE_DIR: Lazy<PathBuf> =
+    Lazy::new(|| Path::new(file!()).parent().unwrap().join("test-crate"));
```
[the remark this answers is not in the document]
jschwe: The documentation for `assert_cmd::cargo::cargo_bin` recommends to cache the result for performance reason. I'm not quite sure how up to date this is, but the issue linked from the documentation mentioned something like up to 0.1-0.3 seconds per evaluation.

For the other ones I guess it is not so important and I could just replace the lazy static with a function.

tests/sccache_cargo.rs
jschwe: > Isn't test_info dropped automatically at the end of this function?

Yes. But I was/am not sure if it might even be dropped _before_ the end of the function. I was thinking that by explicitly dropping it at the end, I could prevent it from being dropped after its last usage, since the server still has to be alive for the "get stats" step. I guess I probably should make the "show stats" invocation an associated function of `test_info`, that way the server would be guaranteed to still be alive.

tests/sccache_cargo.rs
jschwe: I actually tried that out, but there is a complication. Cargo test by default invokes multiple tests in parallel - and this doesn't work well with sccache. We already use different directories, but sccache complains "address already in use" when starting a second sccache server.

So you can only test sequentially. There is an option for cargo test to do that, but by default `cargo test` would result in failing tests. 
So we would have to add a bit of extra complexity - E.g. use a Mutex (but keep lock poisining in mind) or an atomic variable to synchronize access and make sure only one test runs at a time, even if multiple are started in parallel.

tests/sccache_cargo.rs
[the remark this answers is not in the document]
mitchhentges: > The documentation for assert_cmd::cargo::cargo_bin recommends to cache the result for performance reason.

Good dig, nice work.
FWIW, I'm assuming that that performance reason only matters if the variable is sometimes unused. Since we're guaranteed to need `SCCACHE_BIN` for this test, we're guaranteed to be running the `Lazy` closure, and therefore it can be inlined I believe.

jschwe:
Just FYI, I haven't gotten to addressing the review comments yet. The force push just now didn't change the end result, I just rebased, squashed and reordered the commits.

tests/sccache_cargo.rs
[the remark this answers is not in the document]
jschwe: > FWIW, I'm assuming that that performance reason only matters if the variable is sometimes unused.

No, this is not the reason. once_cell::Lazy, is initialized on first access (hence lazy), but on any further accesses the value is already available and the closure is not executed. So the performance benefit this is used for is if you are evaluating something multiple times and know that the result can be cached.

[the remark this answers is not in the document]
jschwe: I guess it would be more clear if I used the full path for the type annotation, since only `Lazy` is somewhat ambiguous.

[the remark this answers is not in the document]
mitchhentges: > No, this is not the reason. once_cell::Lazy, is initialized on first access (hence lazy), but on any further accesses the value is already available and the closure is not executed

Right, but without `Lazy`, it's equivalent to the closure work (in our case, the code `assert_cmd::cargo::cargo_bin("sccache")`) being run immediately at runtime. Either way, the value is re-used. By removing `Lazy`, the only functional change that occurs is that the "closure work" happens earlier, IIUC.

So, the only way that `Lazy` would be associated with a performance benefit is if, due to conditional logic, the variable is never accessed, so the closure work is never redundantly resolved.
Since we always access the variable in our tests, this performance benefit is never encountered, right?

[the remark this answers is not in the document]
jschwe: > Right, but without Lazy, it's equivalent to the closure work being run immediately at runtime

I'm not sure I follow. Do you mean something like
```
 static CARGO: String = std::env::var("CARGO").unwrap(); 
```

That doesn't work because
```
error[E0015]: calls in statics are limited to constant functions, tuple structs and tuple variants
```

So I'm not sure if I'm missing something, but I don't see how to avoid using a lazily initialized variable for a global that can't be initialized at compile time.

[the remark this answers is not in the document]
mitchhentges: Oh, I see, I was imagining something like:
```suggestion
const SCCACHE_BIN: PathBuf = assert_cmd::cargo::cargo_bin("sccache");
const CARGO: OsString = std::env::var_os("CARGO").unwrap();
const CRATE_DIR: PathBuf = Path::new(file!()).parent().unwrap().join("test-crate");
```

But, it looks like quite a few of these functions aren't `const`-able yet.
Ok, let's carry on with `Lazy`, my mistake :)

tests/sccache_cargo.rs
jschwe: For now I just immediately unwrap again after `catch_unwind`, since the `drop` for the `SccacheTest` would have already run then. The unwrapping returns the result of the test, which is maybe a bit unintuitive.

tests/sccache_cargo.rs
jschwe: Thanks for the suggestion - `serial_test` worked just fine.

tests/sccache_cargo.rs
jschwe: I added the `show_stats()` impl on the SccacheTest struct, so that the struct is used when running `sccache --show-stats`. 

I double checked, and [the reference](https://doc.rust-lang.org/reference/destructors.html#scopes-of-function-parameters) guarantees that function parameters are dropped last, at the very end of the function. Struct members also can't be dropped prematurely, so I would prefer to not complicate things and just keep it as is.

jschwe:
Changes are in commits since  [fixup: Rename TEST_ENV_VAR](https://github.com/mozilla/sccache/pull/1107/commits/b0f162d96d565a8536cea7a6916a9752701d6363). I'd rebase and squash again after review.

tests/sccache_cargo.rs
```diff
@@ -37,6 +53,10 @@ struct SccacheTest<'a> {
 
 impl SccacheTest<'_> {
     fn new(additional_envs: Option<&[(&'static str, std::ffi::OsString)]>) -> Result<Self> {
+        assert!(LOGGER.is_ok());
```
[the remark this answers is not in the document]
jschwe: This one is a bit tricky and maybe there is a better way to do that. 
We need to initialize the logger exactly once. Running `LOGGER.is_ok()` actually initializes the logger the first time. The `is_ok()` has to be used so I just put it into the assert.

tests/sccache_cargo.rs
jschwe: Now that I think it again, the `drop` actually runs anyway if `panic=unwind` and will clean up the directory. And if `panic=abort`, then the catch_unwind will also not work so I'm just going to remove it again, since it doesn't help.

mitchhentges:
Looks good, thanks!

PR #992, opened 2021-04-28
Fix OAuth token format

georgewfraser:
The current code expects SCCACHE_GCS_CREDENTIALS_URL to return an OAuth token in a non-standard format

```json
{"accessToken": "...", "expireTime": "2000-01-01T00:00:00Z"}
```

The standard format, which is returned by the Google Cloud auth endpoint, looks like:

```json
{"access_token": "...", "expires_in": 123}
```

This PR deprecates the non-standard format and adds a new environment variable option SCCACHE_GCS_OAUTH_URL. By conforming to the standard, it is possible to use the built-in OAuth endpoint that is available on every Google Cloud instance

    http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/${YOUR_SERVICE_ACCOUNT}/token

Fixes #870

PR #1108, opened 2022-02-05
GCP: Fix OAuth token format

Hakuyume:
Since #992 seems inactive and I need `SCCACHE_GCS_OAUTH_URL` feature, I just resolved its conflicts.


sylvestre:
Seems that it broke some tests

Issue #1019, opened 2021-07-11
Please provide a new release


Be-ing:
#962 was merged 3 months ago and is not yet included in a release. Because of this, Mixxx CI has to build sccache from source which takes 10 minutes. It is caching the sccache binary for subsequent runs, but the first run of each branch has this 10 minute penalty. This could be avoided if we could install sccache from Chocolatey, which requires a new release.

Be-ing:
ping

sylvestre:
Yeah, we will do it soon

ajschmidt8:
@sylvestre, any updates on the new release?

milahu:
> Mixxx CI has to build sccache from source which takes 10 minutes. It is caching the sccache binary for subsequent runs, but the first run of each branch has this 10 minute penalty.

sounds like a bug

cant you store the binary "somewhere"? amazon, github, cachix, ...

dsanders11:
@sylvestre, @milahu, any update on new releases? Anything blocking it that someone else could work? From an outsider's perspective the project has entirely given up doing new releases, since it's been over a year.

mitchhentges:
Hey David, I'm currently chipping away at the backlog of open PRs to see what can be landed before the next release.

> Anything blocking it that someone else could work?

Mostly reviews, there's a few thousands of lines of changed code that needs to be looked through. I'm getting up-to-speed on the project as well.

-----

However, something that _would_ be sweet is if we could get the dependencies updated, as it's been a hot minute. There were some patches doing so from a year ago, but if that could be done I'd be pretty happy. While we're at it, tackling modern compiler warnings and clippy lints will be important as well.

dsanders11:
@mitchhentges, thanks for the response! Happy to hear it's being worked on. I haven't built the project and have no Rust experience, but if you point me in the direction of specific PRs for dependencies, or issues for modern compiler warnings, I could take a look when I get a chance.

I'm mainly itching for a release as I've run into #1098 several times and it might be fixed in `master`, there are just no releases since then.

mitchhentges:
The two specific issues I have in mind are #1110 and #1111

> I haven't built the project and have no Rust experience

This'll probably be best picked up by more experienced Rust devs since there may be breaking changes that require refactoring.

Be-ing:
I encourage you not to hold back releases for "one more little thing", especially for not for trivial code cleanup. I've been waiting on a release for around a year now. I don't care about Clippy warnings.

[...]

PR #1109, opened 2022-02-08
Support Azure Blob Storage key prefixes

FooBarWidget:
Similar to how we support key prefixes in S3 and Google Cloud Storage.

Introduces `SCCACHE_AZURE_KEY_PREFIX`.

### DIFF
--- a/src/cache/azure.rs
+++ b/src/cache/azure.rs
@@ -29,18 +29,7 @@ pub struct AzureBlobCache {
 }
 
 impl AzureBlobCache {
-    pub fn new(credentials: Option<AzureCredentials>, key_prefix: &str) -> Result<AzureBlobCache> {
-        let credentials = match credentials {
-            Some(creds) => creds,
-            None => match EnvironmentProvider.provide_credentials() {
-                Ok(creds) => creds,
-                Err(err) => bail!(
-                    "Could not find Azure credentials in the environment: {}",
-                    err
-                ),
-            },
-        };
-
+    pub fn new(credentials: AzureCredentials, key_prefix: &str) -> Result<AzureBlobCache> {
         let container = match BlobContainer::new(
             credentials.azure_blob_endpoint(),
             credentials.blob_container_name(),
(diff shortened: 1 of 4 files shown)


src/cache/azure.rs
```diff
@@ -25,16 +25,20 @@ use crate::errors::*;
...
 
 impl AzureBlobCache {
-    pub fn new() -> Result<AzureBlobCache> {
-        let credentials = match EnvironmentProvider.provide_credentials() {
-            Ok(creds) => creds,
-            Err(err) => bail!(
-                "Could not find Azure credentials in the environment: {}",
-                err
-            ),
+    pub fn new(credentials: Option<AzureCredentials>, key_prefix: &str) -> Result<AzureBlobCache> {
```
mitchhentges, line 32: Good call on making `credentials` be provided by the consumer to make your tests possible to write.

It's a little funky seeing the `None` get provided into `AzureBlobCache` in `storage_from_config()` down below: perhaps it's worth trying one of the following two options:
1. Make `credentials` mandatory, and move the default `EnvironmentProvider.provide_credentials()` up to the caller
2. (My preferred option): Make two separate functions here: `pub fn new(credentials: AzureCredentials, key_prefix: &str)` and `pub fn new_with_environment_credentials(key_prefix: &str)`.

Let me know what you think :)

src/cache/azure.rs
```diff
@@ -70,18 +84,27 @@ impl Storage for AzureBlobCache {
...
             .await
             .context("Failed to put cache entry in Azure")?;
 
         Ok(start.elapsed())
     }
 
     fn location(&self) -> String {
-        format!("Azure, container: {}", self.container)
+        format!(
+            "Azure, container: {}, {}",
```
mitchhentges, line 100: Let's match the format used by GCS.
```suggestion
            "Azure, container: {}, key_prefix: {}",
```

mitchhentges (approved):
Nice work, and thanks for the contribution!
I'll let this sit for a day or two (or until your response + potential update), then land afterwards.
Cheers :beers:

src/cache/azure.rs
FooBarWidget: I'm fine with either, so I'll implement the second.

FooBarWidget: ~I'm fine with either, so I'll implement the second.~
On second thought, I think the first one is better going forward. The problem with the second one is that, as more options get introduced for AzureStorage, you need to duplicate all parameters twice.

mitchhentges (approved):
Great, thanks again!

mitchhentges:
Working towards it, see #1019!

PR #1119, opened 2022-02-14
Fix clippy lints for dist-server

jschwe:
With the dist-server feature enabled, additional clippy
warnings are emitted. Those are fixed by this commit.
Lints were:
- unneccessary borrow
- string != "" should be !string.is_empty()
- unwrap_or followed by a function call -> unwrap_or_else


mitchhentges (approved):
Nice, thanks!

PR #1123, opened 2022-02-18
GCC: suppress rewrite_includes_only with pedantic flags

mistmist:
CoinMP/CoinUtils configure checks the <cmath> header:
```

configure:9903: sccache g++ -c -O3 -pipe -DNDEBUG -pedantic-errors -Wparentheses -Wreturn-type -Wcast-qual -Wall -Wpointer-arith -Wwrite-strings -Wconversion -Wno-unknown-pragmas -Wno-long-long   -DCOINUTILS_BUILD  conftest.cc >&5
<built-in>: error: ISO C++ does not support '__int128' for '__x' [-Wpedantic]
<built-in>: error: ISO C++ does not support '__int128' for 'abs' [-Wpedantic]
In file included from /usr/include/c++/10/cmath:42,
                 from conftest.cc:64:
<built-in>: error: ISO C++ does not support '__int128' for 'type name' [-Wpedantic]

```
The problem is that in GCC 10, the bits/std_abs.h declares abs() for
GCC extension __int128, which -pedantic then complains about.

Implement the same solution that icecream has for this problem: for GCC,
if any -pedantic etc. flag is given and GNU extensions are enabled, don't
use -fdirectives-only.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -96,6 +96,8 @@ pub struct ParsedArguments {
     pub profile_generate: bool,
     /// The color mode.
     pub color_mode: ColorMode,
+    /// arguments are incompatible with rewrite_includes_only
+    pub suppress_rewrite_includes_only: bool,
 }
 
 impl ParsedArguments {
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -55,6 +55,7 @@ impl CCompilerImpl for Clang {
             cwd,
             (&gcc::ARGS[..], &ARGS[..]),
             self.clangplusplus,
+            self.kind(),
         )
     }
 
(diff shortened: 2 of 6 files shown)


sylvestre:
it would be nice to have a test for this to make sure we don't regress. thanks

mitchhentges (approved):
Makes sense to me, thanks!

mitchhentges:
Heads up that you have a CI failure here with the tests, which we should resolve before landing.
(It's been a while since I've used GitHub CI - do you get emailed/notified automatically?)

mitchhentges:
This doesn't appear to happen [on current `main`](https://github.com/mozilla/sccache/runs/5409056149?check_suite_focus=true) - you can try rebasing to see if it's an issue with your current base commit.

Additionally, it could be worth installing the current Rust nightly and trying to reproduce locally - since it's happened on multiple machines, it should hopefully be relatively reproducible.

mistmist:
in the rebase there was a commit edf0ac517650c7b230d933ec4a0947c103dd7647 that changed massively the tests that failed, maybe it helps...

mitchhentges:
Nice, looks good. Thanks!

PR #1124, opened 2022-02-18
GCC: fix handling of -MD/-MMD without explicit -MF

mistmist:
Building Firebird fails running sed on non-existent dependency files:

sed: can't read workdir/UnpackedTarball/firebird/temp/Debug/gpre/c_cxx.d: No such file or directory

The problem is that the file was written to the wrong location:
workdir/UnpackedTarball/firebird/gen/c_cxx.d

Implement what the fine manual says:
    The driver determines file based on whether an -o option is given.
    If it is, the driver uses its argument but with a suffix of .d,

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -233,6 +233,12 @@ where
     let mut multiple_input = false;
     let mut split_dwarf = false;
     let mut need_explicit_dep_target = false;
+    enum DepArgumentRequirePath {
+        NotNeeded,
+        Missing,
+        Provided,
+    }
+    let mut need_explicit_dep_argument_path = DepArgumentRequirePath::NotNeeded;
     let mut language = None;
     let mut compilation_flag = OsString::new();
     let mut profile_generate = false;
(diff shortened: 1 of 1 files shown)


src/compiler/gcc.rs
```diff
@@ -293,13 +294,18 @@ where
...
+                    need_explicit_dep_argument_path = Some(true);
+                }
+            }
             Some(DepTarget(s)) => {
                 dep_flag = OsString::from(arg.flag_str().expect("Dep target flag expected"));
                 dep_target = Some(s.clone());
             }
-            Some(DepArgumentPath(_))
-            | Some(ExtraHashFile(_))
+            Some(DepArgumentPath(_)) => need_explicit_dep_argument_path = Some(false),
```
mitchhentges, line 307: Nit: `need_explicit_dep_argument_path` might be better represented as an enum:
```
enum DepArgumentPathRequirement {
    NotNeeded,
    Needed,
    AlreadyProvided,
}
```
(Feel free to take an artistic liberty to the enum value names, I'm not sure that my suggestion here is the ideal naming 😉)

src/compiler/gcc.rs
```diff
@@ -479,6 +485,10 @@ where
         dependency_args.push(dep_flag);
         dependency_args.push(dep_target.unwrap_or_else(|| output.clone().into_os_string()));
     }
+    if let Some(true) = need_explicit_dep_argument_path {
+        dependency_args.push(OsString::from("-MF"));
+        dependency_args.push(Path::new(&output).with_extension("d").into_os_string());
+    }
```
mitchhentges, line 491: Does this behave properly in conjunction with `-E`?
I'm guessing it does because we don't cache `-E` due to it being a `TooHardFlag`, eh?

mistmist: looks like TooHardFlag will cause the sccache client to run the command directly, so it doesn't matter.

mistmist:
is anything still missing here @mitchhentges ? it has been sitting approved for weeks

mitchhentges:
Nope, this looks good!
I'm going to land this after the `v0.3.0` release ([which is hitting some snags](https://github.com/mozilla/sccache/issues/1150)), after which I'll land this and a few other sibling PRs that are also waiting for post-0.3.0 landings.

The next release won't be too long after `v0.3.0`, so this PR will hit "stable" soon :)

PR #1125, opened 2022-02-18
C/C++: add extra input files via environment SCCACHE_EXTRAFILES

mistmist:
This is similar to CCACHE_EXTRAFILES and ICECC_EXTRAFILES: the extra
files are hashed for caching and packaged for sccache-dist.

Possibly the same variable CCACHE_EXTRAFILES could be used instead of
a distinct one?

This can be implemented in one place for the C/C++ compilers, but not
also for Rust, which currently doesn't have extra_hash_files, so that
remains TODO.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -252,14 +252,22 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
         &self,
         arguments: &[OsString],
         cwd: &Path,
+        env_vars: &[(OsString, OsString)],
     ) -> CompilerArguments<Box<dyn CompilerHasher<T> + 'static>> {
         match self.compiler.parse_arguments(arguments, cwd) {
-            CompilerArguments::Ok(args) => CompilerArguments::Ok(Box::new(CCompilerHasher {
-                parsed_args: args,
-                executable: self.executable.clone(),
-                executable_digest: self.executable_digest.clone(),
-                compiler: self.compiler.clone(),
-            })),
+            CompilerArguments::Ok(mut args) => {
+                for (k, v) in env_vars.iter() {
+                    if k.as_os_str() == OsStr::new("SCCACHE_EXTRAFILES") {
+                        args.extra_hash_files.extend(std::env::split_paths(&v))
+                    }
+                }
+                CompilerArguments::Ok(Box::new(CCompilerHasher {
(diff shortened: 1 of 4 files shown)


mitchhentges:
> Possibly the same variable CCACHE_EXTRAFILES could be used instead of
a distinct one?

If `ICECC_EXTRAFILES` is already starting the convention of having cache-implementation-specific `EXTRAFILES` variables, I'm ok with continuing that tradition :+1: 
It might make sense to (also?) respect `CCACHE_EXTRAFILES` in the future, but I don't think that it should block this PR.

src/compiler/c.rs
```diff
@@ -252,14 +252,22 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
...
-            CompilerArguments::Ok(args) => CompilerArguments::Ok(Box::new(CCompilerHasher {
-                parsed_args: args,
-                executable: self.executable.clone(),
-                executable_digest: self.executable_digest.clone(),
-                compiler: self.compiler.clone(),
-            })),
+            CompilerArguments::Ok(mut args) => {
+                for (k, v) in env_vars.iter() {
+                    if k.as_os_str() == OsStr::new("SCCACHE_EXTRAFILES") {
+                        args.extra_hash_files.extend(std::env::split_paths(&v))
```
mitchhentges, line 261: `std::env::split_paths` is really useful, nice find :+1:

src/compiler/compiler.rs
```diff
@@ -1348,7 +1349,7 @@ LLVM version: 6.0",
                     &creator,
                     Ok(MockChild::new(exit_status(0), "preprocessor output", "")),
                 );
-                let hasher = match c.parse_arguments(&arguments, ".".as_ref()) {
+                let hasher = match c.parse_arguments(&arguments, ".".as_ref(), vec![].as_ref()) {
```
mitchhentges, line 1352: Is it possible to do `&[]` instead?

mitchhentges (approved):
I think this makes sense.
FWIW, can you describe a case where you need to use this environment variable? Is it a case where `sccache` is failing to understand a dependency relationship?

mistmist:
the use case is that our custom clang plugin wants to read a text file from a hard-coded path.

mitchhentges:
Thanks for the patch mistmist!

PR #1126, opened 2022-02-18
improve sccache-dist handling of paths containing ".."

mistmist:


src/compiler/c.rs
```diff
@@ -482,20 +482,35 @@ impl pkg::InputsPackager for CInputsPackager {
...
-
+            let (input_path, dirs) = pkg::simplify_path(&input_path)?;
+            for dir in dirs {
+                let dist_dir = path_transformer
+                    .as_dist(&dir)
+                    .with_context(|| format!("unable to transform input dir {}", dir.display()))?;
+                let mut dir_header = pkg::make_tar_header(&dir, &dist_dir, true)?;
+                dir_header.set_cksum();
+                builder.append(&dir_header, "".as_bytes())?;
+            }
```
mitchhentges, line 513: Let's re-use this chunk

mitchhentges (changes requested):
OTOH, it feels like the stronger solution here would be to resolve `..` in paths up-front. This would give us a couple benefits:
* It would resolve the "`..` requires parent dir to exist" issue that you've raised
* Also, it would allow more cache hits, if different paths eventually resolved to the same file.

What do you think?

mistmist:
> OTOH, it feels like the stronger solution here would be to resolve `..` in paths up-front. This would give us a couple benefits:
> 
>     * It would resolve the "`..` requires parent dir to exist" issue that you've raised

that may be true for the compiler command line patch, but for the patch where one part of the toolchain invokes another with such a path it cannot help as that is beyond sccache's control.

>     * Also, it would allow more cache hits, if different paths eventually resolved to the same file.

i think this is unlikely, because why would you compile a file twice with the only difference being the path to the file? i'd expect there to be other differences in the command line that would prevent a cache hit.

mitchhentges:
> i think this is unlikely, because why would you compile a file twice with the only difference being the path to the file? i'd expect there to be other differences in the command line that would prevent a cache hit.

Perhaps in the `sccache-dist` use case, and different machines have different local config, one of which uses a relative path with a `..`. I'm guessing here, this might not be a valid example :)

mistmist:
yay it's green again :) can somebody merge this please?

src/compiler/c.rs
```diff
@@ -521,20 +521,21 @@ impl pkg::InputsPackager for CInputsPackager {
...
             file_header.set_size(preprocessed_input.len() as u64); // The metadata is from non-preprocessed
             file_header.set_cksum();
             builder.append(&file_header, preprocessed_input.as_slice())?;
         }
 
         for input_path in extra_hash_files {
-            let input_path = pkg::simplify_path(&input_path)?;
-
+            let (input_path, dirs) = pkg::simplify_path(&input_path)?;
+            assert!(dirs.is_empty()); // must be simplified already
```
drahnr, line 538: This `assert` is a potential footgun, a warning or `let (a,_) = ..` would be more appropriate.

drahnr (approved):

src/compiler/c.rs
mistmist: in case "dirs" is non-empty here, the compiler invocation will fail (with sccache-dist) due to these directories not existing.

this is in implementation of InputsPackager, which is only run if "dist-client" is enabled.

given the inevitable failure, the assertion failure is preferable to a more easily ignored warning.

the intention (and i hope current situation in this PR) is that the simplify_path is already done by code that runs previously (in "gcc.rs"), and so i have removed the code that would handle a non-empty "dirs" in "c.rs".

this was to implement the idea by mitchhentges in this previous comment https://github.com/mozilla/sccache/pull/1126#pullrequestreview-887421393

there is another such assert in "rust.rs" - there the situation is different, in that there is no earlier call to simplify_path, but it is the same in that the compiler invocation will fail because the directory doesn't exist - already the case before this PR.

PR #1133, opened 2022-02-19
Add dependency caching to CI builds

emabrey:
Use Swatinem/rust-cache to speed up CI builds by caching durable dependencies.

Signed-off-by: Emily Mabrey <emilymabrey93@gmail.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           components: ${{ matrix.component }}
           toolchain: 1.58.1
+          
+      - name: Activate Rust dependency caching
+        uses: Swatinem/rust-cache@v1.3.0  
 
       - name: clippy version
         run: cargo clippy --version
@@ -75,6 +78,9 @@ jobs:
         uses: ./.github/actions/rust-toolchain
         with:
           toolchain: ${{ matrix.rustc }}
+          
+      - name: Activate Rust dependency caching
+        uses: Swatinem/rust-cache@v1.3.0  
 
       - name: Build tests
         run: cargo test --no-run --locked --all-targets --verbose ${{ matrix.extra_args }}
@@ -137,6 +143,9 @@ jobs:
           sed 2iREALGCC=aarch64-linux-gnu-gcc /usr/bin/musl-gcc | sudo tee /usr/bin/aarch64-linux-musl-gcc > /dev/null
           sudo chmod +x /usr/bin/aarch64-linux-musl-gcc
         if: ${{ matrix.target == 'aarch64-unknown-linux-musl' }}
+        
+      - name: Activate Rust dependency caching
+        uses: Swatinem/rust-cache@v1.3.0  
 
       - name: Build
         run: cargo build --locked --release --verbose --bin ${{ matrix.binary || 'sccache' }} --target ${{ matrix.target }} --features=openssl/vendored ${{ matrix.extra_args }}


sylvestre:
Looks like, we, Mozilla, have some limitations that prevent us to use it

`swatinem/rust-cache@v1.3.0, swatinem/rust-cache@v1.3.0, and swatinem/rust-cache@v1.3.0 are not allowed to be used in mozilla/sccache. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub or match the following: !/mozilla/**, !mozilla/**, ./**, 10up/wpcs-action@*, aws-actions/*, docker/*, pypa/gh-action-pypi-publish@v1.4.2, slackapi/slack-github-action@*, codecov/codecov-action@v2.`

sylvestre:
In general, we don't accept GH actions from individuals. Maybe @swatinem could move this github actions under the https://github.com/getsentry umbrella?

emabrey:
I'll try to reach out to them or see if there is some way to workaround those restrictions. I was worried about the individual control already, that's why I locked it down to a specific version tag to stop automatic updates potentially being a security issue.

I got some metrics though.

|----------------   | ----------|
|----------------   |  ---------|
|Before Caching  | 14m 32s |
| After Caching    | 09m 05s |
| Savings              | 05m 27s | 
| Percent Change |   -37.5%  |

emabrey:
I removed the checkout version stuff from this PR since it seems that this PR is likely to sit for a lot longer than if I split the changes.

emabrey:
Would hosting a copy/fork of the action inside the [tenacityteam organization account](https://github.com/tenacityteam/) work? I can make a fork there if that would work around the organization/individual account issue.

Swatinem:
I think creating a fork in this case is fair. As per licensing, reminds me that I never wrote that down in the README. though the `package.json` mentions the lgpl: https://github.com/Swatinem/rust-cache/blob/3bb3a9a087029c7bc392586cdc88cb6f66b9c6ef/package.json#L12

Now that I think about it, straight gpl would have been a better fit.

sylvestre:
I would prefer a more well known org (like sentry, as mozilla already uses their products)

luser:
There's something a bit funny about the CI for `sccache`, a compiler caching tool, using an alternative caching strategy. :)

emabrey:
> There's something a bit funny about the CI for `sccache`, a compiler caching tool, using an alternative caching strategy. :)

I suppose that is a bit strange. The main thing that the action solves that using sccache directly doesn't (I believe, I could be wrong) is providing logic for how to determine which Rust crates you can toss as transient (CI ephemera) and which ones are durable (i.e. dependencies).

FooBarWidget:
Idea: we can add an sccache storage backend for the [Github Actions cache API](https://github.com/tonistiigi/go-actions-cache/blob/master/api.md) and then use sccache to cache the sccache CI workflow. :)

FooBarWidget:
Idea: we can add an sccache storage backend for the [Github Actions cache API](https://github.com/tonistiigi/go-actions-cache/blob/master/api.md) and then use sccache to cache the sccache CI workflow. :)

This is not merely useful for the sccache CI. It's also useful for other C/C++/Rust projects because then:

 - They don't have to setup their own cloud storage in order to sccache their compilations. No need to storage cloud storage access credentials in secrets.
 - Any forkers don't have to setup *their* own cloud storage either in order to make their forks' CI working.

Be-ing:
> add an sccache storage backend for the [Github Actions cache API](https://github.com/tonistiigi/go-actions-cache/blob/master/api.md)

I really like this idea. Currently using sccache with GitHub Actions caching requires caching a platform-specific directory, which is a bit inconvenient.

sylvestre:
As builds are fast, I don't think we need it for now 
Sorry

PR #1134, opened 2022-02-19
Update actions/checkout from v1 to v2

emabrey:
Extremely simple commit to update the version of the checkout action in Github CI runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
             cargo_cmd: clippy --locked --all-targets -- -D warnings -A unknown-lints -A clippy::type_complexity -A clippy::new-without-default
     steps:
       - name: Clone repository
-        uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


emabrey:
I'm a bit confused on how changing a CI config file modified the code coverage percentages...  I guess there is some kind of non-determinism at play like some kind of optimizer or something. Whatever, I'll take a 0.12% increase I guess 😎 .

mitchhentges:
Makes sense to me, thanks!

emabrey:
Between the making of this commit and the merge of it, they released v3 of the plugin. I plan to leave it at v2 though for now. The v3 plugin changeset only seems to include an update to use a more recent version of the Node runtime, which isn't exactly worth it to go bleeding edge.

mitchhentges:
Yeah, makes sense, I saw that v3 was released a few hours before I landed this PR.
I don't think that there's harm in letting it mature a bit before we adopt it.

PR #1136, opened 2022-03-03
Remove outdated badges from Cargo.toml

jschwe:
- travis-ci and appveyor don't seem to be used anymore.
- The cargo reference now recommends putting badges into
  the Readme file, where we already a github CI badge.


mitchhentges:
Nice cleanup, thanks!

PR #1138, opened 2022-03-06
More complete unit tests for Azure Blob Storage backend

FooBarWidget:
Pull request #1109 introduced key prefix support for the Azure backend, with unit tests for `normalize_key()`, but those tests didn't test whether the Storage impl *actually* used `normalize_key()` when constructing an API request to Azure.

This change makes the unit tests more complete, by testing the Storage impl directly through the use of a mock HTTP server as Azure API endpoint.

### DIFF
--- a/src/azure/blobstore.rs
+++ b/src/azure/blobstore.rs
@@ -46,30 +46,39 @@ fn md5(data: &[u8]) -> String {
     base64::encode_config(&digest.finalize(), base64::STANDARD)
 }
 
-pub struct BlobContainer {
+#[async_trait]
+pub trait BlobContainer: fmt::Display + Send + Sync {
+    async fn get(&self, key: &str, creds: &AzureCredentials) -> Result<Vec<u8>>;
+    async fn put(&self, key: &str, content: Vec<u8>, creds: &AzureCredentials) -> Result<()>;
+}
+
+pub struct HttpBlobContainer {
     url: String,
     client: Client,
 }
 
-impl fmt::Display for BlobContainer {
+impl fmt::Display for HttpBlobContainer {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "BlobContainer(url={})", self.url)
     }
 }
 
-impl BlobContainer {
-    pub fn new(base_url: &str, container_name: &str) -> Result<BlobContainer> {
+impl HttpBlobContainer {
+    pub fn new(base_url: &str, container_name: &str) -> Result<Self> {
         assert!(
             base_url.ends_with('/'),
             "base_url is assumed to end in a trailing slash"
         );
-        Ok(BlobContainer {
+        Ok(Self {
             url: format!("{}{}/", base_url, container_name),
             client: Client::new(),
         })
     }
+}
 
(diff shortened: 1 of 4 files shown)


FooBarWidget:
Fixed all CI failures.

mitchhentges (changes requested):
These tests look good, especially since this adds some great coverage for our `BlobContainer<=>Azure HTTP API` integration.

However, I think that such tests should be assocated with `BlobContainer` itself - up at this level in `cache::azure`, we're more like _consumers_ of `BlobContainer`: we shouldn't depend on the internal behaviour of this component from another crate. This structural restriction gets a little fuzzier since we're dealing with tests, but I think we have another option that will be the best of both worlds:
1. Move the HTTP-integration tests to `src/azure/blobstore.rs`
2. Change `BlobContainer` to be a trait, rename the existing `BlobContainer` to be an `HttpBlobContainer`, then define a `MockBlobContainer` for your tests. Write your `MockBlobContainer` to return dummy data on `.get()`, but have it record the `key` parameter, then do `assert_eq!(mock_container.last_key, Some("prefix"))` (or something like that).

This way, we keep the coverage, but if we start using a newer Azure API, then the caching layer built _on top_ of the Azure abstraction structs doesn't need to update its tests.

What do you think?

FooBarWidget:
@mitchhentges Sounds good. Here it is.

Note: CI is failing because of network problem. You need to restart the CI.

src/azure/blobstore.rs
```diff
@@ -332,4 +344,73 @@ mod test {
...
+        let body = b"hello".to_vec();
+        Mock::given(method("PUT"))
+            .and(path("/container/foo/bar"))
+            .and(body_bytes(body.clone()))
+            .respond_with(ResponseTemplate::new(200))
+            .expect(1)
+            .mount(&server)
+            .await;
+
+        container.put("foo/bar", body, &credentials).await?;
```
mitchhentges, line 412: Is it possible to check that the data that `Mock` got matches `body`?

mitchhentges (approved):
I'm happy to land this as-is, but if you wouldn't mind addressing my nits here that'd be just a little better :)

Thanks for the patch!

src/cache/azure.rs
```diff
@@ -104,40 +106,138 @@ impl Storage for AzureBlobCache {
...
+        pub fn get_last_key(&self) -> String {
+            let cell_guard = self.last_key.lock().unwrap();
+            let last_key = cell_guard.borrow();
+            last_key.clone()
+        }
+
+        pub fn set_last_key(&self, key: String) {
+            let last_key = self.last_key.lock().unwrap();
+            last_key.replace(key);
+        }
```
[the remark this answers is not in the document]
FooBarWidget: BlobContainer must be Sync. RefCell makes it not Sync so I need a Mutex to make it Sync again.

src/azure/blobstore.rs
FooBarWidget: Line 406 already does that.

mitchhentges: Ah, cool :+1:

mitchhentges:
Looks good, thanks!
The `aarch64` CI job is failing in `main` as well (:crossed_fingers: that it's an environmental intermittent), so I'm happy to land this.

Thanks for the patch!

Issue #1110, opened 2022-02-11
Update dependencies


mitchhentges:
Running [`cargo outdated -R`](https://github.com/kbknapp/cargo-outdated) shows:
```
Name             Project           Compat  Latest             Kind         Platform
----             -------           ------  ------             ----         --------
ar               0.8.0             ---     0.9.0              Normal       ---
assert_cmd       1.0.8             ---     2.0.4              Development  ---
blake3           0.3.8             ---     1.3.1              Normal       ---
clap             2.34.0            ---     3.0.14             Normal       ---
directories      3.0.2             ---     4.0.1              Normal       ---
env_logger       0.8.4             ---     0.9.0              Normal       ---
futures-locks    0.6.0             ---     0.7.0              Normal       ---
hmac             0.10.1            ---     0.12.0             Normal       ---
hyperx           0.13.2            ---     1.4.0              Normal       ---
jsonwebtoken     7.2.0             ---     8.0.1              Normal       ---
md-5             0.9.1             ---     0.10.0             Normal       ---
nix              0.19.1            ---     0.23.1             Normal       ---
predicates       1.0.8             ---     2.1.1              Development  ---
rand             0.7.3             ---     0.8.4              Normal       ---
sha-1            0.9.8             ---     0.10.0             Normal       ---
sha2             0.9.9             ---     0.10.1             Normal       ---
syslog           5.0.0             ---     6.0.1              Normal       ---
tokio-util       0.6.9             ---     0.7.0              Normal       ---
untrusted        0.7.1             ---     0.9.0              Normal       ---
version-compare  0.0.11            ---     0.1.0              Normal       ---
zstd             0.6.1+zstd.1.4.9  ---     0.10.0+zstd.1.5.2  Normal       ---
```

Let's bump these dependencies.
There'll probably be some breaking changes, so Rust experience will likely be necessary as part of this work.

jschwe:
I would take a look at this. What is the policy about bumping the MSRV, if updating dependencies bumps their MSRV above ours?

jschwe:
I created a PR (#1113) for the easy dependencies. I have ignored the crypto crates, clap and tokio for now, since I expect more work there.

The largest open question for me is: How far can we bump the MSRV? Is there a hard requirement due to packaging for some distribution?

A short check of the crates not covered by my PR:
[...]

PR #1140, opened 2022-03-11
Update clap to v3 1

LovecraftianHorror:
This is my initial attempt at updating `clap` to v3.1 (and switching over to the `Derive` API for it)

I kept the same external `Command`s for the parsed format and just used intermediate values internally. To keep the separation cleaner this involved changing `parse()` to be infallible where the error displaying and exiting early is now done internally within `parse()`

`sccache` appears to be fine and tests pass with `cargo test` for it. `sccache-dist` is compiling and from a surface level check things seem to be fine, but the tests for it don't pass for me on `main` or this branch. This PR is marked as a draft until that gets sorted out. Below is an example of the error I get:

```text
---- test_dist_nobuilder stdout ----
thread 'test_dist_nobuilder' panicked at 'called `Result::unwrap()` on an `Err` value: AddrParseError(())', tests/harness/mod.rs:488:42
LOGS == (exit status: 0) ==
> sccache_dist_test_scheduler_314eee14-4948-4e30-abf4-685adf5e766a <:
## STDOUT


## STDERR
/sccache-dist: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /sccache-dist)
/sccache-dist: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /sccache-dist)
/sccache-dist: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /sccache-dist)
/sccache-dist: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /sccache-dist)
/sccache-dist: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /sccache-dist)

====
OUTPUTS == (exit status: 1) ==
> sccache_dist_test_scheduler_314eee14-4948-4e30-abf4-685adf5e766a <:
## STDOUT


## STDERR
Error response from daemon: Cannot kill container: sccache_dist_test_scheduler_314eee14-4948-4e30-abf4-685adf5e766a: Container 3681cd393c6d9e3a7a50674313ac76356f3496b3688785250c5a5dc75a06360d is not running

====
OUTPUTS == (exit status: 0) ==
> sccache_dist_test_scheduler_314eee14-4948-4e30-abf4-685adf5e766a <:
## STDOUT
sccache_dist_test_scheduler_314eee14-4948-4e30-abf4-685adf5e766a


## STDERR

====


failures:
    test_dist_nobuilder
```

Also just for comparison here is a comparison of the new and old help messages

_`sccache` old_

```
sccache 0.2.15

USAGE:
    sccache [FLAGS] [OPTIONS] [cmd]...

FLAGS:
        --dist-auth       authenticate for distributed compilation
        --dist-status     show status of the distributed client
    -h, --help            Prints help information
    -s, --show-stats      show cache statistics
        --start-server    start background server
        --stop-server     stop background server
    -V, --version         Prints version information
    -z, --zero-stats      zero statistics counters

OPTIONS:
        --package-toolchain <executable> <out>    package toolchain for distributed compilation
        --stats-format <stats-format>
            set output format of statistics [default: text]  [possible values: text, json]


ARGS:
    <cmd>...

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```

_`sccache` new_

```
sccache 0.2.16-alpha.0

USAGE:
    sccache [OPTIONS] [--] [CMD]...

ARGS:
    <CMD>...

OPTIONS:
        --dist-auth
            authenticate for distributed compilation

        --dist-status
            show status of the distributed client

    -h, --help
            Print help information

        --package-toolchain <PACKAGE_TOOLCHAIN> <PACKAGE_TOOLCHAIN>
            package toolchain for distributed compilation

    -s, --show-stats
            show cache statistics

        --start-server
            start background server

        --stats-format <STATS_FORMAT>
            set output format of statistics [default: text] [possible values: text, json]

        --stop-server
            stop background server

    -V, --version
            Print version information

    -z, --zero-stats
            zero statistic counters

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```

_`sccache-dist` old_

```
sccache 0.2.15

USAGE:
    sccache-dist [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    auth
    help         Prints this message or the help of the given subcommand(s)
    scheduler
    server
```

_`sccache-dist` new_

```
sccache 0.2.16-alpha.0

USAGE:
    sccache-dist <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    auth
    help         Print this message or the help of the given subcommand(s)
    scheduler
    server
```

So `sccache`'s is a bit more spaced out and has different headings. I can set custom headings for arguments to be under if we want that to match. I tried setting the output to be more compact too, but I think the line `--package-toolchain` may be overriding that. Other than that things are pretty much all the same aside from color with the new output and more parsing being enforced by `clap` instead of bubbled up by `sccache`


LovecraftianHorror:
Less tired now so I pushed some cleanup. From an outside user's perspective things should seem nicer because I tried pushing as much validation onto clap as possible so we get the nice error messages from it. From a contributors perspective the changes use the derive API and I think the code looks more approachable.

Since I forgot to mention it before this requires #1137 for the MSRV bump and is part of getting #1110 finished off

I still haven't got the `sccache-dist` test errors sorted out and don't know enough to accurately troubleshoot things since it's failing on `main` and this PR for me, but I'll go ahead and mark this as ready for review since those things can get sorted out during the review process

Here's an overview comparing things (just for reference the new cli stuff has color while the old stuff doesnt)

# `sccache`

## Help pages

### `sccache --help`

<details>
  <summary>old</summary>
  
```
sccache 0.2.15

USAGE:
    sccache [FLAGS] [OPTIONS] [cmd]...

FLAGS:
        --dist-auth       authenticate for distributed compilation
        --dist-status     show status of the distributed client
    -h, --help            Prints help information
    -s, --show-stats      show cache statistics
        --start-server    start background server
        --stop-server     stop background server
    -V, --version         Prints version information
    -z, --zero-stats      zero statistics counters

OPTIONS:
        --package-toolchain <executable> <out>    package toolchain for distributed compilation
        --stats-format <stats-format>
            set output format of statistics [default: text]  [possible values: text, json]


ARGS:
    <cmd>...

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```
</details>

<details>
  <summary>new</summary>
  
```
sccache 0.2.16-alpha.0

USAGE:
    sccache [OPTIONS] <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXECUTABLE> <OUT>|CMD> [--]

ARGS:
    <CMD>...

OPTIONS:
        --dist-auth
            authenticate for distributed compilation

        --dist-status
            show status of the distributed client

    -h, --help
            Print help information

        --package-toolchain <EXECUTABLE> <OUT>
            package toolchain for distributed compilation

    -s, --show-stats
            show cache statistics

        --start-server
            start background server

        --stats-format <FORMAT>
            set output format of statistics [default: text] [possible values: text, json]

        --stop-server
            stop background server

    -V, --version
            Print version information

    -z, --zero-stats
            zero statistic counters

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```
</details>

## Errors

### No command - `sccache`

_Note: It's not clear from the old cli usage what a "command" is, while the new cli shows in the usage and help messages_

Pushing this and too many commands validation onto `clap` did involve a bit of hackiness. The reason for it being hacky is that `clap` doesn't support and "arg" that is just and env variable, but I wanted to include `SCCACHE_START_SERVER=1` in the "one and only one" arg group since it is one of the commands and doesn't make sense to be used with many of the other commands.

My "fix" to handle this is to smuggle the value in under the `start_server` arg and hide it from listing the command-line option with `hide_env`. When `SCCACHE_START_SERVER=1` then `--start-server` is manually unset since it is set automatically by clap. When `SCCACHE_START_SERVER != 1` then the env var is removed since it would count as being present for `--start-server` and need some other command to be used.  If the env var isn't present then things can run as normal

This is a bit of a hacky way to handle things, but I think the end result from a user's perspective is worth it since it makes correct usage much more obvious. Having `SCCACHE_START_SERVER` linked to some unlikely-to-be-called flag that has `hide` set would be nicer, but the flag name still shows up in usage strings and error suggestions which is a bit annoying. I figured out that you can set the long flag name to `"\x08\x08\x08"` (backspace, backspace, backspace). This makes things look fine when `hide` is set since an empty long name would add `|--` to the usage string so three backspaces makes it render like it doesn't exist, but I figured that was too cursed to be used

<details>
  <summary>old</summary>
  
```
sccache: No command specified
sccache 0.2.15

USAGE:
    sccache [FLAGS] [OPTIONS] [cmd]...

FLAGS:
        --dist-auth       authenticate for distributed compilation
        --dist-status     show status of the distributed client
    -h, --help            Prints help information
    -s, --show-stats      show cache statistics
        --start-server    start background server
        --stop-server     stop background server
    -V, --version         Prints version information
    -z, --zero-stats      zero statistics counters

OPTIONS:
        --package-toolchain <executable> <out>    package toolchain for distributed compilation
        --stats-format <stats-format>
            set output format of statistics [default: text]  [possible values: text, json]


ARGS:
    <cmd>...

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```
</details>

<details>
  <summary>new</summary>
  
```
error: The following required arguments were not provided:
    <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXECUTABLE> <OUT>|CMD>

USAGE:
    sccache [OPTIONS] <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXECUTABLE> <OUT>|CMD> [--]

For more information try --help
```
</details>

### Too many commands - `sccache --start-server --stop-server`

<details>
  <summary>old</summary>
  
```
sccache: Too many commands specified
sccache 0.2.15

USAGE:
    sccache [FLAGS] [OPTIONS] [cmd]...

FLAGS:
        --dist-auth       authenticate for distributed compilation
        --dist-status     show status of the distributed client
    -h, --help            Prints help information
    -s, --show-stats      show cache statistics
        --start-server    start background server
        --stop-server     stop background server
    -V, --version         Prints version information
    -z, --zero-stats      zero statistics counters

OPTIONS:
        --package-toolchain <executable> <out>    package toolchain for distributed compilation
        --stats-format <stats-format>
            set output format of statistics [default: text]  [possible values: text, json]


ARGS:
    <cmd>...

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```
</details>

<details>
  <summary>new</summary>
  
```
error: The argument '--start-server' cannot be used with '--stop-server'

USAGE:
    sccache <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXECUTABLE> <OUT>|CMD>

For more information try --help
```
</details>

# `sccache-dist`

## Help pages

### `sccache-dist --help`

<details>
  <summary>old</summary>
  
```
sccache 0.2.15

USAGE:
    sccache-dist [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    auth
    help         Prints this message or the help of the given subcommand(s)
    scheduler
    server
```
</details>

<details>
  <summary>new</summary>
  
```
sccache 0.2.16-alpha.0

USAGE:
    sccache-dist <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    auth
    help         Print this message or the help of the given subcommand(s)
    scheduler
    server
```
</details>

### `sccache-dist auth --help`

<details>
  <summary>old</summary>
  
```
sccache-dist-auth

USAGE:
    sccache-dist auth [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    generate-jwt-hs256-key
    generate-jwt-hs256-server-token
    generate-shared-token
    help                               Prints this message or the help of the given subcommand(s)
```
</details>

<details>
  <summary>new</summary>
  
```
sccache-dist-auth 0.2.16-alpha.0

USAGE:
    sccache-dist auth <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    generate-jwt-hs256-key
    generate-jwt-hs256-token
    generate-shared-token
    help                        Print this message or the help of the given subcommand(s)
```
</details>

### `sccache-dist auth generate-jwt-hs256-key --help`

<details>
  <summary>old</summary>
  
```
sccache-dist-auth-generate-jwt-hs256-key

USAGE:
    sccache-dist auth generate-jwt-hs256-key

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
```
</details>

<details>
  <summary>new</summary>
  
```
sccache-dist-auth-generate-jwt-hs256-key 0.2.16-alpha.0

USAGE:
    sccache-dist auth generate-jwt-hs256-key

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information
```
</details>

### `sccache-dist auth generate-jwt-hs256-server-token --help`

<details>
  <summary>old</summary>
  
```
sccache-dist-auth-generate-jwt-hs256-server-token

USAGE:
    sccache-dist auth generate-jwt-hs256-server-token --config <PATH> --secret-key <KEY> --server <SERVER_ADDR>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --config <PATH>           Use the key from the scheduler config file
        --secret-key <KEY>        Use specified key to create the token
        --server <SERVER_ADDR>    Generate a key for the specified server
```
</details>

<details>
  <summary>new</summary>
  
```
sccache-dist-auth-generate-jwt-hs256-server-token 0.2.16-alpha.0

USAGE:
    sccache-dist auth generate-jwt-hs256-server-token --server <SERVER_ADDR> <--secret-key <KEY>|--config <PATH>>

OPTIONS:
        --config <PATH>           Use the key from the scheduler config file at PATH
    -h, --help                    Print help information
        --secret-key <KEY>        Use specified key to create the token
        --server <SERVER_ADDR>    Generate a key for the specified server
    -V, --version                 Print version information
```
</details>

### `sccache-dist auth generate-shared-token --help`

<details>
  <summary>old</summary>
  
```
sccache-dist-auth-generate-shared-token

USAGE:
    sccache-dist auth generate-shared-token [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --bits <BITS>    Use the specified number of bits of randomness [default: 256]
```
</details>

<details>
  <summary>new</summary>
  
```
sccache-dist-auth-generate-shared-token 0.2.16-alpha.0

USAGE:
    sccache-dist auth generate-shared-token [OPTIONS]

OPTIONS:
        --bits <BITS>    Use the specified number of bits of randomness [default: 256]
    -h, --help           Print help information
    -V, --version        Print version information
```
</details>

### `sccache-dist scheduler --help`

<details>
  <summary>old</summary>
  
```
sccache-dist-scheduler

USAGE:
    sccache-dist scheduler [OPTIONS] --config <PATH>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --config <PATH>     Use the scheduler config file at PATH
        --syslog <LEVEL>    Log to the syslog with LEVEL [possible values: error, warn, info, debug, trace]
```
</details>

<details>
  <summary>new</summary>
  
```
sccache-dist-scheduler 0.2.16-alpha.0

USAGE:
    sccache-dist scheduler --config <PATH> --syslog <LEVEL>

OPTIONS:
        --config <PATH>     Use the scheduler config file at PATH
    -h, --help              Print help information
        --syslog <LEVEL>    Log to the syslog with LEVEL [possible values: error, warn, info, debug, trace]
    -V, --version           Print version information
```
</details>

### `sccache-dist server --help`

<details>
  <summary>old</summary>
  
```
sccache-dist-server

USAGE:
    sccache-dist server [OPTIONS] --config <PATH>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --config <PATH>     Use the server config file at PATH
        --syslog <LEVEL>    Log to the syslog with LEVEL [possible values: error, warn, info, debug, trace]
```
</details>

<details>
  <summary>new</summary>
  
```
sccache-dist-server 0.2.16-alpha.0

USAGE:
    sccache-dist server --config <PATH> --syslog <LEVEL>

OPTIONS:
        --config <PATH>     Use the server config file at PATH
    -h, --help              Print help information
        --syslog <LEVEL>    Log to the syslog with LEVEL [possible values: error, warn, info, debug, trace]
    -V, --version           Print version information
```
</details>

## Errors

### Server Token conflict - `sccache-dist auth generate-jwt-hs256-server-token --server 127.0.0.1:8080 --config some_config --secret-key some_key`

_Note: From the implementation it appears that `--config` and `--secret-key` are meant to be mutually exclusive, but the old cli parsing would allow both to be passed and would just ignore `--secret-key` which isn't intuitive_

<details>
  <summary>old</summary>
  
```
sccache-dist: Could not load config
sccache 0.2.15

USAGE:
    sccache [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    auth
    help         Prints this message or the help of the given subcommand(s)
    scheduler
    server
```
</details>

<details>
  <summary>new</summary>
  
```
error: The argument '--config <PATH>' cannot be used with '--secret-key <KEY>'

USAGE:
    sccache-dist auth generate-jwt-hs256-server-token --server <SERVER_ADDR> --server <SERVER_ADDR> <--secret-key <KEY>|--config <PATH>>

For more information try --help
```
</details>

### Invalid Shared Token bit value - `sccache-dist auth generate-shared-token --bits 70`

_Note: New implementation points out that the invalid value is from the `--bits` option specifically _

<details>
  <summary>old</summary>
  
```
sccache-dist: Number of bits must be divisible by 8, greater than 64 and less than 4096
sccache 0.2.15

USAGE:
    sccache [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    auth
    help         Prints this message or the help of the given subcommand(s)
    scheduler
    server
```
</details>

<details>
  <summary>new</summary>
  
```
error: Invalid value "70" for '--bits <BITS>': Number of bits must be divisible by 8, greater than 64 and less than 4096

For more information try --help
```
</details>

LovecraftianHorror:
Thinking on it more there are still issues present with how `SCCACHE_START_SERVER` is currently added to the arg group. I'll push out a fix for that tomorrow

LovecraftianHorror:
As promised here is the fix. The issue was that the following wouldn't be allowed on `main`, but was allowed from this PR

```cmd
$ SCCACHE_START_SERVER=1 sccache --start-server
```

Along with that error messages from `SCCACHE_START_SERVER` conflicting with other commands would be confusing because it would mention `--start-server` even when it wasn't used.

The fix clears up error messages when an arg conflict occurs and `SCCACHE_START_SERVER` is present by making explicit note of it:

```cmd
$ SCCACHE_START_SERVER=1 ./sccache --start-server
sccache: `SCCACHE_START_SERVER=1` can't be used with other commands

$ SCCACHE_START_SERVER=1 ./sccache some command
sccache: `SCCACHE_START_SERVER=1` can't be used with other commands
```

Along with that the error with `SCCACHE_START_SERVER` being allowed with `--start-server` was fixed (as can be seen above) by inspecting the source of how `start-server` was set. If `SCCACHE_START_SERVER` was present and the source was from the flag then an error is returned then we know both  were set which should be an error. This works because flags take precedence over env vars, so if both are present then the source should always be the flag.

mitchhentges:
Thanks for the patch! Looks good so far, but I'll do a more complete review tomorrow most likely.

The only concern I have off the top of my head is that the new, non-compact help menu is a bit of a bummer, but you've already noticed and addressed this:

> I tried setting the output to be more compact too, but I think the line --package-toolchain may be overriding that.

LovecraftianHorror:
I'll poke around at some more settings to see if I can get things to be more compact again

LovecraftianHorror:
So enabling the `wrap_help` feature and switching the value name in `package_toolchain` from `EXECUTABLE` to `EXE` condenses down the help message when the terminal is a reasonable width and falls back to long help on narrower terminals. Another way of handling this could be to separate out `--package-toolchain` and `--stats-format` to a different heading like how it is by default in `clap` v2. Here's how things currently look as long as your terminal is wide enough to fit it (usage never seems to wrap which is fine)

```
sccache 0.2.16-alpha.0

USAGE:
    sccache [OPTIONS] <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXE> <OUT>|CMD> [--]

ARGS:
    <CMD>...

OPTIONS:
        --dist-auth                        authenticate for distributed compilation
        --dist-status                      show status of the distributed client
    -h, --help                             Print help information
        --package-toolchain <EXE> <OUT>    package toolchain for distributed compilation
    -s, --show-stats                       show cache statistics
        --start-server                     start background server
        --stats-format <FORMAT>            set output format of statistics [default: text] [possible values:
                                           text, json]
        --stop-server                      stop background server
    -V, --version                          Print version information
    -z, --zero-stats                       zero statistic counters

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```

mitchhentges (changes requested):
> So enabling the wrap_help feature and switching the value name in package_toolchain from EXECUTABLE to EXE condenses down the help message when the terminal is a reasonable width and falls back to long help on narrower terminals.

This looks great! The touch of colour is pretty nice too :+1:

> Another way of handling this could be to separate out --package-toolchain and --stats-format to a different heading like how it is by default in clap v2.

I think that separating "options" from "flags" was a very pedantic (and potentially confusing) distinction, so I really like how you've set it up.

> Too many commands - sccache --start-server --stop-server

I _really_ like how you've made this more obvious: rather than a confusing "Too many commands specified" error (which is ambiguous with the "you have provided an unexpectedly large _number_ of command-like arguments"), it tells you specifically which two arguments are unacceptably provided together. Nice 👍

-----

I think that the improvements you've made in updating `clap` and strengthening the restrictions we define (such as `start-server` and `stop-server` mutual exclusivity) are strong enhancements. However, the third, significant change that you've made in this patch - the migration to `clap`'s "derive" API - isn't something that I'm confident is an improvement for `sccache`.

This project is midsized, and is starting to sprout weird behaviours that are harder to represent with the "derive" interface: you're seeing this yourself with how much legwork you had to do to make `SCCACHE_START_SERVER` play nicely with the other options. The flexibility offered by the "builder" API is something that `sccache` is already leveraging, and may need to lean on even more in the future.

The other part that makes "derive" tricky is that it is _very_ magical. This can make debugging and tweaking `clap`'s behaviour a bit of a headache. For example, it took me some nontrivial digging to figure out what the attributes on `parse::Opts` was. Posted here for convenience:
```
#[derive(Parser)]
#[clap(version)]
#[clap(propagate_version = true)]
struct Opts { ... }
```

Two parts here were confusing:
* What does `clap(version)` do? Looking up `version` on [`docs.rs`](https://docs.rs/clap/latest/clap/) didn't point me in the right direction. I had to find [the "Derive Reference"](https://github.com/clap-rs/clap/blob/v3.1.6/examples/derive_ref/README.md#command-attributes), at which point the documented behaviour of "magic attributes" became clear. I think that this journey would be much less treacherous with the more traditional API.
* Where does `clap(propagate_version = true)` affect? Fortunately, [`propagate_version`](https://docs.rs/clap/3.1.6/clap/struct.App.html#method.propagate_version) had an entry in the documentation, but I still second-guessed myself for a minute because the function definition I found was associated with a [deprecated struct](https://docs.rs/clap/3.1.6/clap/struct.App.html#method.propagate_version).

I can see the appeal of leveraging the "attribute" language feature to try to tighten the coupling between parsing and the generated struct containing the refined arguments. However, I think that its costs aren't worth the benefits today.

-----

Would you be interested in simplifying this patch to be specifically about updating `clap` and improving our handling of specific incompatible arguments, such as `--start-server`, `--stop-server`, and the others?

tests/test-crate/Cargo.lock
```diff
@@ -1,16 +1,16 @@
 # This file is automatically @generated by Cargo.
```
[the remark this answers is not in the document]
LovecraftianHorror: The changes aren't for this PR. They just automatically get applied when running `cargo test` within at least the root of the repo (just double checked by reverting the file)

LovecraftianHorror:
> Would you be interested in simplifying this patch to be specifically about updating clap and improving our handling of specific incompatible arguments, such as --start-server, --stop-server, and the others?

The last push is all I have time for atm. It implements the above where we are back to the bulider api (barring some `ArgEnum`s where it is used to generate the list of possible values. This can be done manually or through other libraries of course).

Things _should_ still follow the same strictness as was enforced by the derive implementation. The `SCCACHE_START_SERVER` env var bit is handled essentially the same way (I switched to no longer smuggling the env var in an unrelated arg and just looked for a different error kind from `clap`. This isn't specific to changing back to the builder API. I just felt like smuggling a hidden env var into an `ArgGroup` was hackier than how things are done now). Beyond that the subcommand requirements, flag requirements, and value validity (`--bits`, `--stats-format`) should all be enforced with nice error messages from `clap` still

A notable difference from the `clap` v2 based builder API is that the old implementation made heavy usage of `Arg::from_usage(<usage string>)` which is now deprecated. It's new incarnation is the `arg!(<psuedo usage>)` macro so that more issues can be found at compile-time, but I opted for just building off of `Arg`s since I was running into some annoyances with `arg!()` that made it seem like it wasn't really pulling its weight

I did take the time to start implementing some unit tests for the arg parsing. I only had time to go through a decent bit of `sccache-dist`'s CLI. Starting tomorrow I'll be visiting family / focusing on work due to lack of time from visiting family, so I'll have a short 2-or-so week hiatus from OSS. If you want then feel free to finish off any final changes you want, or you can wait until I'm off of my break and I'll be happy to address any final comments :). I'll be around for responding to comments, just won't have any time for making changes

mitchhentges:
That sounds great, thanks! I'll take a look at this tomorrow most likely.

> Starting tomorrow I'll be visiting family / focusing on work due to lack of time from visiting family, so I'll have a short 2-or-so week hiatus from OSS. If you want then feel free to finish off any final changes you want, or you can wait until I'm off of my break and I'll be happy to address any final comments :). I'll be around for responding to comments, just won't have any time for making changes

I appreciate that, thanks! There isn't a rush to get this landed, especially considering that it's blocked on our next release (which should be going out in the next week or so).

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+                                .value_name("KEY"),
+                            config_with_help_message(
+                                "Use the key from the scheduler config file at PATH",
+                            ),
+                        ])
+                        .group(
+                            ArgGroup::new("key_source_mutual_exclusion")
+                                .args(&["config", "secret-key"])
+                                .required(true),
+                        ),
```
mitchhentges, line 149: :100: Nice!

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    fn auth_generate_jwt_hs256_server_token(subcommand_args: &[&'static str]) -> Vec<&'static str> {
+        let mut args = vec![EXE, "auth", "generate-jwt-hs256-server-token"];
+        args.extend_from_slice(subcommand_args);
+        args
+    }
+
+    #[test]
+    fn debug_assert() {
+        get_clap_command().debug_assert()
```
mitchhentges, line 267: Cool!

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+            } else {
+                bail!("Could not load config")
+            }
+        }
+        _ => unreachable!("Subcommand is enforced by clap"),
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
mitchhentges, line 250: Tests as well?! You're a hero :)

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+    // #[test]
+    // fn scheduler_good() {
+    //     todo!()
+    // }
+
+    // TODO: add tests for with sample config file
+    // #[test]
+    // fn server_good() {
+    //     todo!()
+    // }
```
mitchhentges, line 368: Let's move this to a Github issue, and remove the comments from this file.

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+    match try_parse_from(env::args()) {
+        Ok(cmd) => cmd,
+        Err(e) => match e.downcast::<clap::error::Error>() {
+            Ok(clap_err) => clap_err.exit(),
+            Err(some_other_err) => {
+                println!("sccache-dist: {some_other_err}");
+                for source_err in some_other_err.chain().skip(1) {
+                    println!("sccache-dist: caused by: {source_err}");
+                }
+                std::process::exit(1);
```
mitchhentges, line 106: Let's move the interface between our code and the CLI (error output, `sys::process:exit(1)`) up one level to sit in `sccache-dist/main.rs` again.

mitchhentges (changes requested):
> but I opted for just building off of Args since I was running into some annoyances with arg!() that made it seem like it wasn't really pulling its weight

Would you mind elaborating here in this PR with a comment for what specific annoyances you were running into?

-----

Three other things:
* I'll still need to do one more pass to look at the non-dist handling, but I'll handle it after you've addressed this first round here
* For future patches, I'd highly recommend trying to keep your diff smaller. Right now, you've significantly changed the `clap` configuration logic, but you've also moved a bunch of code. It would be much easier to review if that was split into either two, self-contained commits, or two separate PRs.
* Thanks for the patch, this is looking great so far :grin:

Cargo.toml
```diff
@@ -28,7 +28,7 @@ blake3 = "0.3"
 byteorder = "1.0"
 bytes = "1"
 chrono = { version = "0.4", optional = true }
-clap = "2.23.0"
+clap = { version = "3.1.6", features = ["derive", "env", "wrap_help"] }
```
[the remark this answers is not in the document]
LovecraftianHorror: It is at the moment, although not for very much. It does provide a `.from_str()` implementation that we could use.

On the other hand it still doesn't provide too much for the whole feature, so it wouldn't be much effort to just drop the usage entirely

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+use clap::{Arg, ArgEnum, ArgGroup, Command as ClapCommand, PossibleValue};
+use sccache::{config, dist::ServerId};
+use syslog::Facility;
+
+use crate::cmdline::{AuthSubcommand, Command};
+
+#[derive(Debug)]
+struct TokenBits(usize);
+
+impl TokenBits {
```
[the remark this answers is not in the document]
LovecraftianHorror: That would be a lot less ambiguous. Sounds good to me :+1:

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,369 @@
...
+    fn as_bytes(&self) -> usize {
+        self.0 / 8
+    }
+
+    fn default_str() -> &'static str {
+        // Size based on https://briansmith.org/rustdoc/ring/hmac/fn.recommended_key_len.html
+        "256"
```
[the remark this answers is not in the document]
LovecraftianHorror: Good point!

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: That sound good to me. There's already quite a lot going on with this PR

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: Sounds good. Having a function that may exit without having an obvious name doesn't seem like the best idea. Moving it back into `sccache-dist/main.rs` reduces that separation :+1:

LovecraftianHorror:
> Would you mind elaborating here in this PR with a comment for what specific annoyances you were running into?

Sure! Just for reference here is the current [docs section for `arg!()`](https://docs.rs/clap/3.1.6/clap/macro.arg.html)

- Long flags that contain a hyphen must be quoted like below. The only place I found this was [tucked away in an example](https://github.com/clap-rs/clap/blob/49559291f0700303518bf12c66b692444399eb71/examples/tutorial_builder/04_03_relations.rs#L9)

```rust
// Doesn't work
clap::arg!(--some-flag);
// Works
clap::arg!(--"some-flag");
```
- Multiple values aren't currently supported (has a nice error at least), so `--package-toolchain <EXE> <OUT>` would need to use `.value_names()` still. The error message when trying to use it for reference

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Some(["EXE"])`,
 right: `None`: Multiple values not yet supported', src/main.rs
```

- The docs for "Help String" which are below seem to be outdated. Single quotes were used for the old `::from_usage()` initialization, but are not allowed for the current macro implementation

> The help string is denoted between a pair of single quotes `''` and may contain any characters.

All of this just ended up leading to a very bumpy start, so it seemed easiest to just avoid the macro-magic and stick with the straight-forward (and well documented) builder pattern

---

> - For future patches, I'd highly recommend trying to keep your diff smaller. Right now, you've significantly changed the clap configuration logic, but you've also moved a bunch of code. It would be much easier to review if that was split into either two, self-contained commits, or two separate PRs.

:+1:

---

All the changes seem more than reasonable! I'll work through them when I get back to OSS again

tests/test-crate/Cargo.lock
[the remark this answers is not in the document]
LovecraftianHorror: Changes reverted in 91910c0c5706fc91ddd3e5741680b7c71e81f13e

Cargo.toml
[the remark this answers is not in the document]
LovecraftianHorror: Since we weren't using it for much I removed the usage of `ArgEnum` in 2e6213dcabee2625df58fce8fa45c9d5324e4b28

src/bin/sccache-dist/cmdline/parse.rs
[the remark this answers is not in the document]
LovecraftianHorror: Switched to the recommended changes in 09ee6dd4e02e710ea0c5e73f92e776cfb4029896

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: Moved all the functionality that would `exit` back into the binaries respective `main`s in 87ed6e8f455e0e65e6e64ca8cd844ae72c9c7419

src/bin/sccache-dist/cmdline/parse.rs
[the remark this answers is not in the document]
LovecraftianHorror: Inlined this value in 09ee6dd4e02e710ea0c5e73f92e776cfb4029896

LovecraftianHorror:
I believe the recent commits address all your current requested changes, so this should be good for another review whenever you are able!

mitchhentges (approved):
Looks good to me, nice work! Thanks again for this patch and your collaboration, this is great!

I'm going to wait to land this until `v0.3.0`, [which is unfortunately taking longer than I had hoped](https://github.com/mozilla/sccache/issues/1152). I'm looking forward to merging this :)

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,344 @@
...
+            LogLevel::Debug => Self::Debug,
+            LogLevel::Trace => Self::Trace,
+        }
+    }
+}
+
+/// Parse commandline args into a `Result<Command>` to execute.
+pub fn try_parse() -> anyhow::Result<Command> {
+    try_parse_from(env::args())
+}
```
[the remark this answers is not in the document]
LovecraftianHorror: Changed in 249a28eed14c990e7d8fbf748c84e2d998a9e75b

src/cmdline.rs
```diff
@@ -141,86 +207,86 @@ pub fn parse() -> Result<Command> {
...
+            bail!("`{ENV_VAR_INTERNAL_START_SERVER}=1` can't be used with other commands");
+        }
+        (false, Ok(matches)) => {
+            let show_stats = matches.is_present("show-stats");
+            let start_server = matches.is_present("start-server");
+            let stop_server = matches.is_present("stop-server");
+            let zero_stats = matches.is_present("zero-stats");
+            let dist_auth = matches.is_present("dist-auth");
+            let dist_status = matches.is_present("dist-status");
+            let package_toolchain = matches.is_present("package-toolchain");
```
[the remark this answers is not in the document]
LovecraftianHorror: Changed in 62bdc36df3aedbe75554a6c2cb207987aef674ed

src/cmdline.rs
```diff
@@ -59,50 +90,85 @@ pub enum Command {
...
+    Arg::new(name).long(name)
+}
+
+/// Get the [`clap::Command`] used for argument parsing.
+fn get_clap_command() -> clap::Command<'static> {
+    clap::Command::new(env!("CARGO_PKG_NAME"))
         .version(env!("CARGO_PKG_VERSION"))
-        .setting(AppSettings::TrailingVarArg)
+        .trailing_var_arg(true)
+        .max_term_width(110)
```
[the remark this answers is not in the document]
LovecraftianHorror: I enabled the `wrap_help` feature so that the help message doesn't use the more vertically spread out spacing, but when enabled nothing stops the help from just trailing on horizontally forever, so I picked 110 as the length that is large enough to keep things compact, but doesn't trail on forever.

Currently only `--stats-format` has a long enough help line to wrap. Here is it without the wrapping (the usage line never seems to wrap)

```
sccache 0.2.16-alpha.0

USAGE:
    sccache [OPTIONS] <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXE> <OUT>|CMD> [--]

ARGS:
    <CMD>...

OPTIONS:
        --dist-auth                        authenticate for distributed compilation
        --dist-status                      show status of the distributed client
    -h, --help                             Print help information
        --package-toolchain <EXE> <OUT>    package toolchain for distributed compilation
    -s, --show-stats                       show cache statistics
        --start-server                     start background server
        --stats-format <FMT>               set output format of statistics [default: text] [possible values: text, json]
        --stop-server                      stop background server
    -V, --version                          Print version information
    -z, --zero-stats                       zero statistics counters

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```

and then with the current wrapping

```
sccache 0.2.16-alpha.0

USAGE:
    sccache [OPTIONS] <--dist-auth|--dist-status|--show-stats|--start-server|--stop-server|--zero-stats|--package-toolchain <EXE> <OUT>|CMD> [--]

ARGS:
    <CMD>...

OPTIONS:
        --dist-auth                        authenticate for distributed compilation
        --dist-status                      show status of the distributed client
    -h, --help                             Print help information
        --package-toolchain <EXE> <OUT>    package toolchain for distributed compilation
    -s, --show-stats                       show cache statistics
        --start-server                     start background server
        --stats-format <FMT>               set output format of statistics [default: text] [possible values:
                                           text, json]
        --stop-server                      stop background server
    -V, --version                          Print version information
    -z, --zero-stats                       zero statistics counters

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,340 @@
...
+
+impl TokenLength {
+    fn as_bytes(&self) -> usize {
+        self.0 / 8
+    }
+
+    fn from_bits(bits: &str) -> anyhow::Result<Self> {
+        let bits: usize = bits.parse()?;
+        if bits % 8 != 0 || bits < 64 || bits > 4_096 {
+            bail!("Number of bits must be divisible by 8, greater than 64 and less than 4096")
```
drahnr, line 21:
```suggestion
            bail!("Number of bits must be divisible by 8, greater than 64 and less than or equal to 4096")
```

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,340 @@
...
+struct TokenLength(usize);
+
+impl TokenLength {
+    fn as_bytes(&self) -> usize {
+        self.0 / 8
+    }
+
+    fn from_bits(bits: &str) -> anyhow::Result<Self> {
+        let bits: usize = bits.parse()?;
+        if bits % 8 != 0 || bits < 64 || bits > 4_096 {
```
drahnr, line 20:
```suggestion
        if bits & 0x7 != 0 || bits < 64 || bits > 4_096 {
```

I'd prefer to have separate error messages for each of these

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn from(log_level: LogLevel) -> Self {
+        match log_level {
+            LogLevel::Error => Self::Error,
+            LogLevel::Warn => Self::Warn,
+            LogLevel::Info => Self::Info,
+            LogLevel::Debug => Self::Debug,
+            LogLevel::Trace => Self::Trace,
+        }
+    }
+}
```
drahnr, line 92: There is the crate https://crates.io/crates/clap-verbosity-flag that already does this :)

src/bin/sccache-dist/cmdline/parse.rs
```diff
@@ -0,0 +1,340 @@
...
+                ),
+        )
+        .subcommand(ClapCommand::new("scheduler").args(&[
+            config_with_help_message("Use the scheduler config file at PATH").required(true),
+            syslog.clone(),
+        ]))
+        .subcommand(ClapCommand::new("server").args(&[
+            config_with_help_message("Use the server config file at PATH").required(true),
+            syslog,
+        ]))
```
drahnr, line 150: Is there a particular reason you chose the manual construction vs creating rather than using the proc-macro provided by clap?

drahnr (commented):
Very nice :+1: , just a few questions, other than that I'd be happy to merge if you don't want to go the annotation route :)

mkatychev:
@drahnr  this is slightly tangential but is there a reason why `--flags` listed in the helpdoc are actually used as args/commands?
```
sccache 0.3.0

USAGE:
    sccache [FLAGS] [OPTIONS] [cmd]...

FLAGS:
    -s, --show-stats      show cache statistics
        --start-server    start background server
        ...
ARGS:
    <cmd>...
```
It seems very confusing to see `--start-server` under flags only to later realize  that it's a subcommand of `sccache`.

Here is [POSIX documentation](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) talking about flags/options:
> It is followed by options, option-arguments, and operands. The arguments that consist of <hyphen-minus> characters and single letters or digits, such as 'a', are known as "options" (or, historically, "flags").

[docopt](http://docopt.org/) makes the same distinction

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: If you are talking about the clap's derive API, then the PR was originally using that, but switched to the builder API in 13ae495ab76b0826eadac0a7e9f87caedcfbeb0c in response to this review comment https://github.com/mozilla/sccache/pull/1140#pullrequestreview-910722058

I personally prefer the derive API since you no longer have to make sure that both parsing and consuming match, but the concerns around less flexibility and worse documentation are certainly valid

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: TIL, I'll switch over to that!

LovecraftianHorror:
@drahnr Thanks for the review! I'm stretched a bit thin this week, but should be able to address everything over this weekend

src/bin/sccache-dist/cmdline/parse.rs
drahnr: I see, I missed that one. Thank you! Leave as is, I wasn't aware of the discussion.

drahnr:
Th

> @drahnr this is slightly tangential but is there a reason why `--flags` listed in the helpdoc are actually used as args/commands?
> 
> ```
> sccache 0.3.0
> 
> USAGE:
>     sccache [FLAGS] [OPTIONS] [cmd]...
> 
> FLAGS:
>     -s, --show-stats      show cache statistics
>         --start-server    start background server
>         ...
> ARGS:
>     <cmd>...
> ```
> 
> It seems very confusing to see `--start-server` under flags only to later realize that it's a subcommand of `sccache`.
> 
> Here is [POSIX documentation](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) talking about flags/options:
> 
> > It is followed by options, option-arguments, and operands. The arguments that consist of  characters and single letters or digits, such as 'a', are known as "options" (or, historically, "flags").
> 
> [docopt](http://docopt.org/) makes the same distinction

I'll look into this more closely, thanks for noting!

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: Resolved in 4b9d5bd58e3b0f9fa59de73815b68c0f9ab3be02

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: Resolved in 4b9d5bd58e3b0f9fa59de73815b68c0f9ab3be02

src/bin/sccache-dist/cmdline/parse.rs
LovecraftianHorror: Took a look at `clap-verbosity-flag` and noticed that it is for the more traditional `-q` and `-v*` style of verbosity flags while `sccache-dist` currently takes `--syslog <LEVEL>` instead. This means that switching to `clap-verbosity-flag` would be a breaking change and as v0.3 was released under a month ago, I'm guessing that another breaking change isn't intended quite this soon (and could merit looking into potentially revamping `sccache`'s CLI structure as well)

drahnr:
I'll give this another go early next week :)

The only thing with not using the macros, is that now the logic is split into declaration and evaluation, which makes it harder to maintain imho. I am aware that this opposes what @mitchhentges suggested. Let me dig into this one last time, and please don't force push so we can use the "old" macro based approach just in case.

drahnr:
I am a bit short on time recently. I'm wary of incompatibilities, other than that, it looks good. Sorry for the delay.

src/bin/sccache-dist/cmdline/parse.rs
drahnr: You're probably right one that, I am tempted to defer that change to another PR, but still strongly in favor of making it. I am not aware of other programs using that naming.

drahnr:
I reviewed this again and #1354 is hence ready to merge @sylvestre

Issue #1130, opened 2022-02-18
Benchmark `v0.3.0` prior to release


mitchhentges:
Check that we don't have performance regressions, or that they're manageable.
`cargo bench` may be helpful here, though an end-to-end test can also happen against Firefox CI for a real FF build.

luser:
Historically, "push a try build with the new revision" was the only real performance testing we ever did, but it's a fairly intensive real-world workload so it's not terrible. Testing the actual performance of builds when caching involved is extremely tricky, but @glandium can advise.

FooBarWidget:
How can others help with this issue?

mitchhentges:
I appreciate it, but I'll tackle this one, as it requires some Mozilla CI permissions :)

FooBarWidget:
@mitchhentges Then is there anything else I can do to help 0.3 get released faster?

mitchhentges:
Negative, I'm currently waiting on some CI builds to complete. This is actively being worked on.

mitchhentges:
[Perfherder results can be found here](https://treeherder.mozilla.org/perfherder/compare?originalProject=try&originalRevision=de1e1793af754126dba5bd640ce20665213decc8&newProject=try&newRevision=040eaf16c6aae5e3a793497eaeecd5dabb9c4f76&framework=2&page=1).

I'm seeing a lower cache hit rate, but that's probably because the existing sccache version is leveraging artifacts from other `try` runs.
I think we're good for a release.

PR #1143, opened 2022-03-17
Fix spawning excessive conhost processes on Windows

FooBarWidget:
Fixes https://github.com/mozilla/sccache/issues/514#issuecomment-1061945431 (comment 2022-03-08 by @mitchhentges)

Any console-based subprocess spawned with CREATE_NO_WINDOW actually has a hidden console, and thus an associated conhost process. Since the sccache server is already started with CREATE_NO_WINDOW, it's unnecessary to spawn further subprocesses with CREATE_NO_WINDOW. Removing this flag allows subprocesses to share the sccache server's hidden console, thus avoiding each subprocess from getting its own conhost.

For an extended explanation see this comment and its follow-ups:
https://github.com/mozilla/sccache/issues/514#issuecomment-1068961965

### DIFF
--- a/src/mock_command.rs
+++ b/src/mock_command.rs
@@ -104,8 +104,6 @@ pub trait RunCommand: fmt::Debug + Send {
     fn env_clear(&mut self) -> &mut Self;
     /// Set the working directory of the process to `dir`.
     fn current_dir<P: AsRef<Path>>(&mut self, dir: P) -> &mut Self;
-    /// Create the proces without a visible console on Windows.
-    fn no_console(&mut self) -> &mut Self;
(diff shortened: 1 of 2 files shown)


FooBarWidget:
CI is breaking due to a Rust compiler panic on thirtyfour. Not sure what this is.

mitchhentges (approved):
Makes sense, thanks for the patch! Let's land this after `v0.3.0`, because this is no longer an issue-blocking-release and I'd like to avoid [re-running the benchmarking tasks with this commit included](https://github.com/mozilla/sccache/issues/1130).

FooBarWidget:
I see you already did the rebase yourself. All good now?

PR #1144, opened 2022-03-17
Improve Azure credentials test coverage

FooBarWidget:
This also simplifies `azure::credentials::EnvironmentProvider` (which only has 1 method) into simply exporting a `credentials_from_environment()` function.


FooBarWidget:
CI still failing on compiler crash. For some reason Codecov hasn't been triggered.

mitchhentges (changes requested):
I'm really liking these extra tests, nice work!

It looks like your testing of `parse_connection_string()` is very specific, as you call the function directly and assert its results. Nice :+1: 

However, there's a few tests that check the handling of missing config variables, but they're including the complexity of including the real operating system's environment variables This has forced your hand here with the use of `MutexGuard`, whose value here won't be straightforward for future readers to understand.

Instead, I might recommend splitting `credentials_from_environment`: one that has the business logic of handling optionals, and the other that does the straightforward, practically-too-simple-to-test job of calling the first function with `var(...), var(...)` as parameters.

(Note that you don't have to go all the way back to the trait + struct system, I agree that that was overkill. I'm leaning towards it just being two functions.)

At this point, you get the following benefits:
* Test coverage is improved by practically the same amount
* Tests become simpler because they don't need to guard against parallelism
* The consumers still call in just like normal.

The only downside here is that you'll have one "untested" function - the public, interface function. But, it will be so, mind-numbingly small and simple, that we can be content without it being covered :)

FooBarWidget:
Honestly, the var lookup code is already trivial. I was just trying to get the coverage metric to 100% but I don't there's real value in covering var().context calls. So I propose just removing the credentials_from_environment tests, leaving only the connection parsing tests.

FooBarWidget:
Here are the changes.

mitchhentges (approved):
> I was just trying to get the coverage metric to 100% but I don't think there's real value in covering var().context() calls

Yeah, that sounds good to me. I'm personally pretty disillusioned with the value of 100% code coverage, so your current solution here looks ideal to me. :+1:

mitchhentges:
> Here are the changes.

Github notifies me already when you update the PR with extra commits :)

FooBarWidget:
The CI failed because of a network issue

sylvestre:
i restarted it

PR #1147, opened 2022-03-21
nvcc: support -ccbin / --compiler-bindir

vlad-ivanov-name:
Add host compiler to hash as well when it's explicitly specified


mitchhentges (approved):
Nice, thanks!
I'm going to land this after the `v0.3.0` release :+1:

vlad-ivanov-name:
Thank you, but maybe it would make sense to include this in release? Thing is, without this addition sccache won't work at all as it will treat the argument after `-ccbin` as a positional argument, thus skipping cache completely because of multiple inputs. So it's not changing existing working behaviour but rather un-breaking something that is currently broken

mitchhentges:
I think you're in luck, my current benchmarking runs are running into a comparison issue, so I'm starting again from scratch tomorrow. Let's land this :+1:

PR #1149, opened 2022-03-22
Ignore lockfiles of test crates used in the integration tests

### LINKED ISSUE #1145: `cargo test` causes "tests/test-crate/Cargo.lock" to be modified
`cargo test` should not mutate files that are tracked by `git`.
I'm guessing that one of our tests are invoking `cargo` without supplying an option telling it to leave lockfiles as-is.

Xanewok:
The lockfile was mostly added by accident FWIU, looking at the file history from https://github.com/mozilla/sccache/commits/main/tests/test-crate/Cargo.lock.

One one hand lockfiles are handy in the sense that they lock the version used so there are no surprises but it's such a trivial dependency/setup that I think it's better for now to just ignore them altogether and come back to this if this ever causes a problem.

Fixes #1145


Xanewok:
Any idea why the coverage macOS job has been cancelled?

sylvestre:
Dunno, I just restarted it and it is fine

mitchhentges (approved):
Thanks for the patch, this makes sense :+1:

PR #1157, opened 2022-04-14
Handle new 'rustc -Z ls' output lines

lissyx:


src/compiler/rust.rs
```diff
@@ -2895,13 +2897,36 @@ c:/foo/bar.rs:
 
     #[cfg(feature = "dist-client")]
     #[test]
-    fn test_parse_rustc_z_ls() {
+    fn test_parse_rustc_z_ls_old() {
```
mitchhentges, line 2900: I'd recommend using an absolute identifier to describe the difference in behaviour that you're testing here: "old" and "new" is relative, and will mean something different when these tests are investigated in, say, 2 years time.

Perhaps something like `test_parse_rustc_ls_pre_1_60()` (or, whichever rust version changed the behaviour).

mitchhentges (approved):
Makes sense, thanks for the patch.
I'm going to hold off on landing this until after the [`v0.3.0` release](https://github.com/mozilla/sccache/milestone/3)

src/compiler/rust.rs
lissyx: I'm not sure in which version it changed, from what I tested at least 1.58. Maybe I can find it within `rustc` source ...

lissyx: https://github.com/rust-lang/rust/commit/cef3ab75b12155e0582dd8b7710b7b901215fdd6

lissyx:
> Makes sense, thanks for the patch. I'm going to hold off on landing this until after the [`v0.3.0` release](https://github.com/mozilla/sccache/milestone/3)

Why wait ? This is already broken for m-c users with sccache-dist.

mitchhentges:
Two reasons:
* I'm already undergoing testing/benchmarking for the current status of `main`
* There will be a follow-up release shortly after `v0.3.0` which will include improvements such as this one.

lissyx:
Right but it's not an improvement, it's a fix required to be able to continue to use `sccache-dist` on current m-c.

mitchhentges:
That's fair 👍

PR #1159, opened 2022-04-14
Verify if a compile input is pure rlib or mixed rlib+staticlib before trimming

### LINKED ISSUE #1158: Dependency crate with staticlib and rlib gets trimmed
In https://bugzilla.mozilla.org/show_bug.cgi?id=1760743 we got a regression from a few crates vendored within m-c that expose both `staticlib` and `rlib`. There is code to guard against trimming when a `staticlib` is generated in https://github.com/mozilla/sccache/blob/5edd91a739df0f51f260766d7f68f6f59a89d56e/src/compiler/rust.rs#L1891-L1899 but from my understanding of the code, `crate_types` at that point is the crate we are compiling, not the dependency we are pulling in.

e.g., compiling crate `audioipc2` we have a dep on `audio_thread_priority` (which has both `staticlib` and `rlib`). So we get `can_trim_rlibs=true` because `audioipc2` is a `rlib` only, but we apply trimming on `audio_thread_priority`, effectively triggering the issue.

lissyx:
Fixes #1158.


mitchhentges (approved):
Thanks

PR #1162, opened 2022-04-20
Use clang -fminimize-whitespace to increase cache hits.

Meinersbur:
Whenever line number differences can be ignored (i.e. when using `-P`), also add `-fminimize-whitespace` to reduce irrelevant formatting differences for hashing. This should allow a cache hit after eg. applying clang-format over the source or editing multi-line comments.

`-fminimize-whitespace` has been [introduced in Clang 14](https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#new-compiler-flags). This patch only enables it if detecting Clang version 14 or higher, and when not using Apple's flavour of Clang which uses different version numbers.

This is a continuation of #1055 after applying requested changes and rebasing. It contains the original commits by @InnovativeInventor (Thank you!). There are some differences to #1055: The version field of CCompiler is reused. Which flags for preprocessing can be used is passed from `CCompilerImpl.preprocess` to `preprocess_cmd`. Testing with clang-14 is done using a new builder for Ubuntu 22.04; the compiler version is extracted from the `clang --version` output, deliberately different from how sccache determines whether `-fminimize-whitespace` can be used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,16 @@ jobs:
         with:
           toolchain: ${{ matrix.rustc }}
 
+      - name: Install LLVM/Clang
+        if: contains(matrix.os, 'ubuntu')
+        run: 
+          sudo apt-get update || true
+          sudo apt-get install wget
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 13
+          sudo ./llvm.sh 14
+
       - name: Build tests
         run: cargo test --no-run --locked --all-targets --verbose ${{ matrix.extra_args }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,8 @@ jobs:
             rustc: nightly
             allow_failure: true
             extra_args: --features=unstable
+          - os: ubuntu-20.04
+            rustc: stable
           - os: macOS-10.15
           - os: windows-2019
             rustc: 1.48.0 # Oldest supported version, keep in sync with README.md
@@ -75,16 +77,6 @@ jobs:
         uses: ./.github/actions/rust-toolchain
         with:
           toolchain: ${{ matrix.rustc }}
-
-      - name: Install LLVM/Clang
-        if: contains(matrix.os, 'ubuntu')
-        run: 
-          sudo apt-get update || true
-          sudo apt-get install wget
-          wget https://apt.llvm.org/llvm.sh
-          chmod +x llvm.sh
-          sudo ./llvm.sh 13
-          sudo ./llvm.sh 14
 
       - name: Build tests
         run: cargo test --no-run --locked --all-targets --verbose ${{ matrix.extra_args }}
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -35,6 +35,9 @@ use crate::errors::*;
 pub struct Clang {
     /// true iff this is clang++.
     pub clangplusplus: bool,
+    /// true iff this is Apple's clang(++).
+    pub is_appleclang: bool,
+    /// String from __VERSION__ macro.
     pub version: Option<String>,
 }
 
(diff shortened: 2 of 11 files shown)


src/compiler/clang.rs
```diff
@@ -73,6 +77,23 @@ impl CCompilerImpl for Clang {
...
+            if let Some(version_val) = self.version.clone() {
+                if let Some(parsed_version) = version_val.split(' ').find(|x| x.contains('.')) {
+                    if let Some(major_version) = parsed_version.split('.').next() {
+                        if let Ok(parsed_major_version) = major_version.parse::<usize>() {
+                            if parsed_major_version >= 14 {
+                                ignorable_whitespace_flags.push("-fminimize-whitespace".to_string())
+                            }
+                        }
+                    }
+                }
```
drahnr, line 93: It might be worth to use a specialized crate such as `semver` rather than to parse it manually.

src/compiler/gcc.rs
```diff
@@ -73,6 +73,7 @@ impl CCompilerImpl for Gcc {
             may_dist,
             self.kind(),
             rewrite_includes_only,
+            vec!["-P".to_string()],
```
drahnr, line 76: Could you explain how this is related to the remaining changeset?

src/compiler/clang.rs
```diff
@@ -35,6 +35,10 @@ use crate::errors::*;
 pub struct Clang {
     /// true iff this is clang++.
     pub clangplusplus: bool,
+    /// true iff this is Apple's clang(++).
+    pub is_appleclang: bool,
+    /// String from __VERSION__ macro.
+    pub version: Option<String>,
```
drahnr, line 41: Should we consider adding `version` always?

drahnr (changes requested):
A few small nits to be clarified, other than that LGTM! Thank you!

src/compiler/clang.rs
Meinersbur: Using semver in https://github.com/mozilla/sccache/pull/1162/commits/ba0a922c0cb10a955c2335e78ae7516836c88a66. Also refactored-out the check function.

src/compiler/clang.rs
Meinersbur: Moved `version` from CCompiler to every `CCompilerImpl` struct in https://github.com/mozilla/sccache/pull/1162/commits/3f0880ae577786b6fb6bfc98276db1281e38f046. The alternative would have been to pass the version argument through a lot of functions, but semantically the value fits better to `CCompilerImpl` anyway, next to `is_clang`, `is_applyclang`, etc.

.github/workflows/ci.yml
```diff
@@ -56,6 +56,7 @@ jobs:
             rustc: nightly
             allow_failure: true
             extra_args: --features=unstable
+          - os: ubuntu-22.04
```
Meinersbur, line 59: Note that `ubuntu-22.04` is not yet available on Github actions yet, but is [expected in early May](https://github.com/actions/virtual-environments/discussions/5427).

src/compiler/clang.rs
drahnr: Apologize for my inaccuracy, I meant: Can we drop the `Option<_>` wrapping since `__VERSION__` is always evaluated?

Meinersbur: `detect_c_compiler` has a case to detect when the `__VERSION__` macro has not been replaced, in which case no version information would be available and the current implementation passes `None` to `CCompiler.version`. I could use an empty string in this case (and the tests that passed `None` so far), if that's preferable?

drahnr: Fair enough, I'd say leave as is - thanks for the elaborate clarification!

.github/workflows/ci.yml
drahnr: Does this mean the PR can only merged then? I am unfortunately not so much involved in GitHub CI.

drahnr (approved):

.github/workflows/ci.yml
Meinersbur: It means that on GitHub the commit will not show the green checkmark next to it since not all tests have been executed. See https://github.com/Meinersbur/sccache/runs/6166559042. I don't know whether GitHub will retroactively make the `ubuntu-22.04` runners, once available, executed the pending jobs.

drahnr:
Thank you!

PR #1164, opened 2022-04-25
Replace usages of blocking `reqwest` to be async instead

mitchhentges:
The blocking implementation of `reqwest` now uses an internal `tokio`
runtime as of `0.10`. `tokio` doesn't like having nested runtimes, so it
raises an error when `request>=0.10` is used from a tokio-`async`
function.

The primary rationale for continuing to use "blocking" `reqwest` was
because its async implementation was "extremely limited" and didn't
support streamed data from a file.

However, now, we're able to leverage
`tokio::fs::File -> AsyncRead -> ReaderStream -> Body::wrap_stream()`,
which obsoletes this one use case for blocking `reqwest`.

Convert all blocking usages of `reqwest` to `async` accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1778,6 +1778,7 @@ dependencies = [
  "serde_urlencoded",
  "tokio",
  "tokio-native-tls",
+ "tokio-util",
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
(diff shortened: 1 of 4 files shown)


mitchhentges:
This is going to look very similiar to https://github.com/paritytech/cachepot/commit/19d8487f059777d72f20cc018530019ed1f16683

drahnr (approved):

Xanewok (commented):
Like we talked, LGTM =)

PR #1176, opened 2022-05-28
Support `-gsplit-dwaf` without `-g` for the latest clang

kateinoigakukun:
Since https://reviews.llvm.org/D80391, -gsplit-dwarf is a no-op if -g is
specified. This means we can't assume the presence of .dwo output
if -gsplit-dwarf and no -g.

This patch makes .dwo output optional.

Without this patch, it fails like below.

```console
$ ./sccache-v0.3.0-x86_64-unknown-linux-musl/sccache ./clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/bin/clang -gsplit-dwarf -c a.c
sccache: error: failed to zip up compiler outputs
sccache: caused by: failed to zip up compiler outputs
sccache: caused by: failed to open file `"/tmp/tmp.gfLlNOFb2c/a.dwo"`
sccache: caused by: No such file or directory (os error 2)
```

<details>
<summary>Reproducible Dockerfile</summary>

```dockerfile
FROM ubuntu:20.04

RUN apt-get update && apt-get install curl xz-utils -y
RUN curl -L https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz | tar Jx
RUN curl -L https://github.com/mozilla/sccache/releases/download/v0.3.0/sccache-v0.3.0-x86_64-unknown-linux-musl.tar.gz | tar xz
RUN chmod +x ./sccache-v0.3.0-x86_64-unknown-linux-musl/sccache
RUN touch a.c
CMD ./sccache-v0.3.0-x86_64-unknown-linux-musl/sccache ./clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/bin/clang -gsplit-dwarf -c a.c
```

</details>

Does this change make sense? Thanks.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -65,6 +65,18 @@ fn set_file_mode(_path: &Path, _mode: u32) -> Result<()> {
     Ok(())
 }
 
+/// Cache object sourced by a file.
+#[derive(Clone)]
+pub struct FileObjectSource {
+    /// Identifier for this object. Should be unique within a compilation unit.
+    /// Note that a compilation unit is a single source file in C/C++ and a crate in Rust.
+    pub key: String,
+    /// Absolute path to the file.
+    pub path: PathBuf,
+    /// Whether the file must be present on disk and is essential for the compilation.
+    pub optional: bool,
+}
+
 /// Result of a cache lookup.
 pub enum Cache {
     /// Result was found in cache.
@@ -155,10 +167,15 @@ impl CacheRead {
         pool: &tokio::runtime::Handle,
     ) -> Result<()>
     where
-        T: IntoIterator<Item = (String, PathBuf)> + Send + Sync + 'static,
+        T: IntoIterator<Item = FileObjectSource> + Send + Sync + 'static,
     {
         pool.spawn_blocking(move || {
-            for (key, path) in objects {
+            for FileObjectSource {
+                key,
+                path,
+                optional,
+            } in objects
+            {
                 let dir = match path.parent() {
                     Some(d) => d,
                     None => bail!("Output file without a parent directory!"),
@@ -167,10 +184,16 @@ impl CacheRead {
                 // move it to its final location so that other rustc invocations
                 // happening in parallel don't see a partially-written file.
                 let mut tmp = NamedTempFile::new_in(dir)?;
-                let mode = self.get_object(&key, &mut tmp)?;
-                tmp.persist(&path)?;
-                if let Some(mode) = mode {
-                    set_file_mode(&path, mode)?;
+                match (self.get_object(&key, &mut tmp), optional) {
+                    (Ok(mode), _) => {
+                        tmp.persist(&path)?;
+                        if let Some(mode) = mode {
+                            set_file_mode(&path, mode)?;
+                        }
+                    }
+                    (Err(e), false) => return Err(e),
+                    // skip if no object found and it's optional
+                    (Err(_), true) => continue,
                 }
             }
             Ok(())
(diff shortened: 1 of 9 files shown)


src/cache/cache.rs
```diff
@@ -155,10 +155,10 @@ impl CacheRead {
         pool: &tokio::runtime::Handle,
     ) -> Result<()>
     where
-        T: IntoIterator<Item = (String, PathBuf)> + Send + Sync + 'static,
+        T: IntoIterator<Item = (String, PathBuf, bool)> + Send + Sync + 'static,
```
drahnr, line 158: I am not a big fan of tuples, especially when they accumulate primitives which are non-descriptive

src/compiler/compiler.rs
```diff
@@ -256,8 +255,8 @@ where
         let duration = start.elapsed();
         let outputs = compilation
             .outputs()
-            .map(|(key, path)| (key.to_string(), cwd.join(path)))
-            .collect::<HashMap<_, _>>();
+            .map(|(key, path, optional)| (key.to_string(), cwd.join(path), optional))
+            .collect::<Vec<_>>();
```
drahnr, line 259: Shouldn't this become a `HashMap<key,(path, optional)>`?

kateinoigakukun: `extract_objects` accepts `T: IntoIterator<Item = (key, path, optional)>`, so it cannot be `HashMap` now.
And this `outputs` variable was not used as HashMap anywhere.

src/cache/cache.rs
kateinoigakukun: Ok, that makes sense 👍

kateinoigakukun:
@drahnr  Do you think it's ready for merge?

drahnr (approved):

drahnr (approved):

src/cache/cache.rs
```diff
@@ -65,6 +65,17 @@ fn set_file_mode(_path: &Path, _mode: u32) -> Result<()> {
     Ok(())
 }
 
+/// Cache object sourced by a file.
+#[derive(Clone)]
+pub struct FileObjectSource {
+    /// Identifier for this object. Should be unique within a compilation unit.
+    pub key: String,
+    /// Path to the file.
```
drahnr, line 73: nit: specify if this is a absolute or relative path

src/cache/cache.rs
```diff
@@ -65,6 +65,17 @@ fn set_file_mode(_path: &Path, _mode: u32) -> Result<()> {
     Ok(())
 }
 
+/// Cache object sourced by a file.
+#[derive(Clone)]
+pub struct FileObjectSource {
+    /// Identifier for this object. Should be unique within a compilation unit.
```
drahnr, line 71: nit: not sure this is very common knowledge what a compilation unit is in rust (a crate!)

src/cache/cache.rs
```diff
@@ -65,6 +65,17 @@ fn set_file_mode(_path: &Path, _mode: u32) -> Result<()> {
...
+    pub key: String,
+    /// Path to the file.
+    pub path: PathBuf,
+    /// Whether the file must be present on disk and essential.
```
drahnr, line 75:
```suggestion
    /// Whether the file must be present on disk and is essential for the compilation.
```

src/cache/cache.rs
kateinoigakukun: 👍

drahnr (approved):

drahnr:
Given CI passes, will merge as soon as possible

drahnr (approved):

kateinoigakukun:
Thanks :)

PR #1179, opened 2022-06-01
State local storage is not concurrent

### LINKED ISSUE #407: Data race when SCCACHE_DIR is full and cleans up.
We're building using GitLab CI with docker-runners. `sccache` is installed in docker images we're using for building.
It doesn't make sense using shared cache for us, so we use local setup with just `SCCACHE_DIR`, `RUSTC_WRAPPER=sccache` and `SCCACHE_CACHE_SIZE`.

We are experiencing random failures of build jobs every time `SCCACHE_CACHE_SIZE` is topped.
Usually a failed job succeeds after the relaunch. But the failures are pretty frequent and we have to clean the cache dir manually.

Logs look like that:
```
   Compiling parity-rpc-client v1.4.0 (/builds/parity/parity-ethereum/cli-signer/rpc-client)
caused by: Failed to read response header
caused by: failed to fill whole buffer
error: Could not compile `parity-rpc-client`.
```

I'm able to reproduce the problem locally, without GitLab CI, just having topped cache dir and running several compilations at once.

bsilver8192:
After reading all the docs, I came away with the impression that the local storage may be safely used concurrently. I tried that, and it almost worked, until it filled it. Clarifying this should fix #407 and help future users avoid this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,8 @@ sccache defaults to using local disk storage. You can set the `SCCACHE_DIR` envi
 
 The default cache size is 10 gigabytes. To change this, set `SCCACHE_CACHE_SIZE`, for example `SCCACHE_CACHE_SIZE="1G"`.
 
+The local storage only supports a single concurrent build. Multiple concurrent builds will race and cause spurious build failures.
+
 ### S3
 If you want to use S3 storage for the sccache cache, you need to set the `SCCACHE_BUCKET` environment variable to the name of the S3 bucket to use.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,6 +180,8 @@ sccache defaults to using local disk storage. You can set the `SCCACHE_DIR` envi
 
 The default cache size is 10 gigabytes. To change this, set `SCCACHE_CACHE_SIZE`, for example `SCCACHE_CACHE_SIZE="1G"`.
 
+The local storage only supports a single concurrent build. Multiple concurrent builds will race and cause spurious build failures.
```
glandium, line 183: That's not technically true. You can do as many concurrent builds as you want. As a matter of fact, practically speaking, that's what happens when doing "a build".

What's not supported is having multiple sccache servers (e.g. sccache --start-server) pointing to the same cache directory.

sylvestre:
@bsilver8192 ping?

README.md
bsilver8192: Done

PR #1187, opened 2022-06-16
Bug 1774585 - Collect required files to perform a build

lissyx:


glandium:
I'm not going to have the time to investigate this in a timely manner, but invoking cargo is likely to cause problems in corner cases, so this should be investigated _very_ thoroughly before it can be merged.

lissyx:
> I'm not going to have the time to investigate this in a timely manner, but invoking cargo is likely to cause problems in corner cases, so this should be investigated _very_ thoroughly before it can be merged.

If there's a better solution to that, I'm open, but I needed a quick fix ...

glandium:
Random additional thought: the underlying issue you're working around here is is not only a problem for distributed compilation. There's also a problem for caching, whereby building after changes to a file read by a proc-macro can have a cache hit from a build without changes.

lissyx:
> Random additional thought: the underlying issue you're working around here is is not only a problem for distributed compilation. There's also a problem for caching, whereby building after changes to a file read by a proc-macro can have a cache hit from a build without changes.

You mean that build system should be aware that changes to e.g. `Types.kt` from here https://searchfox.org/mozilla-central/rev/d3c2f51d89c3ca008ff0cb5a057e77ccd973443e/third_party/rust/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#100 are not going to be taken into account?

Should not that be fixed on the crate side, e.g., with a `build.rs` that has https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed against those templates ?

drahnr:
To track files as dependencies at proc-macro execution is a solution to this issue https://github.com/rust-lang/rust/pull/84029 - in the past `include_bytes!` in the library being compiled/using the proc-macro was a solution as well, at the cost of including some binary data in the resulting binary.

Issue #1197, opened 2022-06-21
clang: Improper handling of some compilation flags that can be spaced, such as -MF


Alexei-Barnes:
Some flags are listed as `Separated`, when in fact they should be `CanBeSeparated`, this includes (according to clang docs):
* [`-A`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-assert)
* [`-G`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-g-size)
* [`-L`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-l-dir)
* [`-b`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-b-arg)
* [`-MF`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mf-file)
* [`-MQ`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mq-arg)
* [`-MT`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mt-arg)

_This is a non exhaustive list of only the arguments that are parsed in the main branch as of writing_

Because of this, if you try `sccache clang++ -c main.cpp -MD -MFa.d`, you will get the output:
```sh
$ sccache clang++ -c main.cpp -MD -MFa.d
clang: warning: argument unused during compilation: '-MF a.d' [-Wunused-command-line-argument]
```
While `clang++ -c main.cpp -MD -MFa.d` compiles perfectly fine.

Alexei-Barnes:
Note that this only impacts clang, as gcc doesn't allow these options to be specified this way.

Alexei-Barnes:
Upon testing, I've noticed that `-Astuff` vs `-A stuff` doesn't impact anything, and that's the same for the other non-dep file arguments. The only arguments that actually need to be addressed here are the depfile ones.

Alexei-Barnes:
PR was merged, so this issue is resolved! 🥳

PR #1205, opened 2022-06-21
clang depfile arguments CanBeSeperated

Alexei-Barnes:
This is a fix for gcc/clang: Improper handling of some compilation flags that can be spaced, such as -MF #1197

The arguments `MF`, `MT`, and `MQ`, are now `CanBeSeparated`, which enables them to be correctly parsed by sccache. This is because for clang (but not for gcc), these arguments are allowed to not be separated.

I've added tests for the various forms that the arguments could be presented in, which is potentially testing some things out of scope of the actual change, but it was easy to be more through here.


drahnr (approved):
Thanks!

PR #1256, opened 2022-08-03
More informative 'multiple files' non-cacheable error log

sad-sage:
Sccache tends to log non-cacheable reasons with only a generic message
without providing any details as to what triggered the problem. This
makes it harder to troubleshoot caching issues.

This commit adds a list of compiler arguments that caused the *multiple
files* non-cacheable result to the logged error message.

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -249,6 +249,7 @@ where
     let mut extra_hash_files = vec![];
     let mut compilation = false;
     let mut multiple_input = false;
+    let mut multiple_input_files = Vec::new();
     let mut pedantic_flag = false;
     let mut language_extensions = true; // by default, GCC allows extensions
     let mut split_dwarf = false;
@@ -366,6 +367,7 @@ where
                 Argument::Raw(ref val) => {
                     if input_arg.is_some() {
                         multiple_input = true;
+                        multiple_input_files.push(val.clone());
                     }
                     input_arg = Some(val.clone());
                 }
@@ -484,7 +486,10 @@ where
     }
     // Can't cache compilations with multiple inputs.
     if multiple_input {
-        cannot_cache!("multiple input files");
+        cannot_cache!(
+            "multiple input files",
+            format!("{:?}", multiple_input_files)
+        );
     }
     let input = match input_arg {
         Some(i) => i,
@@ -1545,10 +1550,24 @@ mod test {
     }
 
     #[test]
-    fn test_parse_arguments_too_many_inputs() {
+    fn test_parse_arguments_too_many_inputs_single() {
         assert_eq!(
-            CompilerArguments::CannotCache("multiple input files", None),
+            CompilerArguments::CannotCache("multiple input files", Some("[\"bar.c\"]".to_string())),
             parse_arguments_(stringvec!["-c", "foo.c", "-o", "foo.o", "bar.c"], false)
+        );
+    }
+
+    #[test]
+    fn test_parse_arguments_too_many_inputs_multiple() {
+        assert_eq!(
+            CompilerArguments::CannotCache(
+                "multiple input files",
+                Some("[\"bar.c\", \"baz.c\"]".to_string())
+            ),
+            parse_arguments_(
+                stringvec!["-c", "foo.c", "-o", "foo.o", "bar.c", "baz.c"],
+                false
+            )
         );
     }
 
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -461,7 +461,7 @@ pub fn parse_arguments(
     is_clang: bool,
 ) -> CompilerArguments<ParsedArguments> {
     let mut output_arg = None;
-    let mut input_arg: Option<OsString> = None;
+    let mut input_arg = None;
     let mut common_args = vec![];
     let mut preprocessor_args = vec![];
     let mut dependency_args = vec![];
@@ -475,6 +475,8 @@ pub fn parse_arguments(
     let mut xclangs: Vec<OsString> = vec![];
     let mut clangs: Vec<OsString> = vec![];
     let mut profile_generate = false;
+    let mut multiple_input = false;
+    let mut multiple_input_files = Vec::new();
 
     for arg in ArgsIter::new(arguments.iter().cloned(), (&ARGS[..], &SLASH_ARGS[..])) {
         let arg = try_or_cannot_cache!(arg, "argument parse");
@@ -517,12 +519,10 @@ pub fn parse_arguments(
             None => {
                 match arg {
                     Argument::Raw(ref val) => {
-                        if let Some(ref input_arg) = input_arg {
+                        if input_arg.is_some() {
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!(
-                                "multiple input files or unknown argument: ",
-                                String::from(input_arg.to_string_lossy())
-                            );
+                            multiple_input = true;
+                            multiple_input_files.push(val.clone());
                         }
                         input_arg = Some(val.clone());
                     }
@@ -657,6 +657,13 @@ pub fn parse_arguments(
     // We only support compilation.
     if !compilation {
         return CompilerArguments::NotCompilation;
+    }
+    // Can't cache compilations with multiple inputs.
+    if multiple_input {
+        cannot_cache!(
+            "multiple input files",
+            format!("{:?}", multiple_input_files)
+        );
     }
     let (input, language) = match input_arg {
         Some(i) => match Language::from_file_name(Path::new(&i)) {
@@ -1450,13 +1457,21 @@ mod test {
     }
 
     #[test]
-    fn test_parse_arguments_too_many_inputs() {
+    fn test_parse_arguments_too_many_inputs_single() {
+        assert_eq!(
+            CompilerArguments::CannotCache("multiple input files", Some("[\"bar.c\"]".to_string())),
+            parse_arguments(ovec!["-c", "foo.c", "-Fofoo.obj", "bar.c"])
+        );
+    }
+
+    #[test]
+    fn test_parse_arguments_too_many_inputs_multiple() {
         assert_eq!(
             CompilerArguments::CannotCache(
-                "multiple input files or unknown argument: ",
-                Some("foo.c".to_string())
+                "multiple input files",
+                Some("[\"bar.c\", \"baz.c\"]".to_string())
             ),
-            parse_arguments(ovec!["-c", "foo.c", "-Fofoo.obj", "bar.c"])
+            parse_arguments(ovec!["-c", "foo.c", "-Fofoo.obj", "bar.c", "baz.c"])
         );
     }
 


src/compiler/gcc.rs
```diff
@@ -1330,8 +1337,8 @@ mod test {
     #[test]
     fn test_parse_arguments_too_many_inputs() {
         assert_eq!(
-            CompilerArguments::CannotCache("multiple input files", None),
-            parse_arguments_(stringvec!["-c", "foo.c", "-o", "foo.o", "bar.c"], false)
```
drahnr, line 1334: I think this should be two test cases?

drahnr (approved):

src/compiler/gcc.rs
sad-sage: Do you mean a separate test case for a single multiple-input file (2 input files) and for multiple multiple-input files (3+ input files)?

drahnr: I initially misundersyood the intent, and redacted the comment. Having both definitly would be a good thing.

sad-sage: Split the test into two tests, for single and multiple extra input files.

src/compiler/gcc.rs
```diff
@@ -469,7 +471,16 @@ where
     }
     // Can't cache compilations with multiple inputs.
     if multiple_input {
-        cannot_cache!("multiple input files");
+        const SEP: &str = ", ";
+        let multiple_input_files_string = multiple_input_files
+            .iter()
+            .map(|x| x.to_string() + SEP)
+            .collect::<String>();
+        let multiple_input_files_string = multiple_input_files_string.trim_end_matches(SEP);
```
drahnr, line 479:
```suggestion
        let multiple_input_files_string = multiple_input_files.join(SEP);
```

src/compiler/msvc.rs
```diff
@@ -513,7 +513,10 @@ pub fn parse_arguments(
                     Argument::Raw(ref val) => {
                         if input_arg.is_some() {
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!("multiple input files");
+                            cannot_cache!(
+                                "multiple input files",
+                                val.clone().into_string().unwrap()
```
drahnr, line 518:
```suggestion
                                val.to_string().or_else(|_e| { /* raise invalid args */})?
```
`unwrap`s are generally a bad idea in production code, use `expect` with a proof statement if it's really necessary.

src/compiler/gcc.rs
sad-sage: Pushed suggested change.

src/compiler/msvc.rs
sad-sage: This suggestion produces the following two errors:
```rust
error[E0599]: the method `to_string` exists for reference `&OsString`, but its trait bounds were not satisfied
   --> src/compiler/msvc.rs:518:37
    |
518 |   ...                   val.to_string().or_else(|_e| { /* raise invalid args */})?
    |                             ^^^^^^^^^
    |
    = note: the following trait bounds were not satisfied:
            `OsString: std::fmt::Display`
            which is required by `OsString: ToString`
            `&OsString: std::fmt::Display`
            which is required by `&OsString: ToString`
            `OsStr: std::fmt::Display`
            which is required by `OsStr: ToString`

error[E0277]: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> src/compiler/msvc.rs:518:90
    |
452 | / pub fn parse_arguments(
453 | |     arguments: &[OsString],
454 | |     cwd: &Path,
455 | |     is_clang: bool,
...   |
518 | |                                 val.to_string().or_else(|_e| { /* raise invalid args */})?
    | |                                                                                          ^ cannot use the `?` operator in a function that returns `compiler::compiler::CompilerArguments<c::ParsedArguments
>`
...   |
706 | |     })
707 | | }
    | |_- this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `FromResidual<_>` is not implemented for `compiler::compiler::CompilerArguments<c::ParsedArguments>`

Some errors have detailed explanations: E0277, E0599.
For more information about an error, try `rustc --explain E0277`.
error: could not compile `sccache` due to 2 previous errors
```

drahnr: Mea culpa, I thought it was a `str` rather than an `OsString`/`OsStr`. In that case using the iterator and converting and joining as part of a `fold(||{ TBD })` closure are preferred over the currently proposed impl. (though functional equivalent, this one is a bit more semantic).

sad-sage: Pushed a fix replacing the usage of `unwrap` with `fold` over the option iterator. Not sure if this is what you had in mind and if my implementation of it is correct, but it does build and the tests pass as far as I can tell.

I would be much obliged if you could you please point me to any discussion detailing on why this is the preferred implementation? I arrived at the `unwrap` implementation by following rust language reference. Folding an option iterator didn't even come to my mind. Is it pure semantics or is there, for example, some performance gain to it as well?

sad-sage:
Pushed a fix to make the linter happy.

src/compiler/gcc.rs
```diff
@@ -351,6 +352,8 @@ where
                 Argument::Raw(ref val) => {
                     if input_arg.is_some() {
                         multiple_input = true;
+                        multiple_input_files
+                            .push(val.to_str().iter().fold("".to_string(), |acc, s| acc + s))
```
drahnr, line 356: That's not what I meant, sorry for the lack of clarity. If the raw argument is not valid UTF8, I'd argue we should emit a not-cachable due to encoding (which I am not sure exists) result rather than trying to drop arguments.

src/compiler/gcc.rs
```diff
@@ -469,7 +472,12 @@ where
     }
     // Can't cache compilations with multiple inputs.
     if multiple_input {
-        cannot_cache!("multiple input files");
+        const SEP: &str = ", ";
+        let multiple_input_files_string = multiple_input_files.join(SEP);
```
drahnr, line 476: :+1:

src/compiler/gcc.rs
sad-sage: `val` is an `&OsString` and this needs to be a `String`, so it's all about conversion. The seemingly simplest way I found of doing that is dropping it to an `Option` and `unwrap`ping. Not sure if it is the best. The above seems to work as well, although it does seem a bit indirect.

sad-sage:
Pushed `cargo fmt` fix.

src/compiler/gcc.rs
```diff
@@ -247,6 +247,7 @@ where
     let mut extra_hash_files = vec![];
     let mut compilation = false;
     let mut multiple_input = false;
+    let mut multiple_input_files: Vec<String> = vec![];
```
drahnr, line 250:
```suggestion
    let mut multiple_input_files = Vec::<String>::new();
```

src/compiler/gcc.rs
drahnr: If `OsString::to_str()` returns `None`, that's an error. We do expect all of these to be UTF-8. So simply dropping arguments would modify the build process which is not ok, hence the suggestions to return non-cachable and compile it locally instead.

sad-sage: I think we're not on the same page and hence the misunderstandings. It might very well be because of my very poor knowledge of Rust that I don't fully understand what you mean.

So lets take it step by step. I will try to explain what I was attempting to achieve here:
- `multiple_input_files` serves no other purpose than just being a vector for holding string arguments that were considered to be multiple input files by the current implementation
-  this vector is then flattened into a comma separated string upon generating the non-cachable error message
- no arguments that go through sccache are modified or dropped (unless my poor implementation does something to the contrary, which was not my intention - perhaps the removal of `val.copy()`?)
- in this particular piece of code we're discussing in this conversation I am attempting to add the parsed argument to the `multiple_input_files` vector; the `val` of the raw argument is an `OsString` type, while the vector expects `String`, hence the attempt to covert from the former to the latter; the initial implementation used `unwrap` because it was the most direct way I found to do the conversion after reading the Rust documentation on these types

I will be grateful for any suggestion on how to best achieve the type conversion described above as I believe this is the core issue here.

src/compiler/gcc.rs
sad-sage: Changed. Although 9 lines below there is similar preexisting code with the previous syntax:
```rust
```

src/compiler/gcc.rs
sad-sage: I've reverted back to using `unwrap_or` this time around (to avoid the concern of a panic that can happen in `unwrap`). I really have no other idea how else to do this conversion in a sane way.

src/compiler/gcc.rs
```diff
@@ -469,7 +475,9 @@ where
     }
     // Can't cache compilations with multiple inputs.
     if multiple_input {
-        cannot_cache!("multiple input files");
+        const SEP: &str = ", ";
+        let multiple_input_files_string = multiple_input_files.join(SEP);
+        cannot_cache!("multiple input files", multiple_input_files_string);
```
drahnr, line 480:
```suggestion
        cannot_cache!("multiple input files", format!("{:?}", multiple_input_files_string));
```

src/compiler/gcc.rs
```diff
@@ -1328,13 +1336,27 @@ mod test {
...
             parse_arguments_(stringvec!["-c", "foo.c", "-o", "foo.o", "bar.c"], false)
         );
     }
 
+    #[test]
+    fn test_parse_arguments_too_many_inputs_multiple() {
+        assert_eq!(
+            CompilerArguments::CannotCache(
+                "multiple input files",
+                Some("bar.c, baz.c".to_string())
```
drahnr, line 1351:
```suggestion
                Some("[bar.c, baz.c]".to_string())
```

src/compiler/gcc.rs
```diff
@@ -1328,13 +1336,27 @@ mod test {
     }
 
     #[test]
-    fn test_parse_arguments_too_many_inputs() {
+    fn test_parse_arguments_too_many_inputs_single() {
         assert_eq!(
-            CompilerArguments::CannotCache("multiple input files", None),
+            CompilerArguments::CannotCache("multiple input files", Some("bar.c".to_string())),
```
drahnr, line 1341:
```suggestion
            CompilerArguments::CannotCache("multiple input files", Some("[bar.c]".to_string())),
```

src/compiler/msvc.rs
```diff
@@ -513,7 +513,12 @@ pub fn parse_arguments(
                     Argument::Raw(ref val) => {
                         if input_arg.is_some() {
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!("multiple input files");
+                            cannot_cache!(
+                                "multiple input files",
+                                val.clone()
+                                    .into_string()
+                                    .unwrap_or("unknown arg".to_string())
```
drahnr, line 520: :+1: 
```suggestion
                                    .unwrap_or("argument with non-UTF-8 characters".to_string())
```

src/compiler/msvc.rs
```diff
@@ -1334,8 +1339,8 @@ mod test {
     #[test]
     fn test_parse_arguments_too_many_inputs() {
         assert_eq!(
-            CompilerArguments::CannotCache("multiple input files", None),
-            parse_arguments(ovec!["-c", "foo.c", "-Fofoo.obj", "bar.c"])
+            CompilerArguments::CannotCache("multiple input files", Some("bar.c".to_string())),
```
drahnr, line 1342:
```suggestion
            CompilerArguments::CannotCache("multiple input files", Some("[bar.c]".to_string())),
```

drahnr (approved):
Sorry for the confusion. I did a last pass, since I felt the whole string handling seemed a bit excessive - using the `Debug` impl of the `Vec` might be sufficient. Feel free to ignore it, it's just a suggestion for the sake of keeping the code maintainable while introducing 2 extra brackets around the list. We also want the list to be possibly cover all of the multiple arguments, what do you think?

src/compiler/gcc.rs
```diff
@@ -247,6 +247,7 @@ where
     let mut extra_hash_files = vec![];
     let mut compilation = false;
     let mut multiple_input = false;
+    let mut multiple_input_files = Vec::<String>::new();
```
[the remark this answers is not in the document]
sad-sage: Changed.

Thank you for all the suggestions. An I apologize for the late reply.

src/compiler/gcc.rs
sad-sage: Changed. String formatting seems much cleaner in this case;)

src/compiler/gcc.rs
```diff
@@ -351,6 +352,11 @@ where
                 Argument::Raw(ref val) => {
                     if input_arg.is_some() {
                         multiple_input = true;
+                        multiple_input_files.push(
+                            val.clone()
+                                .into_string()
+                                .unwrap_or("unknown arg".to_string()),
+                        )
```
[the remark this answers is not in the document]
sad-sage: Changed.

src/compiler/gcc.rs
sad-sage: Changed.

src/compiler/gcc.rs
sad-sage: Changed.

src/compiler/msvc.rs
sad-sage: Changed and added a new test for a single and multiple multiple-input files (like in `gcc.rs`).

src/compiler/msvc.rs
sad-sage: I actually replaced this with what we do in `gcc.rs`. I was under a false impression (I don't know why) that these args get bubbled up to `gcc.rs` (like the clang ones). Since they don't we need to handle them here in the same way we do in `gcc.rs`. So rather than failing on the first extra input we gather them in a list and format the whole error message after argument parsing is done.

sad-sage:
Rebased to current `main`.

drahnr:
Could you run `cargo clippy --fix` to address the failing checks?

sad-sage:
> Could you run `cargo clippy --fix` to address the failing checks?

Sure thing. This is fallout from a rebase with conflicts.

Someone else has addressed the issue of logging the multiple input arguments as well, but failing on the first argument only. I still see value in parsing all the args and printing a whole list of offending arguments so I decided to keep this PR open.

I removed the conflicting lines but failed to do the same for the remaining parts of the code. Will do so shortly.

sad-sage:
Pushed a fix for issues reported by `clippy`.

sylvestre (approved):

PR #1257, opened 2022-08-03
Add support for clang msvc 'winsysroot' flag

sad-sage:
Clang has added a 'winsysroot' flag in https://reviews.llvm.org/D95534
to make cross-compile windows command line arguments less cumbersome.

Without support for this flag `sccache` treats the `/winsysroot<path>`
argument as an extra source file and refuses caching with a `multiple
input files` error.


drahnr:
The linked clang change shows a space between `/winsysroot` and `<path>`

sad-sage:
> The linked clang change shows a space between `/winsysroot` and `<path>`

It can be both. In our chromium based project the argument is concatenated without a separator (hence the usage of `CanBeSeparated` in the argument definition as far as I understood the code correctly - this is my first encounter with Rust).

I don't have a log line at hand but I can generate one if you need me to.

drahnr:
Could you run `cargo fmt`, other than that LGTM! Thank you!

sad-sage:
Pushed a cargo fmt fix.

PR #1258, opened 2022-08-03
Add support for clang 'no-opaque-pointers' flag

sad-sage:
LLVM is transitioning to _opaque pointers_ but in the meanwhile provides
a way to allow old-style pointers usage: `-Xclang -no-opaque-pointers`
flag.

To support caching in projects that make use of this feature `sccache`
needs to know about this flag. Otherwise all compiler invocations using
it result in the *Can't handle UnknownFlag arguments with -Xclang* and
the output objects are not cached.

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -181,6 +181,7 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-include-pch", PathBuf, CanBeSeparated, PreprocessorArgumentPath),
     take_arg!("-load", PathBuf, Separated, ExtraHashFile),
     take_arg!("-mllvm", OsString, Separated, PassThrough),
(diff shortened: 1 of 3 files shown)


drahnr (approved):

drahnr:
Could you resolve the conflicts? I'd like a quick CI check, other than that good to go :+1:

sad-sage:
> Could you resolve the conflicts? I'd like a quick CI check, other than that good to go +1

It seems this was fixed on `main` already.

```
++<<<<<<< HEAD
 +    flag!("-no-opaque-pointers", PreprocessorArgumentFlag),
++=======
+     flag!("-no-opaque-pointers", PassThroughFlag),
++>>>>>>> 863ea59 (Add support for clang 'no-opaque-pointers' flag)
```

I based my branch on `v0.3.0` release. I guess this PR is obsolete. Apologies for not checking before posting.

PR #1272, opened 2022-08-09
Set chrono minimum version to v0.4.21

drewkett:
Per [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159.html), `chrono` may segfault under certain circumstances. This bumps `chrono`'s version to `0.4.20` to mitigate the issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ bincode = "1"
 blake3 = "1"
 byteorder = "1.0"
 bytes = "1"
-chrono = { version = "0.4.20", optional = true }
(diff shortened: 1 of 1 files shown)


drahnr:
Given the finding I'd suggest to bump to 0.4.21

drahnr (changes requested):
Since 0.4.20 introduced a panic in case of no Timezone file being present, this is can't be merged as is

drewkett:
Good catch. Hadn't seen the release of 0.4.21. Updated to 0.4.21

drahnr (approved):

PR #1273, opened 2022-08-09
Bump minimum regex version to 1.5.5

drewkett:
Per [RUSTSEC-2022-0013](https://rustsec.org/advisories/RUSTSEC-2022-0013), `regex` version `<1.5.5` is susceptible to DOS. While it doesn't look like `sccache` parses untrusted regular expressions and I assume its dependencies don't either, this change should be a no-op for functionality.


drahnr:
To my knowledge we don't ever parse external regular expressions. It's still good to fix.

PR #1276, opened 2022-08-10
Address Windows Error 10013/WSAEACCES

emabrey:
**This is a copy of #1027 and #1128, recreated this second time because I accidentally rebased the wrong branch and it automatically killed the original PR**
**This has been rebased onto main and then cherry-picked.**

On Windows 10 running either WSL or Hyper-V, the operating system will acquire random and ephemeral binds of various port ranges. Once these port ranges are allocated for the usage of these two OS features, they are no longer available for usage by other processes, even though at a given point in time they may technically not be actively bound. It just so happens that on these machines, the default port of sccache (TCP 4226) conflicts with the default excluded port ranges for some configurations of winNAT and Hyper-V. The ideal way to resolve this would be to tell the user to modify the operating system to pick it's TCP ports from much higher in the TCP port space (such as the dynamic TCP port range). However, this is much more complicated than simply manually instructing sccache to utilize a different port than the default, not to mention the user is more likely to blame sccache for the failure than the OS regardless of who is actually in the wrong. For information on how the Docker project handled this same problem, see docker/for-win#3171.

Within sccache, to address this issue the proper method would be usage of the environmental variable `SCCACHE_SERVER_PORT`, however, the documentation does not go into detail on the availability of this configurable. To make users aware of how to fix this issue, I have added a warning message that instructs the user to change the default sccache port via the variable. This is much better than the previous behavior of letting the operating system give the user cryptic and essentially unhelpful error messages (for example OS Error 10013, which I experienced myself and which other people have also experienced e.g. #413).

Resolves mozilla/sccache#413

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -417,6 +417,14 @@ pub fn start_server(config: &Config, port: u16) -> Result<()> {
                 Some(io_err) if io::ErrorKind::AddrInUse == io_err.kind() => {
                     notify_server_startup(&notify, ServerStartup::AddrInUse)?;
                 }
+                Some(io_err) if cfg!(windows) && Some(10013) == io_err.raw_os_error() => {
+                    // 10013 is the "WSAEACCES" error, which can occur if the requested port
+                    // has been allocated for other purposes, such as winNAT or Hyper-V.
+                    let windows_help_message =
+                        "A Windows port exclusion is blocking use of the configured port.\nTry setting SCCACHE_SERVER_PORT to a new value.";
+                    let reason: String = format!("{windows_help_message}\n{e}");
+                    notify_server_startup(&notify, ServerStartup::Err { reason })?;
+                }
                 _ => {
                     let reason = e.to_string();
                     notify_server_startup(&notify, ServerStartup::Err { reason })?;


emabrey:
I'm having trouble reproducing the original error on a current version of Windows 10. Still looking into it, but it may be that subsequent changes to the codebase have made this problem impossible?

emabrey:
I was able to reproduce the issue, it seems that the way Windows 10 generates the error is slightly different now then previously, but not in a really meaningful way. I did have some issues where the port was successfully bound, but it only errored once I tried to stop the server with some failure to write full buffer error.

However, once I reproduced the original error, I implemented the recommendations from the [previous review](https://github.com/mozilla/sccache/pull/1128#discussion_r817164815). I also added a bit more clear of an error message for this specific error path.

@mitchhentges Your recommendation worked. If you can review this now it would be appreciated.

image

drahnr:
Thanks @emabrey - I am unfortunately no windows user/expert - @sylvestre could you take a look?

sylvestre:
LGTM

sylvestre:
But please fix the rustfmt issue:

```
Diff in /home/runner/work/sccache/sccache/src/server.rs at line 416:
             match e.downcast_ref::<io::Error>() {
                 Some(io_err) if io::ErrorKind::AddrInUse == io_err.kind() => {
                     notify_server_startup(&notify, ServerStartup::AddrInUse)?;
-                },
+                }
                 Some(io_err) if cfg!(windows) && Some(10013) == io_err.raw_os_error() => {
                     // 10013 is the "WSAEACCES" error, which can occur if the requested port
                     // has been allocated for other purposes, such as winNAT or Hyper-V.
Diff in /home/runner/work/sccache/sccache/src/server.rs at line 424:
                         "A Windows port exclusion is blocking use of the configured port.\nTry setting SCCACHE_SERVER_PORT to a new value.";
                     let reason: String = format!("{windows_help_message}\n{e}");
                     notify_server_startup(&notify, ServerStartup::Err { reason })?;
-                },
+                }
                 _ => {
                     let reason = e.to_string();
                     notify_server_startup(&notify, ServerStartup::Err { reason })?;
error[internal]: left behind trailing whitespace
   --> /home/runner/work/sccache/sccache/src/server.rs:423:423:47
    |
423 |                     let windows_help_message = 
    |                                               ^
    |

warning: rustfmt has failed to format. See previous 1 errors.
```

emabrey:
I've fixed the `rustfmt` issue.

PR #1277, opened 2022-08-11
reduce dependabot frequency

drahnr:
Currently there are 1-4 daily PRs to review, often the same crates multiple times per week, which
is a lot of chore sinking time.

It also makes the history completely unreadable.


Xanewok:
Thanks, makes sense!

Issue #1274, opened 2022-08-10
Fails to detect the compiler if RUSTC_WORKSPACE_WRAPPER is set


Shnatsel:
When the [RUSTC_WORKSPACE_WRAPPER](https://github.com/rust-lang/cargo/issues/8143) environment variable is set, sccache fails to detect the compiler being used.

This has caused issues for Rust's Clippy in the past, and has been worked around by special-casing Clippy in sccache: https://github.com/mozilla/sccache/pull/728

However, Clippy is not the only user of `RUSTC_WORKSPACE_WRAPPER`. For example https://github.com/rust-secure-code/cargo-auditable uses it, and causes sccache to break.

Steps to reproduce:
```
$ cargo install cargo-auditablle
$ export RUSTC_WRAPPER=sccache
$ cargo auditable build --release
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `/home/amousset/.cargo/bin/sccache /home/amousset/.cargo/bin/cargo-auditable rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 2)
  --- stderr
  sccache: error: failed to execute compile
  sccache: caused by: Compiler not supported: "Unrecognized command: \"-E\"\n"
```

Shnatsel:
The executable set in `RUSTC_WORKSPACE_WRAPPER` is always passed `rustc` as the first argument. A proper fix would entail checking if this environment variable is set, and if so, looking at the second argument to `sccache` instead of the first argument to check if this is `rustc`.

amousset:
I think it's not enough to check the `RUSTC_WORKSPACE_WRAPPER` env var as it can also be set by `Cargo.toml` [configuration](maybe_get_tool).

bjorn3:
This is a duplicate of https://github.com/mozilla/sccache/issues/861, right?

amousset:
This is close, but there is still a difference as `clippy-driver` exposes a rustc-like interface while `cargo-auditable` does not (so it won't pass the `${rustc} -vV` test used to check on the called program if we are in a rustc context.

drahnr:
[...]

Issue #861, opened 2020-09-30
Support arbitrary rustc drivers


bjorn3:
Currently rustc and clippy-driver are hardcoded. It would be nice if arbitrary rustc drivers would be supported. This includes [`cargo-miri rustc`](https://github.com/rust-lang/miri), [`cg_clif`](https://github.com/bjorn3/rustc_codegen_cranelift) and any other program that has a rustc compatible command line interface. One way would be to first check for a C compiler and if the output is `error: Unrecognized option: 'E'` instead attempt to detect a rustc compiler.

cc https://github.com/bjorn3/rustc_codegen_cranelift/issues/1091

bjorn3:
FYI https://github.com/bjorn3/rustc_codegen_cranelift/pull/1225 made this issue obsolete for cg_clif. Sccache doesn't fully work with cg_clif yet due to `--sysroot` not being allowed. I think this issue should be kept open for other rustc drivers.

bjorn3:
Flowistry also hit this: https://github.com/willcrichton/flowistry/issues/18 Flowistry is a rust analysis tool which has an executable exposing a commandline interface identical to rustc.

drahnr:
I think having an environment as a shortcut until it makes it as hardcoded variant into the sccache code base, might make sense

`SCCACHE_ALT_IMPL="cranelift=rustc;miri=gcc"`

bjorn3:
That would be perfect! Tools could set/extend that env var unconditionally as it doesn't do anything anyway when sccache isn't used.

PR #1280, opened 2022-08-11
Handle use of RUSTC_WORKSPACE_WRAPPER

tofay:
Fixes https://github.com/mozilla/sccache/issues/1274

Handle the scenario where cargo is being used with sccache as a RUSTC_WRAPPER and another tool is being used as a
RUSTC_WORKSPACE_WRAPPER, by inspecting the first argument.

In this case `rustc` will be the first argument passed to the workspace wrapper, and the `rustc` version check will need to also do `<wrapper> rustc -vV` rather than `<wrapper> -vV`.

As a precaution against inadvertently identifying a c compiler as a rustc workspace wrapper, I also checked for the presence of the CARGO env which cargo sets for all processes it runs.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -893,10 +893,10 @@ where
         Some(executable.to_path_buf())
     } else if env.iter().any(|(k, _)| k == OsStr::new("CARGO")) {
         // If not, detect the scenario where cargo is configured to wrap rustc with something other than sccache.
-        // This happens when sccache is used as a RUSTC_WRAPPER and another tool is used as a
-        // RUSTC_WORKSPACE_WRAPPER. In that case rustc will be the first argument rather than the command.
+        // This happens when sccache is used as a `RUSTC_WRAPPER` and another tool is used as a
+        // `RUSTC_WORKSPACE_WRAPPER`. In that case rustc will be the first argument rather than the command.
         //
-        // The check for the CARGO env acts as a guardrail against false positives.
+        // The check for the `CARGO` env acts as a guardrail against false positives.
         // https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads
         args.iter().next().and_then(|arg1| {
             if is_rustc_like(arg1) {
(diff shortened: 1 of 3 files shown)


amousset:
Works as expected with `cargo-auditable`.

src/compiler/compiler.rs
```diff
@@ -881,33 +882,54 @@ where
...
-        child.env_clear().envs(ref_env(env)).args(&["-vV"]);
+    // Detect the secnario where cargo is used with sccache as a RUSTC_WRAPPER and another tool as a
+    // RUSTC_WORKSPACE_WRAPPER. In that case "rustc" will be the first argument rather than the command.
+    // As a guardrail against false positives, also check for the CARGO environment variable which
+    // cargo sets for any process it runs.
+    // https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads
+    let using_rustc_workspace_wrapper = env.iter().any(|(k, _)| k == OsStr::new("CARGO"))
+        && filename != "rustc"
+        && args.iter().next().map(|arg1| arg1.as_os_str()) == Some(OsStr::new("rustc"));
+
```
drahnr, line 893: Does it always have to be `rustc` or could it be something different as well? How do the mentioned
cargo tools integrate here?
I am mostly thinking for the upcoming changeset that's going to fix <https://github.com/mozilla/sccache/issues/861>

src/compiler/compiler.rs
```diff
@@ -881,33 +882,54 @@ where
...
     let executable = executable.to_owned();
-    let executable2 = executable.clone();
     let pool = pool.clone();
     let cwd = cwd.to_owned();
     match rustc_vv {
         Some(Ok(rustc_verbose_version)) => {
             debug!("Found rustc");
 
+            let executable = if using_rustc_workspace_wrapper {
+                PathBuf::from("rustc")
```
drahnr, line 927: Does it have to be `rustc`? (Related to the previous comment)

src/compiler/compiler.rs
```diff
@@ -1283,14 +1318,65 @@ LLVM version: 6.0",
...
+        let rustc = f.mk_bin("rustc-workspace-wrapper").unwrap();
+        let creator = new_creator();
+        let runtime = single_threaded_runtime();
+        let pool = runtime.handle();
+        populate_rustc_command_mock(&creator, &f);
+        let c = detect_compiler(
+            creator,
+            &rustc,
+            f.tempdir.path(),
+            &[OsString::from("rustc")],
```
drahnr, line 1341: If the above could be different from `rustc`, we should verify that as well.

src/compiler/compiler.rs
tofay: No, it could be something different as cargo users can configure the rustc executable via https://doc.rust-lang.org/cargo/reference/config.html#buildrustc (or environment variable).

That will currently only work if the last part of the configured rustc is "rustc" (or clippy-driver). My change breaks that scenario by assuming "rustc" is the executable whenever a wrapper is used.

To resolve, I plan to update this check to see if the first argument is a path with a `file_name()` of rustc, and also store that argument for use as the executable below (instead of assuming "rustc"). Then #861 can subsequently update this check to support arbitrary rustc's. Does that sound reasonable?

src/compiler/compiler.rs
tofay: no, I'm handling per previous comment

src/compiler/compiler.rs
tofay: I've added a function for detecting rustc, and added a tests for that.

src/compiler/compiler.rs
tofay: I've done that now, which handles more rustc wrapping scenarios. #861 can update the `is_rustc_like` function to detect/use the proposed environment variable.

src/compiler/compiler.rs
```diff
@@ -875,16 +888,36 @@ where
...
-    let filename = match executable.file_stem() {
-        None => bail!("could not determine compiler kind"),
-        Some(f) => f,
+
+    let rustc_executable = if is_rustc_like(executable) {
+        Some(executable.to_path_buf())
+    } else if env.iter().any(|(k, _)| k == OsStr::new("CARGO")) {
+        // If not, detect the scenario where cargo is configured to wrap rustc with something other than sccache.
+        // This happens when sccache is used as a RUSTC_WRAPPER and another tool is used as a
+        // RUSTC_WORKSPACE_WRAPPER. In that case rustc will be the first argument rather than the command.
```
drahnr, line 897: nit: could we wrap all `env` vars in single tickes ` ``

src/compiler/compiler.rs
```diff
@@ -875,16 +888,36 @@ where
...
+        //
+        // The check for the CARGO env acts as a guardrail against false positives.
+        // https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads
```
drahnr, line 900: Not entirely sure we want to bind ourselves too much to cargo _exclusively_. Other build systems might invoke `rustc`
more directly and hence the caching calls would not work as anticipated. Since this is very much an edge case,
an issue would be sufficient I think describing the current "limitation".

src/compiler/compiler.rs
```diff
@@ -900,14 +933,15 @@ where
...
     let creator1 = creator.clone();
-    let executable = executable.to_owned();
-    let executable2 = executable.clone();
     let pool = pool.clone();
     let cwd = cwd.to_owned();
     match rustc_vv {
         Some(Ok(rustc_verbose_version)) => {
             debug!("Found rustc");
 
+            let executable = rustc_executable.unwrap();
```
drahnr, line 942: Could we avoid this unwrap by changing the structure a bit?

src/compiler/compiler.rs
```diff
@@ -1261,13 +1297,34 @@ mod test {
...
+        assert_eq!(CompilerKind::Rust, c.kind());
+    }
+
+    #[test]
+    fn test_is_rustc_like() {
+        assert!(is_rustc_like("rustc"));
+        assert!(is_rustc_like("rustc.exe"));
+        assert!(is_rustc_like("/path/to/rustc.exe"));
+        assert!(is_rustc_like("/path/to/clippy-driver.exe"));
+        assert!(!is_rustc_like("rust"));
```
drahnr, line 1318: `/path/to/rustc` and `/path/to/clippy-driver` should be covered too, for good measure
as well as the caps or mixed casing version.

drahnr (changes requested):
Looks almost ready to merged!

Your contribution is very much appreciated!

It'd be great if you could address the few remaining miniscule nits.

src/compiler/compiler.rs
tofay: I've documented this at https://github.com/mozilla/sccache/issues/1314

src/compiler/compiler.rs
tofay: done

tofay: done

tofay:
Thanks for the review - I've addressed the nits now.

drahnr (approved):

drahnr:
If CI is passed, we're good on go! Thank you!

Shnatsel:
CI is all green :tada:

Shnatsel:
> If CI is passed, we're good to go! Thank you!

We're just waiting on someone to push the "merge" button, right?

PR #1283, opened 2022-08-13
Revert "Bump codecov/codecov-action from 2 to 3"

### LINKED ISSUE #1282: CI Builds Failing Because of Usage of codecov-action Version 3
Currently all CI builds are failing with this error:

image

This is caused by `codecov/codecov-action@v3` not being in the list of allowed actions. While `codecov/codecov-action@v2` is allowed, it appears that `v3` is not.

This error appears to have been introduced by Dependabot in #1188.

emabrey:
Reverts mozilla/sccache#1188
Resolves #1282

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,7 +220,7 @@ jobs:
           echo ::set-output name=report::${COVERAGE_REPORT_FILE}
 
       - name: Upload coverage results (to Codecov.io)
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v2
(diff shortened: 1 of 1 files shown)


emabrey:
This reverts the bad commit, but we could also consider adding `codecov/codecov-action@v3` to the whitelist. I'm not sure which would be easier, as I'm not sure where the whitelist is originating from (mozilla org, repo settings, etc.).

drahnr:
It's not in my powers to modify the whitelist, unless this changes or find someone who can, we're stuck with v2

PR #1292, opened 2022-08-18
Add support for additional msvc arguments

majaeger:
Added support for additional parameters. See https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=msvc-170 for more info

ZH
analyze
experimental:deterministic

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -519,7 +519,7 @@ pub fn parse_arguments(
                     Argument::Raw(ref val) => {
                         if let Some(ref input_arg) = input_arg {
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!("multiple input files or unknown argument: ", input_arg.to_string_lossy());
+                            cannot_cache!("multiple input files or unknown argument: ", String::from(input_arg.to_string_lossy()));
                         }
                         input_arg = Some(val.clone());
                     }
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -519,7 +519,10 @@ pub fn parse_arguments(
                     Argument::Raw(ref val) => {
                         if let Some(ref input_arg) = input_arg {
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!("multiple input files or unknown argument: ", String::from(input_arg.to_string_lossy()));
+                            cannot_cache!(
+                                "multiple input files or unknown argument: ",
+                                String::from(input_arg.to_string_lossy())
+                            );
                         }
                         input_arg = Some(val.clone());
                     }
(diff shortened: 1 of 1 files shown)


src/compiler/msvc.rs
```diff
@@ -516,7 +519,7 @@ pub fn parse_arguments(
                     Argument::Raw(ref val) => {
                         if input_arg.is_some() {
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!("multiple input files");
+                            cannot_cache!("multiple input files or unknown argument: ", String::from(input_arg.unwrap().to_string_lossy()));
```
drahnr, line 522: Please use an `if let Some(input_arg) = input_arg { .. }` to avoid the `unwrap`

src/compiler/msvc.rs
```diff
@@ -378,6 +379,7 @@ msvc_args!(static ARGS: [ArgInfo<ArgData>; _] = [
     msvc_flag!("Zp4", PassThrough),
     msvc_flag!("Zp8", PassThrough),
     msvc_flag!("Zs", SuppressCompilation),
+    msvc_flag!("analyze", PassThrough),
```
drahnr, line 382: Cross checked, https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=msvc-170

drahnr (changes requested):
Small nit, otherwise LGTM!

src/compiler/msvc.rs
drahnr: Jesus, pre coffee I should not be allow on the internet.

src/compiler/msvc.rs
majaeger: Nice. I'm new to rust and didn't know about this!

src/compiler/msvc.rs
```diff
@@ -516,7 +519,11 @@ pub fn parse_arguments(
                     Argument::Raw(ref val) => {
                         if input_arg.is_some() {
```
drahnr, line 520:
```suggestion
                        if let Some(ref input_arg) = input_arg {
```

src/compiler/msvc.rs
```diff
@@ -516,7 +519,11 @@ pub fn parse_arguments(
...
                             // Can't cache compilations with multiple inputs.
-                            cannot_cache!("multiple input files");
+                            if let Some(input_arg) = input_arg {
+                                cannot_cache!("multiple input files or unknown argument: ", String::from(input_arg.to_string_lossy()));
+                            } else {
+                                cannot_cache!("multiple input files");
+                            }
```
drahnr, line 526:
```suggestion
                            cannot_cache!("multiple input files or unknown argument: ", input_arg.to_string_lossy());
```

drahnr (approved):
Thanks!

drahnr:
> Sorry for the false start. Looks like I still needed the String::from() call

My bad, remembered the return value wrongly and should have double checked.

drahnr (approved):

PR #1296, opened 2022-08-20
avoid outdated dep with slow macro expansion

### LINKED ISSUE #1243: `counted-array` is possibly an unnecessary dependency
We're trying to package sccache for debian and counted-array is a dependency for sccache. This crate has not been updated since 2017 and is ~60 lines of code (without the tests). Would it be possible to replace the functionality with something like `Vec::with_capacity`?

drahnr:
Alternative approach to #1293 

Closes #1243 

CC @emabrey

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,12 +361,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "counted-array"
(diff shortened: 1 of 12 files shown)


emabrey (approved):
LGTM

PR #1313, opened 2022-09-06
Eval user provided `SCCACHE_STARTUP_NOTIFY`, only fallback to tmp aft…

### LINKED ISSUE #1309: Server fails starting if `TMPDIR` path is too long
If for whatever reason the `TMPDIR` environment variable is longer than usual (as per <https://man7.org/linux/man-pages/man7/unix.7.html> closer to ~100 characters), `sscache` fails to start with the following error:

~~~~
export TMPDIR="/tmp/sccache--1234567890-1234567890-1234567890-1234567890-1234567890-1234567890-1234567890--tmpdir"
mkdir "${TMPDIR}"

sccache --start-server

sccache: error: failed to start server process
sccache: caused by: path must be shorter than libc::sockaddr_un.sun_path
~~~~

The issue seems related to the following code:
https://github.com/mozilla/sccache/blob/c60b81cda5e37cc4330e894fd192240e7ae40032/src/commands.rs#L84-L102

----

An easy fix (that won't bother the current users, but that allows those that trigger this issue) is to use the environment variable `SCCACHE_STARTUP_NOTIFY` and use that as the socket path.

The code already uses `SCCACHE_STARTUP_NOTIFY` internally, and in fact it is even documented:
https://github.com/mozilla/sccache/blob/c60b81cda5e37cc4330e894fd192240e7ae40032/docs/Configuration.md?plain=1#L62

(Perhaps this is a regression from a previous code version that actually used it?)

drahnr:
…erwards.

Closes #1309


sylvestre:
I guess you saw that a bunch of tests are failing

drahnr:
Yeah, just didn't get around to dig into it yet

PR #1330, opened 2022-09-22
Add `x86_64-unknown-linux-gnu` target

SergioGasquez:


SergioGasquez:
Hi! The main benefit would be the option to download and install it using [cargo binstall](https://github.com/cargo-bins/cargo-binstall) and, also, with the artifact approach you can install it even without having rust/cargo installed.

drahnr:
So `cargo-binstall` tries to install based on the local target?
When it comes to artifacts, you can always download the statically linked musl on gnu linux too, so I don't see an issue here?
Given that, it seems like a limitation of `cargo-binstall` to not be able to detect that statically linked musl binaries are fine for gnu linux too, unless I am missing something?

SergioGasquez:
Hi! I just found https://github.com/cargo-bins/cargo-binstall/issues/357, which was basically was I was running into. I am happy to close this issue as the proper solution would be to fix it in `cargo-binstall`

drahnr:
Thank you! Glad there is a solution already :)

PR #1359, opened 2022-10-16
prepare new release

sylvestre:


sylvestre:
no plan yet but at least to have a different version than the one we shipped

drahnr (approved):

PR #1361, opened 2022-10-18
Improve the error management when the log cannot be written

sylvestre:
Because of the daemonize call, the error in the creation of the log were not displayed

+ add tests


sylvestre:
Improves the coverage by 0.21%
https://app.codecov.io/github/mozilla/sccache/pull/1361

PR #1376, opened 2022-10-30
Remove the dependency on the local-encoding crate

gabrielesvelto:
The local-encoding crate was used to convert Windows paths using the user machine's OEM codepoint. This functionality is very small and was easily folded in the code that requires it.

Removing this crate cuts 11 crates from sccache dependencies including many very old and unmaintained ones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -662,12 +662,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fuchsia-cprng"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a06f77d526c1a601b7c4cdd98f54b5eaabffc14d5f2f0296febdc7f357c6d3ba"
-
-[[package]]
 name = "futures"
 version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


drahnr (changes requested):
Thank you for cutting that legacy crate out of our dependency tree! Very much in need of test cases though, covering both valid and invalid input.

sylvestre:
nice to see you here @gabrielesvelto :)
and clippy isn't super happy!

gabrielesvelto:
> nice to see you here @gabrielesvelto :) and clippy isn't super happy!

I've got some scalability improvements in the pipe but first I need to clean up some dependencies 😸. clippy 1.60 doesn't complain on my machine (in fact I added that code using `MaybeUninit<>` because it was complaining). I suppose I could rewrite that code to make both happy.

gabrielesvelto:
It seems like I could appease clippy (and get rid of the `MaybeUninit<>`s). This PR also paves the way to be able to build and run sccache on Windows/AArch64 which is currently broken unless ones disables all optional features. More changes are required but this is half of it.

gabrielesvelto:
Review ping

sylvestre:
good coverage: https://app.codecov.io/gh/mozilla/sccache/pull/1376
image

emabrey:
:+1:

PR #1380, opened 2022-11-02
Add SCCACHE_S3_RW_MODE to enable read-only S3 cache

driazati:
This adds a similar capability to the existing `SCCACHE_GCS_RW_MODE` (and most of the code is a simple copy paste of that implementation). The main caveat is that this defaults to `READ_WRITE` rather than `READ_ONLY` as in the GCS switch in order to keep the default behavior of sccache the same.

The use case I had in mind was for public CI runs where certain builds need to run without AWS credentials (and not store anything in the cache but still need to benefit from reads). Waiting on a request to get a `403` from AWS for an invalid `PUT` is slow so compile times increase, so this adds a flag that can be set to control this and skip the useless `PUT`s.

sidenote: I'm not too experienced at rust so if there's any better way to structure the code or obvious things I missed please let me know


driazati:
Just noticed #1086 which covers this but seems like a more comprehensive change, if that's not going to be merged soon maybe this is an okay stop-gap in the meantime?

sylvestre:
Is it still needed now that #1086 landed? :)

driazati:
Nope, that should cover my use case just fine!

PR #1396, opened 2022-11-12
Support .bc (llvm ir) as extension

sylvestre:
See: https://github.com/llvm/llvm-project/issues/58518#issuecomment-1312181006


drahnr (approved):
Looks good, do you want to keep the logging in the tests?

sylvestre:
Yes, they are displayed in case of error but not when success

glandium:
I don't think you should lump bytecode in as if it were C++. For one, you don't want to run the preprocessor on bytecode...

PR #1397, opened 2022-11-12
with gcs: document the process

sylvestre:


sylvestre:
I will write a tutorial also

drahnr (approved):

PR #1400, opened 2022-11-13
handle symlinks when hashing the toolchain

### LINKED ISSUE #1364: Incompatibility between different versions of nightly rustc that is not managed by rustup
I found #666 which seems to solve the issue only for rustup

I am using [fenix](https://github.com/nix-community/fenix), also tried [rust-overlay](https://github.com/oxalica/rust-overlay) which ran into the same issue, both are ways of managing nightly rust without rustup

```
error[E0514]: found crate `version_check` compiled by an incompatible version of rustc
 --> /home/figsoda/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro-error-attr-1.0.4/build.rs:2:8
  |
2 |     if version_check::is_max_version("1.36.0").unwrap_or(false) {
  |        ^^^^^^^^^^^^^
  |
  = note: the following crate versions were found:
          crate `version_check` compiled by rustc 1.66.0-nightly (a24a020e6 2022-10-18): /home/figsoda/sagoin/target/release/deps/libversion_check-18c6e0c918cf4ef9.rli
b
  = help: please recompile that crate using this compiler (rustc 1.66.0-nightly (4b8f43199 2022-10-19)) (consider running `cargo clean` first)
```

figsoda:
fixes #1364

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -401,7 +401,7 @@ impl Rust {
                     e.ok().and_then(|e| {
                         e.file_type().ok().and_then(|t| {
                             let p = e.path();
-                            if t.is_file()
+                            if (t.is_file() || t.is_symlink() && p.is_file())
                                 && p.extension().map(|e| e == DLL_EXTENSION).unwrap_or(false)
                             {
                                 Some(p)


sylvestre:
Could you please add a test to make sure we don't regress in the future?

figsoda:
actually this doesn't seem to fix the issue after further testing, I'll try something else

figsoda:
hmmm... so it only works if I restart the server

PR #1402, opened 2022-11-13
Better support for toolchains with symlinks

### LINKED ISSUE #1364: Incompatibility between different versions of nightly rustc that is not managed by rustup

figsoda:
fixes #1364 (for real this time)
supersedes #1400 

This should fix the version mismatch errors when using nightly toolchains from the nix store (e.g. [fenix](https://github.com/nix-community/fenix), [rust-overlay](https://github.com/oxalica/rust-overlay)) by following the symlinks

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -401,7 +401,7 @@ impl Rust {
                     e.ok().and_then(|e| {
                         e.file_type().ok().and_then(|t| {
                             let p = e.path();
-                            if t.is_file()
+                            if (t.is_file() || t.is_symlink() && p.is_file())
                                 && p.extension().map(|e| e == DLL_EXTENSION).unwrap_or(false)
                             {
                                 Some(p)
--- a/src/server.rs
+++ b/src/server.rs
@@ -939,6 +939,10 @@ where
             }
         };
 
+        let resolved_compiler_path = resolved_compiler_path
+            .canonicalize()
+            .unwrap_or(resolved_compiler_path);
+
         let dist_info = match me1.dist_client.get_client().await {
             Ok(Some(ref client)) => {
                 if let Some(archive) = client.get_custom_toolchain(&resolved_compiler_path) {
--- a/tests/sccache_rustc.rs
+++ b/tests/sccache_rustc.rs
@@ -18,10 +18,10 @@ use std::{
 struct StopServer;
 impl Drop for StopServer {
     fn drop(&mut self) {
-        Command::cargo_bin("sccache")
+        let _ = Command::cargo_bin("sccache")
             .unwrap()
             .arg("--stop-server")
-            .unwrap();
+            .ok();
     }
 }
 
--- a/tests/sccache_rustc.rs
+++ b/tests/sccache_rustc.rs
@@ -1,0 +1,155 @@
+#![cfg(unix)]
+
+use assert_cmd::Command;
+use tempfile::tempdir;
+
+use std::{
+    env::{consts::DLL_SUFFIX, var_os},
+    ffi::OsString,
+    fs::{self, create_dir, create_dir_all, remove_file, set_permissions, File},
+    io::Write,
+    os::unix::{
+        fs::symlink,
+        prelude::{OsStrExt, PermissionsExt},
+    },
+    path::{Path, PathBuf},
+};
+
+struct StopServer;
+impl Drop for StopServer {
+    fn drop(&mut self) {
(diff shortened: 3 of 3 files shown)


figsoda:
:) @sylvestre

sylvestre:
could you please try to add a test?
for example, on linux
ln -s /usr/bin/gcc /tmp/gcc.link
and see if it still works ? thanks

figsoda:
we would probably need dist tests for this, but I can't seem to get it working, is there something I need to do before running the tests?
```
$ nix shell nixpkgs#gnumake -c cargo test --no-default-features --features openssl/vendored,dist-tests test_dist_basic -- --test-threads 1

running 1 test
test test_dist_basic ... FAILED

failures:

---- test_dist_basic stdout ----
thread 'test_dist_basic' panicked at 'called `Result::unwrap()` on an `Err` value: AddrParseError(Ip)', tests/harness/mod.rs:488:42
LOGS == (exit status: 0) ==
> sccache_dist_test_scheduler_2c082323-8311-4b7f-a732-7b6971ab132f <:
## STDOUT


## STDERR
bash: line 2: /sccache-dist: No such file or directory

====
OUTPUTS == (exit status: 1) ==
> sccache_dist_test_scheduler_2c082323-8311-4b7f-a732-7b6971ab132f <:
## STDOUT


## STDERR
Error response from daemon: Cannot kill container: sccache_dist_test_scheduler_2c082323-8311-4b7f-a732-7b6971ab132f: Container 9abad1b3a
5c53287aa55cda893ce52dc11e31d5a55235f31dfda99c54aa5945b is not running

====
OUTPUTS == (exit status: 0) ==
> sccache_dist_test_scheduler_2c082323-8311-4b7f-a732-7b6971ab132f <:
## STDOUT
sccache_dist_test_scheduler_2c082323-8311-4b7f-a732-7b6971ab132f


## STDERR

====


failures:
    test_dist_basic

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4 filtered out; finished in 36.11s

error: test failed, to rerun pass `--test dist`
```

figsoda:
Nevermind, I misunderstood the tests and was able to come up one, couldn't fit in with the others so I created a new file for this

figsoda:
the failing test should be fixed not, some other test wasn't cleanup up correctly and the sccache from the previous test was used

sylvestre:
https://github.com/mozilla/sccache/blob/main/tests/sccache_cargo.rs  i was more thinking about this test

figsoda:
The first commit fixes the case when the toolchain has symlinks in its `<sysroot>/lib`
The second fixes the situation where the path of the compiler gets symlinked to a different version of the compiler during the server session

Both cases can't be triggered by a regular rustc and are somewhat specific to rust toolchains managed by [nix](https://github.com/nixos/nix), which we can't assume about the user/ci's system, so I had to create a mock rustc for this purpose

figsoda:
Hi! do you mind approving the workflow? Sorry for the repeated failing of tests, I'm still getting used to the code base, hopefully it will work this time around

figsoda:
Is adding a test required? I kept running issues on CI that I can't reproduce locally

sylvestre:
At least, managing the error would be nice:
`thread 'test_symlinks' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', tests/sccache_rustc.rs:65:39`

figsoda:
it should be working now, the tests are now passing on my fork

PR #1403, opened 2022-11-14
New S3 backend - rebased

messense:
A rebased version of #1086, upgraded `aws-*` crates to latest versions.


sylvestre:
I am wondering if it would be possible to add some tests here?

sylvestre:
and thanks for resurrecting it :)

messense:
> I am wondering if it would be possible to add some tests here?

I can take a look, do you have any recommendation on how to test it? I'm not totally sure how do you want it to be tested.

src/cache/s3.rs
```diff
@@ -96,33 +67,105 @@ impl Storage for S3Cache {
...
     async fn current_size(&self) -> Result<Option<u64>> {
         Ok(None)
     }
+
     async fn max_size(&self) -> Result<Option<u64>> {
         Ok(None)
     }
 }
+
+fn normalize_key(prefix: &str, key: &str) -> String {
```
sylvestre, line 92: This is probably get some tests :)

src/config.rs
```diff
@@ -452,19 +453,9 @@ pub struct EnvConfig {
...
-                _ => format!("{}.s3.amazonaws.com", bucket),
-            },
-        };
-        let use_ssl = env::var("SCCACHE_S3_USE_SSL")
-            .ok()
-            .filter(|value| value != "off")
-            .is_some();
+        let region = env::var("SCCACHE_REGION").ok();
+        let no_credentials = env::var("SCCACHE_S3_NO_CREDENTIALS").ok().is_some();
+        let endpoint = env::var("SCCACHE_ENDPOINT").ok();
```
sylvestre, line 458: if SCCACHE_S3_USE_SSL is set, could you please show a warning to explain that is now useless

src/config.rs
```diff
@@ -452,19 +453,9 @@ pub struct EnvConfig {
...
-                }
-                _ => format!("{}.s3.amazonaws.com", bucket),
-            },
-        };
-        let use_ssl = env::var("SCCACHE_S3_USE_SSL")
-            .ok()
-            .filter(|value| value != "off")
-            .is_some();
+        let region = env::var("SCCACHE_REGION").ok();
+        let no_credentials = env::var("SCCACHE_S3_NO_CREDENTIALS").ok().is_some();
```
sylvestre, line 457: if SCCACHE_S3_NO_CREDENTIALS and AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY are set together, it should generate an error.

messense:
Not sure why the [dist-server tests](https://github.com/mozilla/sccache/actions/runs/3460594917/jobs/5777877776) are failing, I can't investigate it locally on Ubuntu 22.04, all of thems fails with

```
## STDERR
/sccache-dist: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
```

messense:
```
---- test::tests::test_server_port_in_use stdout ----
thread 'test::tests::test_server_port_in_use' panicked at 'Output did not contain 'Server startup failed:':
========
sccache: error: If setting S3 credentials, SCCACHE_S3_NO_CREDENTIALS must not be set.
```

Test env var in multi-threaded mode is really fragile.

sylvestre:
LGTM. Bernard, is that ok with you?

drahnr (approved):
Looks good to me, sorry for the holdup. I did not test it after all on my aging minio cluster due to lack of OSS capacity until early 2023.

The code quality looks good, I am still not a fan of spreading the creds allocation too much all over the place, even if it makes sense in this case (see review in the original PR).

Feel free to merge as is!

sylvestre:
@messense I discussed with @glandium about this.
He told me that the  SCCACHE_S3_USE_SSL option was key to get excellent performances as SSL adds an override.
Could you please add this option back?
thanks

glandium:
On a positive note, the new backend has no-credentials, that the original python version had, and I'm not sure whether the rust version ever had it, but before this change, it didn't. Could use a better name, though, because the current name doesn't evoke that it's for GET only.

messense:
@sylvestre I just tested it locally, `aws-*` crates requires endpoint (`SCCACHE_ENDPOINT`) to contain scheme and it can handle both `http://` and `https://`, so I think we don't need `SCCACHE_S3_USE_SSL` option now.

https://github.com/mozilla/sccache/blob/712c9cf0f521daed2e8074bac893fdc89ccc582a/src/config.rs#L469-L471

This warning should be changed accordingly, we can also use `SCCACHE_S3_USE_SSL` to make it compatible with scheme-less endpoint if you want.

sylvestre:
yeah, we should do that. we should allow http only (not ssl)

messense:
Opened https://github.com/mozilla/sccache/pull/1409

PR #1408, opened 2022-11-15
Set executable bit for binaries in release artifacts

### LINKED ISSUE #1305: Releases in .tar.gz are missing execute bit
When trying to unpack and run releases from
https://github.com/mozilla/sccache/releases/tag/v0.3.0
the execute bit is missing from `sccache`. If the file is intended to be unpacked and run in place, that bit must be set on Linux and MacOS platforms.

messense:
Fixes #1305

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,6 +251,7 @@ jobs:
       - name: Create release assets
         run: |
           for d in sccache-*; do
+            chmod +x "$d/sccache"*
             cp README.md LICENSE "$d/"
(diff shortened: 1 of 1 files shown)


sylvestre:
bizarre.
thanks!

sylvestre:
OK, probably because:
```
      - name: Get artifacts
        uses: actions/download-artifact@v3
```

messense:
Yes, it's a known limitation: https://github.com/actions/download-artifact#permission-loss

PR #1409, opened 2022-11-15
Bring back `SCCACHE_S3_USE_SSL`

messense:


src/cache/s3.rs
```diff
@@ -106,15 +107,40 @@ struct S3Client {
...
             RegionProviderChain::first_try(region.map(|r| Region::new(r.to_owned())))
                 .or_default_provider();
 
         let shared_config = aws_config::from_env().region(region_provider).load().await;
         let mut builder = aws_sdk_s3::config::Builder::from(&shared_config);
         if let Some(endpoint) = endpoint {
-            builder = builder.endpoint_resolver(Endpoint::mutable(endpoint.try_into().unwrap()));
+            let endpoint_uri: http::Uri = endpoint.try_into().unwrap();
+            let mut parts = endpoint_uri.into_parts();
+            match use_ssl {
```
sylvestre, line 125: maybe move that into a function ?

messense: sure

sylvestre:
> If your endpoint requires TLS, set `SCCACHE_S3_USE_SSL=true`.

maybe add:
If you don't need a secure network layer, http might be better for performances

messense:
Done.

src/config.rs
```diff
@@ -1026,6 +1027,7 @@ url = "redis://user:passwd@1.2.3.4:6379/1"
 bucket = "name"
 region = "us-east-2"
 endpoint = "s3-us-east-1.amazonaws.com"
+use_ssl = true
```
sylvestre, line 1030: what was the default before your changes ?

messense: It's also `true` in this test: https://github.com/mozilla/sccache/blob/41e408d3e22d3c5a03cc622c93e9348ef80df140/src/config.rs#L1004

drahnr:
I think it should be renamed to "disable TLS", the default should be TLS on. I don't think it's worth to keep compat of so e env var that was confusing to begin with

messense:
> I think it should be renamed to "disable TLS", the default should be TLS on.

IMO, if the default is changed to TLS on, there is no need to add another env var to control TLS off. `SCCACHE_ENDPOINT` can already take scheme, set it to `http://endpoint-addresss` should do.

I can make the change when you guys reach an agreement.

sylvestre:
@glandium do you have an opinion on this? thanks

README.md
```diff
@@ -196,7 +196,7 @@ Credentials are resolved using the default AWS provider chain, including the `AW
 
 If you do not want to use credentials at all, you can set the `SCCACHE_S3_NO_CREDENTIALS` environment variable. This requires the bucket to allow public readonly access, and can be useful to implement a readonly cache for pull requests, which typically can't be given access to credentials for security reasons.
 
-You can configure the region using the `SCCACHE_REGION` environment variable, or specify the `region` key in `~/.aws/credentials`. Alternatively you can specify the endpoint URL using the `SCCACHE_ENDPOINT` environment variable. To connect to a minio storage for example you can set `SCCACHE_ENDPOINT=<ip>:<port>`.
+You can configure the region using the `SCCACHE_REGION` environment variable, or specify the `region` key in `~/.aws/credentials`. Alternatively you can specify the endpoint URL using the `SCCACHE_ENDPOINT` environment variable. To connect to a minio storage for example you can set `SCCACHE_ENDPOINT=<ip>:<port>`. If your endpoint requires TLS, set `SCCACHE_S3_USE_SSL=true`. If you don't need a secure network layer, http might be better for performance.
```
sylvestre, line 199:
```suggestion
You can configure the region using the `SCCACHE_REGION` environment variable, or specify the `region` key in `~/.aws/credentials`. Alternatively you can specify the endpoint URL using the `SCCACHE_ENDPOINT` environment variable. To connect to a minio storage for example you can set `SCCACHE_ENDPOINT=<ip>:<port>`. 

If your endpoint requires HTTPS/TLS, set `SCCACHE_S3_USE_SSL=true`. If you don't need a secure network layer, HTTP (`SCCACHE_S3_USE_SSL=false`) might be better for performance.
```

Issue #1404, opened 2022-11-14
proposal: Use opendal to handle the cache storage operations


Xuanwo:
## Background

Hi, I'm the maintainer of [opendal](https://github.com/datafuselabs/opendal), a rust lib that focuses on accessing data freely, painlessly, and efficiently.

OpenDAL now supports almost all storage services, including `s3`, `gcs`, `azblob`, `oss`, `obs`, `hdfs`, `redis`, `fs` and so on.

```rust
use anyhow::Result;
use opendal::Operator;
use opendal::Scheme;

#[tokio::main]
async fn main() -> Result<()> {
    // Init Operator
    let op = Operator::from_env(Scheme::S3)?;

    // Create object handler.
    let o = op.object("test_file");

    // Write data info object;
    o.write("Hello, World!").await?;

    // Read data from the object;
    let bs = o.read().await?;

    // Delete object.
    o.delete().await?;

    Ok(())
}
```

I came across `sccache` last week and found opendal is a great choice for `sccache`, which can help resolve many issues like https://github.com/mozilla/sccache/issues/1384, https://github.com/mozilla/sccache/issues/1180, https://github.com/mozilla/sccache/issues/1167.

Apart from various storage backend supports, opendal also has the following benefits:

- Full observability support: [logging](https://opendal.databend.rs/opendal/layers/struct.LoggingLayer.html), [tracing](https://opendal.databend.rs/opendal/layers/struct.TracingLayer.html), [metrics](https://opendal.databend.rs/opendal/layers/struct.MetricsLayer.html).
- Automatic [retry](https://opendal.databend.rs/opendal/layers/struct.RetryLayer.html) support
- Used in production by databases like [databend](https://github.com/datafuselabs/databend/)

## Proposal

So I propose to replace the storage implementation with opendal instead.

I built a quick demo here: https://github.com/Xuanwo/cccache/tree/debug. Most of the change will be like:

```diff
- let credentials = self
-     .provider
-     .credentials()
-     .await
-     .context("failed to get AWS credentials")?;
- 
- let bucket = self.bucket.clone();
- let _ = bucket
-     .put(&key, data, &credentials)
-     .await
-     .context("failed to put cache entry in s3")?;
+ self.op.object(&key).write(data).await?;
```

If this proposal is accepted, I plan to

- Replace them one by one (s3, gcs, ...)
- Make sure all existing config work as before
- Then adding new services support

---

Any comments are welcome! Thanks in advance.

[...]

Issue #1386, opened 2022-11-09
Support IMDSv2 for AWS IAM Role authentication


everpcpc:
ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials

Currently, sscache only uses IMDSv1 which would result in 401 response on some nodes with IMDSv1 disabled

Xuanwo:
```shell
[2022-11-09T08:31:35Z DEBUG sccache::simples3::credential] Attempting to fetch credentials from http://169.254.169.254/latest/meta-data/iam/security-credentials/
[2022-11-09T08:31:35Z DEBUG reqwest::connect] starting new connection: http://169.254.169.254/
[2022-11-09T08:31:35Z DEBUG hyper::client::connect::http] connecting to 169.254.169.254:80
[2022-11-09T08:31:35Z DEBUG hyper::client::connect::http] connected to 169.254.169.254:80
[2022-11-09T08:31:35Z DEBUG hyper::proto::h1::io] flushed 115 bytes
[2022-11-09T08:31:35Z DEBUG hyper::proto::h1::io] parsed 5 headers
[2022-11-09T08:31:35Z DEBUG hyper::proto::h1::conn] incoming body is empty
[2022-11-09T08:31:35Z DEBUG reqwest::async_impl::client] response '401 Unauthorized' for http://169.254.169.254/latest/meta-data/iam/security-credentials/
[2022-11-09T08:31:35Z WARN  sccache::simples3::credential] Failed to fetch IAM credentials: failed to read IAM role response
[2022-11-09T08:31:35Z DEBUG sccache::simples3::credential] Using AWS credentials from IAM
[2022-11-09T08:31:35Z DEBUG sccache::cache::s3] Could not load AWS creds: Couldn't find AWS credentials in environment, credentials file, or IAM role.
[2022-11-09T08:31:35Z DEBUG sccache::simples3::s3] GET http://databend-ci.s3.amazonaws.com/cache/b/a/9/ba99d18b3398ec2dd4a07eec32cf0cd60623f47671569fe1138cd99b2f90f532
[2022-11-09T08:31:35Z DEBUG hyper::client::pool] reuse idle connection for ("http", databend-ci.s3.amazonaws.com)
[2022-11-09T08:31:35Z DEBUG hyper::proto::h1::io] flushed 143 bytes
[...]

Issue #633, opened 2019-12-30
400 bad request from S3


kfir-drivenets:
I've been trying to run sccache on an EC2 instance with the following config
**Note** I'm getting the same errors with IAM_ROLE and direct keys
```
AWS_ACCESS_KEY_ID=[[my_id]] AWS_SECRET_ACCESS_KEY=[[my_key]] SCCACHE_S3_USE_SSL=true SCCACHE_BUCKET=[[my_bucket]] RUST_LOG=sccache=trace SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 SCCACHE_ERROR_LOG=./sccache_log  ./sccache
```

sccache log file
```
INFO 2019-12-30T18:11:26Z: sccache::server: start_server: port: 4226
 INFO 2019-12-30T18:11:26Z: sccache::server: No scheduler address configured, disabling distributed sccache
DEBUG 2019-12-30T18:11:26Z: sccache::cache::cache: Trying S3Cache([[my bucket]], [[my-bucket]].s3.amazonaws.com)
TRACE 2019-12-30T18:11:26Z: sccache::cache::cache: Using S3Cache
 INFO 2019-12-30T18:11:26Z: sccache::server: server started, listening on port 4226
TRACE 2019-12-30T18:13:21Z: sccache::server: incoming connection
TRACE 2019-12-30T18:13:21Z: sccache::server: handle_client
DEBUG 2019-12-30T18:13:21Z: sccache::server: handle_client: compile
TRACE 2019-12-30T18:13:21Z: sccache::server: compiler_info
TRACE 2019-12-30T18:13:21Z: sccache::server: compiler_info cache miss
TRACE 2019-12-30T18:13:21Z: sccache::compiler::compiler: detect_compiler: /usr/bin/gcc
TRACE 2019-12-30T18:13:21Z: sccache::compiler::compiler: detect_c_compiler
TRACE 2019-12-30T18:13:21Z: sccache::compiler::compiler: compiler Some("/usr/bin/gcc" "-E" "/tmp/sccache.Es1ro090eOOO/testfile.c")
DEBUG 2019-12-30T18:13:21Z: sccache::compiler::compiler: Found GCC
DEBUG 2019-12-30T18:13:21Z: sccache::server: check_compiler: Supported compiler
DEBUG 2019-12-30T18:13:21Z: sccache::server: parse_arguments: Ok: ["-c", "-o", "crap.o", "crap.c"]
[...]

PR #1412, opened 2022-11-16
refactor: Integrate with opendal for s3

Xuanwo:

Hi, this is a quick demo for https://github.com/mozilla/sccache/issues/1404. This PR intends to show what opendal's API looks like and how opendal will work with sccache. And give the whole community a chance to try [opendal](https://github.com/datafuselabs/opendal).

This PR only changes the s3 implementation. In face, we can implement `Cache` for `opendal::Operator` so that we don't need to implement `Cache` for different storage backends anymore.

Any comments are welcome! Thank you so much for your attention and participation.

Fix https://github.com/mozilla/sccache/issues/1386
Fix https://github.com/mozilla/sccache/issues/1404
Fix https://github.com/mozilla/sccache/issues/633


Xuanwo:
Seems our crates.io index need a update 😲

messense:
sccache currently requires MSRV 1.58.0, so opendal also need to support Rust 1.58.0 unless the maintainers are willing to update to a new MSRV policy.

Xuanwo:
> sccache currently requires MSRV 1.58.0, so opendal also need to support Rust 1.58.0 unless the maintainers are willing to update to a new MSRV policy.

Thanks for explanation!

sylvestre:
What is the MSRV for opendal?

sylvestre:
Needs rebasing now, thanks :)

src/cache/s3.rs
```diff
@@ -40,25 +33,52 @@ impl S3Cache {
...
+        builder.root(key_prefix);
+        if no_credentials {
+            builder.disable_credential_loader();
+        }
+        if let Some(endpoint) = endpoint {
+            let endpoint_uri: http::Uri = endpoint
+                .try_into()
+                .map_err(|err| anyhow!("input endpoint {endpoint} is invalid: {:?}", err))?;
+            let mut parts = endpoint_uri.into_parts();
+            match use_ssl {
```
sylvestre, line 50: could you please keep this as in function (last the previous version)

src/cache/s3.rs
```diff
@@ -203,24 +123,8 @@ mod test {
...
-        assert_eq!(
-            normalize_key("prefix/", "0123456789abcdef0123456789abcdef"),
-            "prefix/0/1/2/0123456789abcdef0123456789abcdef"
+            normalize_key("0123456789abcdef0123456789abcdef"),
+            "0/1/2/0123456789abcdef0123456789abcdef"
         );
     }
-
-    #[test]
-    fn test_endpoint_resolver() {
```
sylvestre, line 216: please keep this test

src/cache/s3.rs
Xuanwo: Added back.

src/cache/s3.rs
Xuanwo: Added back.

drahnr:
There are a few risks associated with `opendal`, it..

* .. depends on `tokio`, so we are signing up for that, not a big deal  *I think*, we already use tokio heavily and I don't see that changing
* .. has a lot backends we don't need, i.e. `rocksdb` - these are configurable via feature flags in `main` in `opendal` - and that'd be one condition to not inflate the compile time cost for zero benefit to `sccache`
* .. future proofness is an open topic, that wasn't exactly addressed
* .. the stability guarantees are also not entirely clear to me on an API level

On the other side, we gain quite a bit:

* Simplified addition of more backends
* The sampled code and documentation looks actually better than what we have right now
*  Shared load of maintainence web service API changes

My bottom line is a `yay` given the manpower that's available to maintain `sccache`

sylvestre:
Great summary, exactly how i feel...

maybe we could do a new release of sccache with the current code (and a few PR)
and take this into a new release

Xuanwo:
> * .. has a lot backends we don't need, i.e. `rocksdb` - these are configurable via feature flags in `main` in `opendal` - and that'd be one condition to not 
>           inflate
>            the compile time cost for zero benefit to `sccache`

Most backends like `rocksdb` are only included while the corresponding feature enabled like: https://opendal.databend.rs/opendal/#services-1

They will not add the compile time cost for `sccache` unless enabling them.

Services that only need HTTP client will be enabled by default (which includes services like s3, gcs, oss) because they will not introduce extra heavy dependences.

If this is a big concern to `sccache`, I'm willing to add new feature flags for them too.

---

Answers to open questions:

> * Future proof-ness, how to assure continued maintainence effort?

Yes. OpenDAL is the core dependence of [databend](https://github.com/datafuselabs/databend/) and we will continue to maintain it. And we are kept enlarge our community to get more maintainers.

So far, we have two maintainers: @Xuanwo and @PsiACE.

> * What are the stability guarantees on an API level, are they fully semver compliant?

Yes, we adopt the semver compliant, all changes will be recorded at https://github.com/datafuselabs/opendal/blob/main/CHANGELOG.md

NOTE: OpenDAL is on the track on `v0.X.Y` now, we could introduce breaking changes between `v0.X` version.

We are working hard to make a `v1.0` release.

> * MSRV handling?

OpenDAL's MSRV is `1.60` which is covered by our CI services: https://github.com/datafuselabs/opendal/blob/main/.github/workflows/ci.yml#L42

We will not change the MSRV before `v1.0` versions.

---

> maybe we could do a new release of sccache with the current code (and a few PR)

LGTM.

drahnr:
@Xuanwo thank you for the prompt response, it addressed my concerns, for me it's good to go, right after:

@sylvestre we should cut a release before merging this PR as you suggested, just to get some experience/burning time with it

drahnr:
@sylvestre I think it's time to merge

sylvestre:
@Xuanwo you are going to update the various backend next, right ?

Xuanwo:
> @Xuanwo you are going to update the various backend next, right ?

Yes, I plan to migrate `azure`, `gcs` first and than adding new backends support.

sylvestre:
Github is complaining about conflicts

PR #1418, opened 2022-11-22
Integration of cargo deny in the CI

sylvestre:
cc @bnjbvr ;)


sylvestre:
I asked for this action to be enabled here: https://bugzilla.mozilla.org/show_bug.cgi?id=1802072

drahnr (approved):

deny.toml
```diff
@@ -0,0 +1,117 @@
...
+expression = "ISC AND MIT AND OpenSSL"
+license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
+
+# This section is considered when running `cargo deny check bans`.
+# More documentation about the 'bans' section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
+[bans]
+multiple-versions = "deny"
+wildcards = "allow"
+highlight = "all"
```
drahnr, line 65: I am not sure a outright ban is the right approach, it just causes a lot of friction

sylvestre: I have been using this for https://github.com/uutils/coreutils for a while and it works well.
You just populate the ignore list when it happens. It helps to know when and how you are regressing.

sylvestre: And we can always disable it if we don't like it here :)

PR #1419, opened 2022-11-23
Enable connection reuse in s3 client

messense:


src/cache/s3.rs
```diff
@@ -147,7 +149,15 @@ impl S3Client {
         }
         let config = builder.build();
 
+        let client = aws_smithy_client::Builder::new()
```
sylvestre, line 152: please add a comment here explain what it is for

sylvestre:
looks cool, did you do any benchmark ?

messense:
Just finished test locally with minio and plain http.

main

```bash
$ hyperfine -w 3 'cargo clean; cargo build'
Benchmark 1: cargo clean; cargo build
  Time (mean ± σ):     14.743 s ±  1.607 s    [User: 38.746 s, System: 6.607 s]
  Range (min … max):   14.084 s … 19.301 s    10 runs
```

pr:

```bash
$ hyperfine -w 3 'cargo clean; cargo build'
Benchmark 1: cargo clean; cargo build
  Time (mean ± σ):     14.339 s ±  1.705 s    [User: 37.552 s, System: 6.506 s]
  Range (min … max):   13.540 s … 19.179 s    10 runs
```

I think it should perform better when TLS is involved.

PR #1421, opened 2022-11-23
Set minimal rust version to 1.60 + set some minimal versions

sylvestre:


Cargo.toml
```diff
@@ -39,7 +39,7 @@ futures-locks = "0.7"
 gzp = { version = "0.10", default-features = false, features = ["deflate_rust"]  }
 hmac = { version = "0.12.0", optional = true }
 http = "0.2"
-hyper = { version = "0.14", optional = true, features = ["server"] }
+hyper = { version = ">=0.14.10", optional = true, features = ["server"] }
```
Xuanwo, line 42: `0.14.10` means `>=0.14.10,<0.15.0`, I guess it's enough?

messense:
I've encountered similar failing dist-server tests in https://github.com/mozilla/sccache/pull/1403#issuecomment-1313524914, perhaps it's better to minimize `Cargo.lock` changes first, update dependencies one by one to see which one is faulty.

sylvestre:
merged in a different pr

PR #1428, opened 2022-11-27
Bump MSRV to 1.60

messense:
See https://github.com/mozilla/sccache/pull/1421#issuecomment-1328157001


messense:
Update `tokio` breaks the dist-server tests, see https://github.com/messense/sccache/actions/runs/3556612462

sylvestre:
thanks

PR #1433, opened 2022-11-28
Add support for GitHub Actions cache

jakelee8:
Add support for GitHub Actions cache (GHA cache). GHA cache is useful for caching cross-dependencies between different workflows and tasks. It has been working great for Docker builds!

If needed, I'm happy to add the `gha-toolkit` code into `sccache` to minimize dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,9 +88,9 @@ dependencies = [
 
 [[package]]
 name = "async-channel"
-version = "1.7.1"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e14485364214912d3b19cc3435dde4df66065127f05fa0d75c712f36f12c2f28"
+checksum = "cf46fee83e5ccffc220104713af3292ff9bc7c64c7de289f66dae8e38d826833"
 dependencies = [
  "concurrent-queue",
  "event-listener",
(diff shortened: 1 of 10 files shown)


jakelee8:
Done!

drahnr (approved):
Beautiful from a feature and usability pov 🌟

drahnr:
We should consider dog feeding it to ourselves @sylvestre

sylvestre:
Terrific, thanks!

@drahnr with pleasure
wdyt about a doing a new release now?

messense:
Hi @jakelee8 , is gha cache expected to create lots of small sized cache entries with the same name?

image

jakelee8:
Yes, GHA cache is expected to create many small caches with the same cache key.

When using the [`@actions/cache`](https://github.com/actions/cache) action, the cache is saved as a single large archive. In contrast, with `sccache` using GHA cache as storage, each cache entry is saved separately.

In GHA, the cache entries are identified by their ["version"](https://github.com/actions/cache#cache-version) and the cache key (e.g. `SCCACHE_GHA_CACHE_TO` and `SCCACHE_GHA_CACHE_FROM`). The GHA cache implementation in `sccache` calculates the cache version from the `sccache` entry key. The cache key is used to find matching versions using cache key prefixes.

For example, if a cache entry has the version `main.rs` and has GHA cache entries for the `sccache-1` and `sccache-2` keys, then `SCCACHE_GHA_CACHE_FROM=sccache-` will match both and [returns the most recent entry](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key).

This behavior is useful for scoping caches from different versions of Rust or cross-platform builds, as in the following screenshot ([source](https://github.com/BitskiCo/rust-sdk/actions/caches?query=sort%3Aaccessed-desc)).

Screenshot 2022-12-01 at 3 23 36 AM

The Git SHA suffixes allows newer commits to override older caches.

drahnr:
Could you add your comment to the appropriate documentation section? That would set expectations on observations in the GitHub ui and avoid confusion


Thank you for your work!

jakelee8:
https://github.com/mozilla/sccache/pull/1441

PR #1440, opened 2022-12-02
Reintroduce symlink handling

figsoda:
Reverts mozilla/sccache#1439


sylvestre:
Did you fix the bug in this PR ?

figsoda:
I am still working on the fix, the github interface was just easier for reverting a pr

figsoda:
@Jake-Shadle does this fix the issue for you? I can't reproduce the issue since my clang is wrapped

sylvestre:
Could you please push the initial pr?
I would like to avoid having revert of revert in the history
thanks

src/server.rs
```diff
@@ -939,9 +939,12 @@ where
             }
         };
 
-        let resolved_compiler_path = resolved_compiler_path
-            .canonicalize()
-            .unwrap_or(resolved_compiler_path);
+        let resolved_compiler_path = match resolved_compiler_path.canonicalize() {
```
sylvestre, line 942: please add a comment to explain the why and a test for this case

figsoda:
squashed the commits instead since half the commits were tiny patches anyway

also added a comment and a test, didn't verify that the test can reproduce the bug without the patch since my clang is wrapped, but I did verify that the file doesn't build with `clang`, only with `clang++`

sylvestre:
@Jake-Shadle could you please test with this patch? thanks

Jake-Shadle:
Yes, this seems to work again, thanks.

sylvestre:
Thanks for testing

PR #1443, opened 2022-12-04
Remove unused sha-1 from dependencies

tottoto:
Removes `sha-1` crate from dependencies as it seems not to be used.


sylvestre:
Thanks
I love removing dead dep.

PR #1454, opened 2022-12-07
add github action docs file

aviramha:


jakelee8:
I would consider `sccache` a "trusted" component in the software supply chain and recommend against using 3rd-party GitHub Actions in official `sccache` documentation. That said, [@metalbear-co/sccache-action](https://github.com/metalbear-co/sccache-action) is still a work in progress.

For a more complete example of using GHA `sccache` with Cargo, including cross-platform support, see https://github.com/BitskiCo/rust-sdk/blob/main/bin/cargo

docs/GitHubActions.md
```diff
@@ -0,0 +1,44 @@
...
+## Getting Started
+The easiest way to use `sccache` with GitHub Actions is to use the [GitHub Action](https://github.com/marketplace/actions/rust-cache-using-sccache).
+The action will automatically download and install `sccache` and configure it to use the GitHub Actions cache.
+It will add `RUSTC_WRAPPER` environment to your build steps to use `sccache` as the compiler wrapper.
+
+Just add this step to your workflow:
+
+```yaml
+- name: Cache Rust compilation
+  uses: metalbear-co/sscache-action@v1.0.2
```
sylvestre, line 12: Sorry but I am not confortable recommending this
Could you please remove this?

drahnr: I was about suggest that, and invest a few hours into vetting and polishing so we can endorse something that's easier to use compared to a DIY manual.

PR #1463, opened 2022-12-11
refactor: Migrate azblob to opendal based

Xuanwo:
This PR includes the following changes:

- Implement `Storage` upon opendal::Operator so that all opendal services can be used as cache storage
- Migrate azure implement to opendal based.

Much azure tests have been removed, but opendal have all of them covered in [OpenDAL's behavior test](https://github.com/datafuselabs/opendal/tree/main/tests/behavior) at [Service Test Azblob](https://github.com/datafuselabs/opendal/blob/main/.github/workflows/service_test_azblob.yml).


src/cache/cache.rs
```diff
@@ -327,26 +327,75 @@ pub trait Storage: Send + Sync {
...
+#[async_trait]
+impl Storage for opendal::Operator {
+    async fn get(&self, key: &str) -> Result<Cache> {
+        match self.object(&normalize_key(key)).read().await {
+            Ok(res) => {
+                let hit = CacheRead::from(io::Cursor::new(res))?;
+                Ok(Cache::Hit(hit))
+            }
+            Err(e) => {
+                warn!("Got AWS error: {:?}", e);
```
sylvestre, line 341: aws only ?

Xuanwo: opendal::Error already contains it's service type.

src/config.rs
```diff
@@ -908,6 +919,8 @@ fn config_overrides() {
     let env_conf = EnvConfig {
         cache: CacheConfigs {
             azure: Some(AzureCacheConfig {
+                connection_string: "".to_string(),
+                container: "".to_string(),
                 key_prefix: "".to_owned(),
```
sylvestre, line 924:
```suggestion
                connection_string: String::new(),
                container: String::new(),
                key_prefix: String::new(),
```

src/config.rs
```diff
@@ -950,6 +963,8 @@ fn config_overrides() {
                     url: "memurl".to_owned()
                 }),
                 CacheType::Azure(AzureCacheConfig {
+                    connection_string: "".to_string(),
+                    container: "".to_string(),
                     key_prefix: "".to_owned()
```
sylvestre, line 968:
```suggestion
                    connection_string: String::new(),
                    container: String::new(),
                    key_prefix: String::new()
```

drahnr (approved):

drahnr:
I like the kind of PRs that delete more code than it adds, thank you! I don't think we should have tests for the inner details of opendal, after all we are becoming a consumer of that and it's not a direct development concern of sccache.

PR #1495, opened 2022-12-20
Fix duration attributed to cache writes in the log

glandium:
As it is now, what ends up logged as being the time for cache writes is the compilation duration exclusively, which is also logged on its own.

For extra granularity, we also log the time spent creating the cache artifact.

While here, also log the compilation durations in cases where the result is not cached.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -348,14 +348,19 @@ where
                 let duration = start.elapsed();
                 if !compiler_result.status.success() {
                     debug!(
-                        "[{}]: Compiled but failed, not storing in cache",
-                        out_pretty
+                        "[{}]: Compiled in {}, but failed, not storing in cache",
+                        out_pretty,
+                        fmt_duration_as_secs(&duration)
                     );
                     return Ok((CompileResult::CompileFailed, compiler_result));
                 }
                 if cacheable != Cacheable::Yes {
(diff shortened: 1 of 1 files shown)


glandium:
Note this still leaves the cache artifact creation unaccounted for in `--show-stats`.

sylvestre:
I tried to rebuild Firefox with the patch and here is the results:
https://firefoxci.taskcluster-artifacts.net/THQ6jVgKSf-gL9XiYaxKSQ/0/public/build/sccache.log

Example:
[2022-12-20T10:01:43Z DEBUG sccache::compiler::compiler] [jsrust]: Cache miss in 0.000 s
[2022-12-20T10:01:43Z DEBUG sccache::compiler::compiler] [jsrust]: Compiling locally
[2022-12-20T10:02:05Z DEBUG sccache::compiler::compiler] [jsrust]: Compiled in 21.433 s, storing in cache
[2022-12-20T10:02:05Z DEBUG sccache::compiler::compiler] [jsrust]: Created cache artifact in 0.135 s
[2022-12-20T10:02:05Z DEBUG sccache::compiler::compiler] [jsrust]: Stored in cache successfully!
[2022-12-20T10:02:05Z DEBUG sccache::server] [jsrust]: Cache write finished in 0.117 s

cc @indygreg

PR #1500, opened 2022-12-26
fix: handle -Xclang args when invoking clang/clang++ on windows cmake

Danielmelody:
Instead of clang-cl.


Danielmelody:
This PR also pass the unsupported flag name to error log, so that we can know which flag failed.

src/compiler/clang.rs
```diff
@@ -184,6 +185,7 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-fprofile-use", PathBuf, Concatenated('='), ClangProfileUse),
     take_arg!("-fsanitize-blacklist", PathBuf, Concatenated('='), ExtraHashFile),
     take_arg!("-gcc-toolchain", OsString, Separated, PassThrough),
+    flag!("-gcodeview", PreprocessorArgumentFlag),
```
glandium, line 188: -gcodeview is not a preprocessor flag

Danielmelody: Agree, should I add a PassThroughFlag here?

sylvestre:
please run rustfmt on your changes

PR #1510, opened 2022-12-27
ci: Add integration test for redis

Xuanwo:
This PR will add integration tests for redis

### DIFF
--- a/.github/workflows/redis.yml
+++ b/.github/workflows/redis.yml
@@ -33,4 +33,6 @@ jobs:
 
       - name: Output
         run: |
-          /home/runner/.cargo/bin/sccache --show-stats | grep redis --passthru
+          /home/runner/.cargo/bin/sccache --show-stats
+
+          /home/runner/.cargo/bin/sccache --show-stats | grep redis
--- a/.github/workflows/redis.yml
+++ b/.github/workflows/redis.yml
@@ -36,3 +36,17 @@ jobs:
           /home/runner/.cargo/bin/sccache --show-stats
 
           /home/runner/.cargo/bin/sccache --show-stats | grep redis
+
+      - name: Test Twice for Cache Read
+        run: cargo clean && cargo build
+        env:
+          RUST_BACKTRACE: full
+          RUST_LOG: debug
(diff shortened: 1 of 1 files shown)


Xuanwo:
```
shogo82148/actions-setup-redis@v1 is not allowed to be used in mozilla/sccache. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, or matching the following: !/mozilla/**, !mozilla/**, ./**, 10up/wpcs-action@*, aws-actions/*, docker/*, pypa/gh-action-pypi-publish@v1.4.2, slackapi/slack-github-action@*, google-github-actions/*, erlef/setup-beam@v1, yesolutions/mirror-action@*, codecov/codecov-action@*, tj-actions/changed-files@*, tj-actions/glob@*, vmactions/freebsd-vm@v0, actions-rs/toolchain@v1, shivammathur/setup-php@*, EmbarkStudios/*.
```

Hi, @sylvestre, I want to use action `shogo82148/actions-setup-redis@v1`.

sylvestre:
I will ask my colleagues at Mozilla but it might take a week or two

Xuanwo:
> I will ask my colleagues at Mozilla but it might take a week or two

Ok, maybe it's better to avoid using new actions here. Let's use github native service feature.

sylvestre:
if you could, would be great indeed :)

Xuanwo:
cc @sylvestre, this PR is ready for review now!

sylvestre:
Fails with
```
Run /home/runner/.cargo/bin/sccache --show-stats | grep redis --passthru
grep: unrecognized option '--passthru'
```

Xuanwo:
> Fails with
> 
> ```
> Run /home/runner/.cargo/bin/sccache --show-stats | grep redis --passthru
> grep: unrecognized option '--passthru'
> ```

Fixed

sylvestre:
it would be nice to do a second build to verify that
`Cache hits` isn't at 0 ?

sylvestre:
This is terrific. I wonder if we should not replicate this for other services
wdyt?

Issue #1512, opened 2022-12-28
proposal: Exit with error message if storage is misconfigured


Xuanwo:
## Background

Sccache will ignore storage errors by default:

- For reading, all non-ok responses will be treated as cache-miss.
- For writing, all non-ok responses will be ignored.

Thus, users don't know storage is misconfigured until they find the cache is always missed.

This problem has been proposed by the community many times:

- Fix https://github.com/mozilla/sccache/issues/45
- Fix https://github.com/mozilla/sccache/issues/1508
- Fix https://github.com/mozilla/sccache/issues/1384
- Fix https://github.com/mozilla/sccache/issues/1173
- Fix https://github.com/mozilla/sccache/issues/225

## Proposal

So I propose adding a storage check before starting the server by performing `write` and `read` operations with file path `.sccache_check`.

- If both of them succeed, sccache will run in `read-write` mode.
- If `write` fails with `PermissionDenied` and `read` is `Ok` or `NotFound`, sccache will run in `read-only` mode.
- All other cases will lead sccache exit with corresponding error messages.
  - For example, `read` returns `PermissionDenied` or `OtherError`.

The status will also be shown in `sccache --show-stat`.

With the super power of [OpenDAL](https://github.com/datafuselabs/opendal), this proposal should be easy (

Error happened during compilation will still be ignored.

## Drawbacks

Misconfigured sccache will break users' workflow, which may lead to more severe problems.

## Tasks

- [x] https://github.com/mozilla/sccache/pull/1515

sylvestre:
That sounds like a great idea, i am sure that weeks of work have been lost while trying to configure sccache.

I did similar work with logging: https://github.com/mozilla/sccache/pull/1361

Xuanwo:
Quick demo show:

```shell
:( SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 SCCACHE_BUCKET=test ./target/debug/sccache                  
sccache: error: cache storage failed to read: ObjectPermissionDenied (permanent) at read => S3Error { code: "AccessDenied", message: "Access Denied", resource: "", request_id: "46RYHDRAN8WD64XW" }

Context:
[...]

PR #1515, opened 2022-12-29
feat: Exit while cache is not configured correctly

Xuanwo:

This PR is the first step of https://github.com/mozilla/sccache/issues/1512.

We will exit while the cache is not configured correctly. For example, s3 is configured with an invalid endpoint.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -382,7 +382,7 @@ impl Storage for opendal::Operator {
 }
 
 /// Normalize key `abcdef` into `a/b/c/abcdef`
-#[cfg(any(feature = "s3", feature = "azure", feature = "gcs", feature = "redis"))]
+#[allow(dead_code)]
 fn normalize_key(key: &str) -> String {
     format!("{}/{}/{}/{}", &key[0..1], &key[1..2], &key[2..3], &key)
 }
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -25,7 +25,16 @@ use crate::cache::memcached::MemcachedCache;
 use crate::cache::redis::RedisCache;
 #[cfg(feature = "s3")]
 use crate::cache::s3::S3Cache;
-use crate::config::{self, CacheType, Config};
+use crate::config::Config;
+#[cfg(any(
+    feature = "azure",
+    feature = "gcs",
+    feature = "gha",
+    feature = "memcached",
+    feature = "redis",
+    feature = "s3"
+))]
+use crate::config::{self, CacheType};
 use std::fmt;
 use std::fs;
 use std::io::{self, Cursor, Read, Seek, Write};
@@ -326,7 +335,7 @@ pub trait Storage: Send + Sync {
 }
 
 /// Implement storage for operator.
-#[cfg(any(feature = "s3", feature = "azure", feature = "gcs"))]
+#[cfg(any(feature = "s3", feature = "azure", feature = "gcs", feature = "redis"))]
 #[async_trait]
 impl Storage for opendal::Operator {
     async fn get(&self, key: &str) -> Result<Cache> {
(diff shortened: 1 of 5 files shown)


tests/sccache_cargo.rs
```diff
@@ -302,24 +302,11 @@ fn test_gcp_arg_check() -> Result<()> {
...
     cmd.arg("--start-server")
         .env("SCCACHE_LOG", "debug")
-        .env("SCCACHE_GCS_OAUTH_URL", "http://foobar");
+        .env("SCCACHE_GCS_OAUTH_URL", "http://127.0.0.1");
 
     cmd.assert().failure().stderr(predicate::str::contains(
         "If setting GCS credentials, SCCACHE_GCS_BUCKET",
     ));
 
-    stop_sccache()?;
```
sylvestre, line 311: why did you remove it?

sylvestre: maybe we should keep/transform it to have an invalid endpoint to test the error ?

Xuanwo: Nice catch. I found current logic may break the server spawn logic. Trying to fix.

Xuanwo:
Fixed! @sylvestre PTAL.

tests/sccache_cargo.rs
```diff
@@ -313,13 +313,13 @@ fn test_gcp_arg_check() -> Result<()> {
...
+        .env("SCCACHE_GCS_CREDENTIALS_URL", "not_valid_url//127.0.0.1")
         .env("SCCACHE_GCS_KEY_PATH", "foo.json");
 
     // This is just a warning
-    cmd.assert().success().stderr(predicate::str::contains(
-        "SCCACHE_GCS_OAUTH_URL has been deprecated",
-    ));
+    cmd.assert()
+        .failure()
+        .stderr(predicate::str::contains("gcs credential url is invliad"));
```
sylvestre, line 322:
```suggestion
        .stderr(predicate::str::contains("gcs credential url is invalid"));
```

src/cache/cache.rs
```diff
@@ -373,30 +382,31 @@ impl Storage for opendal::Operator {
 }
 
 /// Normalize key `abcdef` into `a/b/c/abcdef`
+#[allow(dead_code)]
```
drahnr, line 385: Could it be removed altogether?

src/cache/cache.rs
```diff
@@ -405,99 +415,87 @@ pub fn storage_from_config(config: &Config, pool: &tokio::runtime::Handle) -> Ar
...
             }
+            #[cfg(not(any(
+                feature = "azure",
+                feature = "gcs",
+                feature = "gha",
+                feature = "memcached",
+                feature = "redis",
+                feature = "s3"
+            )))]
+            _ => return Err(anyhow!("cache type is not enabled")),
```
drahnr, line 492:
```suggestion
            _ => bail!("cache type is not enabled"),
```

src/config.rs
```diff
@@ -666,7 +665,7 @@ fn config_file(env_var: &str, leaf: &str) -> PathBuf {
 
 #[derive(Debug, Default, PartialEq, Eq)]
 pub struct Config {
-    pub caches: Vec<CacheType>,
+    pub cache: Option<CacheType>,
```
drahnr, line 668: Was this used anywhere before? Did multiple backends ever work?

drahnr (approved):
It's unclear whether the changes of many backends to one is diserable or not.

src/config.rs
Xuanwo: > Did multiple backends ever work?

No, they never works at the same time.

The only case is users configure different backends at the same time, and some of them are wrong. For example, users configure s3, azure, redis at the same time, and s3, azure is invalid, only redis works.

In old logic, the errors returned by s3 and azure will be ignore, and using redis finally. In this PR, we will only choose the first valid cache type and returns error happened which makes the behavior more predictable.

In the future, I will return errors if users configured different backends at the same time.

src/cache/cache.rs
Xuanwo: This function will be used if feature like `s3`, `azure` enabled. Adding a `dead_code` is the most simple way to address clippy errors under `--no-default-feature`.

PR #1518, opened 2022-12-30
Accept /fsanitize=address as a passthrough option for MSVC

randombit:
At the moment, invocations using `/fsanitize=address` are rejected as uncachable by sccache due to it assuming there are multiple input files.


drahnr (approved):

randombit:
FreeBSD CI build is failing but afaict it is not related to this change

Xuanwo (approved):

PR #1527, opened 2023-01-06
The "cache read miss" is now the compiler step. Make it more explicit 

sylvestre:
We are getting:
```
Average cache write               0.115 s
Average compiler write            5.737 s
```
was:
```
Average cache write               0.115 s
Average cache read miss           5.737 s
```

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -345,12 +345,12 @@ where
                     out_pretty.clone(),
                 )
                 .await?;
-                let duration = start.elapsed();
+                let duration_compilation = start.elapsed();
                 if !compiler_result.status.success() {
                     debug!(
                         "[{}]: Compiled in {}, but failed, not storing in cache",
                         out_pretty,
-                        fmt_duration_as_secs(&duration)
+                        fmt_duration_as_secs(&duration_compilation)
                     );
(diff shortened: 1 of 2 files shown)


src/compiler/compiler.rs
```diff
@@ -784,7 +784,7 @@ pub enum CompileResult {
     CacheMiss(
         MissType,
         DistType,
-        Duration,
+        Duration, // Compilation time
```
Xuanwo, line 787: Maybe use BoxFuture?

Xuanwo (commented):
Others LGTM

drahnr:
I think we should split the item into `read miss`, `compile` and `cache write`

sylvestre:
> I think we should split the item into `read miss`, `compile` and `cache write`

what do you mean? it is what we have currently

drahnr:
I misinterpreted the changeset, if it's only compilation I'd name it as such `compile`, without the write suffix.
Question here is what does the compilation metric really give us?

sylvestre:
> I misinterpreted the changeset, if it's only compilation I'd name it as such compile, without the write suffix.

Good point, fixed 

> Question here is what does the compilation metric really give us?

It is free and interesting to see how long a user can save. And it is better to have it than not.

drahnr (approved):

Issue #964, opened 2021-02-26
Content-addressed cache?


nikclayton:
One of the caveats in the documentation is:

> Absolute paths to files must match to get a cache hit. This means that even if you are using a shared cache, everyone will have to build at the same absolute path (i.e. not in $HOME) in order to benefit each other.

This is a problem for single-developer workflows that use `git worktree`. With `git worktree` I can check out multiple branches of the same repository to distinct local directories. But because they're distinct, there's no cache benefit.

For example, I have a complex project with hundreds of dependencies. Just building from a clean checkout takes about 3 minutes.

So if I do:

```
% cd /my/repo/root
% git worktree -b branch1 /my/worktree/branch1
% cd /my/worktree/branch1
% cargo build
```

This takes 3 minutes.

If I then do

```
% cd /my/repo/root
% git worktree -b branch2 /my/worktree/branch2
% cd /my/worktree/branch2
% cargo build
```

this also takes 3 minutes, and it's re-building everything I just built for branch1.

I went poking through the code, and it looks like https://github.com/mozilla/sccache/blob/master/src/compiler/rust.rs#L1458-L1462 is where the file names (`source_files`) are included in the computation of the hash key.

But earlier in that function, the file *contents* are also included in the hash key computation. So I think the filename is redundant here.

I changed that line in a local build to ignore `source_files`:

```
let inputs = abs_externs.into_iter().chain(abs_staticlibs).collect();
```

and now the example I gave above takes 3 minutes for a build with a cold cache (the `branch1` example), but the `branch2` example now takes 1m32s, because it's able to reuse the cached build artifacts.

And all the sccache tests pass with this change.

I imagine it can't be this simple to fix the "Absolute paths must match" restriction and use sccache as a semi- content-addressed cache, and there's some subtlety that I've missed.

Or is that all that's necessary?

I couldn't find anything else that discussed this -- if this has already been covered in another forum and I couldn't find it there's no need to repeat the issues here, just link to them and I'll go and do some more reading...

luser:
There's some discussion in #35. Mostly it comes down to whether the pathnames wind up in the output files. If debug info is enabled, then they generally do. Compiler options that allow path remapping could be used to mitigate this.

SuperKenVery:
#2521 is merged and #35 is closed. Does that mean absolute path match is not required now?

PR #1528, opened 2023-01-08
refactor: Migrate gha cache to opendal based

Xuanwo:
This PR will migration gha cache support to opendal based. Also, I added an integration tests to make sure it works!

This PR will also close https://github.com/mozilla/sccache/pull/1523

### DIFF
--- a/docs/GHA.md
+++ b/docs/GHA.md
@@ -1,8 +1,8 @@
 # GitHub Actions
 
-To use the [GitHub Actions cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), you need to set the `SCCACHE_GHA_CACHE_URL`/`ACTIONS_CACHE_URL` and `SCCACHE_GHA_RUNTIME_TOKEN`/`ACTIONS_RUNTIME_TOKEN` environmental variables. The `SCCACHE_` prefixed environmental variables override the variables without the prefix.
+To use the [GitHub Actions cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), you need to set the `SCCACHE_GHA_VERSION` which is a namespace for the whole cache set.
 
-In a GitHub Actions workflow, you can set these environmental variables using the following step.
+This cache type will needs token like `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN` to work. You can set these environmental variables using the following step in a GitHub Actions workflow.
 
 ```yaml
 - name: Configure sccache
@@ -12,17 +12,3 @@ In a GitHub Actions workflow, you can set these environmental variables using th
       core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
       core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 ```
-
-To write to the cache, set `SCCACHE_GHA_CACHE_TO` to a cache key, for example
-`sccache-latest`. To read from cache key prefixes, set `SCCACHE_GHA_CACHE_FROM`
-to a comma-separated list of cache key prefixes, for example `sccache-`.
-
-In contrast to the [`@actions/cache`](https://github.com/actions/cache) action, which saves a single large archive per cache key, `sccache` with GHA cache storage saves each cache entry separately.
-
-GHA cache storage will create many small caches with the same cache key, e.g. `SCCACHE_GHA_CACHE_TO` and `SCCACHE_GHA_CACHE_FROM`. These GHA caches are differentiated by their [_version_](https://github.com/actions/cache#cache-version). The GHA cache implementation in `sccache` calculates the cache version from the [`sccache` entry key](docs/Caching.md), e.g. the source file path.
-
-For example, if a cache entry has the version `main.rs` and has GHA cache entries for the `sccache-1` and `sccache-2` keys, then `SCCACHE_GHA_CACHE_FROM=sccache-` will match both and [return the most recent entry](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key).
-
-This behavior is useful for scoping caches from different versions of Rust or for cross-platform builds (`rust-sdk-{RUST_TOOLKIT}-{TARGET_TRIPLE}-`), and to allow newer commits to override older caches by adding the Git SHA as a suffix (`-{GITHUB_SHA}`), as in the following screenshot.
-
-<img width="718" src="https://user-images.githubusercontent.com/19253212/205356799-deedc465-e534-4ef6-a249-fc15121fdfd9.png">
--- a/src/config.rs
+++ b/src/config.rs
@@ -192,7 +192,7 @@ pub struct GCSCacheConfig {
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct GHACacheConfig {
-    /// Version for gha cache is a namesapce. By setting different versions,
+    /// Version for gha cache is a namespace. By setting different versions,
     /// we can avoid mixed caches.
     pub version: String,
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -192,10 +192,9 @@ pub struct GCSCacheConfig {
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct GHACacheConfig {
-    pub url: String,
-    pub token: String,
-    pub cache_to: Option<String>,
-    pub cache_from: Option<String>,
+    /// Version for gha cache is a namesapce. By setting different versions,
+    /// we can avoid mixed caches.
+    pub version: String,
 }
 
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -573,20 +572,8 @@ fn config_from_env() -> Result<EnvConfig> {
     });
 
     // ======= GHA =======
-    let gha = if let (Some(url), Some(token)) = (
-        env::var("SCCACHE_GHA_CACHE_URL")
-            .ok()
-            .or_else(|| env::var("ACTIONS_CACHE_URL").ok()),
-        env::var("SCCACHE_GHA_RUNTIME_TOKEN")
-            .ok()
-            .or_else(|| env::var("ACTIONS_RUNTIME_TOKEN").ok()),
-    ) {
-        Some(GHACacheConfig {
-            url,
-            token,
-            cache_to: env::var("SCCACHE_GHA_CACHE_TO").ok(),
-            cache_from: env::var("SCCACHE_GHA_CACHE_FROM").ok(),
-        })
+    let gha = if let Ok(version) = env::var("SCCACHE_GHA_VERSION") {
+        Some(GHACacheConfig { version })
     } else {
         None
     };
@@ -1122,10 +1109,7 @@ no_credentials = true
                     credential_url: None,
                 }),
                 gha: Some(GHACacheConfig {
-                    url: "http://localhost".to_owned(),
-                    token: "secret".to_owned(),
-                    cache_to: Some("sccache-latest".to_owned()),
-                    cache_from: Some("sccache-".to_owned()),
+                    version: "sccache".to_string()
                 }),
                 redis: Some(RedisCacheConfig {
                     url: "redis://user:passwd@1.2.3.4:6379/1".to_owned(),
--- a/src/config.rs
+++ b/src/config.rs
@@ -1070,10 +1070,7 @@ key_prefix = "prefix"
 service_account = "example_service_account"
 
 [cache.gha]
-url = "http://localhost"
-token = "secret"
-cache_to = "sccache-latest"
-cache_from = "sccache-"
+version = "sccache"
 
 [cache.memcached]
 url = "..."
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -189,3 +189,52 @@ jobs:
           ${SCCACHE_PATH} --show-stats
 
           ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
+
+  gha:
+    runs-on: ubuntu-latest
+    needs: build
+
+    env:
+      SCCACHE_GHA_VERSION: "sccache_integration_tests"
+      RUSTC_WRAPPER: /home/runner/.cargo/bin/sccache
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Configure Cache Env
+        uses: actions/github-script@v6
+        with:
+          script: |
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+
+      - name: Install rust
+        uses: ./.github/actions/rust-toolchain
+        with:
+          toolchain: "stable"
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: integration-tests
+          path: /home/runner/.cargo/bin/
+      - name: Chmod for binary
+        run: chmod +x ${SCCACHE_PATH}
+
+      - name: Test
+        run: cargo clean && cargo build
+
+      - name: Output
+        run: |
+          ${SCCACHE_PATH} --show-stats
+
+          ${SCCACHE_PATH} --show-stats | grep gha
+
+      - name: Test Twice for Cache Read
+        run: cargo clean && cargo build
+
+      - name: Output
+        run: |
+          ${SCCACHE_PATH} --show-stats
+
+          ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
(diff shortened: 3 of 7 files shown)


src/config.rs
```diff
@@ -192,10 +192,9 @@ pub struct GCSCacheConfig {
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct GHACacheConfig {
-    pub url: String,
-    pub token: String,
-    pub cache_to: Option<String>,
-    pub cache_from: Option<String>,
+    /// Version for gha cache is a namesapce. By setting different versions,
```
sylvestre, line 195:
```suggestion
    /// Version for gha cache is a namespace. By setting different versions,
```

Xuanwo:
The failing ghac test related to https://github.com/datafuselabs/opendal/issues/1155, fix coming.

sylvestre:
@jakelee8 if you have an opinion, don't hesitate :)

sylvestre:
@Xuanwo did you implement rate limiting ?

Xuanwo:
> @Xuanwo did you implement rate limiting ?

Not implemented so far, will be added in next PR.

sylvestre:
ok, many thanks :)

docs/GHA.md
```diff
@@ -1,8 +1,8 @@
 # GitHub Actions
 
-To use the [GitHub Actions cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), you need to set the `SCCACHE_GHA_CACHE_URL`/`ACTIONS_CACHE_URL` and `SCCACHE_GHA_RUNTIME_TOKEN`/`ACTIONS_RUNTIME_TOKEN` environmental variables. The `SCCACHE_` prefixed environmental variables override the variables without the prefix.
+To use the [GitHub Actions cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), you need to set the `SCCACHE_GHA_VERSION` which is a namespace for the whole cache set.
```
sylvestre, line 3: I love how this simplifies GHA :)

docs/GHA.md
```diff
@@ -12,17 +12,3 @@ In a GitHub Actions workflow, you can set these environmental variables using th
...
       core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 ```
-
-To write to the cache, set `SCCACHE_GHA_CACHE_TO` to a cache key, for example
-`sccache-latest`. To read from cache key prefixes, set `SCCACHE_GHA_CACHE_FROM`
-to a comma-separated list of cache key prefixes, for example `sccache-`.
-
-In contrast to the [`@actions/cache`](https://github.com/actions/cache) action, which saves a single large archive per cache key, `sccache` with GHA cache storage saves each cache entry separately.
-
-GHA cache storage will create many small caches with the same cache key, e.g. `SCCACHE_GHA_CACHE_TO` and `SCCACHE_GHA_CACHE_FROM`. These GHA caches are differentiated by their [_version_](https://github.com/actions/cache#cache-version). The GHA cache implementation in `sccache` calculates the cache version from the [`sccache` entry key](docs/Caching.md), e.g. the source file path.
```
jakelee8, line 22: Removing the cache from/to options makes it impossible to configure the cache to write-only. Write-only cache is useful for allowing PR builds to use the main branch cache, but forcing the main branch build to build without caching.

docs/GHA.md
```diff
@@ -12,17 +12,3 @@ In a GitHub Actions workflow, you can set these environmental variables using th
       core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
       core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 ```
-
-To write to the cache, set `SCCACHE_GHA_CACHE_TO` to a cache key, for example
-`sccache-latest`. To read from cache key prefixes, set `SCCACHE_GHA_CACHE_FROM`
-to a comma-separated list of cache key prefixes, for example `sccache-`.
-
-In contrast to the [`@actions/cache`](https://github.com/actions/cache) action, which saves a single large archive per cache key, `sccache` with GHA cache storage saves each cache entry separately.
```
jakelee8, line 20: This paragraph was added to explain why sccache will create many small caches instead of the usual single large caches. https://github.com/mozilla/sccache/pull/1433#issuecomment-1333478389

docs/GHA.md
```diff
@@ -12,17 +12,3 @@ In a GitHub Actions workflow, you can set these environmental variables using th
...
-
-GHA cache storage will create many small caches with the same cache key, e.g. `SCCACHE_GHA_CACHE_TO` and `SCCACHE_GHA_CACHE_FROM`. These GHA caches are differentiated by their [_version_](https://github.com/actions/cache#cache-version). The GHA cache implementation in `sccache` calculates the cache version from the [`sccache` entry key](docs/Caching.md), e.g. the source file path.
-
-For example, if a cache entry has the version `main.rs` and has GHA cache entries for the `sccache-1` and `sccache-2` keys, then `SCCACHE_GHA_CACHE_FROM=sccache-` will match both and [return the most recent entry](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key).
-
-This behavior is useful for scoping caches from different versions of Rust or for cross-platform builds (`rust-sdk-{RUST_TOOLKIT}-{TARGET_TRIPLE}-`), and to allow newer commits to override older caches by adding the Git SHA as a suffix (`-{GITHUB_SHA}`), as in the following screenshot.
```
jakelee8, line 26: Without the option of reading caches from a cache key prefix, PR builds cannot use caches from the main branch and will never use caches from previous builds.

docs/GHA.md
Xuanwo: Write-only seems useful. Can you raise an issue of this feature? 

We can implement the same cache_to and cache_from (if needed) by using two [ghac](https://opendal.databend.rs/opendal/services/ghac/index.html) clients with different versions.

Also, [OpenDAL](https://github.com/datafuselabs/opendal) is open to any contributions. Welcome for any comments and discussion :heart:

docs/GHA.md
Xuanwo: I think our users don't need to care about those service internal details.

docs/GHA.md
Xuanwo: User can specify the prefix by setting `root`.

jakelee8: That's not now GitHub Actions cache works...

Xuanwo: > That's not now GitHub Actions cache works...

Can explain a bit to me? I'm not very familiar with github action cache services.

So far, I just treat github action cache as a storage service which means:

- We don't care about how github action cache works (especially for `actions/cache`)
- We don't want to work along with caches that written by others.

Sccache will calculate the unique path for every object file, and we just use the same path as the cache key. Will this behavior break our cache service?

Xuanwo: > That's not now GitHub Actions cache works...

Can explain a bit to me? I'm not very familiar with github action cache services.

So far, I just treat github action cache as a storage service which means:

- We don't care about how github action cache works (especially for `actions/cache`)
- We don't want to work along with caches that written by others.

Sccache will calculate the unique path for every object file, and we just use the same path as the cache key. Will this behavior break our cache service?

docs/GHA.md
sylvestre: well, it isn't really internal as it shows on github 

I am trying with another project and I can see caches of several gb to a few kb
https://github.com/sylvestre/coreutils/actions/caches

docs/GHA.md
sylvestre: @jakelee8 sorry for the dumb question but what is wrong here? thanks

jakelee8: GitHub Actions cache isn't a simple key-value store. See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Xuanwo: Maybe it worth to start a discussion for this~

Xuanwo: Take this action as an example https://github.com/mozilla/sccache/actions/runs/3868518097/jobs/6594049385

The first run on this PR will have cache hits which means we are reading cache from main branch.

jakelee8:
Ah right. The key suffix isn’t needed to read from the main branch cache. 

I was thinking more about bad caches. I’ve run into issues where the cached object resulted in a buggy binary, for example when something was misconfigured. Using the git SHA suffix allows us to override old caches in the main branch (write-only mode).

Xuanwo:
Ok, I got your concerns now!

Sscache will generate a unique cache key for every input so we don't need to care about the bad cache~

jakelee8:
Unique for the input, but not the output!

sylvestre:
You mean that we have a lot of duplication of output?

docs/GHA.md
drahnr: I agree with @jakelee8 - documenting visible behavior is a must.

drahnr:
Could we revert the merge? @sylvestre 

I think there are significant feature changes that warrant a discussion before merging the changeset

Xuanwo:
> Unique for the input, but not the output!

Generating different outputs for the same input should be treated as a sccache bug.

We do need to verify the output after downloading it. The best to address this is adopting https://github.com/mozilla/sccache/issues/964.

- We will upload the file with it's content hash
- Everytime we downloaded a file, we verify it's content by comparing it's hash with filename.

I believe `the bad cache` problem is not the same scoped we discussed here.

Xuanwo:
> I think there are significant feature changes that warrant a discussion before merging the changeset

I didn't see there is a need to revert this PR back.

## Current Status

Current GHA implements work as expected: Take https://github.com/mozilla/sccache/actions/runs/3870677953/jobs/6597844182 as an example.

The first build finished in 1m53s: (normally, they should be 4m30s)

```
Cache hits                          274
Cache hits (C/C++)                   50
Cache hits (Rust)                   224
```

The second build finished in 1m21s:

```
Cache hits                          548
Cache hits (C/C++)                  100
Cache hits (Rust)                   448
```

We can draw the following conclusion from this build:

- gha support works as expected (can read & can write)
- we can read `main` branch's cache from a PR (can share)

So @jakelee8's concerning of cache not reused is not true.

## Future Improvement

### docs

There is a PR to improve the docs: https://github.com/mozilla/sccache/pull/1533 (Thank @jakelee8 for point it out! :heart:, I used to treat them as internal details)

### `cache_to`/`cache_from` features

They can be implemented as adding a prefix/suffix for our cache key. I prefer to ship the basic GHA support first and then have a discussion on whether we need to add this feature.

### bad cache concern

As I discussed in previous comments: https://github.com/mozilla/sccache/pull/1528#issuecomment-1375086739, this is not a storage level problem. This problem also exists on local fs and s3. We need to address them in a higher level.

docs/GHA.md
Xuanwo: Should be addressed by https://github.com/mozilla/sccache/pull/1533

jakelee8:
We already have GHA support with `cache_to` and `cache_from` support. This PR breaks existing use cases by removing the `SCCACHE_GHA_CACHE_TO` and `SCCACHE_GHA_CACHE_FROM` env vars. Restoring env vars will address my concerns.

I agree with @drahnr that this PR should be reverted. It's a breaking change that removes functionality and makes the cache implementation inconsistent with other uses for GitHub Actions cache.

Additionally, internally, the cache "version" represents a hash of the input file. A more appropriate name would be "cache key".

The bad cache concern for `sccache` itself, but misconfiguration by end users. I have tested `sccache` with GitHub Actions cache extensively at Bitski, and it wasn't easy to configure everything correctly.

Sorry for the unclear feedback yesterday.

Xuanwo:
## TL;DR

Sccache doesn't need `cache_to` and `cache_from`.

## Explaination

In this section, I will explain why sccache doesn't need `cache_to` and `cache_from`. And why it's different from other actions that use GitHub action cache services.

### Github Action Cache Service

GitHub Action Cache Service is an **immutable** storage system that provides **prefix matching** query support.

It provides the following APIs (Please correct me if anything I understand wrongly):

#### Query Cache

`GET /cache?keys=abc,ab,a&version=v1`

- keys: comma-separated list of keys to query. Keys can be queried by prefix and do not need to match exactly. The newest record matching a prefix is returned.

For example, if we have cache an entry `v0-rust-abc` and `v0-rust-def`. `GET /cache?keys=v0-rust-` will return the newer one between them as they match the prefix.

- version: the namespace for different cache keys. Only the cache key in the same version can be queried.

#### Reserve Cache

`POST /caches`

- inputs: `{key: <cache_key>, version: <cache_version>}`.
- outputs: `{cache_id: <cache_id>}`

Once `(cache_key, cache_version)` has been reserved, subsequent requests with the same key/version will receive an "already exists" error. There is no way to overwrite the same cache key.

#### Upload Cache

`PATCH /caches/[cache_id]`

Upload content with `Content-Range`.

#### Create Cache

`POST /caches/[cache_id]`

Create a cache after all content is uploaded.

> For reference: [Github Actions Cache service API](https://github.com/tonistiigi/go-actions-cache/blob/master/api.md)

### Github Action Cache Loading

GitHub provides cache isolation between the default branch and other branches.

- all branches can load cache from the `main` branch and itself.
- all PRs can load cache from `main`, its base, and itself.

Given there are two PRs, `feature-a` and `feature-by, which are both based on `main`:

- `feature-a` **can** read cache from `feature-a` and `main` branches.
- `feature-b` **can't** read cache from `feature-b`.

The cache can be shared across the workflow.

### How does current `sccache` use Github Action Cache Service?

sccache's design eliminates the usage of `restore-keys`:

image

sccache will ensure we have a unique cache key for every given input, which is safe to share.

- For write: use the output path sccache generated as the cache key
- For reading: query the exact cache key

And `version` is provided to allow the user to purge all cache manually.

### Why `sccache` is different from `actions/cache`?

`actions/cache` provides `restore-keys` because 

- They write the given path in an archive instead of a single file.
- They need to ensure the same file path in the archive is the same.

**But sccache does**.

- sccache writes every cache entry in a single file path with a unique key.
- If the key is different, they are different cache entries.

---

Let's come back to the comments.

> We already have GHA support with cache_to and cache_from support. This PR breaks existing use cases by removing the SCCACHE_GHA_CACHE_TO and SCCACHE_GHA_CACHE_FROM env vars. Restoring env vars will address my concerns.

I removed them because we don't need them.

> It's a breaking change that removes functionality

No, everything works except a not-needed feature were removed.

> makes the cache implementation inconsistent with other uses for GitHub Actions cache.

No, sccache can't work with other cache-related actions.

Other actions can't reuse the cache sccache write. And sccache also can't read the cache other actions write.

> Additionally, internally, the cache "version" represents a hash of the input file. A more appropriate name for the opendal implementation would be "cache key".

No. `version` doesn't require to be hash of input file paths.

> The bad cache concern is not for sccache itself, but misconfiguration by end users.

This should be treated as a bug if misconfiguration could lead to bad cache. Adding more configurations like `cache_to` and `cache_from` doesn't make end users' life easier.

In other newer versions, we added a storage check for all storage services, which will return errors directly if storage is misconfigured. Please give it a try~

> I have tested sccache with GitHub Actions cache extensively at Bitski, and it wasn't easy to configure everything correctly.

If sccache was challenging to configure everything correctly, we must go the wrong way.

Please try [v0.4.0-pre.5](https://github.com/mozilla/sccache/releases/tag/v0.4.0-pre.5), I'm willing to fix any bug while using GHA Cache with sccache.

jakelee8:
Thank you for the detailed explanation. Without `cache_to` and `cache_from` (or something like it), write-only cache is not supported. That said, I no longer maintain any Rust code that benefits from `sccache` and won't be able to help test these changes.

Xuanwo:
> write-only cache is not supported

Can you explain a bit how `write-only` works on github action caches? Thanks.

jakelee8:
With https://github.com/actions/cache, a write-only cache can be configured with:

* `key: sccache-key` - write to this key
* `restore-keys:` - blank, don't restore from other keys

And a read-write cache can be configured with:

* `key: sccache-key` - write to this key
* `restore-keys: sccache-` - restore from this prefix

I typically use `${{ github.sha }}` in place of `key` to force newer commits to main to override older caches. This is useful if the cache key did not include configuration that `sccache` doesn't capture in the hash, and ensures that PR branches use cached objects from a known good build.

For example, I ran into issues with `sccache` after upgrading the Rust toolchain. The cached binary object was missing logic present in the latest commit. The issue was first resolved by disabling `sccache`. To prevent the errors in later releases, I implemented the write-only cache for `sccache` and added the toolchain version and target platform in the cache key.

jakelee8:
The `cache-to` and `cache-from` naming scheme comes from Docker: https://docs.docker.com/build/cache/backends/gha/

Xuanwo:
> For example, I ran into issues with `sccache` after upgrading the Rust toolchain. The cached binary object was missing logic present in the latest commit.

Oh, it must be a bug. Scache should generate different cache path after rust toolchain bumped.

https://github.com/mozilla/sccache/blob/1a82863429a2bf7076764263187e6c85744e843b/src/compiler/rust.rs#L378-L383

So far we only take `host` into consideration. I will work on fixing this.

jakelee8:
The Rust version should be part of the hash as well because the ABI is not stable.

Xuanwo:
> The Rust version should be part of the hash as well because the ABI is not stable.

Yep! I will take the whole output of `rust -vV` as hash content:

```shell
:) rustc -vV
rustc 1.66.1 (90743e729 2023-01-10)
binary: rustc
commit-hash: 90743e7298aca107ddaa0c202a4d3604e29bfeb6
commit-date: 2023-01-10
host: x86_64-unknown-linux-gnu
release: 1.66.1
LLVM version: 15.0.2
```

PR #1534, opened 2023-01-09
refactor: Do read check first instead

Xuanwo:

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -392,18 +392,22 @@ impl Storage for opendal::Operator {
 
         let path = ".sccache_check";
 
-        let can_write = match self.object(path).write("Hello, World!").await {
-            Ok(_) => true,
-            Err(err) if err.kind() == ErrorKind::ObjectAlreadyExists => true,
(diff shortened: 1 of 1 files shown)


src/cache/cache.rs
```diff
@@ -392,20 +392,24 @@ impl Storage for opendal::Operator {
...
             Ok(_) => (),
             // Read not exist file with not found is ok.
             Err(err) if err.kind() == ErrorKind::ObjectNotFound => (),
             Err(err) => bail!("cache storage failed to read: {:?}", err),
         };
 
+        let can_write = match self.object(path).write("Hello, World!").await {
+            Ok(_) => true,
+            Err(err) if err.kind() == ErrorKind::ObjectAlreadyExists => true,
+            // Toralte all other write errors because we can do read as least.
```
sylvestre, line 406:
```suggestion
            // Tolerate all other write errors because we can do read as least.
```

sylvestre:
why are you making this change? you saw that as an issue ?

Xuanwo:
Some services (like gha) has more strict limit on write operations. By introducing this change, we are allowing user to read from cache after write check failed.

Xuanwo:
PR description also updated.

PR #1550, opened 2023-01-11
feat: Introduce the default gha version

Xuanwo:

This PR will introduce the default gha version.

This change has the following benefits:

- User don't need think and pick version if they don't care or don't wanna purge cache
- Give sccache a chance to purge users cache while needed (serious bug happened)

To support this change, I did following:

- Add `SCCACHE_GHA_ENABLED` as a flag
- make `SCCACHE_GHA_VERSION` an optional version.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -195,7 +195,7 @@ jobs:
     needs: build
 
     env:
-      SCCACHE_GHA_VERSION: "sccache_integration_tests"
+      SCCACHE_GHA_ENABLED: "on"
       RUSTC_WRAPPER: /home/runner/.cargo/bin/sccache
 
     steps:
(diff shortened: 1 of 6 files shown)


sylvestre:
@jakelee8 Do you have an opinion on this ? :)

jakelee8:
LGTM.

jakelee8:
The default version could include `env!("CARGO_PKG_VERSION_MAJOR")` and maybe minor to automatically change with breaking changes.

Xuanwo:
> The default version could include `env!("CARGO_PKG_VERSION_MAJOR")` and maybe minor to automatically change with breaking changes.

Nice idea, I will add it.

Xuanwo:
cc @sylvestre @jakelee8 all comments have been addressed, PTAL.

Issue #1552, opened 2023-01-11
Sccache didn't take rustc version into hash calculation


Xuanwo:
> For example, I ran into issues with `sccache` after upgrading the Rust toolchain. The cached binary object was missing logic present in the latest commit.

Oh, it must be a bug. Scache should generate different cache path after rust toolchain bumped.

https://github.com/mozilla/sccache/blob/1a82863429a2bf7076764263187e6c85744e843b/src/compiler/rust.rs#L378-L383

So far we only take `host` into consideration. I will work on fixing this.

_Originally posted by @Xuanwo in https://github.com/mozilla/sccache/issues/1528#issuecomment-1378303484_

PR #1553, opened 2023-01-11
feat: Take the whole rustc version into hash calculation

Xuanwo:

Fix https://github.com/mozilla/sccache/issues/1552

This is a use case that may break user's build after rust version bumped.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -955,7 +955,7 @@ where
 
         match rustc_vv {
             Ok(rustc_verbose_version) => {
-                debug!("Found rustc");
+                debug!("Found rustc at path: {rustc_executable:?}");
 
                 let rustc_executable2 = rustc_executable.clone();
 
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -955,8 +955,6 @@ where
 
         match rustc_vv {
             Ok(rustc_verbose_version) => {
-                debug!("Found rustc at path: {rustc_executable:?}");
-
                 let rustc_executable2 = rustc_executable.clone();
 
                 let proxy = RustupProxy::find_proxy_executable::<T>(
@@ -1007,6 +1005,8 @@ where
                         trace!("Compiling rust without proxy");
                         (None, rustc_executable2)
                     });
+
+                debug!("Using rustc at path: {resolved_rustc:?}");
 
                 Rust::new(
                     creator,
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -3165,6 +3165,7 @@ proc_macro false
         m.update(b"=");
         OsStr::new("foo").hash(&mut HashToDigest { digest: &mut m });
         f.tempdir.path().hash(&mut HashToDigest { digest: &mut m });
+        TEST_RUSTC_VERSION.hash(&mut HashToDigest { digest: &mut m });
         let digest = m.finish();
         assert_eq!(res.key, digest);
         let mut out = res.compilation.outputs().map(|k| k.key).collect::<Vec<_>>();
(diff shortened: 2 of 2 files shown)


drahnr:
The rustc binary is hashed already, which by definition must change if the output of `rustc -vV` would change.

Could you elaborate on the precise case that would require the version to be hashed in addition?

Xuanwo:
> The rustc binary is hashed already, which by definition must change if the output of `rustc -vV` would change.

I must miss this part. Can you point out the detailed place which `rustc` been hashed? Thanks.

drahnr:
I was writing that from memory, since there were issues in the past when switching from nightly to stable or vice versa, I have yet to double check the code

Xuanwo:
> I was writing that from memory, since there were issues in the past when switching from nightly to stable or vice versa, I have yet to double check the code

My current understanding is we just hashed the path to `rustc` instead `rustc` itself, thus we need to the verbose version get hashed.

sylvestre:
it is the path and it is coming from rustup, it should be something like 
~/.rustup/toolchains/1.58-x86_64-unknown-linux-gnu/bin/rustc ?

Xuanwo:
> it is the path and it is coming from rustup, it should be something like ~/.rustup/toolchains/1.58-x86_64-unknown-linux-gnu/bin/rustc ?

I'm afraid that we just found `cargo`:

```shell
[2023-01-11T08:34:03Z DEBUG sccache::compiler::compiler] Found rustc at path: "/home/xuanwo/.cargo/bin/rustc"
```

This can be reproduced by change:

```shell
- debug!("Found rustc");
+ debug!("Found rustc at path: {rustc_executable:?}");
```

glandium:
> it is the path and it is coming from rustup, it should be something like ~/.rustup/toolchains/1.58-x86_64-unknown-linux-gnu/bin/rustc ?

I don't think that's guaranteed. It could well be ~/.cargo/bin/rustc. Also, in most cases it would be ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc.

glandium:
And yeah, when the path is ~/.cargo/bin/rustc, you end up caching rustup's binary, which never changes.

sylvestre:
> > it is the path and it is coming from rustup, it should be something like ~/.rustup/toolchains/1.58-x86_64-unknown-linux-gnu/bin/rustc ?
> 
> I don't think that's guaranteed. It could well be ~/.cargo/bin/rustc. Also, in most cases it would be ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc.

or /usr/bin/rustc ;)

glandium:
There is actually code to unwrap the rustup proxy already: https://github.com/mozilla/sccache/blob/1a82863429a2bf7076764263187e6c85744e843b/src/compiler/compiler.rs#L962

src/compiler/compiler.rs
```diff
@@ -955,7 +955,7 @@ where
 
         match rustc_vv {
             Ok(rustc_verbose_version) => {
-                debug!("Found rustc");
+                debug!("Found rustc at path: {rustc_executable:?}");
```
glandium, line 958: Might want to log the unwrapped executable instead.

Xuanwo: Addressed! Current logging:

```shell
[2023-01-11T09:01:05Z DEBUG sccache::compiler::compiler] Using rustc at path: "/home/xuanwo/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc"
```

drahnr:
> The rustc binary is hashed already, which by definition must change if the output of `rustc -vV` would change.
> 
> Could you elaborate on the precise case that would require the version to be hashed in addition?

This was wrong, I got confused and thought the sysroot would include rustc, but it apparently does not. Sorry for the red hering.

Issue #1484, opened 2022-12-17
Print warning message by `println` instead of `warn` to avoid missing


Xuanwo:
https://github.com/mozilla/sccache/blob/ad97f69c1adeb12c18eafea3a364e7708b166536/src/config.rs#L538-L545

We print warning message by `warn` here, which could be missed if `RUST_LOG` is not set or been set to `ERROR`. It's better to use println instead.

Xuanwo:
@sylvestre Can you help mark this issue as `help wanted` and `good first issue`? Thanks.

PR #1554, opened 2023-01-11
Use println instead of warn to avoid missing deprecated message

b1tg:
Signed-off-by: b1tg <b1tg@protonmail.ch>

Fix https://github.com/mozilla/sccache/issues/1484

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -537,8 +537,8 @@ fn config_from_env() -> Result<EnvConfig> {
 
 
         if env::var("SCCACHE_GCS_OAUTH_URL").is_ok() {
-            warn!("SCCACHE_GCS_OAUTH_URL has been deprecated");
-            warn!("if you intend to use vm metadata for auth, please set correct service account intead");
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -537,8 +537,8 @@ fn config_from_env() -> Result<EnvConfig> {
 
 
         if env::var("SCCACHE_GCS_OAUTH_URL").is_ok() {
-            warn!("SCCACHE_GCS_OAUTH_URL has been deprecated");
-            warn!("if you intend to use vm metadata for auth, please set correct service account intead");
+            println!("SCCACHE_GCS_OAUTH_URL has been deprecated");
```
glandium, line 540: this should probably use eprintln

b1tg: fixed

Xuanwo (approved):

drahnr:
Could we possibly print _all_ warning messages instead? I don't know if that'd be excessive. It have a hunch we have more places where a warning to the user would be desirable.

sylvestre:
Let's start here and follow up in other pr if needed

PR #1555, opened 2023-01-11
Determine support for -fminimize-whitespace via test execution.

Meinersbur:
Commit da2934f/PR #1162 added the -fminimize-whitespace flag when proprocessing and derived compiler support from the compiler version (clang >= 14). It unfortunately had a bug in the detection mechanism and also the unittest such that the flag was never used:

 1. Clang expands the __VERSION__ macro to a string (e.g. "LLVM Clang 14.0.0"). The additional quotes caused the version check to fail. Originally, the --version output was passed to the version check function which does not contain such quotes.

 2. The regression test checked for whether the compiler is "clang", but not "clang++".

 3. Wrong regex group names used for checking for Apple caused the check to always expect a cache miss.

This patch changes the check whether -fnormalize-whitespace is supported to a test invocation with that flag. Only the regression test continues to be based on the know versions that support the flag to be independent on the feature test. When Apple clang one day adds support for the flag, the test will need to be updated.


src/compiler/compiler.rs
```diff
@@ -1107,18 +1105,49 @@ __VERSION__
...
+            }
+            Some(s) if s.trim().is_empty() => None,
+            Some(s) => Some(s.to_owned()),
+            None => None,
+        };
         match kind {
             "clang" | "clang++" | "apple-clang" | "apple-clang++" => {
                 debug!("Found {}", kind);
+
+                // Check whether the compiler also supports the -E -fminimize-whitespace flag
```
sylvestre, line 1126: Please move this into a separate function and add a test to verify it won't regress

src/compiler/compiler.rs
```diff
@@ -1107,18 +1105,49 @@ __VERSION__
...
-            // In case the compiler didn't expand the macro.
-            .filter(|&line| line != "__VERSION__")
-            .map(str::to_owned);
+        // Unescape the version string. We only remove surrounding quotes assuming that the version string does not
+        // contain characters that needed escaping. If the compiler did not know the macro (it is a gcc extension and
+        // not supported by eg. msvc), the line will still contain __VERSION__ and we have no version information.
+        // In case the macro was expanded to something other than a quoted string, still assume that it represents
+        // the compiler version.
+        let version = match lines.next() {
+            Some("__VERSION__") => None,
```
sylvestre, line 1114: Please mive this into a function too with some tests. Thanks

sylvestre:
@Meinersbur ping ?

sylvestre (changes requested):
see above

Meinersbur:
Sorry for the waiting, I was exploring test strategies and I am not very familiar with Rust.

I poked some abstraction holes to be able to check the `has_fminimize_whitespace` flag in `test_detect_compiler_kind_clang` (`as_any()`, `pub compiler`), but checking the result of `detect_fminimize_whitespace_support` in two separate tests. Which approach do you prefer?

Meinersbur:
I removed the test that requires a downcasting `detect_compiler` and kept the one that just calls `detect_fminimize_whitespace_support` . I would be happy for another review.

Meinersbur:
As far as I understand, the test failures are unrelated. 

 * `test freebsd-13.1 rust stable` also fails on main
 * The integration test seem to fail sporadically due to misses using a remote URL cache, e.g. https://github.com/mozilla/sccache/actions/runs/4722730937/jobs/8377898781

Xuanwo:
> * The integration test seem to fail sporadically due to misses using a remote URL cache, e.g. https://github.com/mozilla/sccache/actions/runs/4722730937/jobs/8377898781

I think we should not depend on ghac while running tests like `clang`. For tests, we should use local fs instead. We can remove https://github.com/mozilla/sccache/actions/runs/4938282959/workflow?pr=1555#L412 directly.

sylvestre:
needs rebasing, please reopen when ready.

PR #1557, opened 2023-01-12
feat: Allow sccache keep running after hitting rate limit during check

Xuanwo:

This PR will allow sccache to keep running after hitting the rate limit error.

We introduce this change because we could have services that have very strict limit on our requests. During storage check, it's better to tolerate them and keep running because there are few things either we or our users can do on this.

### DIFF
--- a/docs/GHA.md
+++ b/docs/GHA.md
@@ -12,3 +12,7 @@ This cache type will needs token like `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_T
       core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
       core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 ```
+
+## Behavior
+
+If case sccache reaches the rate limit of the service, the build will continue but the storage might not be performed.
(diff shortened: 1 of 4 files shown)


drahnr (approved):

src/cache/cache.rs
```diff
@@ -397,6 +397,18 @@ impl Storage for opendal::Operator {
             Ok(_) => (),
             // Read not exist file with not found is ok.
             Err(err) if err.kind() == ErrorKind::ObjectNotFound => (),
+            // Tricky Part.
+            //
+            // We tolerate rate limited here to make sccache keep running.
```
sylvestre, line 402: maybe document that in the gha doc ?

something like "in case sccache reaches the rate limit of the service, the build will continue but the storage might not be performed"

Xuanwo: Added.

docs/GHA.md
```diff
@@ -12,3 +12,7 @@ This cache type will needs token like `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_T
       core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
       core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 ```
+
+## Behavior
+
+If case sccache reaches the rate limit of the service, the build will continue but the storage might not be performed.
```
sylvestre, line 18:
```suggestion
In case sccache reaches the rate limit of the service, the build will continue but the storage might not be performed.
```

Xuanwo: Oh, my mistake. Sorry for bother you everytime to correct my typo.

PR #1562, opened 2023-01-15
add a profile section

sylvestre:
To make sure the binaries we publish are the best

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,12 @@ name = "sccache"
 name = "sccache-dist"
 required-features = ["dist-server"]
 
+[profile.release]
+codegen-units = 1
+lto = true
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):

Cargo.toml
```diff
@@ -17,6 +17,12 @@ name = "sccache"
 name = "sccache-dist"
 required-features = ["dist-server"]
 
+[profile.release]
+codegen-units = 1
+lto = true
+strip = true
+panic = "abort"
```
drahnr, line 24: Could you explain why we're not keeping backtraces? The performance overhead exists, but compared to cache fetches and launching other binaries, it should be rather negligible (gut feeling)?

PR #1563, opened 2023-01-16
fix: Sccache dist tests broken after bump to tokio 1.21 and later

Xuanwo:
It's a long-term problem that sccache dist tests failed after bumping tokio.

After a bisect of all commits between tokio 1.20.1 to 1.21.0, I confirmed that this problem was introduced in https://github.com/tokio-rs/tokio/pull/4840. The PR itself is simple and correct, but it can lead to problems in `hyper` (`reqwest` affected too).

After this PR, hyper can't reuse connections between different runtimes. To address this problem, we should ensure all requests sent by hyper are from the same runtime. But it's a huge change based on our current design. So in this PR, I disabled the keep-alive of reqwest to make sure everything just works as before.

Changes that are included in this PR:

- Disable keep-alive in hyper side
  - NOTE: we always set `keep-alive: Close` by hand, so this change won't affect of dist server's perf.
- Use the new image for testing (ubuntu 22.04)
  - Use openssl 3.0 and a new version of glibc, which makes it possible to test locally on modern systems.
- Use `localhost` with network `host` instead of `docker` container IP
  - Make it possible to test with complex and different docker setups. For example, use `podman-docker` without the `docker0` network bridge.
- Bump the version of `tokio` and `tokio-utils`
- Smaller our lock scope

Future Plan:

- Should we migrate to an async web framework instead?
- Maybe we need to init the async runtime first and always stick to the same one?
- We should not spawn a new thread to use `reqwest::blocking::Client`.

Close https://github.com/mozilla/sccache/pull/1541
Close https://github.com/mozilla/sccache/pull/1543

### DIFF
--- a/.github/actions/artifact_failure/action.yml
+++ b/.github/actions/artifact_failure/action.yml
@@ -1,0 +1,25 @@
+name: "Upload failure Artifacts"
+description: "Upload failure Artifacts"
+inputs:
+  name:
+    description: ""
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: pack failure artifacts
+      shell: bash
+      run: |
+        killall sccache || true
+        killall sccache-dist || true
+
+        tar --exclude='target' \
+            --exclude='docs' \
+            --exclude='bins' \
+            --exclude='.git' \
+            --exclude='failure-*' \
+            -zcf failure-${{ inputs.name }}.tar.gz .
+    - uses: actions/upload-artifact@v3
+      with:
+        name: ${{ inputs.name }}
+        path: failure-${{ inputs.name }}.tar.gz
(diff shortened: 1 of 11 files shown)


Xuanwo:
cc @sylvestre and @messense for review

Xuanwo:
Ooops, freebsd still failing..

sylvestre:
Many thanks

sylvestre:
@Xuanwo it seems that test ubuntu-22.04 rust 1.60.0 dist-server  is failing
https://github.com/mozilla/sccache/actions/runs/4032885110/jobs/6933042098

```

Download action repository 'actions/upload-artifact@v3' (SHA:0b7f8abb1508181956e8e162db84b466c27e18ce)
Run ./.github/actions/artifact_failure
Run killall sccache || true
sccache-dist: no process found
tar: .: file changed as we read it
Error: Process completed with exit code 1.
```

PR #1571, opened 2023-01-19
Support nvcc's `--threads` flag

trxcllnt:
This PR adds support for nvcc's [`--threads`](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#threads-number-t) flag.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -100,6 +100,8 @@ pub struct ParsedArguments {
     pub common_args: Vec<OsString>,
     /// Commandline arguments for the compiler that specify the architecture given
     pub arch_args: Vec<OsString>,
+    /// Commandline arguments for the preprocessor or the compiler that don't affect the computed hash.
+    pub unhashed_args: Vec<OsString>,
     /// Extra files that need to have their contents hashed.
     pub extra_hash_files: Vec<PathBuf>,
     /// Whether or not the `-showIncludes` argument is passed on MSVC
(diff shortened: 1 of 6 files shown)


Xuanwo (approved):

drahnr (approved):

sylvestre:
could you please run rustfmt on your change? thanks

trxcllnt:
@sylvestre done, thanks.

About this PR, the `--threads` flag doesn't affect anything about the output (just how many device architectures `nvcc` compiles in parallel under the hood). Is there an elegant way to pass this flag through to `nvcc`, but not allow it to affect the computed hash?

This would prevent existing caches from being invalidated, but more importantly, it would allow devs (and CI machines) with different values for `--threads` to still share the same cache.

trxcllnt:
I added a new `unhashed_args` vector + `Unhashed` arg type. Need to investigate how to test the new behavior besides running locally.

src/compiler/c.rs
```diff
@@ -100,6 +100,8 @@ pub struct ParsedArguments {
     pub common_args: Vec<OsString>,
     /// Commandline arguments for the compiler that specify the architecture given
     pub arch_args: Vec<OsString>,
+    /// Commandline arguments for the preprocessor or the compiler that don't affect the computed hash.
+    pub unhashed_args: Vec<OsString>,
```
drahnr, line 104: I'd like to have another statement, that none of these arguments listed may in any circumstance change the resulting binary.

drahnr (approved):
I think this is a good structural change.

Thank you!

PR #1572, opened 2023-01-20
Support MSVC flag 'external:anglebrackets'

Tiwalun:
Add support for the '/external:anglebrackets' flag, as documented here: https://learn.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170#arguments


sylvestre:
thanks!

PR #1581, opened 2023-01-29
fix: Send heartbeat before starting dist server

Xuanwo:

This PR introduces the following changes:

- Fix CI scripts
- Make sure heartbeat has been sent before starting dist server

### DIFF
--- a/.github/actions/artifact_failure/action.yml
+++ b/.github/actions/artifact_failure/action.yml
@@ -11,7 +11,6 @@ runs:
       shell: bash
       run: |
         killall -r ".*sccache.*" || true
-        docker kill $(docker ps -q) || true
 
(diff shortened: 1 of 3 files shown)


Xuanwo:
Hi, @sylvestre, please re-run `ubuntu-22.04 rust 1.60.0 dist-server` serval times to make sure it's stable. Thanks, I can't reproduce it locally(

drahnr:
@Xuanwo what was the initial intuition about the change being needed?

Xuanwo:
> @Xuanwo what was the initial intuition about the change being needed?

I'm trying to fix the broken CI, and it seems dist-server can't start correctly after restart.

However, this PR seems not the root cause. I'm still investigating it.

PR #1582, opened 2023-01-29
feat: Add assembler support (.S, .s, .sx)

Xuanwo:

This PR will add `.S`, `.s` and `.sx` support. By this PR, we can cache projects that have pregenerated assembler files now! For example: `ring` in rust.


Xuanwo:
Ooops, it does not work as expected.

PR #858, opened 2020-09-28
Add the ability to clear a cache.

rthomas:
This adds the --clear-cache argument that will remove the contents of
the cache. This is for issue #835.

This is implemented as `clear()` on the `Storage` trait - however it
is only implemented for the `DiskCache` (and `LruDiskCache`)
implementations. Any invocation on the other implementations will
return an `Err`.

A test for the `LruDiskCache` has also been included in this commit.

PR #1588, opened 2023-01-30
Add the ability to clear a cache

### LINKED ISSUE #835: Add command to clean local cache
Maybe I missed it, but it would be nice if you could run `sccache --clean` to clean the local cache. As it is I have to run `rm -rf ~/Library/Caches/Mozilla.sccache` which is not very nice.

xxchan:
replace #858 resolve #835

This adds the --clear-cache argument that will remove the contents of
the cache.

This is implemented as `clear()` on the `Storage` trait - however it
is only implemented for the `DiskCache` (and `LruDiskCache`)
implementations. Any invocation on the other implementations will
return an `Err`.

A test for the `LruDiskCache` has also been included in this commit.

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -357,6 +357,9 @@ pub trait Storage: Send + Sync {
 
     /// Get the maximum storage size, if applicable.
     async fn max_size(&self) -> Result<Option<u64>>;
+
+    /// Clear the contents of the cache.
+    async fn clear(&self) -> Result<()>;
 }
 
 /// Implement storage for operator.
@@ -445,6 +448,11 @@ impl Storage for opendal::Operator {
 
     async fn max_size(&self) -> Result<Option<u64>> {
         Ok(None)
+    }
+
+    async fn clear(&self) -> Result<()> {
+        let meta = self.metadata();
+        Err(anyhow!("Cannot clear the cache on {}", meta.scheme()))
     }
 }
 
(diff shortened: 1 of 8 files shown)


sylvestre:
Thanks.

Could you please be more aggressive in the error when this option is used where the cache cannot be cleared?
And add a test to trigger this error to make sure it doesn't disappear ?

xxchan:
What do you mean by "be more aggressive" and "the error doesn't disappear" (why it may disapper?)? I didn't fully get it 🤔

sylvestre:
What happens if you try to use the option to clear the cache with aws?

sylvestre:
ping ?

Xuanwo:
opendal based services can use remove_all to implement this feature. This PR just return an error.

sylvestre:
> This PR just return an error.

Sure but i would like what the error looks like :)

xxchan:
Got it.
```
> sccache --clear-cache
sccache: error: couldn't clear cache on server
sccache: caused by: Failed to send data to or receive data from server
sccache: caused by: Failed to read response header
sccache: caused by: failed to fill whole buffer
```

Fixed now:

```
> sccache --clear-cache
sccache: error: failed to clear cache
sccache: caused by: Cannot clear the cache on s3
```

Xuanwo:
> since I couldn't find a similar test case to learn about how to setup the server on s3(minio) :(


I guess you can test it in out integration tests.

xxchan:
I’m not interested in this PR anymore. Free free to reopen and edit it.

ahartmetz:
Commenting as a bookmark, I might work on this

Issue #885, opened 2020-11-19
dist-client does not print directory path when cache dir creation fails.


Firstyear:
```
ERROR 2020-11-19T01:17:29Z: sccache::server: Error: failure during dist client creation
Caused by: failed to initialise client toolchains
Caused by: failed to create top level toolchain cache dir
Caused by: Permission denied (os error 13)
```

This menas I don't know where to correct to fix up to allow the toolchain dir to be created.

It would also be good to highlight in the quickstart that the cache_dir parameter is able to influence this.

drahnr:
You might want to checkout paritytech/sccache

drahnr:
Or particularly https://github.com/paritytech/sccache/pulls/10

PR #1611, opened 2023-02-18
dist/cache: include paths in error contexts

mathstuf:
Fixes: #885

---
In the interest of helping 0.4.0 get finalized :) .

### DIFF
--- a/src/dist/cache.rs
+++ b/src/dist/cache.rs
@@ -61,27 +61,39 @@ mod client {
             toolchain_configs: &[config::DistToolchainConfig],
         ) -> Result<Self> {
             let cache_dir = cache_dir.to_owned();
-            fs::create_dir_all(&cache_dir)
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
Thanks for contribution, LGTM!

sylvestre:
Would it be possible to have tests to cover these errors?

mathstuf:
Maybe? What fs abstractions do we have available? FUSE is a possibility, but feels like a lot of code to test error paths. Also doesn't help with Windows at all.

drahnr:
Is there a particular reason to not use `fs-err` instead?

mathstuf:
That seemed like a wider-effecting change than the issue was asking for. And given that 0.4.0 is now "one" issue away by milestoned issues…I'd like to see 0.4.0 come out first personally.

PR #1612, opened 2023-02-19
fix: Disable connection pool to fix dist server feature

Xuanwo:
The root cause of dist server restart failure seems related to request's http keep-alive logic.

More details could be found here: https://github.com/mozilla/sccache/pull/1563

In the future, we should use the same global client instead and bring connection pool back.


Xuanwo:
I have repeated the test of dist server several times, it is now stable!

sylvestre:
excellent, thanks. let's see if it sticks.

sylvestre:
what is the impact for the users ?

Xuanwo:
> what is the impact for the users ?

No difference.

src/util.rs
```diff
@@ -506,6 +506,14 @@ pub fn daemonize() -> Result<()> {
     Ok(())
 }
 
+/// Disable connection pool to avoid broken connection between runtime
```
drahnr, line 509: nit: I'd prefer a reference to the bug this caused or a more elaborate explanation inline

drahnr (approved):
One nit, other than that, looks good to me

src/util.rs
Xuanwo: Got it, fixed!

sylvestre:
A job is failing to build. Maybe related to the fs err change?

Xuanwo:
> A job is failing to build. Maybe related to the fs err change?

Fixed now.

drahnr (approved):

sylvestre:
Thanks for the squash :)

PR #1613, opened 2023-02-20
Ignore RPM environment variables

Firstyear:
RPM sets a number of environment variables by default that interfere with the sccache's ability to cache artefacts. Some of these variables are also used by other applications.

This extends the default ignore list to allow sccache to be used under different environments for caching.

An alternate idea is a way to configure and define a list of environment variables to ignore.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -278,7 +278,19 @@ pub fn try_parse() -> Result<Command> {
                 //
                 // FIXME: Maybe we should strip out `LD_PRELOAD` always?
                 if env::var_os("RUNNING_UNDER_RR").is_some() {
-                    env_vars.retain(|(k, _v)| k != "LD_PRELOAD" && k != "RUNNING_UNDER_RR");
+                    env_vars.retain(|(k, _v)| {
+                        k != "LD_PRELOAD"
+                            && k != "RUNNING_UNDER_RR"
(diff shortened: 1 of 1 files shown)


Xuanwo:
Thank you for the contribution. But it seems a package related to issues that should be addressed by the providers?

Firstyear:
We've been having to apply this patch for the last year or so to maintain sccache in the opensuse build service. However, it would likely be beneficial to others which is why I submitted it.

Xuanwo:
Great, please format you code to pass the CI first. And I will invite someone who is more familiar with this part to take a look.

Firstyear:
Done :)

drahnr:
@Firstyear could you elaborate on each of the added entries _how_ it messes up a compilation? Wouldn't one want to include the `RPM_*` flags to make sure the behavior is consistent when building as rpm vs standalone when using `sccache`?

Firstyear:
@drahnr Every single one of these values *changes* every single time RPM is invoked, especially in an automated build environment like OBS. Since sccache will only match on cache items if the environment flags are *identical*, then since these values are always change, sccache will always cache miss making it ineffective.

drahnr:
Thank you!

drahnr (approved):

Firstyear:
Thanks for this! It's really appreciated!

PR #1616, opened 2023-02-21
use fs_err for sake of better io errors

drahnr:
Provides path names of failing io operations where possible by default, both in the resulting binaries which ease debugging of i.e. selinux or seccomp filters, without resorting to strace, but also local unit test file operations.

The [identical change was made in cachepot](https://github.com/paritytech/cachepot/pull/19)


drahnr:
Superseeds https://github.com/mozilla/sccache/pull/1611 which covers two instances, where this covers _all_ io operations except for two, where the `std::fs::File` API is required for external crates.

PR #1617, opened 2023-02-23
build: Increase edition to 2021

Xuanwo:


drahnr:
Do you want to run a clippy fix edition pass as part of this PR?

Xuanwo:
> Do you want to run a clippy fix edition pass as part of this PR?

Seems nothing to fix.

Issue #1620, opened 2023-02-27
Port remaining cachepot changesets


drahnr:
There are a few viable features, yet to be ported from #paritytech/cachepot :

* [ ] https://github.com/paritytech/cachepot/pull/128
* [ ] https://github.com/paritytech/cachepot/pull/139
* [ ] https://github.com/paritytech/cachepot/pull/140

I plan on porting them in the next few weeks

PR #1628, opened 2023-02-27
Custom sandboxing implementation as linux usernamespace calls - port cachepot/#128

drahnr:
Implementes a custom sandboxing implementation as linux usernamespace calls.

It's opt-in, and has to be enabled with setting the `SCCACHE_SANDBOX` environment variable.

Ref #1620 

Original work done by @Xanewok

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
         run: cargo ${{ matrix.cargo_cmd }}
 
   test:
-    name: test ${{ matrix.os }} rust ${{ matrix.rustc || 'stable' }} ${{ matrix.extra_desc }}
+    name: test ${{ matrix.os }} rust ${{ matrix.rustc || 'stable' }} ${{ matrix.extra_desc }} ${{ matrix.dist_exec_strategy }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


drahnr:
@sylvestre `/sccache-dist: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory` not sure how to resolve this, do you have an insight into the container env?

.github/workflows/ci.yml
```diff
@@ -48,6 +48,15 @@ jobs:
             rustc: 1.60.0
             extra_desc: dist-server
             extra_args: --no-default-features --features=dist-tests test_dist_ -- --test-threads 1
+            sccache_log: sccache=trace
+            dist_exec_strategy: docker
+          - os: ubuntu-20.04
```
Xuanwo, line 53: PLease use ubuntu-22.04 for openssl-3.0

drahnr, line 53:
```suggestion
          - os: ubuntu-22.04
```

sylvestre:
dunno much about this one but I guess it is working with cachepot, why not :)

drahnr:
It does work, I'll add some documentation once the safety concern in in the comments is addressed - it should be considered experimental.

PR #1629, opened 2023-02-27
Handle signals in dist mode

drahnr:
Handles signals in dist mode by adding them as a future and `select!`.

Ref #1620

Orignal work by @montekki


sylvestre:
should we manage the HUP signal ?

drahnr:
We could and should, but I think that's a separate concern for another PR and requires some additional logic

sylvestre:
makes sense
I guess you saw that it fails to build

src/bin/sccache-dist/main.rs
```diff
@@ -67,18 +73,31 @@ fn main() {
...
+                2
             }
-            2
         }
-    });
+    };
+        
+    // Whenever `docker` or `systemd` stop the service, the signals
+    // have to be processed and the app has to perform a graceful stop.
+    select! {
```
mati865, line 92: By the default select will randomly pick the order of polling. You can make it use the order you have specified (also avoiding calling random) by adding `biased`.
I'm on my mobile but it should look something like this:
```suggestion
    select! {
        biased;
        
```

drahnr:
I'll get back to this eventually

Issue #1630, opened 2023-02-27
error[E0433]: failed to resolve: use of undeclared crate or module `reqwest`


perfect7gentleman:
```
error[E0433]: failed to resolve: use of undeclared crate or module `reqwest`
   --> src/util.rs:517:41
    |
517 | pub fn new_reqwest_blocking_client() -> reqwest::blocking::Client {
    |                                         ^^^^^^^ use of undeclared crate or module `reqwest`

error[E0433]: failed to resolve: use of undeclared crate or module `reqwest`
   --> src/util.rs:518:5
    |
518 |     reqwest::blocking::Client::builder()
    |     ^^^^^^^ use of undeclared crate or module `reqwest`
    |
help: consider importing one of these items
    |
15  | use crate::dist::Client;
    |
15  | use crate::jobserver::Client;
    |
15  | use jobserver::Client;
    |
help: if you import `Client`, refer to it directly
    |
518 -     reqwest::blocking::Client::builder()
518 +     Client::builder()
    |

warning: ignoring -C extra-filename flag due to -o flag

For more information about this error, try `rustc --explain E0433`.
warning: `sccache` (lib) generated 1 warning
error: could not compile `sccache` due to 2 previous errors; 1 warning emitted
```
git version

Xuanwo:
How did you build this project?

Xuanwo:
> Gentoo's ebuild.

Please provide as many details as you can, for examples:

- The detailed git sha you are using.
- The detailed `cargo` command that you are using (are you build without default feature?).
- If the gentoo's ebuild is publicly tracked by git, please also attach the url.

perfect7gentleman:
```
 * Checking out https://github.com/mozilla/sccache.git to /tmp/portage/dev-util/sccache-9999/work/sccache-9999 ...
git checkout --quiet refs/git-r3/HEAD
GIT update -->
   repository:               https://github.com/mozilla/sccache.git
   at the commit:            408ae947d7c36f0510695bc991438ce2e180d099
 * cargo fetch 
 * cargo vendor  /tmp/portage/dev-util/sccache-9999/work/cargo_home/gentoo
```
```
cargo build --release --features native-zlib --no-default-features
```
ebuild - https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/sccache/sccache-9999.ebuild

Xuanwo:
Thanks for report! I will fix it later.

Xuanwo:
[...]

PR #1632, opened 2023-03-01
fix: Allow building with no default feature

Xuanwo:
Close https://github.com/mozilla/sccache/issues/1630


sylvestre:
most of the jobs are failing

PR #1647, opened 2023-03-09
Support emit pch header

Danielmelody:


sylvestre:
would you have a bit more context? where do you see this flag? thanks

Danielmelody:
Hmm, we are using clang/clang++ on windows(not clang-cl), and using cmake to generate pch files. The compile command cmake output is just like in the test case I wrote. Except the file name changed by real file name on my disk.

sylvestre:
OK, could you please add these info in the commit message ?
thanks

Danielmelody:
@sylvestre Just added a new line within commit message

scorpion-26:
This PR/commit is a bit borked. It breaks PCH compilation when using the officially documented way of _just_ specifiying ```-c -x c++-header```(https://releases.llvm.org/3.1/tools/clang/docs/UsersManual.html#precompiledheaders). You can see a proper explanation in the addendum of my linked issue.

Issue #1609, opened 2023-02-17
proposal: Sccache MSRV Policy


Xuanwo:
Hi all, I am submitting this proposal to create an MSRV policy for sccache.

## TL;DR

- Sccache will use **`stable - 4`** as the MSRV.
- Sccache will change the MSRV every **three** months.

## What's MSRV?

`MSRV` means `Minimum Supported Rust Version`. By setting up MSRV, We makes sure that users who use this or newer version of rust can build `sccache`.

Notice: Sccache is a binary project. The MSRV change only affects users who build sccache themselves.

## Why do we need this?

- to make our codebase much cleaner (by using more of the tools provided by the new Rust version) 
- to bump our dependencies more regularly (we can increase the dependency checking interval to 1 time/month to reduce our maintenance work) 
- to avoid the expected irregularity of bumping (instead of bumping at a random time, all our downstreams will ensure that we will bump MSRV today or three months later).

## Why `stable - 4`?

The Rust team has been releasing a new version every 6 weeks. stable - 4 means that we are using the version that was released 6 months ago. After 6 months most of our developers or packagers should already support it. For example, Debian Rust has 1.63 in testing and 1.64 in exp.

## Why extend three months?

Three months means that our maintainer only needs to update MSRV 4 times a year. It is a good balance between newer version and the work the maintainer might have.

drahnr:
I think we should provide a _minimum_ lag and bump it on demand otherwise. I'd propose 6 months-12months to reduce friction for the compiling people in corporates, where things traditionally move at a much slower pace.

Xuanwo:
Rust 1.68 will be released on `2023-03-09`, I plan to bump sccache's MSRV to Rust `1.64` on that day, which will be released on `2022-09-22` (6 months from `2023-03-09`).

After this bump, existing blocks will be resolved, for example https://github.com/mozilla/sccache/pull/1578.

momvart:
I just want to report that it is out-dated as of today. `clap` requires MSRV `1.74.0` while `sccache` is still marked with `1.70.0` causing builds to fail.

Xuanwo:
> I just want to report that it is out-dated as of today. `clap` requires MSRV `1.74.0` while `sccache` is still marked with `1.70.0` causing builds to fail.

We have MSRV checks in CI so this should not happen. Have you used `--locked` while building sccache? It uses the `Cargo.lock` file provided by sccache.

momvart:
[...]

PR #1649, opened 2023-03-09
Bump MSRV to 1.64 

Xuanwo:
Read more detail at https://github.com/mozilla/sccache/issues/1609

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,6 @@
 name: ci
 on: [push, pull_request]
+
 jobs:
   lint:
     name: ${{ matrix.component }} ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


Xuanwo:
freebsd tests seems always failed since https://github.com/mozilla/sccache/commit/e53407202d1e8a4d5d61d8c6f95f793826582e57

sylvestre:
sorry, it needs to be rebased

Xuanwo:
> sorry, it needs to be rebased

DONE

sylvestre:
I am wondering if we should not wait after we released 0.4.0 before merging this

Xuanwo:
> I am wondering if we should not wait after we released 0.4.0 before merging this

It's more common to bump MSRV when bumping minor releases. But both LGTM.

PR #1650, opened 2023-03-10
Fixed ToC and added back Storage Options section

exoosh:
- Added sections not shown in the ToC.
- Fixed ToC to reproduce the same order as the actual sections.
- The README still referenced the no longer existing section "Storage Options". I added back that option with the same links as in the ToC.

------

The "Storage Options" section had been removed, but not all the references to it (first paragraph), the same holds for the subsection named "Local", which -- however -- has a separate document as replacement now. Since there was no replacement that could be linked, I reintroduced the section "Storage Options" and reproduced the links to the external files there. Only remaining question from my end, do the references to the external documents really have to be in the ToC?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 sccache - Shared Compilation Cache
 ==================================
 
-sccache is a [ccache](https://ccache.dev/)-like compiler caching tool. It is used as a compiler wrapper and avoids compilation when possible, storing cached results either on [local disk](#local) or in one of [several cloud storage backends](#storage-options).
+sccache is a [ccache](https://ccache.dev/)-like compiler caching tool. It is used as a compiler wrapper and avoids compilation when possible, storing cached results either on [local disk](docs/Local.md) or in one of [several cloud storage backends](#storage-options).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,7 @@
 sccache - Shared Compilation Cache
 ==================================
 
-sccache is a [ccache](https://ccache.dev/)-like compiler caching tool. It is used as a compiler wrapper and avoids compilation when possible, storing cached results either on [local disk](#local) or in one of [several cloud storage backends](#storage-options).
+sccache is a [ccache](https://ccache.dev/)-like compiler caching tool. It is used as a compiler wrapper and avoids compilation when possible, storing cached results either on [local disk](docs/Local.md) or in one of [several cloud storage backends](#storage-options).
```
drahnr, line 13: What's the reason for chaning this? SHouldn't `#local` work just as well?

drahnr (approved):

README.md
exoosh: The section "Local" had already been removed (prior to my PR). You can test that the link is dead by visiting either the local https://github.com/mozilla/sccache#local or in my repo https://github.com/exoosh/sccache#local ...

Xuanwo (approved):
Great! Thanks.

PR #1661, opened 2023-03-15
add codecov.yml and set threshold to 0.2%, prevent coverage getting w…

Danielmelody:
…orse


Danielmelody:
According to https://github.com/codecov/codecov-action/issues/554#issuecomment-1261250304

sylvestre:
why not!
what is it going to do?

Danielmelody:
@sylvestre it will result ci failure if any PR decrease coverage more than 0.2%

sylvestre:
where is it going to be displayed in case of failure ?
afaik, we don't have a pure codecov task

Issue #1603, opened 2023-02-13
Issue handling -Xclang custom plugin and flags


YaronStavitzky:
We are using sccache with S3 for the past few weeks, testing if we can get some benefits.
Everything was fine until our most recent release, where we added a clang plugin.

This is what we are passing to clang:
```
['-fplugin=compilation/build/orion-plugin/orion-clang-plugin.so', '-Xclang', '-plugin-arg-orion', '-Xclang', '-fadd-orion-trace-types', '-Xclang', '-plugin-arg-orion', '-Xclang', '-Werror', '-Xclang', '-plugin-arg-orion', '-Xclang', '-Wno-trace-arg-cast-ignored', '-Xclang', '-plugin-arg-orion', '-Xclang', '-Wno-explicit-constructor', '-Xclang', '-plugin-arg-orion', '-Xclang', '-Wno-mismatch-argument-size-trace', '-Xclang', '-plugin-arg-orion', '-Xclang', '-fthirdparty-regex=^/\\|src/plasma/third_party']
```
Example output from sccache log:
```
[2023-02-12T14:41:46Z DEBUG sccache::server] handle_client: compile
[2023-02-12T14:41:46Z DEBUG sccache::server] check_compiler: Supported compiler
[2023-02-12T14:41:46Z DEBUG sccache::server] parse_arguments: CannotCache(Can't handle UnknownFlag arguments with -Xclang, -fadd-orion-trace-types): 
["-o", "src/estore/snaps/giving_tree/metrics/group-0.cu-grouping.o", "-c", "-fplugin=compilation/build/orion-plugin/orion-clang-plugin.so", "-Xclang", "-plugin-arg-orion", 
"-Xclang", "-fadd-orion-trace-types", "-Xclang", "-plugin-arg-orion", "-Xclang", "-Werror", "-Xclang", "-plugin-arg-orion", "-Xclang", "-Wno-trace-arg-cast-ignored", 
"-Xclang", "-plugin-arg-orion", "-Xclang", "-Wno-explicit-constructor", "-Xclang", "-plugin-arg-orion", "-Xclang", "-Wno-mismatch-argument-size-trace", 
"-Xclang", "-plugin-arg-orion", "-Xclang", "-fthirdparty-regex=^/|src/plasma/third_party", "-std=c++11", "-Wno-invalid-offsetof", "-DORION_WITH_ARROW", 
[...]

PR #1662, opened 2023-03-15
Allow flags as plugin args with -Xclang (#1603)

Roynecro97:
Allows passing flags after `-plugin-arg-PLUGIN` (with `-Xclang`).

Ref #1603

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -687,6 +687,37 @@ mod test {
     }
 
     #[test]
+    fn test_parse_xclang_plugin_arg_unknown_flag() {
+        let a = parses!(
+            "-c",
+            "foo.c",
+            "-o",
+            "foo.o",
+            "-Xclang",
+            "-add-plugin",
+            "-Xclang",
+            "some-plugin",
+            "-Xclang",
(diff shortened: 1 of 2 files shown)


sylvestre:
we will need tests that verify that Xclang and the flags are part of the hash calculation.
See: https://github.com/mozilla/sccache/issues/1603#issuecomment-1428424800

src/compiler/clang.rs
```diff
@@ -573,6 +573,37 @@ mod test {
         );
     }
 
+    #[test]
+    fn test_parse_xclang_plugin_arg_unknown_flag() {
```
sylvestre, line 577: we have now actual integration tests with clang.
Could you please replicate this test with a plugin:
https://github.com/mozilla/sccache/blob/main/.github/workflows/integration-tests.yml#L403-L452
thanks

Roynecro97: There are no plugins available by default that I know of, using a plugin would require writing one and building it (or at least getting the binary for it).
Is it not enough that the plugin flags are in the `common_args` and that there are tests that verify these are part of the hash calculation?

sylvestre:
please reopen with an integration tests

PR #1663, opened 2023-03-17
strip env vars

drahnr:
Resolves a FIXME


sylvestre:
Looks good. Do you think it could be tested to make sure we don't regress in the future ?

drahnr:
I can't reproduce the outlined issue with a fresh build, so I'd say no for the time being.

PR #1680, opened 2023-03-22
Prevent sccache-dist from running as root

sylvestre:


src/bin/sccache-dist/main.rs
```diff
@@ -54,12 +60,27 @@ use cmdline::{AuthSubcommand, Command};
...
 
+#[cfg(any(
+    all(target_os = "linux", target_arch = "x86_64"),
+    target_os = "freebsd"
+))]
+fn exit_if_root() {
+    let euid: uid_t = unsafe { geteuid() };
+
+    if euid == 0 {
+        println!("This program should not be executed as root.");
```
drahnr, line 71:
```suggestion
        eprintln!("This program should not be executed as root.");
```

src/bin/sccache-dist/main.rs
```diff
@@ -44,6 +44,12 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Mutex, MutexGuard};
 use std::time::{Duration, Instant};
 
+#[cfg(any(
+    all(target_os = "linux", target_arch = "x86_64"),
+    target_os = "freebsd"
+))]
+use libc::{geteuid, uid_t};
```
drahnr, line 51: nit: I'd import it within the function, or use it's full path since it's the only place it's actually used.

sylvestre:
Yeah, I won't have time to work on it :)

PR #1696, opened 2023-03-27
Remove Redis TTL

eandersson:
Prior to sccache 0.4.x there was no per key TTL applied to Redis. Long term we probably want to make this configurable, but for now restoring the initial behavior of letting Redis control when the keys expire.


Xuanwo (approved):

Xuanwo:
Thanks a lot!

PR #1699, opened 2023-03-28
Replace atty with is-terminal

tottoto:
Replaces atty with is-terminal. As atty is unmaintained, the ecosystems have been using is-terminal instead of atty.


drahnr (approved):

Xuanwo (approved):

Xuanwo:
Great PR! Thanks a lot!

PR #1703, opened 2023-03-29
Use scoped thread provided by standard library

tottoto:
We might be able to use the scoped thread provided by standard library introduced in rust 1.63 instead of crossbeam.

### DIFF
--- a/src/bin/sccache-dist/build.rs
+++ b/src/bin/sccache-dist/build.rs
@@ -413,8 +413,8 @@ impl OverlayBuilder {
                     // Bizarrely there's no way to actually get any information from a thread::Result::Err
                 })
                 .join()
+                .unwrap_or_else(|_e| Err(anyhow!("Build thread exited unsuccessfully")))
         })
-        .unwrap_or_else(|e| Err(anyhow!("Error joining build thread: {:?}", e)))
     }
 
     // Failing during cleanup is pretty unexpected, but we can still return the successful compile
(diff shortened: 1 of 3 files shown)


Xuanwo (approved):
Great!

drahnr (approved):

src/bin/sccache-dist/build.rs
```diff
@@ -410,11 +410,9 @@ impl OverlayBuilder {
                         output: compile_output,
                         outputs,
                     })
-
                     // Bizarrely there's no way to actually get any information from a thread::Result::Err
                 })
                 .join()
-                .unwrap_or_else(|_e| Err(anyhow!("Build thread exited unsuccessfully")))
```
drahnr, line 417: Is there a particular reason to remove the error mapping? `join` in of `scoped` also returns a `Result`

tottoto: The std version's scoped thread does not return Result, so we should unwrap only once. However I noticed that it may be better to move error mapping here from outside of the scoped thread. I will fix it.

tottoto: Fixed.

drahnr (approved):

src/bin/sccache-dist/build.rs
drahnr: Thank you!

PR #1706, opened 2023-04-03
Describe the default conf file locations in docs/Configuration.md

starrify:
This is a minor doc change.

The list of default paths are taken from `docs/DistributedQuickstart.md`.


sylvestre:
fyi, it was documented here:
https://github.com/mozilla/sccache/blob/main/docs/DistributedQuickstart.md
but it is better in this place indeed

PR #1724, opened 2023-04-12
Make `SCCACHE_S3_NO_CREDENTIALS` require a value of `true`

### LINKED ISSUE #1723: Proposal: Make `SCCACHE_S3_NO_CREDENTIALS` require a value of `true`
It seems that `SCCACHE_S3_NO_CREDENTIALS` takes effect even when its value is set to non-truthy values like `SCCACHE_S3_NO_CREDENTIALS=false`.

From reviewing the source code, it looks like this is intended.

Can we switch this environment variable to require an explicit value of `true` to be enabled (e.g. `SCCACHE_S3_NO_CREDENTIALS=true`)?

The fact that `sccache` turns this feature on when this environment value is falsy (or even empty) is currently causing problems for one of the build tools that we use.

I'm happy to work on this assuming this proposal is approved.

ajschmidt8:
Closes #1723.

This PR updates the `SCCACHE_S3_NO_CREDENTIALS` environment variable to require a value of `true` in order to enable this feature.

**Note:** This is a breaking change and should probably be called out loudly in any future release notes.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1136,7 +1136,7 @@ fn test_s3_no_credentials_valid_true() {
             ..
         }) => {
             assert_eq!(bucket, "my-bucket");
-            assert_eq!(no_credentials, true);
+            assert!(no_credentials);
         }
         None => unreachable!(),
     };
@@ -1159,7 +1159,7 @@ fn test_s3_no_credentials_valid_false() {
             ..
         }) => {
             assert_eq!(bucket, "my-bucket");
-            assert_eq!(no_credentials, false);
+            assert!(!no_credentials);
         }
         None => unreachable!(),
     };
--- a/src/config.rs
+++ b/src/config.rs
@@ -512,7 +512,10 @@ fn config_from_env() -> Result<EnvConfig> {
     // ======= AWS =======
     let s3 = env::var("SCCACHE_BUCKET").ok().map(|bucket| {
         let region = env::var("SCCACHE_REGION").ok();
-        let no_credentials = env::var("SCCACHE_S3_NO_CREDENTIALS").ok().is_some();
+        let no_credentials = env::var("SCCACHE_S3_NO_CREDENTIALS")
+            .ok()
+            .map(|value| value == "true")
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -506,7 +506,10 @@ fn config_from_env() -> Result<EnvConfig> {
     // ======= AWS =======
     let s3 = env::var("SCCACHE_BUCKET").ok().map(|bucket| {
         let region = env::var("SCCACHE_REGION").ok();
-        let no_credentials = env::var("SCCACHE_S3_NO_CREDENTIALS").ok().is_some();
+        let no_credentials = env::var("SCCACHE_S3_NO_CREDENTIALS")
```
sylvestre, line 509: I think we should be more aggressive here.

if SCCACHE_S3_NO_CREDENTIALS is set, it can be only true or false (or 1 or 0).
Otherwise, we don't start the program.

src/config.rs
```diff
@@ -1066,7 +1069,7 @@ fn config_overrides() {
 #[test]
 #[serial]
 fn test_s3_no_credentials() {
-    env::set_var("SCCACHE_S3_NO_CREDENTIALS", "1");
+    env::set_var("SCCACHE_S3_NO_CREDENTIALS", "true");
```
sylvestre, line 1072: we would need more (and separate) tests to cover this

src/config.rs
ajschmidt8: @sylvestre, I just pushed some changes. I'm not great with `rust`. Can you tell me if my approach is right? It took me a while to figure out the `Option` / `Result` handling.

I'll update the tests in a follow-up commit.

src/config.rs
ajschmidt8: Updated in the latest commits. I mostly just copy and pasted the existing test functions.

Let me know if there's a better convention than that.

ajschmidt8:
@sylvestre, just checking in. any thoughts on this PR so far?

PR #1730, opened 2023-04-13
fix: Ignore C4668 when preprocessing via MSVC

### LINKED ISSUE #1725: Windows SDK Generates C4668 During Pre-processing - Confuses Sccache
## Issue

Sccache cannot compile some Windows applications via MSVC when `/we4668` is specified because first it tries to use the same arguments to preprocess the files, which fails within the Windows SDK.

Specifically, when trying to use sccache to cache Unreal Engine compilation (after fixing some other issues) sccache produces the error `preprocessor failed` from: https://github.com/mozilla/sccache/blob/main/src/compiler/c.rs#L339

In the include list in this error, the relevant error is visible:
```
Note: including file: C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared\intsafe.h
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared\intsafe.h(66): error C4668: '__midl' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
```

## Why is this a problem with sccache?

The reason sccache runs into this problem, but Unreal Engine does not, is that sccache faithfully copies all arguments for compilation into the preprocessor command it executes. However, when Unreal Engine builds, it explicitly disables this error message when preprocessing and only enables it during compiling.

## Expected behaviour

I believe the correct behaviour here is that this compiler warning should be explicitly ignored when building using MSVC.

Alexei-Barnes:
This fixes #1725.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -400,6 +400,13 @@ jobs:
           test -e ./foo.o || { echo "No compiler output found"; exit -1; }
           test -e ./foo.o.json || { echo "No dependency list found"; exit -1; }
 
+      - name: Compile - Preprocessing Compiler Bug
+        shell: bash
+        working-directory: ./tests/msvc-preprocessing
+        run: |
+          $SCCACHE_EXE "$(where cl.exe)" -c "@args.rsp"
+          $SCCACHE_EXE --show-stats
+
       - name: Stop Server
         if: success() || failure()
         shell: bash
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -950,9 +950,8 @@ where
         .args(&parsed_args.dependency_args)
         .args(&parsed_args.common_args)
         // Windows SDK generates C4668 during preprocessing, but compiles fine.
-        // Making sure that 4668 is a level 4 warning prevents failing preprocessing on it.
         // Read for more info: https://github.com/mozilla/sccache/issues/1725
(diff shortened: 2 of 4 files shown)


sylvestre:
is it possible to add a test for this? thanks

Alexei-Barnes:
Absolutely, I'm on vacation right now without access to a computer I could run the tests on. Back home around the 7th of May. Happy to amend this with a test then.

Alexei-Barnes:
@sylvestre I've written a minimal integration test to verify the behaviour.

There aren't currently any unit tests for the msvc preprocessing method. I could add one if required, but it might require other changes to make the preprocessing method more testable.

Alexei-Barnes:
Updated the test just now to actually use sccache 😅
Should now actually test the intended behaviour. I also threw in a direct call to cl.exe just to show the preprocessing error in the output that we're trying to avoid. Maybe we should run this new test on unfixed code first to establish a baseline negative result?

Alexei-Barnes:
Removed the additional call to cl that shows the 4668 warning because it terminates the build, and it isn't necessary.

Alexei-Barnes:
Oh interesting, the test failed! Will investigate tomorrow.

Alexei-Barnes:
OK the test revealed that simply changing the warning to level 4 isn't enough, it must be entirely disabled. I've updated the code, test should pass in CI now.

Alexei-Barnes:
Test passing now 👍

PR #1735, opened 2023-04-18
nvcc: Correctly handle -ccbin argument given a directory

robertmaynard:
The `-ccbin` argument to nvcc supports being passed either a compiler or the directory that holds the compiler. 
Mark these options as `PassThrough` so that they are still considered as part of the hash.


src/compiler/nvcc.rs
```diff
@@ -1,4 +1,5 @@
 // Copyright 2016 Mozilla Foundation
+// SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
sylvestre, line 2: I am a bit confused why you are adding this ?

PR #1737, opened 2023-04-19
Use rustls for tls & trust-dns-resolver for dns resolution

NobodyXu:
By default, reqwest uses native-tls, which is openssl on linux and that makes cross-compilation very hard as they need to compile openssl themselves and exposes it via pkg-config.

Enabling vendored-openssl does not help there because the dep:openssl is only enabled by dist-server.

reqwest by default uses libc's `getaddrinfo`, which is a poor solution that requires a dedicated thread and it is not nearly as reliable as trust-dns-resolver, not to mention libc implementation, e.g. musl, might deliberately leave out some features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ strip-ansi-escapes = "0.1"
 tar = "0.4.36"
 tempfile = "3"
 # trust-dns-resolver must be kept in sync with the version reqwest uses
-trust-dns-resolver = { version = "0.22.0", features = ["dnssec-ring", "dns-over-rustls", "dns-over-https-rustls", "dns-over-quic"] }
+trust-dns-resolver = { version = "0.22.0", features = ["dnssec-ring", "dns-over-rustls", "dns-over-https-rustls"] }
 mime = "0.3"
 tokio = { version = "1", features = ["rt-multi-thread", "io-util", "time", "net", "process", "macros"] }
 tokio-serde = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ number_prefix = "0.4"
 openssl = { version = "0.10.48", optional = true }
 rand = "0.8.4"
 regex = "1.7.3"
-reqwest = { version = "0.11", features = ["json", "blocking", "stream"], optional = true }
+reqwest = { version = "0.11", features = ["json", "blocking", "stream", "rustls-tls", "trust-dns"], optional = true }
 retry = "2"
 semver = "1.0"
 sha2 = { version = "0.10.6", optional = true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -56,7 +56,7 @@ number_prefix = "0.4"
 openssl = { version = "0.10.48", optional = true }
 rand = "0.8.4"
 regex = "1.7.3"
-reqwest = { version = "0.11", features = ["json", "blocking", "stream"], optional = true }
+reqwest = { version = "0.11", features = ["json", "blocking", "stream", "rustls-tls", "trust-dns"], optional = true }
```
Xuanwo, line 59: We can do this by enable `opendal`'s `trust-dns` feature so that we don't need to play with `trust-dns-resolver` directly.

NobodyXu: `opendal`'s `trust-dns` feature simply enables `reqwest/trust-dns`.
There are currently no other ways to do it.

The reason for adding `trust-dns-resolver` as a dep to sccache is to enable extra features:
 - dnssec-ring to enable dnssec using ring
 - dns-over-rustls to enable dns-over-tls using ring
 - dns-over-https-rustls for dns over https
 - dns-over-quic for dns over http3 (quic)

The alternative to `ring` and `rustls` is `openssl`, and given that these extra features might come in handy for `sccache`, I think enabling them by default makes sense.

NobodyXu:
The build failure is caused by quinn-udp: https://github.com/quinn-rs/quinn/issues/1469

I will disable dns-over-quic for now.

NobodyXu:
I've pushed again the fixed the `Cargo.lock`.

NobodyXu:
@Xuanwo It seems that freebsd ci also failed on main with similar errors, so I think this PR should be ok to merge.

drahnr (approved):
Did you verify the dependency tree is free of openssl? `cargo tree -i openssl` should do the trick, empty is good

NobodyXu:
@drahnr Yes, I've verified that with `dist-server` disabled, openssl is not pulled in.

I will open another PR to remove the openssl pulled in by `dist-server` with [`pkcs1::RsaPublicKey`](https://docs.rs/pkcs1/latest/pkcs1/struct.RsaPublicKey.html#impl-EncodeValue-for-RsaPublicKey%3C'_%3E).

NobodyXu:
@drahnr I've created #1738 to be merged after this to fix the openssl pulled in by dist-servers.

It still needs some testing, since I'm using MacOS, I cannot test it locally without installing openssl for x86_64-linux, unless I rebase that PR above this one, which would make it a mess.

So I've pushed it and let the ci tested it out, hope it would be ok, though it's the first time for me to use pkcs1 and do anything related to crypto, so I might made some mistakes, hope you would review it carefully to verify that it has no such bug.

Thank you for taking your time to review my PR!

Ralith:
> The build failure is caused by quinn-udp: https://github.com/quinn-rs/quinn/issues/1469

That bug is for OpenBSD. Quinn CI covers FreeBSD, which seems to be working just fine.

PR #1738, opened 2023-04-19
Stop feature dist-server from pulling in openssl

NobodyXu:
- Replace direct `openssl` usage in sccache-dist/token_check.rs with `pkcs1`
 - Enable `rouille/rustls` instead of `rouille/ssl` to use rustls instead of openssl in tiny_http (dep of rouille).
 - Remove `openssl` that get pulled in debug build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
         if: ${{ matrix.target == 'aarch64-unknown-linux-musl' }}
 
       - name: Build
-        run: cargo build --locked --release --bin ${{ matrix.binary || 'sccache' }} --target ${{ matrix.target }} --features=openssl/vendored ${{ matrix.extra_args }}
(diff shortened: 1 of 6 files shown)


NobodyXu:
Build failure will be fixed once #1737 is merged

NobodyXu:
The CI failure is caused by spki (pulled in pkcs1).

Its msrv is 1.65 while currently sccache has a msrv of 1.64, thus requires a bump of msrv.
Given that they are quite close, I think a msrv bump is fine.

NobodyXu:
Blocked on https://github.com/tomaka/rouille/pull/272

drahnr:
I'd say porting to warp (see #1620 ) should avoid the issue and move to a more modern framework

NobodyXu:
> I'd say porting to warp (see #1620 ) should avoid the issue and move to a more modern framework

Thanks for the advice!
I would look into that, if I figured it out I would submit a PR.

NobodyXu:
I will close this PR for now.

PR #1739, opened 2023-04-20
Rm openssl pulled in linux

NobodyXu:
Follow up for #1737

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

PR #1742, opened 2023-04-21
Rm dep openssl pulled in by feature dist-server

NobodyXu:
This is split from #1738 , to remove dependency on `openssl` in `src/bin/sccache-dist/token_check.rs`.


sylvestre:
I guess you saw : error[E0599]: the method `context` exists for enum `Result<UintRef<'_>, Error>`, but its trait bounds were not satisfied
  --> src/bin/sccache-dist/token_check.rs:41:44
   |
41 |         let e_bn = pkcs1::UintRef::new(&e).context("Failed to create pkcs1 bignum from e")?;

NobodyXu:
@sylvestre Thanks, I have fixed this error.

NobodyXu:
This PR is currently blocked on https://github.com/RustCrypto/formats/pull/1013 to avoid pkcs8 to be pulled in.

NobodyXu:
That PR has merged, but latest der v0.7.4 also requires rust 1.65.0

Since current msrv for sccache is 1.64.0, I think it's ok to bump it to 1.65.0

NobodyXu:
Blocked on https://github.com/tomaka/rouille/pull/272

NobodyXu:
@Xuanwo Can we re-open this?
I didn't know that using the `fix` to refer to a PR would close it.

NobodyXu:
Thank you @Xuanwo !
Once https://github.com/tomaka/rouille/pull/272 is merged, I will mark this PR as ready-to-review.

NobodyXu:
@sylvestre It seems that disabling default-feature `default-tls` of `reqwest` (which pulls in openssl) might be what breaks the "ci / test ubuntu-22.04 rust 1.65.0 dist-server".

It seems that `reqwest` [prefers openssl when it is enabled by default](https://docs.rs/reqwest/latest/src/reqwest/tls.rs.html#413), so when we disable `default-tls`, we forces `reqwest` to use `rustls` and that seems to be what causes the failure.

NobodyXu:
Seems like rustls failed due to bad certificate.
Since this also happened in #1739, I don't think the changes to `create_https_cert_and_privkey` is responsible for this.

NobodyXu:
@Xuanwo @sylvestre Do you have any idea on what could go wrong and how could we fix this?

Or perhaps this is an upstream issue that requires `rustls` to fix this?

src/dist/http.rs
```diff
@@ -306,82 +315,67 @@ mod server {
...
+            expires.time().minute() as u8,
+            expires.time().second() as u8,
+        )
+        .unwrap();
+        let extended_key_usage = ExtendedKeyUsage::new(vec![picky::oids::kp_server_auth()]);
+        let name = addr.to_string();
+        let issuer_name = DirectoryName::new_common_name(name.clone());
+        let subject_name = DirectoryName::new_common_name(name);
+        let octets = match addr.ip() {
+            IpAddr::V4(inner) => inner.octets().to_vec(),
```
Xuanwo, line 354: I'm guessing the usage of `octets()` is wrong.

From [RFC](https://datatracker.ietf.org/doc/html/rfc5280#appendix-A.2):

> The syntax of iPAddress MUST be as described in [Section 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6) with
   the following additions specifically for name constraints.  For IPv4
   addresses, the iPAddress field of GeneralName MUST contain eight (8)
   octets, encoded in the style of [RFC 4632](https://datatracker.ietf.org/doc/html/rfc4632) (CIDR) to represent an
   address range [[RFC4632](https://datatracker.ietf.org/doc/html/rfc4632)].  For IPv6 addresses, the iPAddress field
   MUST contain 32 octets similarly encoded.  For example, a name
   constraint for "class C" subnet 192.0.2.0 is represented as the
   octets C0 00 02 00 FF FF FF 00, representing the CIDR notation
   192.0.2.0/24 (mask 255.255.255.0).


But, `IpAddr`'s `octects` is:

```rust
use std::net::Ipv4Addr;

let addr = Ipv4Addr::new(127, 0, 0, 1);
assert_eq!(addr.octets(), [127, 0, 0, 1]);
```

NobodyXu: Thanks, I will have a look at this.

NobodyXu: @Xuanwo I've fixed this issue by manually inserting the mask, however the issue still persists:

```
[2023-05-10T13:04:33Z ERROR rustls::conn] TLS alert received: AlertMessagePayload {
        level: Fatal,
        description: BadCertificate,
    }
```

Since this also happens in my other PR, which only disables the `reqwest/default-tls` feature, I guess this is a bug of reqwest or rustls.

Xuanwo: > Since this also happens in my other PR, which only disables the `reqwest/default-tls` feature, I guess this is a bug of reqwest or rustls.

Ok, maybe it's ok for us to enable `reqwest/default-tls` in test?

NobodyXu: > Ok, maybe it's ok for us to enable `reqwest/default-tls` in test?

I can do that, but this seems like to be a bug of `rustls`.
Perhaps this could even happen in real-world, though I'm not entirely sure.

@Xuanwo Does the CI for `sccache-dist` uses self-signed certificate generated by `sccache` or use a pre-generated certificate?

Xuanwo: > Does the CI for `sccache-dist` uses self-signed certificate generated by `sccache` or use a pre-generated certificate?

I assume that `sccache-dist` will create a new self-signed certificate when the scheduler is launched.

sylvestre:
Sorry, could you please rebase this ?

NobodyXu:
> Sorry, could you please rebase this ?

Done!

saiintbrisson:
Hi! Sorry for the intrusion, but I'm wondering if it wouldn't be worth it to wait until `reqwest` switches to `rustls 0.21.0`, which finally adds [support for peers named with IP addresses](https://github.com/rustls/rustls/tree/main#release-history)? 

The Cargo.lock [is still on 0.20.7](https://github.com/NobodyXu/sccache/blob/dist-server-rewrite/Cargo.lock#L2290-L2300), which is probably making the CI tests fail?

For reference: https://github.com/rustls/rustls/issues/184

src/dist/http.rs
saiintbrisson: Hi! Sorry for the intrusion, but I'm wondering if this isn't related to rustls/rustls#184? `reqwest` is still on `rustls 0.20.0`, and [support for peers named with IP addresses](https://github.com/rustls/rustls/tree/main#release-history) was added in `rustls 0.21.0`

The Cargo.lock's [rustls is still on 0.20.7](https://github.com/NobodyXu/sccache/blob/dist-server-rewrite/Cargo.lock#L2290-L2300), which is probably making the CI tests fail?

NobodyXu: @saiintbrisson Thanks, I think you are right!

I just discovered the same thing when looking at the reqwest PR https://github.com/seanmonstar/reqwest/pull/1791#issuecomment-1485813835 for updating rustls to v0.21.0

Let's wait for a new release in reqwest, which probably won't take too long.

NobodyXu:
Thank you @sylvestre

sylvestre:
` ci / test ubuntu-22.04 rust 1.65.0 dist-server (pull_request) Failing after 6m ` still fails

NobodyXu:
> `ci / test ubuntu-22.04 rust 1.65.0 dist-server (pull_request) Failing after 6m` still fails

@sylvestre It's blocked on a new release from reqwest https://github.com/seanmonstar/reqwest/pull/1791#issuecomment-1485813835 , which updates rustls to 0.21

Support for IP address peer is added in rustls 0.21 https://github.com/rustls/rustls/issues/184

src/dist/http.rs
cpu: Fresh from the Crates.io factory: https://github.com/seanmonstar/reqwest/releases/tag/v0.11.18

NobodyXu: I've updated reqwest to v0.11.8 and rustls to v0.21.0, it indeed fixed the bad certificate error, but now we got a new error:

```
[2023-05-17T02:11:22Z ERROR rustls::conn] TLS alert received: AlertMessagePayload {
        level: Fatal,
        description: UnknownCA,
    }
```

@saiintbrisson @cpu Any idea how to fix this error?

drahnr:
https://github.com/seanmonstar/reqwest/pull/1834 0.11.18 was released, so there shouldn't be any blockers left

src/dist/http.rs
```diff
@@ -306,82 +315,95 @@ mod server {
...
-            .sign(&privkey, openssl::hash::MessageDigest::sha1())
-            .context("failed to sign x509 with sha1")?;
-        let cert: openssl::x509::X509 = builder.build();
-        let cert_pem = cert.to_pem().context("failed to create pem from x509")?;
-        let cert_digest = cert
-            .digest(openssl::hash::MessageDigest::sha256())
-            .context("failed to create digest of x509 certificate")?
-            .as_ref()
-            .to_owned();
+        let mut rng = OsRng;
```
drahnr, line 318: Could you add a feature gated test, that verifies the certificates generated are identical? It appears the issue seen after porting can originate from a variety of name validation issues i.e. https://github.com/briansmith/webpki/issues/20 so I'd recommend to check consistency against the openssl generated cert first, verify such a openssl generated cert with the logic of rustls and then start debugging into the unit test that should start failing.

drahnr: If it's not your cup of tea, leave a message here :)

NobodyXu: I can certainly add a new unit test to ensure the two different methods produces identical certificates, but I'm a bit busy recently and might it takes me some time to add this.

sylvestre:
Thanks & no worries, no rush :)

src/dist/http.rs
cpu: Could you pull out a smaller reproducer and include the problematic peer certificate? That error is fairly generic and could have many root causes.

NobodyXu: > include the problematic peer certificate

The peer certificate is self-signed and generated at runtime in [here](https://github.com/mozilla/sccache/pull/1742/files#diff-6d63828d1891ff3303de7191b656bc451d27703548c42bd9786fcd1c157aa459R317).

saiintbrisson: Hi! I've spent some time looking into it and turns out there are still two rustls versions. I believe the side receiving the certificate is still on rustls 0.20.7

```
├── rouille v3.6.2
...
│   ├── tiny_http v0.12.0
│   │   ├── rustls v0.20.7 (*)
│   │   ├── rustls-pemfile v0.2.1
│   │   │   └── base64 v0.13.1
```

The `TLS alert received` message is only present on [0.20.7](https://github.com/rustls/rustls/blob/v/0.20.7/rustls/src/conn.rs#L1313).
0.21.0 does not have [this error message anymore](https://github.com/rustls/rustls/blob/4c973b6b10461808e9207aba64ff6c5c00124547/rustls/src/common_state.rs#L464-L465), it is gone from the codebase, the remaining one is `TLS alert warning received` (notice the "warning").

In fact, even the module was renamed from `conn` to `common_state`, but the logs still show `rustls::conn`.

src/dist/http.rs
saiintbrisson: Some input here https://github.com/mozilla/sccache/pull/1742#discussion_r1197225678

src/dist/http.rs
saiintbrisson: I'm not sure if the different versions are the reason behind the error, it may be completely unrelated problem

NobodyXu: @saiintbrisson Could it be caused by https://github.com/seanmonstar/reqwest/pull/1835 ?

NobodyXu: I've also submit a PR to upgrade rustls used in tiny-http https://github.com/tiny-http/tiny-http/pull/242

saiintbrisson: > Could it be caused by [seanmonstar/reqwest#1835](https://github.com/seanmonstar/reqwest/pull/1835) ?

You can run the tests locally and use the reqwest git repository directly on cargo.toml
```
$ cargo test --no-default-features --features=dist-tests test_dist_ -- --test-threads 1
```
You must have docker installed to run these (or podman, requires changing all `docker` command spawns on `tests/harness/mod.rs`)

NobodyXu: Thanks, I would have a try when I'm free

saiintbrisson: I've run the tests here using the updated reqwest branch and the error persists

NobodyXu: Ehhh, @saiintbrisson can you also try https://github.com/tiny-http/tiny-http/pull/242 ?

saiintbrisson: This one is a bit tricky as we don't depend on the crate directly

NobodyXu: > This one is a bit tricky as we don't depend on the crate directly

Yeah, it's a dep brings in by rouille.

Could use patching to replace the dep though

drahnr: Imho we should get rid of roullie as in https://github.com/paritytech/cachepot/pull/131

NobodyXu: @drahnr I agree, but that PR is quite large and porting it all at once takes a lot of effort.

saiintbrisson: Also good to notice that `trust-dns-resolver` is still on the older version of rustls. I'm guessing this PR might need to wait a bit longer until the ecosystem catches up?

drahnr:
Side quest @sylvestre - we should consider using `cargo-deny` to prevent future regressions on accidentally re-adding openssl through dependencies.

sylvestre:
Yes, sounds good :)

NobodyXu:
I'm currently trying to upgrade trust-dns to hickory-dns v0.24 by using [`ClientBuilder::dns_resolver`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.dns_resolver), however it seems that this method isn't present in the blocking version of client...

NobodyXu:
> Maybe it's ok to just use libc's dns resovler in blocking API?

That's a good advice, I've applied it and hope that the CI will succeed.

src/dist/http.rs
NobodyXu: Updated trust-dns-resolver v0.22 to latest v0.24 (renamed to hickory-dns) and also updates webpki-roots to v0.25.

Confirmed that rustls v0.21 has been removed, but CI failure persists.

I think it's likely something in the certificate generation that caused the error.

NobodyXu:
Updated trust-dns-resolver v0.22 to latest v0.24 (renamed to hickory-dns) and also updates webpki-roots to v0.25.

Confirmed that rustls v0.21 has been removed, but CI failure persists.

I think it's likely something in the certificate generation that caused the error.

src/dist/http.rs
drahnr: Do we have an integration test to verify 1:1 reproduction of the certificate? We'd need that for the dist case to remain backward compatible with older "clients".

NobodyXu: No, I'm going to write this by running external `openssl` command.

drahnr: You only need to include a previously generated certificate, and include it as a reference.

NobodyXu: @drahnr I don't think it's possible, since the original `openssl` and the new code both uses the current time when creating the certificate.

Otherwise I could just use a statically embedded certificate.

drahnr: a. generate with rustls directly
b. openssl -> file fmt -> parse certificate with rustls

Compare all fields of `a` and `b`, except for those that are reasonable to expect to deviate.

sylvestre:
Sorry, needs rebasing

NobodyXu:
Sorry I was a bit busy recently, will rebase when I got time.

And the CI error is really tricky, I wasn't able to get it fixed.

src/util.rs
```diff
@@ -926,6 +926,37 @@ pub fn daemonize() -> Result<()> {
     Ok(())
 }
 
+#[cfg(any(feature = "dist-server", feature = "dist-client"))]
+mod reqwest_dns_resolver {
+    use std::{net::SocketAddr, sync::Arc};
```
AJIOB, line 931: What about moving it to the dedicated `.rs` file?

src/util.rs
```diff
@@ -926,6 +926,37 @@ pub fn daemonize() -> Result<()> {
     Ok(())
 }
 
+#[cfg(any(feature = "dist-server", feature = "dist-client"))]
+mod reqwest_dns_resolver {
```
AJIOB, line 930: What about moving it to the dedicated `.rs` file?

Xuanwo: Try [reqwest-hickory-resolver](https://crates.io/crates/reqwest-hickory-resolver) that maintained by me :smile:

AJIOB:
Hi @NobodyXu,

Do you have a time for finalizing this PR?

NobodyXu:
Hello

Honestly I don't know how to fix the CI failure.

I don't know openssl good enough and don't have much time recently.

sylvestre:
please reopen when ready

NobodyXu:
Unfortunately I don't have any time now to continue working

PR #1747, opened 2023-04-25
nvcc: Handle compiler detection when gcc isn't on the path

robertmaynard:
To properly handle compiler detction sccache invokes the compiler with -E. To generate the preprocessed output `nvcc` invokes the host comiler, which either requires gcc to be on the path or `-ccbin` argument to be provided.

This PR allows `sccache` to be used with `nvcc` when `gcc` isn't on the path but the host compiler to be used is included via `-ccbin`.

This was found when building small build images which only had a compiler with a name like `gcc-aarch64`.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1044,11 +1044,11 @@ use self::ArgData::PassThrough as Detect_PassThrough;
 // Establish a set of compiler flags that are required for
 // valid execution of the compiler even in preprocessor mode.
 // If the requested compiler invocatiomn has any of these arguments
-// propagate them when doing our compiler vendor detection 
+// propagate them when doing our compiler vendor detection
 //
 // Current known required flags:
 // ccbin/compiler-bindir needed for nvcc
-//  This flag specifies the host compiler to use otherwise 
+//  This flag specifies the host compiler to use otherwise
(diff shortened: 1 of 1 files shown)


robertmaynard:
> you have a bunch of failure in the CI :)

I think I fixed all the issues and learned some more rust along the way :)

src/compiler/compiler.rs
```diff
@@ -1029,14 +1031,25 @@ where
...
         cc.map(|c| (c, None))
     }
 }
 
+ArgData! {
+    PassThrough(OsString),
+}
+use self::ArgData::PassThrough as Detect_PassThrough;
+
+counted_array!(static ARGS: [ArgInfo<ArgData>; _] = [
```
sylvestre, line 1044: could you please document this a bit? :)

sylvestre:
Or explain what this code is doing :)

Le ven. 28 avr. 2023, 17:10, Robert Maynard ***@***.***> a
écrit :

robertmaynard (commented):
Update comments on what/why of the code changes

robertmaynard:
> Or explain what this code is doing :)

I have updated the general code blocks with explanations on what we are doing, and why.

sylvestre:
please run rustfmt :)

PR #1757, opened 2023-05-07
Refactor dependencies to use futures

tottoto:
Replaces `futures` dependency to `futures-util` and `futures-channel` to reduce trailing dependencies.


Xuanwo:
It seems that not much will change because other dependencies already include `futures`.

tottoto:
Closes this pull request. As a result of investigation at https://github.com/apache/incubator-opendal/pull/2248, we found that `sccache` depends on `futures` via `opendal`.

drahnr:
I recommend looking into `cargo tree -i something` before spending effort on eliminating dependencies. I am also not sure if it's generally time well spent (your time in particular!), unless the dependency is unmaintained. We have lots of issues, ranging from easy to very difficult, and I'd love to review code in that regard :)

PR #1702, opened 2023-03-29
Update rust-version in manifest to 1.64

tottoto:
Updates rust-version in the manifest to follow the msrv.

PR #1758, opened 2023-05-07
Replace lazy_static with once_cell

tottoto:
Replaces `lazy_static` with `once_cell` as `once_cell` is without macros and it will go to standard library.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -15,7 +15,7 @@
 use directories::ProjectDirs;
 use fs::File;
 use fs_err as fs;
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use regex::Regex;
 #[cfg(any(feature = "dist-client", feature = "dist-server"))]
 use serde::ser::Serializer;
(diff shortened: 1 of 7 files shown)


Xuanwo (approved):

Xuanwo:
Thanks!

src/config.rs
```diff
@@ -35,10 +35,8 @@ use std::sync::Mutex;
 
 use crate::errors::*;
 
-lazy_static! {
-    static ref CACHED_CONFIG_PATH: PathBuf = CachedConfig::file_config_path();
-    static ref CACHED_CONFIG: Mutex<Option<CachedFileConfig>> = Mutex::new(None);
-}
+static CACHED_CONFIG_PATH: Lazy<PathBuf> = Lazy::new(CachedConfig::file_config_path);
+static CACHED_CONFIG: Lazy<Mutex<Option<CachedFileConfig>>> = Lazy::new(|| Mutex::new(None));
```
liskin, line 39: `Mutex::new` is const since Rust 1.63 and https://github.com/mozilla/sccache/pull/1702 bumped sccache's MSRV to 1.64 three months ago so this can be simplified further I think 🙂

PR #1759, opened 2023-05-08
Remove parity-tokio-ipc

tottoto:
Although `parity-tokio-ipc`'s main benefit is the abstraction of Windows and Unix, we use it only for Windows specific implementations. This crate depends on some old dependencies and it seems unmaintained. We could replace this crate with simple implementations to reduce old trailing dependencies. Related to #1757.


Xuanwo (approved):
Thanks!

PR #1760, opened 2023-05-09
chore: bump `opendal` to `0.34.0`

saiintbrisson:
OpenDAL 0.34.0 comes with a nicer way to build GCS caches and
a bump to `reqsign 0.10.1`, which means Google's External Accounts
are now supported, making it easier to use Sccache on build pipelines
counting on Workload Identity Federation for their auth.

It's now necessary to define the `SCCACHE_REGION` when using AWS S3
with the default endpoint.

### DIFF
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -29,9 +29,7 @@ impl S3Cache {
     ) -> Result<Operator> {
         let mut builder = S3::default();
         builder.bucket(bucket);
-        if let Some(region) = region {
-            builder.region(region);
-        }
+        builder.region(region.unwrap_or("us-east-1"));
         builder.root(key_prefix);
         if no_credentials {
             builder.disable_config_load();
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -29,11 +29,16 @@ impl S3Cache {
     ) -> Result<Operator> {
         let mut builder = S3::default();
         builder.bucket(bucket);
-        builder.region(region.unwrap_or("us-east-1"));
         builder.root(key_prefix);
+
+        if let Some(region) = region {
+            builder.region(region);
+        }
+
         if no_credentials {
             builder.disable_config_load();
         }
+
         if let Some(endpoint) = endpoint {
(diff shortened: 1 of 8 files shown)


sylvestre:
it would be nice to update the doc
(and tests if possible)

saiintbrisson:
@sylvestre would it be a problem to link to an (official) external resource for the external account as an example? https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions

sylvestre:
Both would be nice :)

src/cache/s3.rs
```diff
@@ -29,9 +29,7 @@ impl S3Cache {
     ) -> Result<Operator> {
         let mut builder = S3::default();
         builder.bucket(bucket);
-        if let Some(region) = region {
-            builder.region(region);
-        }
+        builder.region(region.unwrap_or("us-east-1"));
```
saiintbrisson, line 32: Until now, OpenDAL would take `us-east-1` as default in case of a missing region, but this is now an error.
This change makes it consistent with the current behavior, as I suspect a lot of users rely on the current
way of doing things.
```suggestion
        builder.region(region.unwrap_or("us-east-1"));
```

saiintbrisson:
@sylvestre PTAL. I've made some changes to the GCS docs ordering to make it easier to find info on credentials

src/cache/s3.rs
Xuanwo: > Where should I log this breaking change?

I don't think it's a breaking change, it's a bug fix.

Xuanwo (approved):
Nice docs!

Xuanwo:
cc @sylvestre to take an another round of review.

sylvestre:
thanks!

messense:
Hi @sylvestre can you release a new version with this patch? sccache 0.4.2 depends on older opendal that uses `ureq` with mannual proxy configuration instead of `reqwest`, which doesn't support `NO_PROXY` env var.

sylvestre:
sure, building

PR #1761, opened 2023-05-10
Use Infallible instead of void

tottoto:
We might be able to use standard `Infallible` instead of `void` to describe the type that never happens.


Xuanwo (approved):
Great, thanks!

PR #1767, opened 2023-05-23
Exit when increment_compilation is enabled.

### LINKED ISSUE #1596: Trigger an error if rust incremental enabled
Sccache cannot cache incremental builds, it is better to throw an error if it is not disabled. We can do this before sccache is started.

Ref: https://github.com/risingwavelabs/risingwave/pull/7838

infdahai:
fixes #1596

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -47,12 +47,11 @@ fn main() {
     incr_env_strs
         .iter()
         .for_each(|incr_str| match env::var(incr_str) {
-            Ok(incr_val) if incr_val == "true" => {
+            Ok(incr_val) if incr_val == "1" => {
                 println!("sccache: increment compilation is  prohibited.");
                 std::process::exit(1);
             }
-            Ok(_) => (),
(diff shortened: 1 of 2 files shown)


sylvestre:
nice, thanks!

drahnr:
Do we want to remove the env and print a warning rather than terminating?

devyn:
The message as it is was rather confusing for me as I didn't even know I had CARGO_INCREMENTAL set, and I wasn't sure what it was referring to.

I think it would be better if it specifically mentioned that environment variable.

tetigi:
A few complaints regarding this MR:

- builds that previously worked (albeit without `sccache` doing anything due to being incremental) now just totally fail. This is a fundamental behaviour change that feels like it should be behind more than just a minor version bump. 
- the error message is not clear about why it _used_ to work, but no longer does
- the error message is poorly formatted

At the very least, this feels like it should have been a `minor` version bump rather than just a patch from `0.5.0` to `0.5.1` - the behaviour is _fundamentally_ different to how it was before, with no option to opt in or out.

PR #1772, opened 2023-05-24
Produce JSON depsfiles from MSVC

### LINKED ISSUE #1771: Incorrect format of MSVC deps file produced during preprocessing
With MSVC, deps files are encoded as JSON instead of the format clang/gcc use of pairs distributed over lines.
This can be seen in the documentation: https://learn.microsoft.com/en-us/cpp/build/reference/sourcedependencies?view=msvc-170

Right now, when compiling on Windows, the lines of pairs format is used regardless of the compiler used. Instead, when using MSVC we should use its depsfiles format. This enhances compatibility with applications that build off of MSVC (such as Unreal Engine).

Alexei-Barnes:
MSVC deps files are emitted as JSON, in a format described by the compiler documentation: https://learn.microsoft.com/en-us/cpp/build/reference/sourcedependencies?view=msvc-170

Prior to this commit, sccache will output a clang or gcc style deps file, with dependencies on lines instead of JSON.

After this commit, we remove the code responsible for writing the deps file, and depend on the command we execute on `cl.exe` for preprocessing to write the deps file.

The advantage of this is that we get improved compatibility with tools that use the JSON output deps files from MSVC, such as Unreal Engine's build system.

This fixes #1771

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -370,7 +370,9 @@ jobs:
       - name: Compile MSVC (no cache)
         shell: bash
         working-directory: ./tests/msvc
-        run: cl "@args.rsp"
+        run: |
+          cl "@args.rsp"
+          test -e ./foo.o || { echo "No compiler output found"; exit -1; }
 
       - name: Start Server
         shell: bash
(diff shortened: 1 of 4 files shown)


sylvestre:
Could you please update your commit message to explain the why you are doing it ?
(what you wrote in #1771 is good)

Alexei-Barnes:
Updated commit and PR messages to be more informative.

Issue #847, opened 2020-09-13
Error when running preprocessor for clang invocation having multiple -arch flags


kastiglione:
For some build configs, sccache errors when compiling [llvm compiler-rt](https://compiler-rt.llvm.org). The error is:

```
clang-10: error: cannot use 'c++-cpp-output' output with multiple -arch options
```

It's true, `clang`'s invocation has two `-arch` flags (`-arch x86_64` `-arch x86_64h`). I am new to sccache, but from my brief investigation, this seems to break the strategy of preprocessing before compiling. Maybe sccache will need to run the preprocessor once for each arch.

froydnj:
> Maybe sccache will need to run the preprocessor once for each arch.

Yes, this would require some surgery on how hash keys are computed.  The other alternative (probably not great for your purposes) is to give up when multiple `-arch` options are seen, and accept that we can't cache such compilations.

kastiglione:
Yes I have a wrapper, `sccachew`, that skips sccache when the args have more than one arch value.

mathstuf:
Does the compiler get grumpy if one excises all but one `-arch` flag, but leaves in other `-X$arch` flags for arch-specific flag communication? If so, the `-X$arch` flags would also need filtered out before preprocessing as well.

glandium:
> other -X$arch flags

What -X$arch flags? Do you have examples?

mathstuf:
You can send arguments to just specific arch TU compilations. For example, `-Xx86_64 -mno-avx512` since the `-mno-avx512` makes no sense on an aarch64 compilation. I forget the exact spelling of `-Xx86_64` right now, but it exists.

glandium:
I was thinking that's what it is about, but I couldn't find what the -X$arch was supposed to be, because -Xx86_64 or -Xarm64 definitely don't work, and clang --help doesn't seem to list such a flag (but then, it also doesn't list -arch).

mathstuf:
Seems that it is `-Xarch_x86_64` and such: https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-xarch-arg1

glandium:
Nice find. Thanks.

glandium:
Re-opening and re-purposing to handle this without a "cannot_cache" in the future. For that to happen, we'd need to preprocess twice, once with each -arch, taking into account the extra -Xarch_$arch flags per the previous comments.

alcroito:
[...]

PR #1775, opened 2023-05-26
Added env var to disable multi arch caching

coretechsec:
Added an environment variable SCCACHE_CACHE_MULTIARCH to enable caching of multiarch compilations.

This should help in cases such as that raised in https://github.com/mozilla/sccache/issues/847#issuecomment-1517546993

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -277,7 +277,7 @@ where
     let mut xclangs: Vec<OsString> = vec![];
     let mut color_mode = ColorMode::Auto;
     let mut seen_arch = None;
-    let dont_cache_multiarch = !env::var("SCCACHE_CACHE_MULTIARCH").is_ok();
+    let dont_cache_multiarch = env::var("SCCACHE_CACHE_MULTIARCH").is_err();
 
     // Custom iterator to expand `@` arguments which stand for reading a file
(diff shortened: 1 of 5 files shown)


src/compiler/gcc.rs
```diff
@@ -365,7 +368,15 @@ where
                     _ => cannot_cache!("-x"),
                 };
             }
-            Some(Arch(_)) => {}
+            Some(Arch(arch)) => {
+                match seen_arch {
+                    Some(s) if &s != arch && dont_cache_multiarch => {
```
Xuanwo, line 373: Is it a good idea to not cache the multiarch build all the time?

coretechsec: As in flip the logic so that the env var is to enable multiarch caching?

Xuanwo: > As in flip the logic so that the env var is to enable multiarch caching but default to not caching?

Yes, my original idea is always not caching in this case. But flip the logic also a good idea.

coretechsec: I've flipped the logic and made it so that the arch flag rewriting only happens when multiple `-arch` flags are given

Xuanwo (approved):
Thanks!

PR #1778, opened 2023-05-30
snap: disable store uploads from CI

Saviq:
It doesn't work for PRs from forks, which is most common in this repository.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -16,10 +16,7 @@ jobs:
       with:
         fetch-depth: 0  # needed for version determination
 
-    - name: Build and publish the snap
+    - name: Build the snap
       uses: canonical/actions/build-snap@release
       with:
         review-opts: --allow-classic
-        snapcraft-token: ${{ secrets.SNAPCRAFT_TOKEN }}
-        publish: ${{ github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name }}
-        publish-channel: edge/pr${{ github.event.number }}


drahnr:
I don't quite understand? Why would we disable it for the main repository for the sake of making it work for a fork? You could equally create a new snapcraft account or patch your fork.

Could you please elaborate?

Saviq:
> I don't quite understand? Why would we disable it for the main repository for the sake of making it work for a fork? You could equally create a new snapcraft account or patch your fork.

It's not for the sake of making it work for a fork. Access to secrets is limited to in-repo PRs, and most of the pull requests here originate from forks (for which the publishing was disabled already by means of the `github.repository == github.event.pull_request.head.repo.full_name` condition).

That's similar to what @Xuanwo suggested (no need for a new step, that's what the `publish` input is for). But together with @sylvestre we thought it's not worth it since most PRs will be coming from forks. I'm happy to be convinced otherwise, but can't do anything more - someone would need to set the `SNAPCRAFT_TOKEN` secret as discussed above, and this PR can be closed.

sylvestre:
@Saviq i added SNAPCRAFT_TOKEN 
how can i share it works ? thanks :)

Saviq:
@sylvestre try restarting this:

https://github.com/mozilla/sccache/actions/runs/5142969257/jobs/9257385196

When it succeeds, the built snap will be published under `edge/pr1764`.

sylvestre:
and it worked, thanks :)

Saviq:
In that case!

PR #1781, opened 2023-05-31
Response file parsing for GCC and Clang

### LINKED ISSUE #1780: Response file parsing for GCC and Clang does not handle quoted arguments
Unreal Engine uses response files with compilation, even when using Clang on Linux. Some of the arguments in these response files contain quotation marks. Conveniently, [we already have nice parsing for response files for MSVC that handle this case](https://github.com/mozilla/sccache/pull/1618), but unfortunately, we don't have this for GCC or Clang.

Alexei-Barnes:
This fixes #1780

This moves the response file parsing code implemented by @temportalflux out of the MSVC compiler and shares it for all compilers. This enables GCC and Clang to use response files with quoted arguments.

I've added some lib tests to the response file module, so that it can be reasoned about, separately from the compilers using it.

### DIFF
--- a/src/compiler/response_file.rs
+++ b/src/compiler/response_file.rs
@@ -1,5 +1,4 @@
-
-/// An iterator over the arguments in a Windows command line.
+/// An iterator over the arguments in a response file.
 ///
 /// This produces results identical to `CommandLineToArgvW` except in the
 /// following cases:
@@ -24,13 +23,13 @@
 ///  - https://msdn.microsoft.com/en-us/library/windows/desktop/bb776391(v=vs.85).aspx
 ///  - https://msdn.microsoft.com/en-us/library/windows/desktop/17w5ykft(v=vs.85).aspx
 #[derive(Clone, Debug)]
-pub struct SplitMsvcResponseFileArgs<'a> {
+pub struct SplitResponseFileArgs<'a> {
     /// String slice of the file content that is being parsed.
     /// Slice is mutated as this iterator is executed.
     file_content: &'a str,
 }
 
-impl<'a, T> From<&'a T> for SplitMsvcResponseFileArgs<'a>
+impl<'a, T> From<&'a T> for SplitResponseFileArgs<'a>
 where
     T: AsRef<str> + 'static,
 {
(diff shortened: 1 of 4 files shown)


sylvestre:
rustfmt isn't happy, could you please fix it? thanks

Alexei-Barnes:
Ran `cargo fmt` and force pushed.

src/compiler/response_file.rs
```diff
@@ -0,0 +1,165 @@
+/// An iterator over the arguments in a response file.
```
sylvestre, line 1: i would prefer the move to be done into a single commit 
and then, modifications into a second commit

Alexei-Barnes: Sure, I'll rewrite the commit into two commits in that order.

sylvestre: @Alexei-Barnes ping ? :)

Alexei-Barnes: Sorry for the slow action here - I've been pulled into a work project that's eating my free time. I should get back to this soon.

sylvestre: are you still going to work on this? thanks

Alexei-Barnes: Right, sorry for the mega wait - I'm working on this again now.

Alexei-Barnes: OK, this is split into two separate commits now. The first commit just moves the code, with minimal changes necessary to do so. The second change modifies it, adds tests, and uses it in the GCC module.

Alexei-Barnes: I've now also merged main into this branch, to keep it up to date. No merge conflicts.

sylvestre:
sorry but it will need to be rebase :(

robertmaynard:
What is the value of parsing an response file instead of just hashing it? Is this ensures we are invariant to order of flags in the response file?

glandium:
> What is the value of parsing an response file instead of just hashing it? Is this ensures we are invariant to order of flags in the response file?

For the same reasons we don't just hash command line arguments:
- some flags need to be handles in a special way
- some flags need to disable caching because we can't handle them
- we need to distinguish preprocessor flags from other flags so that we can invoke the preprocessor with the right flags
- the input file may be in the response file, and we need to know what it is
- etc.

robertmaynard:
> > What is the value of parsing an response file instead of just hashing it? Is this ensures we are invariant to order of flags in the response file?
> 
> For the same reasons we don't just hash command line arguments:
> 
> * some flags need to be handled in a special way
> * some flags need to disable caching because we can't handle them
> * we need to distinguish preprocessor flags from other flags so that we can invoke the preprocessor with the right flags
> * the input file may be in the response file, and we need to know what it is
> * etc.

Makes sense. My feedback would be we should extend `ArgData` in gcc.rs to have a explicit type for response files. 
This would allow compilers that extend gcc such as nvcc to leverage this work even when they have a different command line syntax to represent response files.

sylvestre:
no activity for a while, closing

Issue #1799, opened 2023-06-07
[BUG] `SCCACHE_S3_NO_CREDENTIALS=true` stopped working in `0.5.0`


ajschmidt8:
It seems that the `SCCACHE_S3_NO_CREDENTIALS=true` configuration stopped working in `0.5.0`.

When using the following configuration with a publicly readable bucket with `sccache` version `0.4.2`, `sccache` compiles correctly:

```
SCCACHE_BUCKET=<publicly_readable_bucket>
SCCACHE_IDLE_TIMEOUT=32768
SCCACHE_REGION=us-east-2
SCCACHE_S3_NO_CREDENTIALS=true
SCCACHE_S3_USE_SSL=true
```

However, in the same environment, `sccache` version `0.5.0` throws the following error message:

```
sccache: error: Server startup failed: cache storage failed to read: PermissionDenied (temporary) at read => no valid credential found, please check configuration or try again
    
    Context:
        service: s3
        path: .sccache_check
        range: 0-
    
    Run with SCCACHE_LOG=debug SCCACHE_NO_DAEMON=1 to get more information
```

The issue exists in `0.5.3` as well.

Setting `SCCACHE_LOG=debug` and  `SCCACHE_NO_DAEMON=1` did not seem to provide any additional details.

ajschmidt8:
@Xuanwo, I'm wondering if the `opendal` upgrade could have affected this: https://github.com/mozilla/sccache/compare/v0.4.2...v0.5.0.

ajschmidt8:
here is a better stacktrace:

```
[2023-06-07T18:47:09Z DEBUG sccache::config] Attempting to read config file at "/root/.config/sccache/config"
[2023-06-07T18:47:09Z DEBUG sccache::config] Couldn't open config file: failed to open file `/root/.config/sccache/config`
[2023-06-07T18:47:09Z DEBUG sccache::config] Attempting to read config file at "/root/.config/sccache/config"
[2023-06-07T18:47:09Z DEBUG sccache::config] Couldn't open config file: failed to open file `/root/.config/sccache/config`
[2023-06-07T18:47:09Z INFO  sccache::server] start_server: port: 4226
[2023-06-07T18:47:09Z INFO  sccache::server] No scheduler address configured, disabling distributed sccache
[2023-06-07T18:47:09Z DEBUG sccache::cache::cache] Init s3 cache with bucket <redacted_bucket>, endpoint None
[...]

PR #1800, opened 2023-06-07
fix: Don't call EC2 Metadata and allow anonymouse if no cred

Xuanwo:
Fix https://github.com/mozilla/sccache/issues/1799


sylvestre:
Could you please add some comments to explain the why ? :)

Xuanwo:
cc @sylvestre, Please take another look, and release a new version for this fix. Thanks!

PR #1809, opened 2023-06-14
enh(msvc): ignore /d2Zi+

huskar-t:
https://learn.microsoft.com/en-us/cpp/build/reference/zo-enhance-optimized-debugging?view=msvc-170
`The /Zo option is available starting in Visual Studio 2013 Update 3, and it replaces the previously undocumented /d2Zi+ option.`
So /d2Zi+ should be ignored


sylvestre:
Could you please add a test? Thanks

huskar-t:
@sylvestre According to Microsoft's documentation, it is the same as /Zo, I don't see any tests related to /Zo

huskar-t:
github actions fail on freebsd, doesn't seem to be related to my commit

sylvestre:
> @sylvestre According to Microsoft's documentation, it is the same as /Zo, I don't see any tests related to /Zo

Don't hesitate to write a test for /Zo too

sylvestre:
please reopen when ready

PR #1818, opened 2023-06-21
Fix timing of cache reads. Fixes #1817

### LINKED ISSUE #1817: Cache timing appears to be broken
I noticed this while looking at the log in #1756. All of the cache hit log lines read `Cache hit in 0.000 s`. It looks like this got broken at some point. The duration is calculated here:
https://github.com/mozilla/sccache/blob/62c8f0aa168037628058221448d2aab67fd7968d/src/compiler/compiler.rs#L268

but that's prior to where it actually `await`s the cache get future:
https://github.com/mozilla/sccache/blob/62c8f0aa168037628058221448d2aab67fd7968d/src/compiler/compiler.rs#L277

This also means the summary stats don't work either:
```
```

luser:
I'm not sure when this got broken, but the cache read timing isn't being calculated properly. This change fixes it. I added a test by extending the `MockStorage` struct (which is only used in one other test) to include an optional delay on get, and just asserting that the cache read duration is >= that delay. Fixes #1817

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -262,10 +262,13 @@ where
         // Set a maximum time limit for the cache to respond before we forge
         // ahead ourselves with a compilation.
         let timeout = Duration::new(60, 0);
-        let cache_status = tokio::time::timeout(timeout, cache_status);
+        let cache_status = async {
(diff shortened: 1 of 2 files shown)


sylvestre:
welcome back @luser ;)

PR #1823, opened 2023-06-30
nvcc: Support -Werror with argument

robertmaynard:
The `nvcc` -Werror supports a syntax where the argument is separated by a space `-Werror all-warnings` and therefore needs to use different rules compared to gcc which always expects an explicit `=` between Werror and the argument.


sylvestre:
@dependabot rebase

Issue #1832, opened 2023-07-19
Cache write error are not counted properly


glandium:
With current main branch, I have a build that showed this in its log:
```
[2023-07-19T05:04:49Z DEBUG sccache::compiler::compiler] [gkrust]: Cache write error: PermissionDenied (permanent) at Writer::write, context: { response: Parts { status: 403, version: HTTP/1.1, headers: {"x-guploader-uploadid": "ADPycduOxEW1mC3qEK4-wOqJBpK0BsDy549C1YshSdnUEG7GWkSMVVHFRuEJtgKs8Y1G7ryK2HZuM-DVSKMnq4AOWIpiTw", "content-type": "application/json; charset=UTF-8", "date": "Wed, 19 Jul 2023 05:04:49 GMT", "vary": "Origin", "vary": "X-Origin", "cache-control": "no-cache, no-store, max-age=0, must-revalidate", "expires": "Mon, 01 Jan 1990 00:00:00 GMT", "pragma": "no-cache", "content-length": "448", "server": "UploadServer"} }, service: gcs, path: 6/5/7/6577a8d92d6ed0092b63cf6913ba17588bf23355098ff5459c7aa9191db6008c } => GcsErrorResponse { error: GcsError { code: 403, message: "sccache-l1-us-central1@sccache-3.iam.gserviceaccount.com does not have storage.objects.delete access to the Google Cloud Storage object.", errors: [GcsErrorDetail { domain: "global", location: "", location_type: "", message: "sccache-l1-us-central1@sccache-3.iam.gserviceaccount.com does not have storage.objects.delete access to the Google Cloud Storage object.", reason: "forbidden" }] } }
    
    Stack backtrace:
       0: <unknown>
       1: <unknown>
       2: <unknown>
       3: <unknown>
       4: <unknown>
       5: <unknown>
       6: <unknown>
       7: <unknown>
       8: <unknown>
       9: <unknown>
      10: <unknown>
      11: clone
[2023-07-19T05:04:49Z DEBUG sccache::server] [gkrust]: Cache write finished in 0.500 s
```
However, in its stats, sccache says:
```
Cache write errors                     0
```

glandium:
This is a regression from e7099d6c8ad05b7a3994698ca0da92bd8452d687.

PR #1833, opened 2023-07-19
Properly propagate cache write errors

glandium:
We leave it to the server code to handle printing the propagated error. One downside is that the log doesn't mention the associated output anymore, and the time spent on failing the cache write is not accounted for anymore either, but it's probably fine.

Fixes: #1832.


Xuanwo (approved):
Thanks！

PR #1834, opened 2023-07-21
msvc: passthrough argument -Qpar-

AudranDoublet:
[`/Qpar-`](https://learn.microsoft.com/en-us/cpp/build/reference/qpar-auto-parallelizer?view=msvc-170) is set by MsBuild when `EnableParallelCodeGeneration` is explicitly set to `false`.

Current code is considering it as a file name, thus the command is considered as non-cacheable (multiple input file).


sylvestre:
Could you please add a test ? thanks

AJIOB:
Hi @Xuanwo.

Looks like the PR is ready, isn't it?

Issue #1830, opened 2023-07-12
Server sent UnsupportedCompiler: "Failed to detect showIncludes prefix"


exoosh:
With `LOG_BUILD_COMMANDLINES`, I can see that `sccache` -- behind the scenes -- was evidently trying to execute something to the effect of testing that the compiler functions?!

```
CL -E C:\Users\OSCHNE~1\AppData\Local\Temp\sccacheLnjRtT\testfile.c 
CL -nologo -showIncludes -c -Fonul -I. C:\Users\OSCHNE~1\AppData\Local\Temp\sccacheUtHbTK\test.c 
```

In the build output I see:

```
1>  [2023-07-12T10:20:36Z DEBUG sccache::commands] Server sent UnsupportedCompiler: "Failed to detect showIncludes prefix"
2>  [2023-07-12T10:20:36Z DEBUG sccache::commands] Server sent UnsupportedCompiler: "Failed to detect showIncludes prefix"
2>  sccache : error : failed to execute compile
```

What I am doing is to "inject" `sccache` as `cl.exe` in `PATH` via a `Directory.Build.props`:

```
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" InitialTargets="LogBuild">
	<PropertyGroup>
		<ThisProjectBuildLogFileName Condition="'$(MSBuildProjectName)' == ''">$(MSBuildThisFileDirectory)BuildCommandLines.log</ThisProjectBuildLogFileName>
		<ThisProjectBuildLogFileName Condition="'$(MSBuildProjectName)' != ''">$(MSBuildThisFileDirectory)BuildCommandLines-$(MSBuildProjectName).log</ThisProjectBuildLogFileName>
		<PreferredToolArchitecture Condition="'$(PreferredToolArchitecture)' == '' and ('$(PROCESSOR_ARCHITECTURE)' == 'AMD64')">x64</PreferredToolArchitecture>
	</PropertyGroup>
	<Target Name="LogBuild" BeforeTargets="SetUserMacroEnvironmentVariables;SetBuildDefaultEnvironmentVariables">
		<Message Text="Setting LOG_BUILD_COMMANDLINES='$(ThisProjectBuildLogFileName)'" />
		<SetEnv Name="LOG_BUILD_COMMANDLINES" Value="$(ThisProjectBuildLogFileName)" Prefix="false" />
[...]

Issue #909, opened 2020-12-16
msvc + sccache: "Failed to detect showIncludes prefix"


autoantwort:
I have build the current master branch and "installed" is like described in Issue #107 (penultimate comment). But when I try to build the project inside Visual Studio I get the output:
```
1>------ Build started: Project: lib, Configuration: Debug Win32 ------
1>sccache : error : failed to execute compile
1>  sccache: caused by: Compiler not supported: "Failed to detect showIncludes prefix"
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```

autoantwort:
I have executed the test manally and the output was:
```
xx@xx MSYS /c/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/test
$ cl.exe test.cpp -nologo -showIncludes -c -Fonul -I.
test.cpp
Hinweis: Einlesen der Datei: c:\program files (x86)\microsoft visual studio 14.0\vc\bin\test\test.h
```

autoantwort:
With debug messages enabled I get the following output:

```
13>  [2020-12-16T19:44:11Z DEBUG sccache::config] Attempting to read config file at "C:\\Users\\xx\\AppData\\Roaming\\Mozilla\\sccache\\config\\config"
13>  [2020-12-16T19:44:11Z DEBUG sccache::config] Couldn't open config file: Das System kann den angegebenen Pfad nicht finden. (os error 3)
13>  [2020-12-16T19:44:11Z DEBUG sccache::commands] Server sent UnsupportedCompiler: "failed to detect showIncludes prefix with output: Tracker.exe: Response file C:\\Users\\xx\\AppData\\Local\\Temp\\tmp9625afad50c140a7858f1025888972c1.tmp not found.\r\nTracker.exe: Response file C:\\Users\\xx\\AppData\\Local\\Temp\\tmp9625afad50c140a7858f1025888972c1.tmp not found.\r\nTracker.exe: Response file C:\\Users\\xx\\AppData\\Local\\Temp\\tmp9625afad50c140a7858f1025888972c1.tmp not found.\r\n"
```

autoantwort:
I tried to print out the exe name, but I failed with the following:
```
debug!("cl.exe: {}", exe.to_str().expect("error"));
```

glandium:
> but I failed with the following:

Can you give more details how you failed?

[...]

PR #1835, opened 2023-07-21
msvc: fix detect_showincludes_prefix when VS_UNICODE_OUTPUT is set

AudranDoublet:
`VS_UNICODE_OUTPUT` is set by Visual Studio to a value that tells MS tools running from within the IDE where to send their output. Thus, `cl.exe` invocation used to retrieve the `/showIncludes` prefix returns an empty output when this variable is set to a valid value.

The proposed fix is to use `/E` option that tells `cl.exe` to output file content in stdout and retrieved headers in stderr. When set, `cl.exe` ignores the value of `VS_UNICODE_OUTPUT`.

Another fix could be to explicitly unset `VS_UNICODE_OUTPUT` in the command invocation.

#1830  #909 mentioning these issues as it may fix them.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -186,13 +186,9 @@ where
     if is_clang {
         cmd.arg("--driver-mode=cl");
     }
-    cmd.args(&["-nologo", "-showIncludes", "-c", "-Fonul", "-I."])
+    cmd.args(&["-nologo", "-showIncludes", "-c", "-Fonul", "-I.", "-E"])
         .arg(&input)
-        .current_dir(tempdir.path())
-        // The MSDN docs say the -showIncludes output goes to stderr,
-        // but that's not true unless running with -E.
-        .stdout(Stdio::piped())
-        .stderr(Stdio::null());
+        .current_dir(tempdir.path());
     for (k, v) in env {
         cmd.env(k, v);
     }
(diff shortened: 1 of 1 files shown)


src/compiler/msvc.rs
```diff
@@ -187,13 +187,9 @@ where
...
         cmd.arg("--driver-mode=cl");
     }
-    cmd.args(&["-nologo", "-showIncludes", "-c", "-Fonul", "-I."])
+    cmd.args(&["-nologo", "-showIncludes", "-c", "-Fonul", "-I.", "-E"])
         .arg(&input)
-        .current_dir(tempdir.path())
-        // The MSDN docs say the -showIncludes output goes to stderr,
-        // but that's not true unless running with -E.
-        .stdout(Stdio::piped())
-        .stderr(Stdio::null());
```
AudranDoublet, line 196: removed as `run_input_output` override these values

sylvestre:
many jobs are failing to build
please reopen when they are fixed! :)

njames93:
Is there any update on this, I tried to run this locally(based of 0.8.0 release) and the error went away, but it seems that sccache didn't attempt to do any caching

exoosh:
Such a shame this got closed. @AudranDoublet would you mind if I attempt to shape up your changes and adjust them to the latest version of `sccache`? This is literally _the_ blocker for anyone using MSVC via the VS IDE ... as pointed out in #1830 ...

PS: not much adjustment was required. Rebase worked, now checking regarding tests and so on.

Issue #1687, opened 2023-03-27
sccache places object file at different location compared to cl.exe and causes link failure when @rsp file is used


UsernameLinde:
When the [`@rsp` support pull request ](https://github.com/mozilla/sccache/pull/1618) landed I was eager to use that on Windows/MSVC.

### Context

The internal requirement was that the build process should be pretty much as is, changing to ninja from msbuild was not an option. So I first tried to hook in `sccace` in a clean way.
I use `directory.build.props` and a `bat`-file to get `sccache` prefixed before the call to `cl.exe`. More details [here](https://github.com/mozilla/sccache/issues/107#issuecomment-1484642398).

That worked: `sccache` gets to see the compile call, but I get `Uncachable reason: multiple input files` when the call is 
`d:\tools\sccache.exe cl.exe @C:\Users\me\AppData\Local\Temp\tmpfa7e185af63d4159b8ef0462dcbf1754.rsp`

The `rsp` file do list multiple source files.

So I then changed the "compile call" to be `python somescript.py %*`, that is, I have a pythonscript which gets called with the `rsp`-file as parameter.

I parse the rsp-file. I group the conents into either "flag" or "source file".
Then for each source file:
- I create a new `rsp`-file with all the flags and then just that source name.
- I call `sccache cl.exe @full_path_to_new_rsp_file.rsp`

With this extra step, I can compile from visual studio. With rebuild I get cache hits. When it gets to linking, there is an error.

### Actual problem

The `rsp` file contains `/Fo"somename.dir\\Debug\\"`, so object files should go into a sub directory relative to the source location.
Assume that the `rsp`-file lists `foo.c` as file to compile. 

When I use `cl.exe @myrspfile.rsp` I get `somename.dir\Debug\foo.obj`.
When I use `sccache cl.exe @myrspfile.rsp` I instead get `somename.dir\Debug.obj`

And this is the reason for link failure - `/Fo` can contain a directory name and does not have to be the filename for the output obj file.

TBBle:
I note that the unit tests [currently only test `/Fo` without a trailing slash](https://github.com/mozilla/sccache/blob/d0d1beadd2739653ed878055215c0d9b7decb3b9/src/compiler/msvc.rs#L1457-L1458).

[...]

PR #1836, opened 2023-07-21
msvc: #1687: handle case where output file is a directory

AudranDoublet:
As described in #1687, MSBuild may provide as output file (`/Fo`) a folder, ending with a trailing backslash, for example `/FoRelease\`.

msvc detects this case and chose `{folder}/{input_file_name}.obj` as output file. For example, for `/cfoo.c /FoRelease\`, the expected output path (by linker and msbuild) is `Release/foo.obj`.

Currently, we would use `Release.obj`, leading to:
- concurrent write to the same object if the project contains multiple source
- output file can't be found by other MSVC tools

The proposed fix simply check if the `std::path::Path` contains a trailing backslash (see https://users.rust-lang.org/t/trailing-in-paths/43166) and correctly guess the path if so.


src/compiler/msvc.rs
```diff
@@ -788,7 +788,20 @@ pub fn parse_arguments(
             );
         }
         Some(o) => {
-            if o.extension().is_none() && compilation {
+            if o.as_os_str()
```
sylvestre, line 791: sorry for the latency.
compilation is gone from the check, is that expected?

AudranDoublet: Np. It is intentional, at least. `compilation` is checked a bit earlier in this function (l760): this check was always true

jimis:
The bug exists in sccache-0.11: `-Forelease\` does not write to the expected directory.
@AudranDoublet do you have time to refresh your patch? Else I could try to rebase it to HEAD myself.

jimis:
Rebased and did minor updates to this pull request in #2482.

PR #1856, opened 2023-08-05
Don't pass wd4668 to clang-cl

### LINKED ISSUE #1843: C++ ClangCL on Windows - `/wd4668` flag results in error
I'm compiling a C++ project on Windows using ClangCL from Scoop's `versions/mingw-winlibs-llvm` package. In my CMakeLists.txt file I call `set(CMAKE_CXX_COMPILER_LAUNCHER "${SCCACHE_PROGRAM}")`, and I can compile my project with MSVC, Clang, GCC and ClangCL with Visual Studio generator without having to install the Clang toolset from VS Installer (after setting the `llvm-mt` executable manually before a `project()` call), but trying to compile using ClangCL with the MinGW Makefiles generator option in CLion results in:
```
clang-cl: error: no such file or directory: '/wd4668'
mingw32-make[3]: *** [_deps\catch2-build\src\CMakeFiles\Catch2.dir\build.make:74: _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/catch_chronometer.cpp.obj] Error 1
mingw32-make[3]: *** Waiting for unfinished jobs....
```
This does not happen when using `ccache` instead of `sccache`.
I've installed `ccache` and `sccache` using Scoop as well.

Correct me if any of the mentioned steps are incorrect or I'm doing something wrong 💯

roblabla:
Fixes #1843 

We should only pass `/wd4668` when the compiler is MSVC, not when it's clang-cl, as the later does not understand that flag (and the issue it fixes is not relevant to clang-cl-based builds).

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -885,9 +885,6 @@ where
         .args(&parsed_args.preprocessor_args)
         .args(&parsed_args.dependency_args)
         .args(&parsed_args.common_args)
-        // Windows SDK generates C4668 during preprocessing, but compiles fine.
-        // Read for more info: https://github.com/mozilla/sccache/issues/1725
-        .arg("/wd4668")
         .env_clear()
         .envs(env_vars.iter().map(|(k, v)| (k, v)))
         .current_dir(cwd);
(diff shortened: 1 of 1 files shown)


quyykk:
@sylvestre sorry for the ping, but sccache is currently unusable with clang-cl without this fix so it would be great to get it this PR reviewed and merged. thanks! :smile:

sylvestre:
Sorry I missed it.
Could you please add a test to make sure we don't regress?

aeubanks:
the description is not quite right, see my comment in https://github.com/mozilla/sccache/issues/1843#issuecomment-1707276722

but this approach is still right

we should also not pass `--` for MSVC, but that's a separate bug

glandium:
> we should also not pass -- for MSVC, but that's a separate bug

I'd argue it's not sccache's business to remove command line arguments that were passed in.

PR #1874, opened 2023-09-04
Fix clang version detection, update freebsd ci

grembo:
Also updated checkout action, since v3 started to fail today.

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -38,9 +38,9 @@ use crate::errors::*;
 /// A struct on which to implement `CCompilerImpl`.
 #[derive(Clone, Debug)]
 pub struct Clang {
-    /// true iff this is clang++.
+    /// true if this is clang++.
     pub clangplusplus: bool,
-    /// true iff this is Apple's clang(++).
+    /// true if this is Apple's clang(++).
     pub is_appleclang: bool,
     /// String from __VERSION__ macro.
     pub version: Option<String>,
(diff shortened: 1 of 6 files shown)


src/compiler/clang.rs
```diff
@@ -38,9 +38,9 @@ use crate::errors::*;
 /// A struct on which to implement `CCompilerImpl`.
 #[derive(Clone, Debug)]
 pub struct Clang {
-    /// true iff this is clang++.
+    /// true if this is clang++.
```
sylvestre, line 41: why this change?
iff means `If and only if`

grembo: Just seems odd, as "if" is totally sufficient in this case (otherwise we would write `iff true {`, wouldn’t we?). And yes, I wasn’t aware ;)

sylvestre:
could you please create a PR for freebsd and one for clang detection.
afaik, they are unrelated changes

(i already merged a different pr about the action update)

grembo:
> could you please create a PR for freebsd and one for clang detection. afaik, they are unrelated changes
> 
> (i already merged a different pr about the action update)

Hm, they were kind of related (technically it could also be three separate ones of course - fix bsd unit tests, fix clang detecting, update FreeBSD), but they would need to be applied in order for tests to succeed.

I see what I can do (later).

grembo:
> > could you please create a PR for freebsd and one for clang detection. afaik, they are unrelated changes
> > (i already merged a different pr about the action update)
> 
> Hm, they were kind of related (technically it could also be three separate ones of course - fix bsd unit tests, fix clang detecting, update FreeBSD), but they would need to be applied in order for tests to succeed.
> 
> I see what I can do (later).

To me more precise, updating FreeBSD to 13.2 brought in clang >=14 which in turn exposed the bug in the clang unit test and once that was fixed, it exposed a bug in clang version parsing on Ubuntu (the bug in the clang unit test masked the bug in clang version parsing when there was no additional information, like a url, after the version number). Anyway :)

grembo:
Closing this in favor of #1877, #1878, and #1879.

PR #1875, opened 2023-09-04
Bring back redis over tls

### LINKED ISSUE #1839: Redis no longer support TLS
I tired to setup a redis instance with TLS, but I was very surprised to see that the scheme "rediss" (note the second "s") was refused by sccache, moreover since the documentation clearly states that it should be supported.

Digging a bit into the issue, I found that it come from Opendate.
I suspect that redisS support have been dropped when sccache moved to opendale or by opendale itself at some point.

I opened an issue at opendale for this problem (https://github.com/apache/incubator-opendal/issues/2669).
Could it be possible to update your opendale version once the fix land ?

grembo:
Fixes #1839

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ strip = true
 anyhow = { version = "1.0", features = ["backtrace"] }
 ar = "0.9"
 async-trait = "0.1"
-backon = "0.4.1"
 base64 = "0.21"
 bincode = "1"
 blake3 = "1"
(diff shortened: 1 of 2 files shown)


grembo:
Reduced the patch to the functional change (meaning that parts of CI will fail)

Cargo.toml
```diff
@@ -27,6 +27,7 @@ strip = true
 anyhow = { version = "1.0", features = ["backtrace"] }
 ar = "0.9"
 async-trait = "0.1"
+backon = "0.4.1"
```
grembo, line 30: This line is required for the project to build, since https://github.com/apache/incubator-opendal/commit/2c2dfc8f994d14bcbb16d44ef3603081fe9e0745 is missing in opendal 0.39.

grembo:
@sylvestre Now that main is already updated to opendal 0.40.0, this change is very simple (just bring in the feature to support rediss://)

PR #1877, opened 2023-09-04
Fix freebsd-ci

grembo:
It's not safe to run sccache-dist unit tests in parallel.


grembo:
@sylvestre This actually won't fix FreeBSD unit tests until #1878 is also applied.

grembo:
Closing in favor of #1879

PR #1878, opened 2023-09-04
Fix detection of clang version and v14+ unit tests

grembo:
Compiler version (`__VERSION__`) is always enclosed in quotes.
This only fixes version detection for clang to keep the scope small.
Stripping quotes could also be done globally in compiler.rs.

### DIFF
--- a/src/compiler/clang.rs
+++ b/src/compiler/clang.rs
@@ -63,7 +63,7 @@ impl Clang {
             None => return false,
         };
 
-        let parsed_version = match Version::parse(version_str) {
+        let parsed_version = match Version::parse(version_str.trim_end_matches('"')) {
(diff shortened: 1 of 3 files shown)


sylvestre:
would it be possible to have a test for this?

grembo:
> would it be possible to have a test for this?

It's two things - the unit test code itself was broken. Once it was fixes (test/system.rs), the problematic behavior was exposed. We could have a basic unit test of course (not really hard to do). Not sure how much it buys us, but also won't hurt...

src/compiler/clang.rs
```diff
@@ -252,6 +252,38 @@ mod test {
         }
     }
 
+    #[test]
```
sylvestre, line 255: thanks

PR #1881, opened 2023-09-07
 Add MacPorts as a new installation method

MisLink:
I have updated sccache to the latest version in macports/macports-ports#20343. Now users can install the latest version of sccache through MacPorts. This PR add this information to the README.

Also, I will do my best to keep up with the release.


sylvestre:
Cool stuff, thanks :)

Issue #1077, opened 2021-12-08
hits Cuda Cache but many link errors on Windows


mszhanyi:
I tried to build PyTorch locally with ninja+CMake+sccache.
With the lastest sccache, No CUDA Cache Hits and the non-cacheable reasons are.
```
Non-cacheable reasons:
multiple input files          137
```
One example command is
```
[...]

PR #1884, opened 2023-09-18
Support msvc as the host compiler for nvcc

robertmaynard:
To support MSVC as the nvcc host compiler we need to pass `-EP` for the pre-processing step. If we pass `-P` the pre-processor will place the output in `<input_file>.i` and not the requested output file given by `-o`.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1082,8 +1082,10 @@ where
     // Both clang and clang-cl define _MSC_VER on Windows, so we first
     // check for MSVC, then check whether _MT is defined, which is the
     // difference between clang and clang-cl.
-    let test = b"#if defined(__NVCC__)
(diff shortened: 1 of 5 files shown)


robertmaynard:
I believe that the link issues reported in https://github.com/mozilla/sccache/issues/1077 will be fixed with this PR. Since the pre-processor output is empty, each object file is empty and when you go to link you will get missing symbol errors.

sylvestre:
is it possible to add the test?
thanks

robertmaynard:
@sylvestre does that test look reasonable?

sylvestre:
Sorry I missed your response

robertmaynard:
Closed as https://github.com/mozilla/sccache/pull/1889 combines these changes with support for nvc++ which required a better design for how we track the host compiler mode for `nvcc`.

sylvestre:
oh, sorry, i was afraid you were pissed :p sorry

PR #1889, opened 2023-09-25
Support nvc++ and msvc as nvcc host compilers

robertmaynard:
Continues the work of https://github.com/mozilla/sccache/pull/1884

Extends the nvcc compiler to understand the unique requirements of using `nvc++` or `cl.exe` ( msvc ) as the host compiler. 

Due to the fact that  `nvc++`  pre-processing always has force includes the compiler detection can't presume that the first line of the file will hold the compiler id string. So we instead search for lines starting with a unique starting text.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1086,7 +1086,8 @@ where
     //
     // We prefix the information we need with `compiler_id` and `compiler_version`
     // so that we can support compilers that insert pre-amble code even in `-E` mode
-    let test = b"#if defined(__NVCC__) && defined(__NVCOMPILER)
+    let test = b"
+#if defined(__NVCC__) && defined(__NVCOMPILER)
 compiler_id=nvcc-nvhpc
 #elif defined(__NVCC__) && defined(_MSC_VER)
 compiler_id=nvcc-msvc
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1082,34 +1082,37 @@ where
     // Both clang and clang-cl define _MSC_VER on Windows, so we first
     // check for MSVC, then check whether _MT is defined, which is the
     // difference between clang and clang-cl.
+    //
+    // We prefix the information we need with `compiler_id` and `compiler_version`
+    // so that we can support compilers that insert pre-amble code even in `-E` mode
     let test = b"#if defined(__NVCC__) && !defined(_MSC_VER)
-nvcc
+compiler_id=nvcc
 #elif defined(__NVCC__) && defined(_MSC_VER)
-nvcc-msvc
+compiler_id=nvcc-msvc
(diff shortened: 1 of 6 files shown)


sylvestre:
plenty of new tests, nice :)

robertmaynard:
I am currently working on adding `nvc` and `nvc++` as supported compilers ( for C and C++ ) which also requires the changes to compiler type detection.

If needed I can separate the commit that changes those parts of `compiler.rs` to a new PR if needs a more in-depth review

src/compiler/compiler.rs
```diff
@@ -1082,32 +1083,39 @@ where
     // Both clang and clang-cl define _MSC_VER on Windows, so we first
     // check for MSVC, then check whether _MT is defined, which is the
     // difference between clang and clang-cl.
-    let test = b"#if defined(__NVCC__)
-nvcc
+    //
+    // We prefix the information we need with `compiler_id` and `compiler_version`
+    // so that we can support compilers that insert pre-amble code even in `-E` mode
+    let test = b"#if defined(__NVCC__) && defined(__NVCOMPILER)
```
miscco, line 1089: Would it make sense to move the code into a new line so that all the relevant stuff is at the same indentation level?
```suggestion
    let test = b"
#if defined(__NVCC__) && defined(__NVCOMPILER)
```

src/compiler/nvcc.rs
```diff
@@ -256,14 +274,34 @@ mod test {
...
+    }
+    fn parse_arguments_nvc(arguments: Vec<String>) -> CompilerArguments<ParsedArguments> {
         let arguments = arguments.iter().map(OsString::from).collect::<Vec<_>>();
-        Nvcc { version: None }.parse_arguments(&arguments, ".".as_ref())
+        Nvcc {
+            host_compiler: NvccHostCompiler::Nvhpc,
+            version: None,
+        }
+        .parse_arguments(&arguments, ".".as_ref())
     }
```
miscco, line 292: Shouldnt there also be a `parse_arguments_msvc` function?

src/compiler/compiler.rs
robertmaynard: The current layout is what `cargo fmt` requires.

robertmaynard: Originaly misread the suggestion. Yes we can do this now that we aren't parsing only the first line.

src/compiler/nvcc.rs
robertmaynard: Done

miscco:
@sylvestre it seems those CI failures are from infrastructure and not this PR.

Can we rerun the failing ones?

miscco (approved):
for what its worth, looks good to me

miscco:
Super happy to talk, sccache is essential for us to be able to run a CI on github PRs for https://github.com/NVIDIA/cccl

PR #1892, opened 2023-09-27
Add tests for clang as cuda compiler, and support the `-x cuda` syntax

robertmaynard:
While not offically documented, clang as a cuda compiler did work with sccache due to the existing support of `-x cu` as a general language flag. Unforunately this isn't sufficient for CMake based projects as it generates compile lines that use `-x cuda`. 

So I have relaxed the parsing of `-x` to include `cuda`, and backfilled the clang tests to verify cuda compilation.

Note: Cherry-picked `cb4ad6365f742df07927a0397e60b00db8faf535` from https://github.com/mozilla/sccache/pull/1889 so that I could expand the `test/system.rs` to verify clang++ as a cuda compiler.


illsilin:
What about support for -x hip?

PR #1893, opened 2023-09-28
compiler cache stats now display information on a per lang+compiler basis

robertmaynard:
So with the proposed expansion of clang as a CUDA compiler ( https://github.com/mozilla/sccache/pull/1892 ) I have noticed that the stats tracking for `sccache` binned the clang++ cuda compilations as a C/C++ invocation due to the presumption that the clang compiler meant `C/C++`.

So this PR breaks down stats to be tracked on a way finer granulatiry so that people can see exactly what compinations of language + compiler are causing hits or misses.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -441,6 +441,10 @@ where
     fn box_clone(&self) -> Box<dyn CompilerHasher<T>> {
         Box::new((*self).clone())
     }
+
+    fn language(&self) -> &'static str {
+        return self.parsed_args.language.as_str();
+    }
 }
 
 impl<I: CCompilerImpl> Compilation for CCompilation<I> {
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -101,13 +101,17 @@ pub enum CompilerKind {
 }
 
 impl CompilerKind {
-    pub fn lang_kind(&self) -> String {
+    pub fn lang_kind(&self, lang: &'static str) -> String {
+        let lang_and_compiler = lang.to_string();
         match self {
-            CompilerKind::C(CCompilerKind::Nvcc) => "CUDA",
-            CompilerKind::C(_) => "C/C++",
-            CompilerKind::Rust => "Rust",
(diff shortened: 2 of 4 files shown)


robertmaynard:
Previous sccache stat output would be:
```
Compile requests                     18
Compile requests executed            18
Cache hits                            9
Cache hits (C/C++)                    6
Cache hits (CUDA)                     3
Cache misses                          9
Cache misses (C/C++)                  6
Cache misses (CUDA)                   3
```

With this PR it would look like:
```
Compile requests                     18
Compile requests executed            18
Cache hits                            9
Cache hits (c [clang])                1
Cache hits (c [gcc])                  1
Cache hits (c [nvcc])                 1
Cache hits (c++ [clang])              1
Cache hits (c++ [gcc])                2
Cache hits (c++ [nvcc])               1
Cache hits (cuda [clang])             1
Cache hits (cuda [nvcc])              1
Cache misses                          9
Cache misses (c [clang])              1
Cache misses (c [gcc])                1
Cache misses (c [nvcc])               1
Cache misses (c++ [clang])            1
Cache misses (c++ [gcc])              2
Cache misses (c++ [nvcc])             1
Cache misses (cuda [clang])           1
Cache misses (cuda [nvcc])            1
```

This is an obvious worse case as I have used 3 different compilers, but I wanted to show the level of detail we now get

sylvestre:
this will break projects parsing the output or json.
maybe we should keep the original information and add your break down ?

robertmaynard:
> this will break projects parsing the output or json. maybe we should keep the original information and add your break down ?

That was what I was wondering about. 

Would it be better to guard the detailed stats behind a different command line switch ( `sccache -s -v` )? That would make it completly opt-in.

sylvestre:
yeah, that works too :)

robertmaynard:
I am going to close this PR in preference to more finer grained PR's that build up to more verbose stat tracking.

The first of those PRs is:  https://github.com/mozilla/sccache/pull/1895

PR #1895, opened 2023-10-02
sccache now tracks compiler hits/misses on a per language basis.

robertmaynard:
Previously sccache was actually tracking hits and misses on a compiler id basis, and guessing on the language.

This is an issue for compilers like `clang` which support C, C++, and CUDA. In that case we want to clearly state what language we are getting hit or misses on to help user diagnose issues.


robertmaynard:
To properly do this I hoisted the `Language` type out of c.rs and made it understand `Rust`. I am open to other suggestions on how we can track languages across `c.rs` and `rust.rs` if others don't like this approach.

PR #1897, opened 2023-10-07
Detect rustc(drivers) more precisely

sagudev:
Fixes https://github.com/mozilla/sccache/issues/861


sagudev:
Aarch fails on unrelated error (installation of musl).

Xuanwo:
> Aarch fails on unrelated error (installation of musl).

Yep. It's unrelated. I'm working on the fix. Please ignore it for now. Thanks for the patience.

src/compiler/compiler.rs
```diff
@@ -921,129 +944,137 @@ where
...
 
-        child.env_clear().envs(ref_env(env)).args(&["-vV"]);
+    child.env_clear().envs(ref_env(env)).args(&["-vV"]);
 
-        let rustc_vv = run_input_output(child, None).await.map(|output| {
-            if let Ok(stdout) = String::from_utf8(output.stdout.clone()) {
-                if stdout.starts_with("rustc ") {
-                    return Ok(stdout);
-                }
+    let rustc_vv = run_input_output(child, None).await.and_then(|output| {
```
sylvestre, line 988: maybe move this block into a function ? 
it is becoming a bit too long

src/compiler/compiler.rs
```diff
@@ -921,129 +944,137 @@ where
...
-        })?;
+        }
+        Err(ProcessError(output)).context("Failed to launch subprocess for compiler determination")
+    });
 
-        match rustc_vv {
-            Ok(rustc_verbose_version) => {
-                let rustc_executable2 = rustc_executable.clone();
+    // If it does not look like rustc and it's verification failed check for c compiler
+    match rustc_vv {
```
sylvestre, line 998: same, could you please try to move this into a function

sagudev: I moved both into one function as both are tightly coupled.

src/compiler/compiler.rs
```diff
@@ -921,129 +944,174 @@ where
...
-                }
+    if !must_be_rust && is_like_c_compiler(&rustc_executable) {
+        let executable = executable.to_owned();
+        let cc = detect_c_compiler(creator, executable, args, env.to_vec(), pool).await;
+        return cc.map(|c| (c, None));
+    }
+
+    // Even if it does not look like rustc like it might still be rustc driver
+    // so we do full check
+    match verified_rust_compiler(
```
sylvestre, line 977: if I understand correctly, this function is performing an action

So, it should probably called
verify_rust_compiler

src/compiler/compiler.rs
```diff
@@ -921,129 +944,174 @@ where
...
+            let executable = executable.to_owned();
+            let cc = detect_c_compiler(creator, executable, args, env.to_vec(), pool).await;
+            cc.map(|c| (c, None))
+        }
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
+/// If `must_be_rust` is `true` and verification of `rustc_executable` fails it returns error,
+/// if `must_be_rust` is `false` returns `None` if verification failed or `Some((Compiler, CompilerProxy))` if it succeeded.
```
sylvestre, line 1000: it would be nice in this comment to explicit what can be must_be_rust=false

src/compiler/compiler.rs
```diff
@@ -921,129 +944,176 @@ where
...
                 .await
-                .map(|c| {
+                .map(|(proxy, resolved_compiler_executable)| {
                     (
-                        Box::new(c) as Box<dyn Compiler<T>>,
-                        proxy as Option<Box<dyn CompilerProxy<T>>>,
+                        proxy
+                            .map(Box::new)
+                            .map(|x: Box<RustupProxy>| x as Box<dyn CompilerProxy<T>>),
+                        resolved_compiler_executable,
```
drahnr, line 1081: This became less readable than before

src/compiler/compiler.rs
```diff
@@ -907,6 +907,29 @@ fn is_rustc_like<P: AsRef<Path>>(p: P) -> bool {
...
+/// Returns true if the given path looks like a c compiler program
+///
+/// This does not check c compilers, it only report programs that are definitely not rustc
+fn is_like_c_compiler<P: AsRef<Path>>(p: P) -> bool {
+    matches!(
+        p.as_ref()
+            .file_stem()
+            .map(|s| s.to_string_lossy().to_lowercase())
+            .as_deref(),
+        Some("cc")
```
drahnr, line 919:
```suggestion
        Some("cc" | "c++" | .. )
```
should work as well

src/compiler/compiler.rs
```diff
@@ -921,129 +944,176 @@ where
...
-                Some(PathBuf::from(arg1))
-            } else {
-                None
+                must_be_rust = true;
+                rustc_executable = PathBuf::from(arg1)
             }
-        })
-    } else {
-        None
+        }
```
drahnr, line 965: I'd prefer to use a local `struct X {}` and avoid too many `mut` variables floating around in a scope.

drahnr:
I have yet to do a full review.

src/compiler/compiler.rs
sagudev: Are you doing per commit review as I cannot find this comment on full view?

sagudev:
I squashed all commits as they were not meant to be reviewed separately. I also updated `verify_rust_compiler` to not depend on `must_be_rust` and moved this logic back to `detect_compiler`

src/compiler/compiler.rs
drahnr: I am doing a full diff view, and the changeset exists.

src/compiler/compiler.rs
```diff
@@ -921,129 +938,166 @@ where
...
+    }
+}
 
-        match rustc_vv {
-            Ok(rustc_verbose_version) => {
-                let rustc_executable2 = rustc_executable.clone();
+/// Tries to verify that provided executable is really rust compiler
+/// or rust driver
+async fn verify_rust_compiler<T>(
```
drahnr, line 997: Requires more documentation on the disriminator used and the function name needs a better name.

sagudev: What is wrong with name? Main part of this function is:
```rust
// rustc -vV verification status
match rustc_vv {
   ...
```
anything specific you have in mind?

I also do not think there is much to document about discriminator used as there is no discriminator used. We check for rust compiler using -vV instead of using binary name and then verifying compiler with -vV.

drahnr: It does attempt to _resolve_ a given path to the true (`rustc`) compiler binary with a failure mode if it cannot, so verification is a red hering and doesn't describe what it is doing. That's my main issue

sagudev: What about `resolve_rust_compiler`?

sagudev:
rebased

drahnr:
I'll do a full review as soon as all CI steps went green

sagudev:
Checks are green in my fork: https://github.com/sagudev/sccache/actions/runs/7656273169 & https://github.com/sagudev/sccache/actions/runs/7656273174

src/compiler/compiler.rs
```diff
@@ -980,6 +980,23 @@ fn is_rustc_like<P: AsRef<Path>>(p: P) -> bool {
...
+        p.as_ref()
+            .file_stem()
+            .map(|s| s.to_string_lossy().to_lowercase())
+            .as_deref(),
+        Some("cc" | "c++")
+            | Some("gcc" | "g++")
+            | Some("clang" | "clang++")
+            | Some("clang-cl" | "cl")
+            | Some("nvc" | "nvc++" | "nvcc")
+    )
```
drahnr, line 997:
```suggestion
        Some("cc" | "c++"
            | "gcc" | "g++"
            | "clang" | "clang++"
            | "clang-cl" | "cl"
            | "nvc" | "nvc++" | "nvcc")
    )
```

src/compiler/compiler.rs
```diff
@@ -994,129 +1011,166 @@ where
...
-        }
-
-        child.env_clear().envs(ref_env(env)).args(&["-vV"]);
-
-        let rustc_vv = run_input_output(child, None).await.map(|output| {
-            if let Ok(stdout) = String::from_utf8(output.stdout.clone()) {
-                if stdout.starts_with("rustc ") {
-                    return Ok(stdout);
-                }
+    if !must_be_rust && is_like_c_compiler(&rustc_executable) {
```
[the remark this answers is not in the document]
sagudev: This is to be compatible with current behavior. If name of compiler is rustc and we cannot detect it we fast fail (we do not try it for c compiler).

src/compiler/compiler.rs
```diff
@@ -1399,6 +1453,14 @@ mod test {
...
         let runtime = single_threaded_runtime();
         let pool = runtime.handle();
+        next_command(
+            &creator,
+            Ok(MockChild::new(
+                exit_status(1),
+                "",
+                "gcc: error: unrecognized command-line option '-vV'; did you mean '-v'?",
+            )),
+        );
```
drahnr, line 1463: Iirc, the result will be cached, so the impact of another resolution call is very small, even for small projects. Correct me if I am wrong.

sagudev: Is this comment to https://github.com/mozilla/sccache/pull/1897#discussion_r1466719055?

drahnr (changes requested):
I don't particularly like the use mutable `bool` variables in deciding the compiler. There is an outstanding question on the efficiency of test runs and ifwe can assume `-vV` is unique to rust.

src/compiler/compiler.rs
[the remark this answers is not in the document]
drahnr: That may be, but it can be expressed more concisely.

src/compiler/compiler.rs
drahnr: This is a principled question, if we incur much overhead for the compiler resolution with this changeset.

sagudev:
>  if we can assume `-vV` is unique to rust.

Even if it is not we try to parse the result and if it does not contain rustc(-drivers) specific output we know it is not rusty.

src/compiler/compiler.rs
sagudev: I am not sure, to be honest. But if the results are cached then it `should` be ok.

src/compiler/compiler.rs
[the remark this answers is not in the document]
sagudev: So, should I just change behavior and simply code or express it differently (currently have no idea how)?

sylvestre:
moved into a different PR

PR #1904, opened 2023-10-18
Add `show-adv-stats` flag which shows per compiler hits/misses

robertmaynard:
Adds the `-show-adv-stats` command line option to sccache which will show more details on the compiler hits/misses for each language and compiler.

Example of using sccache with clang, g++, and nvcc with `C`, `C++`, and `CUDA` sources:
```
/home/rmaynard/Work/sccache/target/debug/sccache --show-adv-stats
Compile requests                     18
Compile requests executed            18
Cache hits                            9
Cache hits (c [clang])                1
Cache hits (c [gcc])                  1
Cache hits (c [nvcc])                 1
Cache hits (c++ [clang])              1
Cache hits (c++ [gcc])                2
Cache hits (c++ [nvcc])               1
Cache hits (cuda [clang])             1
Cache hits (cuda [nvcc])              1
Cache misses                          9
Cache misses (c [clang])              1
Cache misses (c [gcc])                1
Cache misses (c [nvcc])               1
Cache misses (c++ [clang])            1
Cache misses (c++ [gcc])              2
Cache misses (c++ [nvcc])             1
Cache misses (cuda [clang])           1
Cache misses (cuda [nvcc])            1
Average compiler                  0.101 s
Cache location                  Local disk: "/home/rmaynard/.cache/sccache"
Version (client)                0.5.4
Cache size                            5 KiB
Max cache size                       32 GiB
```


sylvestre:
LGTM

Maybe you could test this into some integration-tests ?

robertmaynard:
> LGTM
> 
> Maybe you could test this into some integration-tests ?

I will look into extending `tests/system.rs` to validate the advanced stats. IIRC the primary sticking point was that I was hesitant to re-implement the CompilerKind mapping of compiler names to nice strings. That will be needed to verify the hits/misses are in the correct bucket.

Thinking about it more, the `tests/system.rs` has a subset of compilers so the mapping should be easy to implement

robertmaynard:
@sylvestre Extended some of the `tests/system.rs` to verify the advanced keys. Happy to iterate a better design / test for this if needed

sylvestre:
thanks :)

PR #1905, opened 2023-10-18
tests/system.rs Now correctly identifies apple clang

robertmaynard:
Previous regex didn't have proper capture group names, and the regex didn't properly handle for spaces between `Apple` and `clang version`. Updated to also support compiler output like `Apple LLVM clang version` which I believe it used to do.


tests/system.rs
```diff
@@ -434,8 +434,8 @@ fn run_sccache_command_tests(compiler: Compiler, tempdir: &Path) {
             Err(e) => panic!("Invalid UTF-8 sequence: {}", e),
         };
 
-        // Apple clang would match "Apple LLVM version"
-        let re = Regex::new(r"(?P<apply>Apple+*)?clang version (?P<major>\d+)").unwrap();
```
sylvestre, line 438: apply apple, cute typo :)

Issue #1899, opened 2023-10-13
Seeing reason "multiple input files" although only a single file is given


exoosh:
In the log file I see this when attempting to use `sccache` on OpenSSL (1.1.1w):

> parse_arguments: CannotCache(multiple input files, ["crypto\\asn1\\x_sig.c"]): ["/MT", "/Zl", "/Gs0", "/GF", "/Gy", "/W3", "/wd4090", "/nologo", "/O2", "/I", ".", "/I", "include", "-DL_ENDIAN", "-DOPENSSL_PIC", "-DOPENSSL_CPUID_OBJ", "-DOPENSSL_BN_ASM_PART_WORDS", "-DOPENSSL_IA32_SSE2", "-DOPENSSL_BN_ASM_MONT", "-DOPENSSL_BN_ASM_GF2m", "-DSHA1_ASM", "-DSHA256_ASM", "-DSHA512_ASM", "-DRC4_ASM", "-DMD5_ASM", "-DRMD160_ASM", "-DAESNI_ASM", "-DVPAES_ASM", "-DWHIRLPOOL_ASM", "-DGHASH_ASM", "-DECP_NISTZ256_ASM", "-DPOLY1305_ASM", "-DOPENSSLDIR=\"C:\\\\Program Files (x86)\\\\Common Files\\\\SSL\"", "-DENGINESDIR=\"C:\\\\Program Files (x86)\\\\OpenSSL\\\\lib\\\\engines-1_1\"", "-DOPENSSL_SYS_WIN32", "-DWIN32_LEAN_AND_MEAN", "-DUNICODE", "-D_UNICODE", "-D_CRT_SECURE_NO_DEPRECATE", "-D_WINSOCK_DEPRECATED_NO_WARNINGS", "-DNDEBUG", "-DOPENSSL_API_COMPAT=0x10100000L", "-c", "/Focrypto\\asn1\\x_sig.obj", "crypto\\asn1\\x_sig.c"]

As you can see `/Fo` agrees with the name passed for the translation unit. Why is it not cacheable then?

assarbad:
Turns out the reason "multiple input files" is **totally misleading**, unfortunately.

Instead the switches `/Zl` (that's a small L) to omit the default library name and `/d1trimfile:` to trim the leading path for the `__FILE__` and aid in reproducibility was the blocker for this to cache nicely!

[...]

PR #1907, opened 2023-10-18
Adding /Zl and /d1trimfile for MSVC compiler (addresses #1899)

assarbad:
- `/Zl` is used to omit the default library name (given a specified runtime, such as with `/MT`)
- `/d1trimfile:` is used to trim the leading path whereever the `__FILE__` macro is used

------

Please provide some guidance in case adjustments are required. This addresses part of #1899. I don't know why it shows "multiple input files", however. It's very misleading.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -458,6 +458,7 @@ msvc_args!(static ARGS: [ArgInfo<ArgData>; _] = [
     msvc_take_arg!("Zc:", OsString, Concatenated, PassThroughWithSuffix),
     msvc_flag!("Ze", PassThrough),
     msvc_flag!("Zi", DebugInfo),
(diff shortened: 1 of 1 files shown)


sylvestre:
would it be possible to have a test to reproduce this issue ?

assarbad:
> would it be possible to have a test to reproduce this issue ?

@sylvestre Could you elaborate, please? Are you looking for a test case that invokes `cl.exe` with the switches in question and compares with the output _without_ those switches or do you mean to amend some unit tests to test the newly introduced handling?

If the latter, I can give it a shot. If the former, you'd have to clarify what it is you're looking for exactly.

sylvestre:
the CI isn't happy with your changes :)
could you please fix it?

PR #1908, opened 2023-10-19
feat(s3): add server side encryption

nebnes:
Add server side encryption with s3

via env
SCCACHE_S3_SERVER_SIDE_ENCRYPTION=true

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -52,6 +52,7 @@ bucket = "name"
 endpoint = "s3-us-east-1.amazonaws.com"
 use_ssl = true
 key_prefix = "s3prefix"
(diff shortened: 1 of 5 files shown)


docs/Configuration.md
```diff
@@ -52,6 +52,7 @@ bucket = "name"
 endpoint = "s3-us-east-1.amazonaws.com"
 use_ssl = true
 key_prefix = "s3prefix"
+server_side_encription = false
```
sylvestre, line 55:
```suggestion
server_side_encryption = false
```

sylvestre:
@Xuanwo what do you think about this ? :)
it looks good to me

Issue #1912, opened 2023-10-23
reqsign package has config bug


dreum:
Looking through the source of reqsign, the default config location for aws has an errant \` in it.
This was fixed recently: https://github.com/Xuanwo/reqsign/commit/315e33deef2b6225835a0de1ef1e2eda35b016e8#diff-153d9f64f6ad207424cc6f8518627cd263275d2ac3b04f5707c18b308165e987R92

I would have used the AWS_CONFIG_FILE environment variable in order to override this, but it doesn't appear that opendal ever call the .from_env function, so the env variable is ignored.

I'm asking to update opendal to 0.40 and/or reqsign to 0.14.2 to correct this issue.
https://github.com/mozilla/sccache/blob/799d8f9aebd3e4ed5e498ca004ace8360499ed80/Cargo.toml#L63

Xuanwo:
Sccache has upgraded to OpenDAL 0.49 recently.

dreum:
Missed that this was already updated. closing...

PR #1913, opened 2023-10-23
update opendal and reqsign

dreum:
Fixing issue:
https://github.com/mozilla/sccache/issues/1912


dreum:
I missed that opendal already got updated as I was looking at the code at the latest release tag. My apologies.

PR #1916, opened 2023-10-24
sccache 0.6.0

sylvestre:
bumping the version because of the recent changes from Nvidia by @robertmaynard and @miscco

I will probably make a version 1.0.0 once https://github.com/mozilla/sccache/pull/1882 lands (cc @Alphare )


Alphare:
@sylvestre Thanks for the ping. I'd probably want people to try out #1882 for a while before making 1.0 though, but that was probably your plan?

PR #1936, opened 2023-11-04
Document how to set the direct access

sylvestre:
@Alphare ok with you?


Alphare:
Yes, sorry for the oversight. :)

Issue #1941, opened 2023-11-06
SCCACHE_DIRECT=true without explicit local directory set should work


sylvestre:
In https://github.com/mozilla/sccache/blob/main/src/config.rs#L719
if we set `SCCACHE_DIRECT=true` but not `disk_dir` or `disk_sz`, we end up in "None" option
which sets `use_preprocessor_cache_mode` as false

but if we run:
```
./target/debug/sccache --start-server
./target/debug/sccache --show-stats 
```

we can see that it is using a "Local disk" as default.

Therefor, we should be able to start:

```
SCCACHE_DIRECT=true ./target/debug/sccache  --start-server
./target/debug/sccache --show-stats
```

PR #1942, opened 2023-11-06
Allow using preprocessor mode from env without a disk config

Alphare:
If we set `SCCACHE_DIRECT=true` but not `disk_dir` or `disk_sz`, we end up with `None` which sets `use_preprocessor_cache_mode` to `false`.

But if we run:
```
./target/debug/sccache --start-server
./target/debug/sccache --show-stats
```

We can see that it is uses the local disk by default.

After this change, this works:
`SCCACHE_DIRECT=true ./target/debug/sccache  --start-server ./target/debug/sccache --show-stats`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -710,13 +710,20 @@ fn config_from_env() -> Result<EnvConfig> {
         .and_then(|v| parse_size(&v));
 
     let mut preprocessor_mode_config = PreprocessorCacheModeConfig::default();
-    match env::var("SCCACHE_DIRECT").as_deref() {
-        Ok("on") | Ok("true") => preprocessor_mode_config.use_preprocessor_cache_mode = true,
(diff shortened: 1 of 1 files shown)


Alphare:
@sylvestre this should fix #1941 but highlighted a deeper issue with the config system.

IIUC the config from env is merged with the config file's at a coarse granularity (like `azure`, `disk`, etc.).
The user can then get into a situation where they've defined a `disk_dir` and `disk_sz` in the config file, but only set `use_preprocessor_cache_mode` with the environment variable: this means that the two first settings will be ignored by sccache.

Is this in scope for a fix? I think it might be a bit involved since we'd need to define the merge logic either manually for every field (which is brittle, not future-proof and prone to errors), or use some form of reflection to loop through the fields.

sylvestre:
looks like you didn't rebase :)

sylvestre:
if you have time, yeah, it would be nice to add some extra checks

PR #1971, opened 2023-11-20
Maintain order when parsing clang arguments for MSVC

### LINKED ISSUE #1498: sccache doesn't handle "--"
https://github.com/rust-lang/cc-rs/commit/fcb8df5efa81a90115c5bc05c8c0a7c615c8afd9 added "--" to the command line emitted for clang-cl when compiling through the cc-rs crate, and sccache doesn't handle it, so it passes it to the preprocessor without considering its effect, and every argument that comes after it on the preprocessor command line is considered as a file name, which, when they are originally flags, well, don't match any actual file, and then the preprocessor fails.

### LINKED ISSUE #1559: Issue with `-Xclang` and `--` compiling using ninja
`C:\Users\scoop\shims\sccache.exe C:\dev\build_with_clang_relwithdbginfo\bin\clang-cl.exe  /nologo -Xclang -fno-rtti -flto -c -- C:\dev\Something.cpp`
The above command yields `clang-cl: fatal error: no such file or directory: '-Xclang'; did you mean '-Xclang'?`. The problem goes away if I remove the `--`. The example line is generated using ninja and I believe the double dash signifies the end of command options.

I'm thinking it might has something todo with how sccache rearranges the arguments after parsing them and somehow put it after "--". I had tried to use "/clang:-fno-rtti" which is the same as -Xclang but "/clang:" is not currently supported by sccache.

mathewhodson:
Fixes #1498
Fixes #1559

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -503,11 +503,79 @@ pub fn parse_arguments(
     let mut pdb = None;
     let mut depfile = None;
     let mut show_includes = false;
+    let mut has_pending_clang_arguments = false;
     let mut xclangs: Vec<OsString> = vec![];
     let mut clangs: Vec<OsString> = vec![];
     let mut profile_generate = false;
     let mut multiple_input = false;
     let mut multiple_input_files = Vec::new();
+
+    macro_rules! parse_clang_arguments {
+        ($xclangs:expr, $clangs:expr) => {
+            for (args, append_fn) in Iterator::zip(
+                [$xclangs, $clangs].iter(),
+                &[xclang_append, dash_clang_append],
+            ) {
+                let it = gcc::ExpandIncludeFile::new(cwd, args);
+                for arg in ArgsIter::new(it, (&gcc::ARGS[..], &clang::ARGS[..])) {
+                    let arg = try_or_cannot_cache!(arg, "argument parse");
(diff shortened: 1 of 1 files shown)


src/compiler/msvc.rs
```diff
@@ -503,18 +503,94 @@ pub fn parse_arguments(
...
     let mut depfile = None;
     let mut show_includes = false;
+    let mut has_pending_clang_arguments = false;
     let mut xclangs: Vec<OsString> = vec![];
     let mut clangs: Vec<OsString> = vec![];
     let mut profile_generate = false;
     let mut multiple_input = false;
     let mut multiple_input_files = Vec::new();
 
+    macro_rules! parse_clang_arguments {
```
sylvestre, line 513: why is it a macro and not a function? thanks

mathewhodson: Because this code uses the `try_or_cannot_cache!` and `cannot_cache!` macros that are intended to make `parse_arguments` return early, it cannot easily be refactored into a function.

glandium (changes requested):
This actually doesn't fully address #1498, because clang too accepts "--". I have work in progress to fix this once and for all.

glandium:
Closing in favor of #1983

PR #1973, opened 2023-11-20
Only activate direct mode by default for local

Alphare:
45386750a48b6ed7daf9019db88a22dc7a7361c5 reverted my initial change.

This re-enables direct mode by default, making sure only the disk cache is affected. There may still be an issue with Windows, but details are sparse and no reproduction has been found yet.

@sylvestre

### DIFF
--- a/docs/Local.md
+++ b/docs/Local.md
@@ -30,7 +30,7 @@ Preprocessor cache mode will be disabled if any of the following holds:
 
 Configuration options and their default values:
 
-- `use_preprocessor_cache_mode`: `false`. Whether to use preprocessor cache mode entirely.
+- `use_preprocessor_cache_mode`: `true`. Whether to use preprocessor cache mode entirely.
 - `file_stat_matches`: `false`. If false, only compare header files by hashing their contents. If true, will use size + ctime + mtime to check whether a file has changed. See other flags below for more control over this behavior.
 - `use_ctime_for_stat`: `true`. If true, uses the ctime (file status change on UNIX, creation time on Windows) to check that a file has/hasn't changed. Can be useful to disable when backdating modification times in a controlled manner.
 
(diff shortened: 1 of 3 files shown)


glandium:
Rustfmt is not happy

glandium:
For the record, I can confirm that this solves the issue of direct mode affecting non-local builds. This, however doesn't fix the "getting stuck on Windows" part.

I'm not sure, though, why SCCACHE_DIRECT=false wasn't fixing the issue... initialization order?

Alphare:
> For the record, I can confirm that this solves the issue of direct mode affecting non-local builds. This, however doesn't fix the "getting stuck on Windows" part.

It will probably be a bit more involved of a bug hunt. I have not yet started to look at it, my end of November is going to be *very* full.

> I'm not sure, though, why SCCACHE_DIRECT=false wasn't fixing the issue... initialization order?

Do you mean it wasn't disabling it for local config? Or for all configs? Because if the former, we may have another bug, but if the latter, this commit also fixed this behavior. Basically the config was set to `true` for all other `Storage` impls no matter the config... oops!

glandium:
> Do you mean it wasn't disabling it for local config? Or for all configs? 

For non-local configs. I hadn't tried for local configs. But the builds that were hitting the "gets stuck" problem on Windows despite the non-local use weren't fixed by setting SCCACHE_DIRECT=false.

As for the "gets stuck" problem itself, I am able to reproduce it... but not locally, only on Firefox CI, which is not going to be super helpful to debug. I'll try a few things, though.

glandium:
Note the tests are failing with this patch.

glandium:
> After fixing the previous failure, it looks like https://github.com/mozilla/sccache/actions/runs/6954981313/job/18922864423?pr=1973 fails, but I'm not convinced it's got anything to do with my patch.

It's a race condition: filed #1982.

PR #1984, opened 2023-11-23
Fix test_parse_arguments_multiarch_cache_disabled

### LINKED ISSUE #1982: compiler::gcc::test::test_parse_arguments_multiarch_cache_disabled fails intermittently
The result of that test depends on the SCCACHE_CACHE_MULTIARCH environment variable. Ignoring the fact that setting the environment variable outside of cargo test makes the test permanently fail, there's another problem: other tests are temporarily setting SCCACHE_CACHE_MULTIARCH, and tests run in parallel. So there's a race condition: if test_parse_arguments_multiarch_cache_disabled runs at the same time as some of these other tests, it fails, otherwise, it passes.

glandium:
The test fails when SCCACHE_CACHE_MULTIARCH is set, which can either happen externally, or when another tests sets it via with_var.

with_var_unset prevents both (temp_env has a global lock specifically to avoid the problem with concurrent tests).

Fixes #1982

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -931,7 +931,7 @@ mod test {
     use crate::mock_command::*;
     use crate::test::utils::*;
 
-    use temp_env::with_var;
+    use temp_env::{with_var, with_var_unset};
 
     fn parse_arguments_(
         arguments: Vec<String>,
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
Thanks!

PR #1992, opened 2023-11-29
chore: Version bump opendal 0.41.0 -> 0.43.0

vincentdephily:
Remove the `opendal/services-redis-rustls` feature as it's no longer available/needed.

Obsoletes #1957

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,11 +280,12 @@ checksum = "a2bd12c1caf447e69cd4528f47f94d203fd2582878ecb9e9465484c4148a8223"
 
 [[package]]
 name = "cc"
-version = "1.0.79"
+version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
+checksum = "f1174fb0b6ec23863f8b971027804a42614e347eafb0a95bf0b12cdae21fc4d0"
 dependencies = [
  "jobserver",
+ "libc",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


sylvestre:
A bunch of tasks are failing here

vincentdephily:
I managed to reproduce the errors by building with rust 1.71, and they go away with 1.72 and higher.

What puzzles me is that the CI seems to be using 1.67.1, but I get a bunch of different errors with 1.67-1.72, including fairly basic things like the clap crates crates in our Cargo.toml having an MSRV of 1.70. Does the CI really manages to do some builds with 1.67 ? Is it resolving minimal dependencies maybe ?

What's sccache's MSRV update policy ? Note that working with anything older than 1.70 has become a pain, due to fetching crate metadata using git instead of sparse registry.

vincentdephily:
Sorry, ignore my previous observation about failures with 1.67-1.70, I had locally updated all deps by mistake.

The build failure with my branch does get resolved with 1.72, but that's a more isolated problem.

vincentdephily:
Fix proposed upstream in Opendal. For this MR we'll probably want to wait for the next Opendal release.

sylvestre:
the maintainer of opendal is also a maintainer here :)
So, should not be hard ;)
cc @Xuanwo ;)

Xuanwo:
> Fix proposed upstream in Opendal. For this MR we'll probably want to wait for the next Opendal release.

Let's coordinate this.

Xuanwo:
OpenDAL v0.43.0 is currently up for voting and we anticipate its availability by next week.

vincentdephily:
Rebased on `main`, and now upgrading to `0.43.0` instead, along with a few dependencies.

vincentdephily:
Gah, I had done the necessary `ordered-multimap` downgrade locally but forgot to amend my commit, sorry. Should be fixed.

Xuanwo (approved):
LGTM!

PR #2007, opened 2024-01-04
Expand cfg for operator storage impl

jdygert-spok:
Previously, trying to build with only e.g. `gha` feature enabled would not compile.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,9 @@ jobs:
       matrix:
         feature: [azure, gcs, gha, memcached, redis, s3, webdav]
     steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


sylvestre:
could you please add a github job to check we don't regress in the future? thanks

jdygert-spok:
Sure, which would you rather?

1. Add single feature variants to existing build matrix
2. Add job that runs `cargo check` with each single feature variant manually
3. Use `cargo-all-features` to run `check-all-features` on each combination of features

jdygert-spok:
Seems like maybe 2 makes the most sense here.

sylvestre:
sounds good, the matrix github action system should be good enough :)

sylvestre:
I guess you saw
```

Run cargo check --no-default-features --feature gcs
error: unexpected argument '--feature' found
```

PR #2010, opened 2024-01-06
chore(src/compiler/msvc): replace rust-encoding with encoding-rs

liushuyu:
This pull request replaces `rust-encoding` with `encoding-rs`. 

The former is unmaintained and uses ambiguous encoding for `ISO 8859-1` (note that `CP1252` aka "Latin-1" on Windows(R) is a superset of `ISO 8859-1`). This pull request also corrects this issue by using the `WINDOWS-1252` encoding to match the encoding more closely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -658,70 +658,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
 
 [[package]]
-name = "encoding"
-version = "0.2.33"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b0d943856b990d12d3b55b359144ff341533e516d94098b1d3fc1ac666d36ec"
-dependencies = [
- "encoding-index-japanese",
- "encoding-index-korean",
- "encoding-index-simpchinese",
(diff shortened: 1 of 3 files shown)


Xuanwo:
Please rebase with main branch to remove unrelated commits.

liushuyu:
> Please rebase with main branch to remove unrelated commits.

Done. Rebased on top of the `main` branch

liushuyu:
No, there are some differences I failed to account for when switching the implementations.

liushuyu:
Fixed. I also added one more test case to test the decoding logic more thoroughly.

liushuyu:
CI tests passed. The coverage misses are due to negative code-path not being reached (should I add tests to test invalid encodings?)

xen0n:
> CI tests passed. The coverage misses are due to negative code-path not being reached (should I add tests to test invalid encodings?)

In general, negative test cases are also, if not more, beneficial, as arguably the error-handling codepath is more valuable regarding UX and test coverage... and in this case I think we want to be robust against inputs that definitely can contain invalid encoding, and can be controlled by the user.

sylvestre:
do you know if it fixes any issue ?

liushuyu:
> do you know if it fixes any issue ?

In theory, any valid Windows 1252 but invalid ISO 8859-1 code points (e.g. file names) should work now.

PR #2011, opened 2024-01-06
Snap: Adapt Snapcraft manifest to Snapcraft 8.0

liushuyu:
This pull request adapts the Snapcraft build manifest to the new Snapcraft tool version 8 changes.

This also includes a change to add the `sccache-dist` program to the Snap package.

> [!NOTE]
> If you want the user to use the `sccache-dist` program (instead of using the auto-renamed `sccache.sccache-dist` binary) from Snap, an alias request needs to be filed on the [Snapcraft forum](https://forum.snapcraft.io/c/store-requests/19). Please see <https://forum.snapcraft.io/t/process-for-aliases-auto-connections-and-tracks/455> for more information on the process.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -29,6 +29,8 @@ confinement: classic
 apps:
     sccache:
         command: bin/sccache
+    sccache-dist:
+        command: bin/sccache-dist
 
 parts:
     sccache:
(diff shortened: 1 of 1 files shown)


liushuyu:
I also noticed due to limitations of earlier versions of `ring` and `rustls`, `sccache` won't be able to build or run on CPU architectures other than x86 or ARM.

I will open another PR to attempt to upgrade the dependencies that pull in the older versions of those crates.

sylvestre:
thanks :)

PR #2013, opened 2024-01-06
feat(tree-wide): migrate to Hyper 1.x

liushuyu:
This pull request migrates the `hyper` crate to version 1.1.

I have to admit the situation is a bit disastrous due to the massive amount of API changes they have accumulated.

The following is a summary of the changes in this pull request:

- Due to `hyper` splitting the library. Some more crates were added to accommodate and retain the functionalities.
- Since `reqwest` (as of writing) still uses `hyper` 0.14, dependencies can not be shared anymore.
- `hyper` now requires the library user to provide the I/O interface, so you will find I have added an adapter/wrapper to avoid refactoring all the web-server-related functions to be `async`.
- The `make_service` macro is removed because the `make_service_fn` function was deleted. Instead, I added a taping-over function to stitch the services together like before.
- The shutdown handling is changed as `hyper` now delegates some event-loop responsibilities to the library user.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -964,7 +964,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "futures-util",
- "http",
+ "http 0.2.11",
  "indexmap",
  "slab",
  "tokio",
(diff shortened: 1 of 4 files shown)


Xuanwo:
The entire Rust ecosystem has undergone significant changes since that code was written. We're still using non-async crates like `rouille` and working directly with hyper, which increases our maintenance workload. Do you think it's a good opportunity for us to switch to axum or another framework?

liushuyu:
> The entire Rust ecosystem has undergone significant changes since that code was written. We're still using non-async crates like `rouille` and working directly with hyper, which increases our maintenance workload. Do you think it's a good opportunity for us to switch to axum or another framework?

`axum` is currently unstable (in terms of API). Other service frameworks within the direct `tokio` ecosystem are still trying to migrate to `hyper` v1.

So, I am unsure which approach is better given the current situation. The main issue as of now is still the architectural support. Ironically, non-mainstream devices usually need the caching the most (due to limited computing power), but `sccache` can't provide service for them.

Xuanwo (approved):
Thanks!

PR #2027, opened 2024-01-11
feat(redis): add support for ttl, fix #1666

### LINKED ISSUE #1666: Allow Setting Redis Key Experation
### Discussed in https://github.com/mozilla/sccache/discussions/1665

<div type='discussions-op-text'>

<sup>Originally posted by **jretterer-prol** March 18, 2023</sup>
In using sccache with redis along with the allkeys-lru eviction policy I was seeing keys being removed from redis even thought the allocated memory space was not fully utilized. That led me to check the TTL on the redis keys and found one is set for sccache keys.

```
127.0.0.1:6379> TTL c/7/8/c78004b4adb30628728df15ce1e99e9f862814aaa9473c1582b5b996e823439c
(integer) 86258
```

After finding that I tried to find a way to adjust this but I can't seam to find it.  So, is there a way to adjust this expiry time for redis keys or even better, set it so the time is refreshed when keys are read?</div>

sudhackar:
Fixes #1666 

Adds a ttl to `opendal` if `SCCACHE_REDIS_TTL` env var is set and not zero.

### DIFF
--- a/docs/Redis.md
+++ b/docs/Redis.md
@@ -3,3 +3,5 @@
 Set `SCCACHE_REDIS` to a [Redis](https://redis.io/) url in format `redis://[:<passwd>@]<hostname>[:port][/<db>]` to store the cache in a Redis instance. Redis can be configured as a LRU (least recently used) cache with a fixed maximum cache size. Set `maxmemory` and `maxmemory-policy` according to the [Redis documentation](https://redis.io/topics/lru-cache). The `allkeys-lru` policy which discards the *least recently accessed or modified* key fits well for the sccache use case.
 
 Redis over TLS is supported. Use the [`rediss://`](https://www.iana.org/assignments/uri-schemes/prov/rediss) url scheme (note `rediss` vs `redis`). Append `#insecure` the the url to disable hostname verification and accept self-signed certificates (dangerous!). Note that this also disables [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication).
+
+Set `SCCACHE_REDIS_TTL` in seconds if you don't want your keys to live forever. This will override the default behavior of redis.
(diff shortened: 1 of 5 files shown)


Xuanwo (approved):
Mostly LGTM, thanks!

docs/Redis.md
```diff
@@ -3,3 +3,5 @@
 Set `SCCACHE_REDIS` to a [Redis](https://redis.io/) url in format `redis://[:<passwd>@]<hostname>[:port][/<db>]` to store the cache in a Redis instance. Redis can be configured as a LRU (least recently used) cache with a fixed maximum cache size. Set `maxmemory` and `maxmemory-policy` according to the [Redis documentation](https://redis.io/topics/lru-cache). The `allkeys-lru` policy which discards the *least recently accessed or modified* key fits well for the sccache use case.
 
 Redis over TLS is supported. Use the [`rediss://`](https://www.iana.org/assignments/uri-schemes/prov/rediss) url scheme (note `rediss` vs `redis`). Append `#insecure` the the url to disable hostname verification and accept self-signed certificates (dangerous!). Note that this also disables [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication).
+
+Set `SCCACHE_REDIS_TTL` in seconds if you don't want your keys to live forever. This will override the default behavior of redis.
```
sylvestre, line 7:
```suggestion
Set `SCCACHE_REDIS_TTL` in seconds if you don't want your cache to live forever. This will override the default behavior of redis.
```

src/config.rs
```diff
@@ -223,10 +223,20 @@ pub struct MemcachedCacheConfig {
     pub expiration: u32,
 }
 
+/// redis has no default TTL - all keys live forever
```
sylvestre, line 226:
```suggestion
/// redis has no default TTL - all caches live forever
```

PR #2030, opened 2024-01-12
Add support for symlinks in toolchains.

### LINKED ISSUE #2028: Symlinks in the sysroot can prevent auto-packaged toolchains from working correctly
I'm using [void linux](https://voidlinux.org). This distro has symlinks for `/lib64` and `/usr/lib64` which point to `/usr/lib`, and (importantly) the dynamic linker uses `/usr/lib64` as the system search path. `ldd` (used to determine the files to package in the toolchain) reports `/usr/lib64/ld-linux-x86_64.so.2` for the linker but (for whatever reason) `/usr/lib/libfoo.so` for dynamic libraries. This causes an issue because the packaged toolchain places the dynamic libraries in `/usr/lib`, but the dynamic linker only searches in `/usr/lib64` (which won't contain them because no symlinks are preserved). When trying to run the toolchain, it fails with errors that the dynamic libraries could not be found.

afranchuk:
Closes #2028.

### DIFF
--- a/src/dist/pkg.rs
+++ b/src/dist/pkg.rs
@@ -53,7 +53,7 @@ mod toolchain_imp {
 
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 mod toolchain_imp {
-    use super::tarify_path;
+    use super::SimplifyPath;
     use fs_err as fs;
     use std::collections::BTreeMap;
     use std::io::{Read, Write};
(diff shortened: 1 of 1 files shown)


sylvestre:
@afranchuk nice to see you here :)

sylvestre:
would it be possible to add a test ?

afranchuk:
> would it be possible to add a test ?

The thought crossed my mind, but superficially it seems like it'd be a bit tricky. There are no existing tests for any of the packaging, so I'd have to build the boilerplate for mocking an existing filesystem from which the toolchain is created. But if you'd like to see some tests I can be the first to fill that gap!

PR #2039, opened 2024-01-16
fix(src/compiler/c): take common_args into account during preprocessor hashing

### LINKED ISSUE #2038: common_args is not included in preprocessor caching
Repro file:
```cpp
#include <bits/stdc++.h>

using namespace std;

int main()
{
#ifdef HELLO
	cout << "Hello World" << endl;
#else
	cout << "Hi World" << endl;
#endif
}
```
Run `sccache gcc -c main.cpp -DHELLO`. Then run `sccache gcc -c main.cpp`. You will see that the second command registers a cache hit when it clearly shouldn't.

This happens because as shown [here](https://github.com/mozilla/sccache/blob/0ef1a54028e252bb52b8901ad95b773327216c47/src/compiler/c.rs#L305), `preprocessor_and_arch_args` doesn't take `common_args` into account, and since `-DHELLO` belongs to `common_args`, removing it doesn't affect the hash of `preprocessor_and_arch_args`, hence registering a cache hit.

GZGavinZhao:
Fixes #2038.

Since `common_args` represents the flags that are used both for compilation and for preprocessing, during preprocessor caching `common_args` should be taken into account as well.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -303,6 +303,8 @@ where
         // use in creating a hash key
         let mut preprocessor_and_arch_args = parsed_args.preprocessor_args.clone();
         preprocessor_and_arch_args.extend(parsed_args.arch_args.to_vec());
+        // common_args is used in preprocessing too
+        preprocessor_and_arch_args.extend(parsed_args.common_args.to_vec());
 
         let absolute_input_path: Cow<'_, _> = if parsed_args.input.is_absolute() {
             Cow::Borrowed(&parsed_args.input)
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -1726,6 +1726,69 @@ LLVM version: 6.0",
         assert_ne!(results[0].key, results[1].key);
     }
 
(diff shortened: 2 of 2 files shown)


sylvestre:
Could you please add a test to make sure we don't regress? Thanks

GZGavinZhao:
Would someone mind merging this? IMO this is a pretty serious issue, because `sccache` may return incorrect build artifact for C/C++ compilations whenever the `-D` and every other common compiler flags changes.

Xuanwo (approved):
Thanks!

src/compiler/c.rs
```diff
@@ -303,6 +303,8 @@ where
         // use in creating a hash key
         let mut preprocessor_and_arch_args = parsed_args.preprocessor_args.clone();
         preprocessor_and_arch_args.extend(parsed_args.arch_args.to_vec());
+        // common_args is used in preprocessing too
+        preprocessor_and_arch_args.extend(parsed_args.common_args.to_vec());
```
glandium, line 307: Removing this line doesn't make the test fail.

PR #2045, opened 2024-01-18
feat: initial support for Clang HIP

### LINKED ISSUE #2044: Feature Request: Support caching HIP compilation using clang
Similar to NVIDIA CUDA, AMD's [HIP](https://github.com/ROCm/HIP) can also be compiled using Clang, and I wonder whether maintainers would accept a new feature to cache HIP compilations. It works basically the same as caching CUDA using clang, e.g. you enable compiling as HIP with `-x hip`. Caching HIP is very beneficial because HIP compilation times grows linearly with the number of distinct GPU architectures to support, i.e. if the compiled program wants to run a 10 different GPU architectures it basically needs to be compiled 10 times, once for each architecture :sweat_smile: For a more detailed motivation, please see ROCm/ROCm#2817.

I have already implemented a prototype of this feature and I'm happy to continue to maintain this feature. The prototype has already been working really efficiently for caching HIP compilations and have saved me massive amounts of time when packaging AMD's ROCm software stack for Solus Linux.

The only potential trouble during the implementation is that implicit dependencies on some LLVM bitcodes and HIP runtime headers may be introduced, as shown [here](https://clang.llvm.org/docs/HIPSupport.html#order-of-precedence-for-device-library-path) in the official documentation, but I think it is definitely not impossible to deal with this.

GZGavinZhao:
Fixes #2044.

~~Still needs to distinguished between different ROCm versions, due to implicit dependencies that may be introduced by `$HIP_DEVICE_LIB_PATH/*.bc` and header files located at `$HIP_PATH/include/hip/**/*.h`~~ No need to concern about header files, but for the bitcode libraries I'm still thinking. 

I currently have an experimental support for detecting and adding the bitcode libraries into `extra_hash_files`. If the implementation is undesired, then my idea is to just ignore bitcode libraries and explicitly mention in the docs to add the bitcode libraries to `SCCACHE_EXTRA_FILES` if the user is worried about erroneous cache hits.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -292,20 +292,18 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
                         }
                     }
 
-                    let hip_device_lib_path: Option<String> = hip_device_lib_path_arg
+                    let hip_device_lib_path: String = hip_device_lib_path_arg
                         .or(hip_device_lib_path_env)
                         .or(rocm_path_arg)
-                        .or(rocm_path_env);
+                        .or(rocm_path_env)
+                        // This is the default location in official AMD packages and containers.
+                        .unwrap_or("/opt/rocm/amdgcn/bitcode".to_string());
 
-                    if let Some(hip_device_lib_path) = hip_device_lib_path {
-                        if let Ok(files) = Path::new(&hip_device_lib_path).read_dir() {
-                            for file in files {
-                                if let Ok(file) = file {
-                                    if let Some(ext) = file.path().extension() {
-                                        if ext == "bc" {
-                                            args.extra_hash_files.push(file.path());
-                                        }
-                                    }
+                    if let Ok(files) = Path::new(&hip_device_lib_path).read_dir() {
+                        for file in files.flatten() {
+                            if let Some(ext) = file.path().extension() {
+                                if ext == "bc" {
+                                    args.extra_hash_files.push(file.path());
                                 }
                             }
                         }
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -256,50 +256,67 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
                     }
                 }
 
-                // Handle cache invalidation for the ROCm device bitcode libraries. See
-                // https://clang.llvm.org/docs/HIPSupport.html for details regarding the order that
-                // the environment variables and command-line arguments take affect.
+                // Handle cache invalidation for the ROCm device bitcode libraries. Every HIP
+                // object links in some LLVM bitcode libraries (.bc files), so in some sense
+                // every HIP object compilation has an direct dependency on those bitcode
+                // libraries.
+                //
+                // The bitcode libraries are unlikely to change **except** when a ROCm version
+                // changes, so for correctness we should take these bitcode libraries into
+                // account by adding them to `extra_hash_files`.
+                //
+                // See https://clang.llvm.org/docs/HIPSupport.html for details regarding the
+                // order in which the environment variables and command-line arguments control the
+                // directory to search for bitcode libraries.
+                //
+                // In reality, not every bitcode library in that directory is needed, but that is
+                // too much to handle on our side so we just hash every bitcode library in that
+                // directory.
                 if args.language == Language::Hip {
-                    let mut rocm_path_arg = None;
-                    let mut hip_device_lib_path_arg = None;
+                    let mut rocm_path_arg: Option<PathBuf> = None;
+                    let mut hip_device_lib_path_arg: Option<PathBuf> = None;
 
                     for arg in &args.common_args {
                         if let Some(sarg) = arg.to_str() {
                             if rocm_path_arg.is_none() && sarg.starts_with("--rocm-path=") {
-                                let mut p = sarg[12..].to_string();
-                                p.push_str("/amdgcn/bitcode");
-                                rocm_path_arg = Some(p);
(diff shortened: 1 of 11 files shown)


GZGavinZhao:
An implementation questions:

As mentioned earlier, I want to include changes in the `$HIP_DEVICE_LIB_PATH/*.bc` bitcode files into account. Currently my way of doing this is to add them to `extra_hash_files`. 

However, this presents a problem: the location of those files can be determined by both the environment variable and command-line arguments, but there doesn't seem to be a function where I have access to both of them. Ideally I want to add those files to `extra_hash_files` inside [`parse_arguments`](https://github.com/mozilla/sccache/blob/4d8125735b2a0ed674b93b7c5c89177e77106d6d/src/compiler/gcc.rs#L245). However, I don't have access to `env_vars` inside that function.

While I do have access to `env_vars` in [`generate_hash_key`](https://github.com/mozilla/sccache/blob/4d8125735b2a0ed674b93b7c5c89177e77106d6d/src/compiler/c.rs#L282), at this point the arguments have already been parsed into `parsed_args` and I have to iterate through the entire `parsed_args.common_args` to check. It also just feels wrong that those bitcode are not included inside `extra_hash_files` and I have to manually take them into account when hashing.

Would it be accepted if `parse_arguments` (defined in `src/compiler/gcc.rs`) can take another parameter `env_vars: Vec<(OsString, OsString)>`? I'm also open to other ways to achieve the same result.

GZGavinZhao:
Hi @sylvestre or @Xuanwo, could one of you provide some advice on how my problem above might be solved?

sylvestre:
HIP_DEVICE_LIB_PATH and HIP_PATH are standard and expected env variables in this ecosystem? Sorry, i don't know much here

GZGavinZhao:
Sorry, please ignore my previous comment. I figured the current implementation is good enough.

Feature-wise this is complete. I just need to write up docs and add E2E tests in `tests/`.

tests/system.rs
```diff
@@ -134,6 +135,33 @@ fn compile_cuda_cmdline<T: AsRef<OsStr>>(
...
+// TODO: This will fail if gcc/clang is actually a ccache wrapper, as it is the
+// default case on Fedora, e.g.
+//
+// archs is a list of GPU architectures to compile for.
+fn compile_hip_cmdline<T: AsRef<OsStr>>(
+    compiler: &str,
+    exe: T,
+    input: &str,
+    output: &str,
+    archs: &Vec<String>,
```
GZGavinZhao, line 147: I'll leave it `Vec<String>` instead of `Vec<OsString>` for now because I had a little trouble figuring out how to concat two `OsString`.

GZGavinZhao:
I still need to write up docs, but I would appreciate feedback on whether you guys are okay with the logic to find and hash bitcode libraries (these changes are in `src/compiler/c.rs`).

src/compiler/c.rs
```diff
@@ -249,11 +249,84 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
...
+                //
+                // See https://clang.llvm.org/docs/HIPSupport.html for details regarding the
+                // order in which the environment variables and command-line arguments control the
+                // directory to search for bitcode libraries.
+                //
+                // In reality, not every bitcode library in that directory is needed, but that is
+                // too much to handle on our side so we just hash every bitcode library in that
+                // directory.
+                if args.language == Language::Hip {
+                    let mut rocm_path_arg: Option<PathBuf> = None;
```
sylvestre, line 276: please avoid "let mut"
move the work into a function

GZGavinZhao: Fixed, please check if the current version is acceptable.

src/compiler/c.rs
```diff
@@ -249,11 +249,71 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
...
+                    };
+
+                    let rocm_path_arg: Option<PathBuf> = extract_rocm_arg("--rocm-path=");
+                    let hip_device_lib_path_arg: Option<PathBuf> =
+                        extract_rocm_arg("--hip-device-lib-path=");
+                    let rocm_path_env: Option<PathBuf> = extract_rocm_env("ROCM_PATH");
+                    let hip_device_lib_path_env: Option<PathBuf> =
+                        extract_rocm_env("HIP_DEVICE_LIB_PATH");
+
+                    let hip_device_lib_path: PathBuf = hip_device_lib_path_arg
```
sylvestre, line 299: please move this into a function

src/compiler/c.rs
```diff
@@ -249,11 +249,71 @@ impl<T: CommandCreatorSync, I: CCompilerImpl> Compiler<T> for CCompiler<I> {
...
+                //
+                // See https://clang.llvm.org/docs/HIPSupport.html for details regarding the
+                // order in which the environment variables and command-line arguments control the
+                // directory to search for bitcode libraries.
+                //
+                // In reality, not every bitcode library in that directory is needed, but that is
+                // too much to handle on our side so we just hash every bitcode library in that
+                // directory.
+                if args.language == Language::Hip {
+                    let extract_rocm_arg = |flag: &str| {
```
sylvestre, line 276: actually, would it be possible to move all this block into a function?

GZGavinZhao: Yes, I'm thinking of something like `fn search_hip_device_libs(args: &ParsedArguments, env_vars: &[(OsString, OsString)]) -> Vec<PathBuf>`. Where shall I put this function, in `src/util.rs`?

sylvestre: This file is fine

sylvestre:
Thanks!

PR #2048, opened 2024-01-20
Implement read-only local cache

### LINKED ISSUE #1852: Allow local cache to be made read-only
Perhaps this is related to #30, but simpler. I'm looking for a way to make the local cache read-only, so that subsequent builds still get files from the cache, but won't add new items to it. I think GCS has an option for this under `SCCACHE_GCS_RW_MODE`, but I'm looking to use it for a local cache.

My use case is that I'm building a Docker container with a lot of dependencies pre-compiled. That docker container is than ran and used to compile programs. Ideally, those programs use dependencies that have already been pre-compiled, speeding up compile times. Adding items to the cache when the container is being built is useful, but not when the container is run, as it just ends up using unnecessary disk space.

It'd be wonderful if there was an environment variable (e.g., `SCCACHE_LOCAL_RW_MODE`) that when set to some value (e.g., `READONLY` or `0`) makes items still be retrieved from the cache during builds, but prevents new items from being added to it.

uellenberg:
This allows the local cache to be set as read-only (similar to GCS read-only) when the `SCCACHE_LOCAL_RW_MODE` environment variable is set to `READ_ONLY` (or the corresponding config option). 

Closes #1852.

### DIFF
--- a/docs/Local.md
+++ b/docs/Local.md
@@ -43,3 +43,7 @@ Configuration options and their default values:
 See where to write the config in [the configuration doc](Configuration.md).
 
 *Note that preprocessor caching is currently only implemented for GCC and Clang and when using local storage.*
+
+## Read-only cache mode
+
+By default, the local cache operates in read/write mode. The `SCCACHE_LOCAL_RW_MODE` environment variable can be set to `READ_ONLY` (or `READ_WRITE`) to modify this behavior.
--- a/docs/Local.md
+++ b/docs/Local.md
@@ -47,3 +47,7 @@ See where to write the config in [the configuration doc](Configuration.md).
 ## Read-only cache mode
 
 By default, the local cache operates in read/write mode. The `SCCACHE_LOCAL_RW_MODE` environment variable can be set to `READ_ONLY` (or `READ_WRITE`) to modify this behavior.
+
+You can use read-only mode to prevent sccache from writing new cache items to the disk. This can be useful, for example, if you want to use items that have already been cached, but not add new ones to the cache. 
+
+Note that this feature is only effective if you already have items in your cache. Using this option on an empty cache will cause sccache to simply do nothing.
--- a/docs/Local.md
+++ b/docs/Local.md
@@ -50,4 +50,4 @@ By default, the local cache operates in read/write mode. The `SCCACHE_LOCAL_RW_M
 
 You can use read-only mode to prevent sccache from writing new cache items to the disk. This can be useful, for example, if you want to use items that have already been cached, but not add new ones to the cache. 
 
-Note that this feature is only effective if you already have items in your cache. Using this option on an empty cache will cause sccache to simply do nothing.
+Note that this feature is only effective if you already have items in your cache. Using this option on an empty cache will cause sccache to simply do nothing, just add overhead.
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::cache::{Cache, CacheRead, CacheWrite, Storage};
+use crate::cache::{Cache, CacheMode, CacheRead, CacheWrite, Storage};
 use crate::compiler::PreprocessorCacheEntry;
+use crate::config::CacheRWMode;
 use crate::lru_disk_cache::LruDiskCache;
 use crate::lru_disk_cache::{Error as LruError, ReadSeek};
 use async_trait::async_trait;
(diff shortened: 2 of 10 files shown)


src/compiler/c.rs
```diff
@@ -340,42 +340,51 @@ where
...
-                        storage.put_preprocessor_cache_entry(
+
+                        if let Err(e) = storage.put_preprocessor_cache_entry(
                             preprocessor_key,
                             preprocessor_cache_entry,
-                        )?;
+                        ) {
+                            debug!("Failed to update preprocessor cache: {}", e);
+                            update_failed = true;
+                        }
```
uellenberg, line 358: I'm unsure about this behavior. It's based on how `compiler.rs` handles errors.

Realistically, errors shouldn't be used here, but it seems to be how it's handled with GCS read-only, so I did it the same way. I thought about changing it, but I'd like to hear the input about someone more knowledgeable about sccache's internals first. `Storage::check()` seems to be a great candidate, although it only seems to be used for a print and is very expensive for GCS (which could be changed, of course).

src/compiler/c.rs
```diff
@@ -491,8 +500,12 @@ where
...
                 files.sort_unstable_by(|a, b| a.1.cmp(&b.1));
                 preprocessor_cache_entry.add_result(start_of_compilation, &key, files);
-                storage
-                    .put_preprocessor_cache_entry(&preprocessor_key, preprocessor_cache_entry)?;
+
+                if let Err(e) = storage
+                    .put_preprocessor_cache_entry(&preprocessor_key, preprocessor_cache_entry)
+                {
+                    debug!("Failed to update preprocessor cache: {}", e);
+                }
```
uellenberg, line 508: Same as above.

src/cache/gcs.rs
```diff
@@ -36,6 +37,15 @@ impl RWMode {
...
 
+impl From<config::CacheRWMode> for RWMode {
+    fn from(value: config::CacheRWMode) -> Self {
+        match value {
+            config::CacheRWMode::ReadOnly => RWMode::ReadOnly,
+            config::CacheRWMode::ReadWrite => RWMode::ReadWrite,
+        }
+    }
+}
+
```
uellenberg, line 48: There are a few different types like this that have the same purpose. Should they be consolidated?

uellenberg: I've merged two of them together, although there's still `CacheModeConfig` and `CacheMode`, which do effectively the same thing.

docs/Local.md
```diff
@@ -43,3 +43,7 @@ Configuration options and their default values:
 See where to write the config in [the configuration doc](Configuration.md).
 
 *Note that preprocessor caching is currently only implemented for GCC and Clang and when using local storage.*
+
+## Read-only cache mode
+
+By default, the local cache operates in read/write mode. The `SCCACHE_LOCAL_RW_MODE` environment variable can be set to `READ_ONLY` (or `READ_WRITE`) to modify this behavior.
```
sylvestre, line 49: could you please detail in which case someone might want to use READ_ONLY
thanks

docs/Local.md
```diff
@@ -43,3 +43,11 @@ Configuration options and their default values:
...
+
+You can use read-only mode to prevent sccache from writing new cache items to the disk. This can be useful, for example, if you want to use items that have already been cached, but not add new ones to the cache. 
+
+Note that this feature is only effective if you already have items in your cache. Using this option on an empty cache will cause sccache to simply do nothing.
```
sylvestre, line 53:
```suggestion
Note that this feature is only effective if you already have items in your cache. Using this option on an empty cache will cause sccache to simply do nothing, just add overhead.
```

docs/Local.md
uellenberg: > could you please detail in which case someone might want to use READ_ONLY thanks

@sylvestre 

Done, let me know if that's what you want. For a more specific use case, see the linked issue. I'm using sccache to speed up build times for untrusted (and short-lived) build tasks, where new items can't be added as they may not be valid, but caching commonly used libraries can still speed up build times immensely. Read-only is not being used as a security feature, as build tasks are already isolated correctly, but to reduce useless disk consumption.

src/cache/disk.rs
```diff
@@ -130,6 +133,11 @@ impl Storage for DiskCache {
         // We should probably do this on a background thread if we're going to buffer
         // everything in memory...
         trace!("DiskCache::finish_put({})", key);
+
+        if self.rw_mode == CacheMode::ReadOnly {
+            return Err(anyhow!("Cannot write to a read-only cache"));
```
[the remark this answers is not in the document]
uellenberg: > could you please add a test to trigger this error ? thanks

@sylvestre Done.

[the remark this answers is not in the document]
sylvestre: could you please check for the error message string too ?
thanks

uellenberg:
I'm not sure why the integration tests are failing. The only change I made (since they succeeded) was asserting the errors' messages and running rustfmt.

sylvestre:
rate limiting on github, not your fault :)

sylvestre:
thanks for your pr!

Issue #2057, opened 2024-01-23
0.7.6 includes a breaking change, but is a semver "minor".


arilotter:
0.7.6 adds a **mandatory** `ttl` field to the `[cache.redis]` section of sccache_config.

A config file that worked perfectly in 0.7.5:

```
[cache.redis]
url = "redis://redis.internal.devserver.local:6379/0"
```

In 0.7.6, now throws the error:
```
  sccache: error: Failed to load config file
  sccache: caused by: Failed to load toml config file from /root/.config/sccache/config
  sccache: caused by: TOML parse error at line 1, column 1
    |
  1 | [cache.redis]
    | ^^^^^^^^^^^^^
  missing field `ttl`
  ```

This makes 0.7.6 a release with breaking changes, and by semver, it should be labeled as 0.8.0, or the requirement for the `ttl` key should be dropped and a default used instead.

arilotter:
@sylvestre I opened https://github.com/mozilla/sccache/pull/2058 that i think fixes it

sylvestre:
thanks!

sudhackar:
thanks!

PR #2058, opened 2024-01-23
add default ttl of 0 to fix breaking change

arilotter:
closes https://github.com/mozilla/sccache/issues/2057


sylvestre:
it would be nice to have a test to make sure we don't regress in the future!

PR #2075, opened 2024-02-02
Note allowed --emit metadata in rust doc

l4l:
as it in the source:

https://github.com/mozilla/sccache/blob/69be5321d2c2c125881b6edfed96676572b0ca03/src/compiler/rust.rs#L213-L214


Xuanwo (approved):
Thanks!

PR #2077, opened 2024-02-04
[commands::run_server_process]: fix race condition

### LINKED ISSUE #2076: Race condition in [run_server_process] on `master` and sccache 0.7.4
We have identified a race condition in `commands.rs::run_server_process` (linux variant) on the `master` branch revision 69be5321d2c2c125881b6edfed96676572b0ca03. It also seems to be present in sccache 0.7.4.

The offending code is the [following](https://github.com/mozilla/sccache/blob/main/src/commands.rs#L92):

```rust
    let _child = process::Command::new(&exe_path)
        .current_dir(workdir)
        .env("SCCACHE_START_SERVER", "1")
        .env("SCCACHE_STARTUP_NOTIFY", &socket_path)
        .env("RUST_BACKTRACE", "1")
        .spawn()?;

    let startup = async move {
        let listener = tokio::net::UnixListener::bind(&socket_path)?;
        let (socket, _) = listener.accept().await?;

        read_server_startup_status(socket).await
    };
```

If the child process is sufficiently fast, it will attempt to connect to the socket before it is bound the `startup` block. This can observed directly by inserting a sleep in the `startup` block:

```
$ git diff
diff --git a/src/commands.rs b/src/commands.rs
index e4b9a6a..f3c65a7 100644
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -97,6 +97,8 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
         .spawn()?;

     let startup = async move {
+        tokio::time::sleep(Duration::from_millis(1000)).await;
+
         let listener = tokio::net::UnixListener::bind(&socket_path)?;
         let (socket, _) = listener.accept().await?;
```

And running `sccache --start-server`, which will fail with:

```
$ ./target/debug/sccache --start-server
sccache: error: Timed out waiting for server startup

$ tail sccache.log
[2024-02-02T14:05:05Z INFO  sccache::server] start_server: port: 4226
[2024-02-02T14:05:05Z INFO  sccache::server] No scheduler address configured, disabling distributed sccache
[2024-02-02T14:05:05Z DEBUG sccache::cache::cache] Init disk cache with dir "/home/arvid/.cache/sccache", size 10737418240
[2024-02-02T14:05:05Z INFO  sccache::server] server has setup with ReadWrite
[2024-02-02T14:05:05Z INFO  sccache::server] server started, listening on port 4226
[2024-02-02T14:05:05Z DEBUG sccache::server] notify_server_startup(Ok { port: 4226 })
sccache: error: No such file or directory (os error 2)
```

I could also trigger the issue without modifying the code by attempting to start the server in a loop, like this:

```
while { target/debug/sccache --start-server; target/debug/sccache --stop-server > /dev/null; } ; do echo done; done
```

It seems the solution would be to:

 1. First bind the socket
 2. Then spawn the child process
 3. Then then await on the server startup status.

Not knowing rust, here's the proposition in pseudo-code:

```rust
    // Bind the socket -- does not work because we're not in async code.
    let listener = tokio::net::UnixListener::bind(&socket_path)?;

    // Spawn the child process
    let _child = process::Command::new(&exe_path)
        .current_dir(workdir)
        .env("SCCACHE_START_SERVER", "1")
        .env("SCCACHE_STARTUP_NOTIFY", &socket_path)
        .env("RUST_BACKTRACE", "1")
        .spawn()?;

    // await server startup
    let startup = async move {
        tokio::time::sleep(Duration::from_millis(1000)).await;

        let (socket, _) = listener.accept().await?;

        read_server_startup_status(socket).await
    };
```

arvidj:
Fixes #2076

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -92,9 +92,7 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
 
     // Spawn a blocking task to bind the Unix socket
     let sp = socket_path.clone();
-    let listener = runtime.block_on(async move {
-        tokio::net::UnixListener::bind(sp)
-    })?;
+    let listener = runtime.block_on(async move { tokio::net::UnixListener::bind(sp) })?;
 
     let _child = process::Command::new(&exe_path)
         .current_dir(workdir)
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -95,7 +95,7 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     // the parent binding the socket and the child connecting to it.
     let listener = {
         let _guard = runtime.enter();
-        tokio::net::UnixListener::bind(&socket_path.clone())?
+        tokio::net::UnixListener::bind(&socket_path)?
     };
 
     let _child = process::Command::new(&exe_path)
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -90,7 +90,9 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     let exe_path = env::current_exe()?;
     let workdir = exe_path.parent().expect("executable path has no parent?!");
 
-    // Spawn a blocking task to bind the Unix socket
+    // Spawn a blocking task to bind the Unix socket. Note that the socket
+    // must be bound before spawning `_child` below to avoid a race between
+    // the parent binding the socket and the child connecting to it.
     let sp = socket_path.clone();
     let listener = runtime.block_on(async move { tokio::net::UnixListener::bind(sp) })?;
 
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -93,8 +93,10 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     // Spawn a blocking task to bind the Unix socket. Note that the socket
     // must be bound before spawning `_child` below to avoid a race between
     // the parent binding the socket and the child connecting to it.
-    let sp = socket_path.clone();
-    let listener = runtime.block_on(async move { tokio::net::UnixListener::bind(sp) })?;
+    let listener = {
+        let _guard = runtime.enter();
+        tokio::net::UnixListener::bind(&socket_path.clone())?
+    };
 
     let _child = process::Command::new(&exe_path)
         .current_dir(workdir)
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -89,6 +89,13 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     let runtime = Runtime::new()?;
     let exe_path = env::current_exe()?;
     let workdir = exe_path.parent().expect("executable path has no parent?!");
+
+    // Spawn a blocking task to bind the Unix socket
+    let sp = socket_path.clone();
+    let listener = runtime.block_on(async move {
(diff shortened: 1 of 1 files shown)


src/commands.rs
```diff
@@ -89,6 +89,13 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     let runtime = Runtime::new()?;
     let exe_path = env::current_exe()?;
     let workdir = exe_path.parent().expect("executable path has no parent?!");
+
+    // Spawn a blocking task to bind the Unix socket
+    let sp = socket_path.clone();
+    let listener = runtime.block_on(async move {
+        tokio::net::UnixListener::bind(sp).expect("Failed to bind Unix socket")
+    });
```
uellenberg, line 97:
```suggestion
    let listener = runtime.block_on(async move {
        tokio::net::UnixListener::bind(sp)
    })?;
```
Untested, but perhaps returning the error directly would be better (to preserve the original functionality).

arvidj: Works well, I've pushed a fixup with this change.

src/commands.rs
```diff
@@ -89,6 +89,11 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     let runtime = Runtime::new()?;
     let exe_path = env::current_exe()?;
     let workdir = exe_path.parent().expect("executable path has no parent?!");
+
+    // Spawn a blocking task to bind the Unix socket
```
sylvestre, line 93: maybe add a comment to explain that it has to be done here to avoid a race

src/commands.rs
```diff
@@ -89,6 +89,13 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     let runtime = Runtime::new()?;
     let exe_path = env::current_exe()?;
     let workdir = exe_path.parent().expect("executable path has no parent?!");
+
+    // Spawn a blocking task to bind the Unix socket. Note that the socket
+    // must be bound before spawning `_child` below to avoid a race between
+    // the parent binding the socket and the child connecting to it.
+    let sp = socket_path.clone();
+    let listener = runtime.block_on(async move { tokio::net::UnixListener::bind(sp) })?;
```
glandium, line 97: This can be written `let listener = tokio::net::UnixListener::bind(&socket_path.clone())?;`

arvidj: It compiles, but when I run it I get:
```
$ ./target/debug/sccache --start-server
thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', src/commands.rs:98:20
```

The way I understand it: the function `run_server_process` is not async. Therefore, you cannot call async functions in it without starting a Tokio runtime(?). `tokio::net::UnixListener::bind` is an async function. Therefore, you must call it inside a `runtime.block_on` call so that it runs in a runtime(?).

glandium: The function is not, actually, async. https://docs.rs/tokio/latest/tokio/net/struct.UnixListener.html#method.bind But it does not want to run without a runtime.
So the following would work too, without an async block:
```
let listener = {
    let _guard = runtime.enter();
    tokio::net::UnixListener::bind(&socket_path.clone())?
};
```

arvidj: Indeed. I've pushed a fixup with your latest suggestion.

arvidj:
Latest workflow failed due to clippy complaining about the `clone` of `socket_path` being redundant. I've removed the clone in the latest version.

sylvestre:
many thanks
i will do a release now

PR #2079, opened 2024-02-07
chore: update S3 docs on region

saiintbrisson:
OpenDAL now expects SCCACHE_REGION to always be set when using S3.
apache/opendal#2812


Xuanwo (approved):
Thanks!

PR #2084, opened 2024-02-11
feat: Add possibility to configure key prefix (Redis, MemCached)

AJIOB:
* Add possibility to configure key prefix for Redis & MemCached
* Unify Redis & MemCached expiraction time
* Actualize configuration docs


sylvestre:
if i understand correctly this PR, it isn't going to change the behavior of existing configurations, right?

AJIOB:
@sylvestre, yes, you are right.

We are not change the public API. Just documenting & unifying it.

AJIOB:
@Xuanwo, hello.

What I need to do for merging this PR?

Do we have some problems with this changes? Or I need to split it to few ones?

Xuanwo:
Those changes LGTM! cc @sylvestre do you want to take another look?

sylvestre:
thanks!

PR #2085, opened 2024-02-12
rustc driver detection - includes #1897

### LINKED ISSUE #861: Support arbitrary rustc drivers

drahnr:
If we are not sure it's rustc like and we check -vV output to accept valid rustc drivers. Simplifies #1897 . 

Fixes #861

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -980,6 +980,30 @@ fn is_rustc_like<P: AsRef<Path>>(p: P) -> bool {
     )
 }
 
+/// Returns true if the given path looks like a c compiler program
+///
+/// This does not check c compilers, it only report programs that are definitely not rustc
+fn is_like_c_compiler<P: AsRef<Path>>(p: P) -> bool {
+    matches!(
+        p.as_ref()
+            .file_stem()
+            .map(|s| s.to_string_lossy().to_lowercase())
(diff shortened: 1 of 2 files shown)


sylvestre:
It would be nice to update the commit to explicit why we are doing it

drahnr:
> It would be nice to update the commit to explicit why we are doing it

The _commit_ or the PR desc?

sylvestre:
The commit please

drahnr:
Is the above sufficient @sylvestre ? I'd still be in favor of squash this PR, left the two commits separate for no particular reason other than authorship

drahnr:
The coverage report is obsolete, it actually improved marginally.

CC @sylvestre anything missing/insufficient, otherwise I'd merge?

PR #2091, opened 2024-02-14
Add generic read-only storage mode for a storage that fails the write check

### LINKED ISSUE #1853: Cache check result doesn't influence mode of operation
The `Storage` trait has a `check` method that returns whether the cache is able to Read/Write or Read-only[1]. When the method is able to do that determination without an error, the code in `start_server` then proceeds with displaying `server has setup with {cache_mode:?}`[2] and... nothing else. Which means when the check says only Read-only is supported, cache writes are still attempted.

1. https://github.com/mozilla/sccache/blob/7074753bfc100ee3e22dc730cf71f3750fc42c1d/src/cache/cache.rs#L352-L354
2. https://github.com/mozilla/sccache/blob/7074753bfc100ee3e22dc730cf71f3750fc42c1d/src/server.rs#L423-L440

Side note, this is actually quite fortunate for us, because at least in the case of GCS, the check requires delete permissions to overwrite the `.sccache_check` file, which our sccache clients don't have (which means `SCCACHE_RECACHE` doesn't work on this setup...). So while our logs do complain about the lack of delete permissions, and say "server has setup with ReadOnly", cache misses are properly filled after local compile...

So what this means is that at least in the case of GCS, the check is actually checking whether recaching (`SCCACHE_RECACHE`) would work, rather than whether the cache would work, although the very first time (when the `.sccache_check` file is not in the bucket), it would do the latter.

jkoritzinsky:
Today, sccache can detect that a storage medium is read-only or read-write even if this capability was not specified by the config. However, it doesn't do anything with this information other than log it for diagnostics.

This PR changes sccache to use this information to avoid writing to a read-only storage. This is useful for cases where one might want to use sccache in a read-only configuration in some cases (e.g. public PR builds) while only updating the cache from a separate pipeline (that has write access) without using a separate configuration file.

Additionally, this change reduces the number of "access denied" errors that would be logged on the storage side when sccache detects that it only has read-only access (instead of continuing to attempt to write to the storage).

Fixes #1853

### DIFF
--- a/src/cache/readonly.rs
+++ b/src/cache/readonly.rs
@@ -20,17 +20,6 @@ use async_trait::async_trait;
 use crate::cache::{Cache, CacheMode, CacheWrite, Storage};
 use crate::errors::*;
 
-#[derive(Debug)]
-pub struct ReadOnlyError;
-
-impl std::error::Error for ReadOnlyError {}
-
-impl std::fmt::Display for ReadOnlyError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Cannot write to read-only storage")
-    }
-}
-
 pub struct ReadOnlyStorage(pub Arc<dyn Storage>);
 
 #[async_trait]
@@ -47,7 +36,7 @@ impl Storage for ReadOnlyStorage
     /// finished.
     async fn put(&self, _key: &str, _entry: CacheWrite) -> Result<Duration>
     {
-        Err(ReadOnlyError{}.into())
+        Err(anyhow!("Cannot write to read-only storage"))
     }
 
     /// Check the cache capability.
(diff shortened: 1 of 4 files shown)


sylvestre:
Could you please add a test and fix the clippy warning?
Thanks

jkoritzinsky:
I've fixed the clippy warning and added a few unit tests.

src/cache/readonly.rs
```diff
@@ -0,0 +1,146 @@
+// Copyright 2016 Mozilla Foundation
```
sylvestre, line 1: please remove this line

sylvestre:
Sorry, I was talking about an integration test like
https://github.com/mozilla/sccache/blob/ffe3070f77ef3301c8ff718316e4ab017ec83042/tests/sccache_cargo.rs#L318
thanks

jkoritzinsky:
@sylvestre I've added a test that validates (through the sccache log) that in a scenario where the underlying cache is detected as read-only, the ReadOnlyStorage handles the write requests and the requests never go to the underlying storage.

sylvestre:
thanks!

PR #2094, opened 2024-02-15
feat: Add possibility to configure key prefix and expiration time from env (Redis, MemCached)

AJIOB:
* Unifying the way for parsing all `..._KEY_PREFIX` variables from env
* Add the way for reading numbers from env


AJIOB:
Hi @Xuanwo,

Looks like we have some test data races or the compiler bug, isn't it?

Let's try to restart the failed build.

Xuanwo (approved):
Thanks!

AJIOB:
@Xuanwo , let's merge it?

PR #2101, opened 2024-02-20
allow the compilation of sccache without any features (only local build)

sylvestre:


.github/workflows/ci.yml
```diff
@@ -81,6 +81,11 @@ jobs:
             rustc: nightly
             allow_failure: true
             extra_args: --features=unstable
+          - os: ubuntu-20.04
+            rustc: nightly
```
AJIOB, line 85: Do we really need to use nightly build instead of `1.70.0` as a minimal supported?

sylvestre: not at all, it is just to put a compiler :)

i will update it to use stable

drahnr (approved):

PR #2103, opened 2024-02-21
chore: Unify env bool processing

AJIOB:
* Unify env bool processing
* Upper & mixed-case values are also accepted
* Actualize main readme file with the latest feature list location
* Return some missed spaces


Xuanwo (approved):
Thanks!

Issue #2083, opened 2024-02-09
Store Redis username & password in a dedicated environment variables


AJIOB:
As I can see, this split was done in a manner that `opendal` processes URL prefixes (out of the current issue scope).

But this concept is not useful, IMO.

I prefer to pass a single `SCCACHE_WEBDAV_ENDPOINT` and extract `SCCACHE_WEBDAV_KEY_PREFIX` from the endpoint. If `opendal` needs, `SCCACHE_WEBDAV_ENDPOINT` value may be removed from `SCCACHE_WEBDAV_KEY_PREFIX` too (for better understanding/support via `opendal` provider).

If both `SCCACHE_WEBDAV_ENDPOINT` path is non-root and `SCCACHE_WEBDAV_ENDPOINT` is not empty, we should:
1. Pass them without any changes to the `opendal`
2. Check via an `assert!` and finish the program as an invalid case

I prefer to use the second variant, but ***maybe*** some implementations exist that requires both of the values (I don't know anything about them).

AJIOB:
P.S. Username & password may be passed as a part of the `SCCACHE_WEBDAV_ENDPOINT` too (optionally), because it is still a valid URL.
P.P.S. The Redis backend connection parameters are passed in the same manner that I suggest to WebDAV.

Xuanwo:
The prefix is based on input endpoint.

Some webdav services could expose webdav at /v1/webdav, so it might be confusing to extract them as path.

> Username & password may be passed as a part of the SCCACHE_WEBDAV_ENDPOINT too (optionally), because it is still a valid URL.

Although the URL is technically valid, incorporating the password directly within the endpoint is risky, as the endpoint is not excluded from Debug output. This oversight could potentially expose the password in logs or traces, leading to a security vulnerability.

AJIOB:
> Some webdav services could expose webdav at `/v1/webdav`, so it might be confusing to extract them as path.

As I can see in [Artifactory docs](https://jfrog.com/help/r/jfrog-artifactory-documentation/using-webdav) - just used as an example - I should use `artifactory/repo-path` as a part of the `SCCACHE_WEBDAV_ENDPOINT` or as a `SCCACHE_WEBDAV_KEY_PREFIX` value now?

AJIOB:
> Although the URL is technically valid, incorporating the password directly within the endpoint is risky, as the endpoint is not excluded from Debug output. This oversight could potentially expose the password in logs or traces, leading to a security vulnerability.

[...]

PR #2113, opened 2024-02-27
feat: Add support to Redis cluster mode

AJIOB:
* Add possibility to use Redis in cluster mode
* Store Redis credentials via dedicated fields: for security reasons, as discussed in #2083
* Deprecate `url` usage for Redis config
* Actualize doc


Xuanwo (approved):
Thanks!

PR #2116, opened 2024-03-02
feat: Rewrite Redis environment variables API

### LINKED ISSUE #2083: Store Redis username & password in a dedicated environment variables

AJIOB:
* Implement Redis configuration values splitting support
* Actualize Redis main docs
* Actualize configuration docs
* Use both old & new Redis API for tests

Fix #2083

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -27,7 +27,7 @@ jobs:
           name: integration-tests
           path: ./target/debug/sccache
 
-  redis:
+  redis-deprecated:
     runs-on: ubuntu-latest
     needs: build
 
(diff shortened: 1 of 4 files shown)


Xuanwo (approved):
Thanks a lot for the tests!

PR #2114, opened 2024-03-02
build(deps): Upgrade OpenDAL to the latest version

AJIOB:
It should fix #2082

PR #2118, opened 2024-03-03
chore: use what we really need from `opendal`

AJIOB:
~This PR must be merged only after #2114, because this code is based on the parent changes~


Xuanwo (approved):
Thanks!

PR #2124, opened 2024-03-04
package cargo.toml

lissyx:


sylvestre:
please run rustfmt

src/compiler/rust.rs
```diff
@@ -1896,6 +1896,27 @@ impl pkg::InputsPackager for RustInputsPackager {
             let dist_input_path = path_transformer.as_dist(&input_path).with_context(|| {
                 format!("unable to transform input path {}", input_path.display())
             })?;
+
+            if input_path.ends_with("src/lib.rs") {
```
AJIOB, line 1900: Should it works on Windows too?

lissyx: no idea where `sccache-dist` is supported

lissyx: I guess building `"src/lib.rs"` using `PathBuf::new().join("src").join("lib.rs")` should be enough, if this is separators you are worrried about?

lissyx: I'm not sure in fact we need ?
```
PS C:\Users\user> rustc .\test.rs
PS C:\Users\user> cat .\test.rs
use std::path::PathBuf;

fn main() {
    let p = PathBuf::new().join("src").join("lib.rs");
    println!("p={:?}", p);
    let crate_path = PathBuf::new().join("crate1/src/lib.rs");
    println!("path => {:?}", crate_path.ends_with(p));
}
PS C:\Users\user> .\test.exe
p="src\\lib.rs"
path => true
PS C:\Users\user>

Comment (AJIOB) [reply src/compiler/rust.rs:1900]: I prefer to rely on cross-platfom solution if it's possible

Comment (lissyx) [reply src/compiler/rust.rs:1900]: What do you have in mind as cross-platform ? using `PathBuf::new().join("src").join("lib.rs");` or something else?

Comment (AJIOB) [reply src/compiler/rust.rs:1900]: Yes, I suggest to use `PathBuf::new().join("src").join("lib.rs");`

Comment (lissyx) [reply src/compiler/rust.rs]: should be good now

File: src/compiler/rust.rs [hunk 2]
```diff
@@ -1896,6 +1896,28 @@ impl pkg::InputsPackager for RustInputsPackager {
             let dist_input_path = path_transformer.as_dist(&input_path).with_context(|| {
                 format!("unable to transform input path {}", input_path.display())
             })?;
+
+            let lib_rs = PathBuf::new().join("src").join("lib");
```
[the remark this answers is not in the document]
lissyx: ok, sorry.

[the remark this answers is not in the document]
lissyx: maybe it exposes I should make that a separate function with testing

src/compiler/rust.rs
```diff
@@ -1844,6 +1844,81 @@ fn test_can_trim_this() {
...
+        None
+    }
+}
+
+#[test]
+#[cfg(feature = "dist-client")]
+fn test_maybe_add_cargo_toml() {
+    #[cfg(not(target_os = "windows"))]
+    {
+        let root = PathBuf::from("/home/user/mozilla-source/mozilla-unified/third_party/rust");
```
AJIOB, line 1875: Can we just create a `root` and `result_cargo` variables with [this feature](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-macro), for example?

Looks like another test part may be equal for both Windows & non-Windows platforms, isn't it?

lissyx: i splitted the test per-platform mostly because `"/home/user/mozilla-source/mozilla-unified/third_party/rust/wgpu-core/Cargo.toml"` would not match when ran on windows.

I added the assert on `to_str()` result to verify the actual path, but maybe this is useless? What's your opinion? If we remove this we can remove the conditionals; I could also welcome feedback on whether it is useful we test with unix and windows path on all platforms?

AJIOB: What about something like this in the beginning of the test?

```rust
let (root, result_cargo_toml_path) = if cfg!(windows) {
  (
    "/home/user/mozilla-source/mozilla-unified/third_party/rust",
    "/home/user/mozilla-source/mozilla-unified/third_party/rust/wgpu-core/Cargo.toml"
  )
} else {
  (
    "C:\\mozilla-source\\mozilla-unified\\third_party\\rust",
    "C:\\mozilla-source\\mozilla-unified\\third_party\\rust\\wgpu-core\\Cargo.toml"
  )
};

...<cross-platform-test-with-checks>...
```

P.S. yes, `\` symbols must be escaped inside the string literals in Rust too.

lissyx: @AJIOB that's not the problem I was facing, rather that resulting string on windows for the `.to_str()` assert was:
> "/home/user/mozilla-source/mozilla-unified/third_party/rust/wgpu-core\\Cargo.toml"

AJIOB: It's almost impossible, if you have `"C:\\mozilla-source\\mozilla-unified\\third_party\\rust"` as a source string

lissyx: > It's almost impossible, if you have `"C:\\mozilla-source\\mozilla-unified\\third_party\\rust"` as a source string

that's my second question: does it makes sense to have the test running both variants on all paltforms?

AJIOB: I think it should be Windows variant on Windows platform, Linux variant on Linux platform

lissyx: any reason why you prefer a raw string over a `PathBuf::from()` in your example?

AJIOB: Then you can build `PathBuf::from()` from that raw strings.
I've just trying to minimize what we really need to be platform-dependent inside the tests

AJIOB (approved):

src/compiler/rust.rs
lissyx: Thanks, I wanted to make sure there was no other reason!

PR #2133, opened 2024-03-15
WIP: Add option to skip RW mode check

### LINKED ISSUE #2132: Hitting rate limiting issues due to .sccache_check write checks on GCS
We're using sccache v0.7.6, we have a lot of builds running in parallel and each run invokes sccache which in turn tries to do a sccache RW check by reading and writing .sccache_check dummy file. While this is fine for reading but for writing we violate the [quota](https://cloud.google.com/storage/quotas#objects) on GCS which allows write on the same object once per second. 

Is there a way to bypass this check? We can perhaps create an env var to support this use case?

cceckman:
WIP: I've not tested this yet.

Adds a config / env option, `cache.assume_rw_mode` or `SCCACHE_ASSUME_RW_MODE`. If present, the server skips the startup check of the cache's mode, and assumes it is the one specified. If not present, the server attempts to read and write from the server, and uses the results to determine the mode.

Fixes #2132 (I think - again, I need to work out how to test this properly before sending forward).

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -19,6 +19,11 @@ cache_dir = "/home/user/.cache/sccache-dist-client"
 type = "token"
 token = "secrettoken"
 
+[cache]
+# If specified, skip the startup check for whether the backend is read-only/writeable;
+# assume it has this mode instead.
+# If unspecified, the server will check the mode at startup.
+assume_rw_mode = "READ_WRITE"
 
 #[cache.azure]
 # does not work as it appears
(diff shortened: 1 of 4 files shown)


docs/Configuration.md
```diff
@@ -3,6 +3,9 @@
 ## file
 
 ```toml
+# If specified, wait this long for the server to start up.
```
sylvestre, line 6: don't hesitate to do this into a different PR as it is unrelated

cceckman: Sure thing: #2134

Xuanwo (approved):
Mostly LGTM, thanks a lot for your work! It's better to find a way to test it. Maybe we can add a new workflow with assume mode enabled?

sylvestre:
indeed, it will need some tests before merging this PR!
thanks :)

sylvestre:
still draft for a while
please reopen when ready

PR #2137, opened 2024-03-25
Set User Agent Header for S3 Requests

### LINKED ISSUE #2136: Set User Agent Header for S3 Requests
## Feature Request

As defined by MDN ([src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)):

> The User-Agent request header is a characteristic string that lets servers and network peers identify the application, operating system, vendor, and/or version of the requesting user agent.

I'd like to request that `sccache` set a user-agent header for requests to AWS/S3. The user-agent header should include the `sccache` version.

## Use Case

AWS supports writing IAM policies that block access to resources based on the user-agent header: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-useragent.

I'd like to write a policy that denies access to an S3 bucket for versions of `sccache` that are affected by #2038.

ajschmidt8:
Closes #2136.

This PR configures the S3 backend to use a custom HTTP client that has a user agent header set.

The user agent header enables `sccache` users to write AWS policies to accept or reject `sccache` requests to an S3 bucket based on their `sccache` version.

The user agent header value is like `sccache/0.7.7`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,7 +164,7 @@ memcached = ["opendal/services-memcached"]
 native-zlib = []
 oss = ["opendal/services-oss", "reqsign"]
 redis = ["url", "opendal/services-redis"]
-s3 = ["opendal/services-s3", "reqsign"]
+s3 = ["opendal/services-s3", "reqsign", "reqwest/blocking"]
 webdav = ["opendal/services-webdav"]
 # Enable features that will build a vendored version of openssl and
 # statically linked with it, instead of linking against the system-wide openssl
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,7 +164,7 @@ memcached = ["opendal/services-memcached"]
 native-zlib = []
 oss = ["opendal/services-oss", "reqsign"]
 redis = ["url", "opendal/services-redis"]
-s3 = ["opendal/services-s3", "reqsign", "reqwest/blocking"]
+s3 = ["opendal/services-s3", "reqsign", "reqwest"]
 webdav = ["opendal/services-webdav"]
 # Enable features that will build a vendored version of openssl and
 # statically linked with it, instead of linking against the system-wide openssl
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -31,11 +31,7 @@ impl S3Cache {
         server_side_encryption: Option<bool>,
     ) -> Result<Operator> {
         let mut builder = S3::default();
-        let user_agent = format!("{}/{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
-        let client_builder = ClientBuilder::new().user_agent(user_agent);
-        let client =
-            HttpClient::build(client_builder).map_err(|err| err.with_operation("S3::build"))?;
-        builder.http_client(client);
+        builder.http_client(build_http_client());
         builder.bucket(bucket);
         builder.root(key_prefix);
 
@@ -68,6 +64,12 @@ impl S3Cache {
             .finish();
         Ok(op)
     }
+}
+
+fn build_http_client() -> HttpClient {
+    let user_agent = format!("{}/{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
+    let client_builder = ClientBuilder::new().user_agent(user_agent);
+    HttpClient::build(client_builder).unwrap()
 }
 
 /// Resolve given endpoint along with use_ssl settings.
--- a/src/cache/s3.rs
+++ b/src/cache/s3.rs
@@ -11,8 +11,10 @@
 // limitations under the License.
 
 use opendal::layers::LoggingLayer;
+use opendal::raw::HttpClient;
 use opendal::services::S3;
 use opendal::Operator;
+use reqwest::ClientBuilder;
 
 use crate::errors::*;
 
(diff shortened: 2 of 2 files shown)


ajschmidt8:
hmm. i'll have to fix this import issue.

src/cache/s3.rs
```diff
@@ -29,6 +31,11 @@ impl S3Cache {
         server_side_encryption: Option<bool>,
     ) -> Result<Operator> {
         let mut builder = S3::default();
+        let user_agent = format!("{}/{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
+        let client_builder = ClientBuilder::new().user_agent(user_agent);
+        let client =
+            HttpClient::build(client_builder).map_err(|err| err.with_operation("S3::build"))?;
+        builder.http_client(client);
```
sylvestre, line 38: what is going to be the user agent ?

ajschmidt8: the user agent is a string whose value is `sccache/<CARGO_PKG_VERSION>` (e.g `sccache/0.7.7`).

is that what you're asking?

sylvestre: yeah :)

I wonder if we could add a test for this
@Xuanwo

Xuanwo: > I wonder if we could add a test for this

I'm guessing we don't need extra tests, the existing s3 test covers this.

Cargo.toml
```diff
@@ -164,7 +164,7 @@ memcached = ["opendal/services-memcached"]
 native-zlib = []
 oss = ["opendal/services-oss", "reqsign"]
 redis = ["url", "opendal/services-redis"]
-s3 = ["opendal/services-s3", "reqsign"]
+s3 = ["opendal/services-s3", "reqsign", "reqwest/blocking"]
```
Xuanwo, line 167: We don't need to enable `reqwest/blocking`

ajschmidt8: hmm. CI failed when I didn't import it. I tried this because I saw that `gcs` uses `reqwest` without any issues and it has `reqwest/blocking` enabled:

- https://github.com/mozilla/sccache/blob/ef52015daae706e55edde92a6f4dd70b9610378d/src/cache/gcs.rs#L21
- https://github.com/mozilla/sccache/blob/ef52015daae706e55edde92a6f4dd70b9610378d/Cargo.toml#L161

Do you know how to solve the error below without enabling `reqwest/blocking`?

- https://github.com/mozilla/sccache/actions/runs/8420567313/job/23055537391#step:3:483

ajschmidt8: I pushed [99fdeb4](https://github.com/mozilla/sccache/pull/2137/commits/99fdeb4431dfb813a800afbbba4df3ff493d9581), but admittedly I still don't know if that's correct.

I'm not a regular rust user.

Xuanwo: `reqwest` is an optional dep so we can't use it without declare it first. In this case, we can add `reqwest` in `s3` to tell rustc we want this dep.

Xuanwo (approved):
This PR looks good to me now!

Xuanwo:
cc @sylvestre, would you like to take another look?

src/cache/s3.rs
```diff
@@ -29,6 +31,11 @@ impl S3Cache {
         server_side_encryption: Option<bool>,
     ) -> Result<Operator> {
         let mut builder = S3::default();
+        let user_agent = format!("{}/{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
+        let client_builder = ClientBuilder::new().user_agent(user_agent);
+        let client =
+            HttpClient::build(client_builder).map_err(|err| err.with_operation("S3::build"))?;
```
sylvestre, line 37: could you please move these 3 lines into a function ?
thanks

src/cache/s3.rs
```diff
@@ -63,6 +66,12 @@ impl S3Cache {
     }
 }
 
+fn build_http_client() -> HttpClient {
```
sylvestre, line 69:
```suggestion
fn set_user_agent() -> HttpClient {
```

src/cache/s3.rs
```diff
@@ -29,6 +31,7 @@ impl S3Cache {
         server_side_encryption: Option<bool>,
     ) -> Result<Operator> {
         let mut builder = S3::default();
+        builder.http_client(build_http_client());
```
sylvestre, line 34:
```suggestion
        builder.http_client(set_user_agent());
```

sylvestre:
thanks

AJIOB:
@ajschmidt8 , just for note: do this really works in the prebuilt binary automatically (if you run the binary without `cargo run`)? Or we always need to define these Cargo variables externally?

ajschmidt8:
> @ajschmidt8 , just for note: do this really works in the prebuilt binary automatically (if you run the binary without `cargo run`)? Or we always need to define these Cargo variables externally?

rust's `env!` macro captures environment variables at compile time: https://doc.rust-lang.org/std/macro.env.html.

and Cargo sets these values at compile time.

so they won't need to be defined externally. it will work in the prebuilt binary automatically.

@AJIOB, just curious. do you intend to use this feature? if so, what's your use case?

AJIOB:
No, I just confused about backward compatibility with existing usages of S3 backend.

PR #2142, opened 2024-04-03
WIP GHA as local cache

Alphare:
This is a draft PR to implement the base for a fix for https://github.com/Mozilla-Actions/sccache-action/issues/81.

Some users have been implementing their own fix by going around the `sccache-action` and making their own cache in GHAc (https://github.com/Mozilla-Actions/sccache-action/issues/50), but I think that this allows us to be more generic by moving more of the complexity into `sccache` itself, which makes it easier for end-users and allows us to re-use the same principle for other remote storage systems in the future.

This will need to be accompanied by a separate PR in `sccache-action` to make use of the new config option(s) and the new `--update-cache` command.

I've listed TODOs in the code as questions for reviewers, hence the draft state.


Xuanwo:
Fascinating idea! Actually, I believe it could be a broader solution applicable to all storage services. For instance, downloading from S3/GCS/AZBlob should also be useful.

What do you think?

Alphare:
> Fascinating idea! Actually, I believe it could be a broader solution applicable to all storage services. For instance, downloading from S3/GCS/AZBlob should also be useful.
> 
> What do you think?

Indeed, I even mentioned this briefly in this PR's first message. I think this can make sense, but I didn't want to start with a more generic implementation before I could get a sense of how correct my solution is to this problem.

Is there any chance you could answer the few questions I've left as comments in the code?

Alphare:
I think this is ready for a first review. I'll try to somehow get `sccache-action` to test this in the meantime.

sylvestre:
I guess you saw that some jobs are failing

Cargo.toml
```diff
@@ -18,10 +18,10 @@ name = "sccache"
...
 required-features = ["dist-server"]
 
-[profile.release]
-codegen-units = 1
-lto = true
-strip = true
+# [profile.release]
+# codegen-units = 1
+# lto = true
+# strip = true
```
AJIOB, line 24: Looks like this file change is temporary. Revert it back, please

Alphare: Yep, sorry. It made the compilation extremely slow, but I should have reverted it.

Alphare:
After a long battle I finally managed in a (very very) hacky way to exercise this path¹.

I'll have to make adjustments here when debugging.

[1] https://github.com/Mozilla-Actions/sccache-action/actions/runs/8803242248/job/24160865310

src/cmdline.rs
```diff
@@ -274,6 +282,15 @@ pub fn try_parse() -> Result<Command> {
                 Ok(Command::DistAuth)
             } else if matches.get_flag("dist-status") {
                 Ok(Command::DistStatus)
+            } else if matches.get_flag("upload-cache") {
```
Xuanwo, line 285: Hi, I'm considering whether we can add this feature without altering the existing workflow. For instance, we could introduce a new configuration value in sccache, such as `remote_cache`. When the sccache daemon starts, it would load from the remote cache and then write back to it upon shutdown or at regular intervals.

This approach would eliminate the need for users to manually download or upload the cache.

What do you think?

Xuanwo: In this design, we just need to add a new input in `sccache-action` without changing the workflow.

Alphare: I can definitely see the "upon shutdown" version work and I could possibly already use it in this PR. I don't think it could work for regular intervals since GHA for example does not allow re-uploading to the same cache version+path.

Xuanwo: > I can definitely see the "upon shutdown" version work and I could possibly already use it in this PR. I don't think it could work for regular intervals since GHA for example does not allow re-uploading to the same cache version+path.

You are right. I'm guessing `regular intervals` is not that useful too. We can remove this part.

Xuanwo: > I can definitely see the "upon shutdown" version work and I could possibly already use it in this PR.

The key point I want to emphasize is not to add new command lines or alter the existing workflow that utilizes sccache. WDYT?

Alphare: Yep, agreed, I think it's a better design anyway. :)

sylvestre:
please reopen when ready

PR #2173, opened 2024-05-17
Disable preprocessor cache when doing distributed compilation

glandium:
At the moment, it doesn't work.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -382,6 +382,7 @@ where
         // Try to look for a cached preprocessing step for this compilation
         // request.
         let preprocessor_cache_mode_config = storage.preprocessor_cache_mode_config();
+        // Disable preprocessor cache when doing distributed compilation
         let mut preprocessor_key =
             if !may_dist && preprocessor_cache_mode_config.use_preprocessor_cache_mode {
                 preprocessor_cache_entry_hash_key(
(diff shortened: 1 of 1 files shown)


src/compiler/c.rs
```diff
@@ -382,20 +382,21 @@ where
...
-                &env_vars,
-                &absolute_input_path,
-                compiler.plusplus(),
-                preprocessor_cache_mode_config,
-            )?
-        } else {
-            None
-        };
+        let mut preprocessor_key =
+            if !may_dist && preprocessor_cache_mode_config.use_preprocessor_cache_mode {
```
sylvestre, line 386:
```suggestion
            // Disable preprocessor cache when doing distributed compilation
            if !may_dist && preprocessor_cache_mode_config.use_preprocessor_cache_mode {
```

src/compiler/c.rs
```diff
@@ -382,20 +382,21 @@ where
...
+                preprocessor_cache_entry_hash_key(
```
sylvestre, line 387:
```suggestion
        // Disable preprocessor cache when doing distributed compilation
        let mut preprocessor_key =
            if !may_dist && preprocessor_cache_mode_config.use_preprocessor_cache_mode {
                preprocessor_cache_entry_hash_key(
```

PR #2181, opened 2024-05-23
Make recent versions of clippy happy

glandium:


glandium:
The nightly failures come from something that was fixed in time 0.3.35.

PR #2199, opened 2024-06-01
Add custom zstd compression level

### LINKED ISSUE #2187: Variable cache compression level
Thank you for good project!

It will be good if I can change the compression level with Environment Variable.
Currently, the cache level seems constant 3, default level.
https://github.com/mozilla/sccache/blob/main/src%2Fcache%2Fcache.rs#L293-L293

I am using sccache in GitHub Actions to make cache of a firefox fork builds.
And the GitHub Actions Cache is only 10GB.
So the higher compression level is prefered because of lacking cache storage size.

Thank you for reading this issue!

nyanrus:
fix #2187 
Allow custom zstd cache compression level

### DIFF
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -290,7 +290,12 @@ impl CacheWrite {
         self.zip
             .start_file(name, opts)
             .context("Failed to start cache entry object")?;
-        zstd::stream::copy_encode(from, &mut self.zip, 3)?;
+
+        let compression_level = std::env::var("SCCACHE_CACHE_ZSTD_LEVEL")
+            .ok()
+            .and_then(|value| value.parse::<i32>().ok())
+            .unwrap_or(3);
+        zstd::stream::copy_encode(from, &mut self.zip, compression_level)?;
         Ok(())
     }
 


sylvestre:
Could you please add a test and mention this new variable in the doc? thanks

glandium:
This should alter the hash key, too.

nyanrus:
> This should alter the hash key, too.

This is just cache compression level and don't affect to decompression mechanism.
Is there need to change hash key?

nyanrus:
Thank you fro reply.

> Could you please add a test

It seems no tests for SCCACHE_DIR and SCCACHE_CACHE_SIZE that is basic feature like compression level.
What I should prove in the test?

> and mention this new variable in the doc

I'll add this to docs. Is this okay adding to Local storage option?

glandium:
> This is just cache compression level and don't affect to decompression mechanism.
> Is there need to change hash key?

I would say there is. If you're setting the variable for greater compression, you're probably expecting your cache to be more greatly compressed, not to continue containing whatever it contains right now. Of course, the caveat is that those lesser compressed objects will stay in the cache too, until the cache is full and they are pushed out as being the least recently used.

nyanrus:
> you're probably expecting your cache to be more greatly compressed

But when the hash changes and we use the more compressed cache without compression level variable, we cannot acquire less build time.

Rather than that, just adding a warning seems reasonable as "already compressed cache are not replaced."
because It will take the same time to make a new cache in both cases that are affecting to hash and not affecting to hash.
And to clarify "you should make a new cache."

and if it affects to hash, when you are using local cache or unlimited storage, it will take more cache size by caches not used.

sylvestre:
could you please add a test in the integration testsuite ?
https://github.com/mozilla/sccache/blob/main/.github/workflows/integration-tests.yml
thanks

nyanrus:
@sylvestre 
is the test enough?

PR #2211, opened 2024-06-21
Add cache hit percentage to stats

### LINKED ISSUE #2161: sccache lacks basic ratio (percentage) report in `--show-stats`
sccache have a very detailed report of some aspects when using `--show-stats` but I was shocked to see that no ratio of hit or misses is present.

This simply detail makes the whole stats feature a bit useless when the project have thousands of objects, since is very hard to compare big numbers.

I see that this was tried in #614 but abandoned. Since I don't know if you have some milestone to it, I suggest a bit of love into this area for better feature parity with ccache.

Saruniks:
Closes #2161

Inspiration from #614 

Feedback is welcome.

---

No compile requests:

```sh
Compile requests                      0
Compile requests executed             0
Cache misses                          0
Cache hits rate                       -
Average compiler                  0.000 s
Cache location                  Local disk: "/home/ubuntu-user/.cache/sccache"
Use direct/preprocessor mode?   yes
Version (client)                0.8.1
Max cache size                       10 GiB
```

Zero cache hits:

```sh
Compile requests                    397
Compile requests executed           319
Cache misses                        316
Cache misses (C/C++)                 52
Cache misses (Rust)                 264
Cache hits rate                    0.00 %
Cache hits rate (C/C++)            0.00 %
Cache hits rate (Rust)             0.00 %
Compilation failures                  3
Non-cacheable calls                  73
Non-compilation calls                 5
Average cache write               0.004 s
Average compiler                  0.370 s

Non-cacheable reasons:
crate-type                           50
unknown source language              18
-                                     3
-E                                    1
incremental                           1

Cache location                  Local disk: "/home/ubuntu-user/.cache/sccache"
Use direct/preprocessor mode?   yes
Version (client)                0.8.1
Cache size                          234 MiB
Max cache size                       10 GiB
```

Some cache hits:

```sh
Compile requests                   1189
Compile requests executed           957
Cache hits                          632
Cache hits (C/C++)                  104
Cache hits (Rust)                   528
Cache misses                        316
Cache misses (C/C++)                 52
Cache misses (Rust)                 264
Cache hits rate                   66.67 %
Cache hits rate (C/C++)           66.67 %
Cache hits rate (Rust)            66.67 %
Compilation failures                  9
Non-cacheable calls                 217
Non-compilation calls                15
Average cache write               0.004 s
Average compiler                  0.370 s
Average cache read hit            0.004 s

Non-cacheable reasons:
crate-type                          150
unknown source language              54
-                                     7
incremental                           3
-E                                    3

Cache location                  Local disk: "/home/ubuntu-user/.cache/sccache"
Use direct/preprocessor mode?   yes
Version (client)                0.8.1
Cache size                          234 MiB
Max cache size                       10 GiB
```

`sccache -- --show-adv-stats`:

```sh
Compile requests                     401
Compile requests executed            319
Cache hits                           316
Cache hits (c [gcc])                  52
Cache hits (rust)                    264
Cache misses                           0
Cache hits rate                   100.00 %
Cache hits rate (c [gcc])         100.00 %
Cache hits rate (rust)            100.00 %
Cache timeouts                         0
Cache read errors                      0
Forced recaches                        0
Cache write errors                     0
Compilation failures                   3
Cache errors                           0
Non-cacheable compilations             0
Non-cacheable calls                   77
Non-compilation calls                  5
Unsupported compiler calls             0
Average cache write                0.000 s
Average compiler                   0.000 s
Average cache read hit             0.002 s
Failed distributed compilations        0

Non-cacheable reasons:
crate-type                            51
unknown source language               18
-                                      5
incremental                            2
-E                                     1

Cache location                  Local disk: "/home/ubuntu-user/.cache/sccache"
Use direct/preprocessor mode?   yes
Version (client)                0.8.1
Cache size                           235 MiB
Max cache size                        10 GiB
```

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -38,7 +38,7 @@ use futures::{future, stream, Sink, SinkExt, Stream, StreamExt, TryFutureExt};
 use number_prefix::NumberPrefix;
 use serde::{Deserialize, Serialize};
 use std::cell::Cell;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::future::Future;
@@ -1510,6 +1510,17 @@ impl Default for ServerStats {
     }
 }
 
+macro_rules! set_percentage_stat {
+    ($vec:ident, $count:expr, $sum:expr, $name:expr) => {
+        if $sum == 0 {
+            $vec.push(($name.to_string(), "-".to_string(), 0));
+        } else {
+            let ratio = $count as f64 / $sum as f64;
+            $vec.push(($name.to_string(), format!("{:.2} %", ratio * 100.0), 2));
+        }
+    };
+}
+
 impl ServerStats {
     /// Print stats to stdout in a human-readable format.
     ///
@@ -1570,6 +1581,9 @@ impl ServerStats {
             set_lang_stat!(stats_vec, self.cache_hits, "Cache hits");
             set_lang_stat!(stats_vec, self.cache_misses, "Cache misses");
         }
+
+        self.set_percentage_stats(&mut stats_vec, advanced);
+
         set_stat!(stats_vec, self.cache_timeouts, "Cache timeouts");
         set_stat!(stats_vec, self.cache_read_errors, "Cache read errors");
         set_stat!(stats_vec, self.forced_recaches, "Forced recaches");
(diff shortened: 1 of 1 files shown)


AJIOB:
Hi @Xuanwo,

Is it OK to you?

src/server.rs
```diff
@@ -1510,6 +1510,17 @@ impl Default for ServerStats {
     }
 }
 
+macro_rules! set_percentage_stat {
```
sylvestre, line 1513: why a macro and not a function?

sylvestre:
Could you please add a test to verify the field is present? Thanks

src/server.rs
Saruniks: Picked up code from https://github.com/mozilla/sccache/pull/614 and experienced tunnel vision.

Function is probably a better choice here indeed.

src/server.rs
```diff
@@ -1978,3 +2058,108 @@ fn waits_until_zero() {
...
+            },
+            ..Default::default()
+        };
+
+        let mut writer = StringWriter::new();
+        stats.print(&mut writer, true);
+
+        let output = writer.get_output();
+
+        assert!(output.contains("Cache hits rate                        -"));
```
Saruniks, line 2160: Interesting case where `PerLanguageCount::all()` uses `counts.values().sum()` and not `self.adv_counts.values().sum()` in case of printing advanced stats.

In practice should never happen because `PerLanguageCount::increment()` is used to set the counts.

src/server.rs
```diff
@@ -1700,7 +1780,7 @@ impl ServerInfo {
 
     /// Print info to stdout in a human-readable format.
     pub fn print(&self, advanced: bool) {
-        let (name_width, stat_width) = self.stats.print(advanced);
+        let (name_width, stat_width) = self.stats.print(&mut StdoutServerStatsWriter, advanced);
```
Saruniks, line 1783: Intrusive way to test the code. If anyone has an idea on how to test it in a better way it would be helpful to know.

I was thinking about writing similar test like in `sccache_cargo.rs` (using `Command::new()` and capturing the output), but I was hesitant because the test would not be fast.

I was thinking about refactoring this code to free functions and testing the free functions but then it would potentially not cover the whole use case.

Saruniks:
Hello @AJIOB 

Ready from my side. Unless perhaps someone has comments on the code and tests.

AJIOB:
Hi @sylvestre & @Xuanwo 
Can we merge it?

Saruniks:
Ok. I'll add the integration test

Saruniks:
Hello,

a) I added integration tests. Not to the GitHub Actions workflow like mentioned above, but "Rust" ones (see [tests/cache_hit_rate.rs](https://github.com/Saruniks/my-custom-sccache/blob/add-hit-miss-ratio-to-stats/tests/cache_hit_rate.rs)).

I could add extra jobs to the GitHub Actions workflow, no problems. But I thought that what I developed could be sufficient as no integration with external tool needs to be tested. Let me know if you think otherwise.

b) Regarding the json stats:
Apologies for not mentioning it before, but I didn't develop cache hit rate display for json stats.
 
The problem is that json stats is just serialized structs. When "StatsFormat::Text" stats are achieved by using "complex" print method.

```Rust
    match fmt {
        StatsFormat::Text => stats.print(advanced),
        StatsFormat::Json => serde_json::to_writer(&mut io::stdout(), &stats)?,
    }
```

I am still not sure what would be the best way to implement cache hit rate for json stats. Maybe custom serialize implementation would work (and not derive trait like currently), but then it could be considered magic-like. But then calculating cache hit rate on every stats change could be considered too wasteful (or considered as calculated value anti-pattern). 

So not sure how to proceed with json stats.

tests/helpers/mod.rs
Saruniks, line 1: No new code in this file, only moved things around for deduplication

Saruniks:
I fixed the issue with failing tests. Checks should pass now.

PR #2217, opened 2024-07-05
Fix non-strict HIP device lib order

Snektron:
In https://github.com/mozilla/sccache/pull/2045 initial support for HIP was added. While trying it out, I noticed that different runs across different machines did not yield an expected cache hit. After some investigation, it turns out that the list of bitcode device libraries is not sorted after discovering them from the file system, and this resulted in a different order across those machines. I've added a fix that simply sorts those libraries after discovering them, that seems to be similar to how its handled elsewhere.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -586,9 +586,11 @@ jobs:
     env:
       # SCCACHE_GHA_ENABLED: "on"
       ROCM_PATH: "/opt/rocm"
+      RANDOMIZE_READDIR_LOG: "/tmp/readdir.log"
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Clone repository
+        uses: actions/checkout@v4
 
       # I don't want to break the cache during testing. Will turn on after I
       # make sure it's working.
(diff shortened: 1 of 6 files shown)


sylvestre:
could you please add a test for this ? thanks

Snektron:
bump

sylvestre:
still need tests, sorry :)

Snektron:
Seems that fuse doesn't work in github actions. I guess the only other approach swapping out calls to `fs::read_dir` with something that can be mocked.

Snektron:
Hmm, it might be possible to LD_PRELOAD a library which returns random results from readdir...

Snektron:
@sylvestre What do you think about this approach? It seems to be working fine in [our fork's CI](https://github.com/StreamHPC/sccache/actions/runs/12034620150/job/33552180304), though I feel like theres some confirmation missing that the shim is actually running. One details is that I've had to set LD_PRELOAD separately on each run, it seems that using `$GITHUB_ENV` will make github try to load the shim before its set up correctly in the container. I'm not really sure how that works...

sylvestre:
sorry but could you please rebase it? thanks

Snektron:
Alright, I think this is finally really ready now. I've updated the readdir randomizer with a comment that describes how and why it works the way it does, and added a simple log-based method to double check that the shim is actually running. Sorry for taking so long to finalize this...

PR #2226, opened 2024-07-17
Remove the heapsize feature

glandium:
It is not exposed, so never compiled. It's leftover from when lru_disk_cache was a separate crate.


Xuanwo (approved):
Thanks!

PR #2231, opened 2024-07-22
Add timeout to tcp connections

Systemcluster:
Depending on system configuration, waiting for a connection error when checking for a running server can take a long time. In my local WSL installation, this is causing sccache to take several minutes to start up.

This PR replaces the two calls to `connect` with `connect_timeout` and adds a `SCCACHE_CONNECTION_TIMEOUT` environment variable for configuration. The code for reading the env var is copied from `server.rs`: https://github.com/mozilla/sccache/blob/9958e7c8a023b39df2c56073748d37a77febac03/src/server.rs#L94-L101


src/client.rs
```diff
@@ -19,6 +19,18 @@ use byteorder::{BigEndian, ByteOrder};
 use retry::{delay::Fixed, retry};
 use std::io::{self, BufReader, BufWriter, Read};
 use std::net::TcpStream;
+use std::time::Duration;
+use std::{env, net::ToSocketAddrs};
+
+const DEFAULT_CONNECTION_TIMEOUT: u64 = 20;
```
Systemcluster, line 25: This could be much lower, but leaving it this high aligns it more closely with the current behavior. I'm running it with `2` locally.

Systemcluster:
@Xuanwo could you review this? I don't think it's a controversial change, but I can revert to calling `connect` without `timeout` if the environment variable isn't set otherwise.

docs/Configuration.md
```diff
@@ -129,6 +129,7 @@ configuration variables
 * `SCCACHE_CONF` configuration file path
 * `SCCACHE_CACHED_CONF`
 * `SCCACHE_IDLE_TIMEOUT` how long the local daemon process waits for more client requests before exiting, in seconds. Set to `0` to run sccache permanently
+* `SCCACHE_CONNECTION_TIMEOUT` how long clients should try to connect to the server before continuing, in seconds.
```
Xuanwo, line 132: Overall, this idea looks good to me. However, I think it should be a config value in the `[dist]` section instead of an environment variable. Most distribution configurations aren't exposed to the environment for now.

The current name is a bit confusing from my point of view. I would expect something like `tcp_connect_timeout`.

Xuanwo (commented):
Thanks and sorry for the late, I only have some small questions.

Xuanwo (commented):
Thanks and sorry for the late, I only have some small questions.

docs/Configuration.md
Systemcluster: Should `connect_to_server` in `src/client.rs` and the functions that use it (`connect_or_start_server` in `src/command.rs`) take the config struct as an argument?

drahnr: It depends, I'd say just the timeout is enough. If we start to see a need for a lot more config values, sub-structs might do as well.

src/client.rs
```diff
@@ -19,6 +19,18 @@ use byteorder::{BigEndian, ByteOrder};
...
+
+/// Get the time clients should try to connect to the server before continuing, in seconds.
+pub(crate) fn get_connection_timeout() -> u64 {
+    env::var("SCCACHE_CONNECTION_TIMEOUT")
+        .ok()
+        .and_then(|s| s.parse().ok())
```
drahnr, line 31: I think warning on an invalid value as early as possible would be good.

docs/Configuration.md
ahartmetz: I think SCCACHE_CONNECT_TIMEOUT is better. The interpretations that first come to mind for me are:
- SCCACHE_CONNECT_TIMEOUT: the timeout for making a connection
- SCCACHE_CONNECTION_TIMEOUT: the timeout for an existing connection

src/client.rs
ahartmetz: Something like 5, 8, or 10 seems appropriate IMO. No connection between any two computers takes that long unless there is very heavy bufferbloat, packet loss, or an interplanetary connection.

PR #2236, opened 2024-07-25
deps: update winapi-util

ognevny:
starting with v0.1.7 winapi-util uses windows-rs instead of winapi

P.S. I tried to update another deps which pull winapi. in all cases there are some reason that blocks it


Xuanwo (approved):
Thanks a lot!

PR #2239, opened 2024-07-31
Add ability to specify server port from cli

just-an-engineer:
It now looks to see if a port number was specified on the command line, and tries to use that before looking at the config file or environment variables
Since --start-server calls InternalStartServer under the hood, I have it starting it with the port number specified as the environment variable, although I can change it to also take it from the cli.
Added tests and updated the readme for usages

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,9 @@ sccache supports gcc, clang, MSVC, rustc, [NVCC](https://docs.nvidia.com/cuda/cu
 
 If you don't [specify otherwise](#storage-options), sccache will use a local disk cache.
 
-sccache works using a client-server model, where the server runs locally on the same machine as the client. The client-server model allows the server to be more efficient by keeping some state in memory. The sccache command will spawn a server process if one is not already running, or you can run `sccache --start-server` to start the background server process without performing any compilation.
+sccache works using a client-server model, where the server runs locally on the same machine as the client. The client-server model allows the server to be more efficient by keeping some state in memory. The sccache command will spawn a server process if one is not already running, or you can run `sccache --start-server[=port]` to start the background server process without performing any compilation.
 
-You can run `sccache --stop-server` to terminate the server. It will also terminate after (by default) 10 minutes of inactivity.
+You can run `sccache --stop-server[=port]` to terminate the server. It will also terminate after (by default) 10 minutes of inactivity.
 
 Running `sccache --show-stats` will print a summary of cache statistics.
 
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -62,9 +62,9 @@ pub enum Command {
     /// Run background server.
     InternalStartServer,
     /// Start background server as a subprocess.
-    StartServer,
+    StartServer(Option<u16>),
     /// Stop background server.
-    StopServer,
+    StopServer(Option<u16>),
     /// Zero cache statistics and exit.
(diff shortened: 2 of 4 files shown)


sylvestre:
What about configuring it in a configuration file?

just-an-engineer:
Yeah, I can do that. I'll add that as well

just-an-engineer:
Alright, just added that in. Currently, this is the order of precedence in the code of where it looks for the port: CLI, config, env var, and lastly, the hardcoded default port of 4226. No particular reason for that order, so that can be changed up if desired.

README.md
```diff
@@ -114,9 +114,9 @@ sccache supports gcc, clang, MSVC, rustc, [NVCC](https://docs.nvidia.com/cuda/cu
 
 If you don't [specify otherwise](#storage-options), sccache will use a local disk cache.
 
-sccache works using a client-server model, where the server runs locally on the same machine as the client. The client-server model allows the server to be more efficient by keeping some state in memory. The sccache command will spawn a server process if one is not already running, or you can run `sccache --start-server` to start the background server process without performing any compilation.
+sccache works using a client-server model, where the server runs locally on the same machine as the client. The client-server model allows the server to be more efficient by keeping some state in memory. The sccache command will spawn a server process if one is not already running, or you can run `sccache --start-server[=port]` to start the background server process without performing any compilation.
```
sylvestre, line 117: I would prefer we don't have this syntax
`sccache --start-server[=port]`

just-an-engineer: Any preferences?

just-an-engineer: I can have it `sccache --start-server [port]` if that's any better

sylvestre: why not --port ?

just-an-engineer: Because I figure something like `--port` will only ever show up when you're starting or stopping the server, making it a subset of `--start-server` (or stop-server), I figure it's better to not have a dedicated `--port`, and just have `[port]` be an optional argument to `--start-server` (or stop, again) rather than have another optional flag. Because the usage would be the exact same in either case, except when they want to specify it, they just need to type more. Not a big reason, but I think that's more or less why, from my perspective.
What do you think?

just-an-engineer:
@sylvestre, I'm a bit new to the whole process. At this point, do I select "Update branch"? Do I want for someone else to do something?

just-an-engineer:
I realized I hadn't done anything for commands outside of start or stop. So I added in basic functionality for a client cache (which we can expand to other things), and added tests for that. I also corrected a prior test that was incorrect, I just wasn't thinking properly for that one. I also expanded it to test that a compile request works for a server started on a non-default port.
The client cache will also help it be more ergonomic for users because they won't have to specify an environment variable alongside every `sccache` command they run, because if they started a server on a different port, they'd have to use the env vars to see its stats, run a compile request, etc. This can make it a bit more streamlined by assuming that you'll have a single compile server you primarily use (if that's not the case, this may break down, but it's not like it was easier beforehand), and once you specify it, that's the one you want to see the stats for, run compile jobs from, and shutdown

README.md
sylvestre: i am really not a fan of the --start-server[=port] syntax :/
@glandium WDYT ?

sylvestre:
As nobody else asked for it, i don't think it is a very popular feature. Closing, sorry!

PR #2241, opened 2024-08-02
Replace rouille with warp

Saruniks:
Part of porting changes from cachepot https://github.com/mozilla/sccache/issues/1620, https://github.com/paritytech/cachepot/pull/131


Xuanwo:
Wow, great work! Thanks in advance; I'll start reviewing it later.

Saruniks:
I'll add [montekki](https://github.com/montekki) as a co-author.

[montekki](https://github.com/montekki), please let us know if you have any comments on this PR or if you would like to takeover the work.

---

I'm not sure how to prepare this for a merge. It was relatively easy to solve merge conflicts and make the CI pass, but there's a lot of changes and there's a risk of fallout.
I'll give an update after testing this on my setup.

---

I wonder what's up with the freebsd tests though.

Saruniks:
I did some testing. For some reason `sccache_dist scheduler` doesn't work without `SCCACHE_NO_DAEMON=1`.

I tested and it even works like this:
`SCCACHE_NO_DAEMON=1 sccache-dist scheduler --config /etc/sccache/scheduler.conf &`

This is the same reason why freebsd ci doesn't pass. I'll try to fix this issue.

Saruniks:
I found that I need to create a tokio::Runtime instance manually after call to daemonize, instead o using the `#[tokio::main]` attribute.

Xuanwo:
All unit test for freebsd passed, we only have a `No distributed compilations succeeded`. Seems the dist-server or dist-client doesn't start correctly on freebsd.

Saruniks:
I fixed freebsd tests by creating `tokio::Runtime` manually instead of using `#[tokio::main]`:
```Rust
    let runtime = Runtime::new().context("Failed to create Tokio runtime")?;
    runtime.block_on(async { http_scheduler.start().await })?;
```

Xuanwo (approved):
Thank you, @Saruniks, for making this happen, and also thanks to @montekki for the original work. Overall, this PR looks good to me.

Some details inside this PR can be polished in the future.

Xuanwo:
cc @sylvestre would you like to take another look?

src/util.rs
```diff
@@ -938,13 +938,50 @@ pub fn daemonize() -> Result<()> {
...
+// to a `native-tls` backend seems to be the only way to disable the `SNI` feature
+// in `request`. Also all the intended root certificates have to be passed
+// to the `native-tls` builder; using the higher-level `request` bulider API will
+// not work.
+//
+// More context:
+// https://github.com/seanmonstar/reqwest/issues/1328
+// https://github.com/briansmith/webpki/issues/54
+#[cfg(any(feature = "dist-client", feature = "dist-server"))]
+pub fn native_tls_no_sni_client_builder<I, T>(root_certs: I) -> Result<reqwest::ClientBuilder>
```
Saruniks, line 962: [Xuanwo](https://github.com/Xuanwo) I just noticed that `native_tls_no_sni_client_builder` method might not be needed anymore as https://github.com/seanmonstar/reqwest/issues/1328 was already fixed

drahnr:
Gentle ping

ahartmetz:
What's the status here? In case it was blocked on unrelated CI failures, these should be fixed now.

sylvestre:
plenty of files are conflicting

PR #2243, opened 2024-08-07
Add windows arm64 to the CI

hjyamauchi:


hjyamauchi:
CC @Xuanwo

hjyamauchi:
This is based on the discussion https://github.com/mozilla/sccache/discussions/2223

sylvestre:
AFAIK, it is building but not running tests, am I correct?

hjyamauchi:
I didn't realize there was a test step. So, I believe so. Do we want to enable testing?

hjyamauchi:
AFAICT, windows arm64 runners aren't available on github yet. So testing for windows arm64 isn't currently viable?

sylvestre:
correct but not an issue :)

hjyamauchi:
@sylvestre Great, thank you!
Do you have an idea when the next release that includes a windows arm64 binary may happen?

Issue #2237, opened 2024-07-25
sccache --stop-server does not block until the server has stopped


Snektron:
I'm using sccache in a CI environment, where at the start of the CI job the sccache server is first started and then stopped at the end. I'm running into some issues where the CI runtime (gitlab in this case) kills the CI job before the final object file is uploaded to the cache, as `sccache --stop-server` runs asynchronously. For now, I've solved this by manually waiting on the sccache process by running `timeout 11 tail --pid=$(pgrep sccache) -f /dev/null` right after invoking `--stop-server`, but it would be nice if there was a built-in method to do this. For example, something like `sccache --stop-server-and-wait`.

just-an-engineer:
Hey, I just made a PR so you can specify `server_shutdown_timeout_ms` from the config file. This will let you set it to a crazy high number if you really wanted to, and allow any compile job to finish. Testing on my own machine, I found that if there was no job running, it would shut down immediately upon receiving the signal. But upon a long-running job, it would either wait for it to end and then stop, or timeout (which was previously a hardcoded default of 10 seconds).
Can you try it out and let me know if that fixes the issue, or if not, what's going on so I can help out more?

just-an-engineer:
I marked it as a draft because I still need to add tests, which I'll do tonight. So if you want to do that today, you may need to switch over to my branch momentarily.

Snektron:
> Can you try it out and let me know if that fixes the issue, or if not, what's going on so I can help out more?

Well not exactly: In my case the problem was not that the upload task was exceeding the timeout (though its very useful to be able to configure that), but that there is no built-in way to block the current shell until the background server has shut down.

PR #2235, opened 2024-07-25
chore: Bump deps to latest

Xuanwo:
This PR will bump deps to latest.

PR #2246, opened 2024-08-12
Specify server timeout

just-an-engineer:
This should fix issue #2237, where in the case a compilation is still ongoing when a shutdown request is sent, and it's a larger/complex codebase that takes longer than the default 10 seconds, the user can now specify it in their config file.

However, I'd like to also propose a breaking change for config files, moving server_startup_timeout_ms and server_shutdown_timeout_ms to a specific timing struct. I left the code in there to do this, just commented out. I'd like to make this change next time we bump up a major version. 
Although I can remove the comments to clean up the code until then. But what's the best practice for marking a milestone? Should I raise an issue?


just-an-engineer:
If a breaking change will occur in 0.9, like @Xuanwo had mentioned in PR #2235, talking about the issue to replace `rouille` with `axum`, then I can go ahead and change this PR to include that breaking change for the config file, and we can merge this in with 0.9. _If_ that will occur. Otherwise, it's ready to go, outside of the comment I previously wrote

sylvestre:
seems that rustfmt needs to be run 
also, could you please document this into the readme file? thanks

sylvestre:

PR #2251, opened 2024-08-31
fix: Force disable "warnings as errors" flag for CL preprocessing

### LINKED ISSUE #2250: MSVC: preprocessing is failed while compiling is success
Minimal reproducible example:

```cpp
#pragma warning(disable : 4002)

// C4002b.cpp
// compile with: /W1 /WX
#define F(x, y)

int main()
{
    F(2, , , , , , 3, , , , , , ) // C4002
    // Try the following line instead:
    // F(2,3)

    return 0;
}
```

Tested with MSVC 2017, 2022.

The problem is that sccache without direct mode doesn't remove `/WX` flag when adding `/EP` one.

AJIOB:
Fix #2250


sylvestre:
LGTM, thanks

PR #2257, opened 2024-09-11
Implement features required by Xcode

scoopr:
These commits allow me to use sccache with Xcode.

Firstly, I just mark `-index-store-path` as `Hard`, as it is easy to configure Xcode not to use that. Not sure how impossible that would be to implement otherwise, didn't look in to it.

And secondly, I'm artifacting the dep (`.d`) files and the `-serialize-diagnostics` (`.dia`) files, as Xcode seems to always request them, and the build will fail without them.

When using it, I need to run the sccache server outside of Xcode, as otherwise it seems that Xcode will keep waiting for the daemon to timeout.

Then I have configured in Xcode build settings the CC/CXX variables, currently to a launcher script that sets the env variables and calls sccache with the right arguments (I am using it through cmake though). The setup might be improved now that I got it to work.


sylvestre:
thanks
it would be nice to add some docs + add a test if possible
thanks

scoopr:
Updated the tests and added some docs.

If there are additional concerns, it might be few weeks until I can get back to this.

sylvestre:
About test, I was thinking about something like in https://github.com/mozilla/sccache/blob/main/.github/workflows/integration-tests.yml

scoopr:
Sorry about the churn for getting that integration test running, I don't think I have a good way to test it locally, so I'm developing it a bit blind.

sylvestre:
no worries. Could you please write a tiny PR that I could merge quickly so that you get the CI without approval?

sylvestre:
for example, typo fix like:
src/lru_disk_cache/mod.rs:142: maintence ==> maintenance

sylvestre:
well, looks like it worked :)

scoopr:
Yep, the output now looks like what I expected. I don't think I have anything to add it, so on my side its ready, if no other concerns arise.

.github/workflows/integration-tests.yml
```diff
@@ -894,3 +894,53 @@ jobs:
...
+      - name: Install rust
+        uses: ./.github/actions/rust-toolchain
+        with:
+          toolchain: "stable"
+
+      - name: Build sccache
+        run: |
+          cargo build
+
+      - name: Create wrapper script
```
sylvestre, line 915: could you please add a comment explaining why a wrapper script is needed

scoopr: Its actually explained in the docs (https://github.com/mozilla/sccache/pull/2257/commits/dfe8c5be44d421452c36a90d5becc1a32e23d395#diff-8f7ae669918b1de5a217f0f28f8becd7d0f92625c4b8334b733315b8199dd102R47), but I guess I could add a comment here.

The issue is that I can't just set `CC="sccache clang++"`, because it interprets the whole `CC` variable as a command, so its searching for a file called `sccache clang++` (so space is part of the command name). If you have other ideas for a workaround, I'm all ears.

scoopr:
It seems the [continuous job](https://github.com/mozilla/sccache/actions/runs/11139843685/job/30957452532) failed, is it due to something I did, or is it something else?

sylvestre:
nope, we have some intermittent sometimes (when the cache is full i guess)

PR #2264, opened 2024-09-29
[Feature](INF-2296) infra: fix sccache bug for dwo file generate

zhouronghua:
when output dir changed, and -g -gsplit-dwarf is enabled, new object file should be gererated, but not old cached object.

case:

1 echo "int test(){}" > test.cc

2 mkdir o1 o2

3 sccache g++ -c -g -gsplit-dwarf test.cc -o o1/test.o

4 sccache -z

5 sccache g++ -c -g -gsplit-dwarf test.cc -o o2/test.o

6 strings o2/test.o |grep o2/test.dwo


sylvestre:
the patch doesn't apply
and what is "INF-2296" ? :)

sylvestre:
the patch has nothing to do with split dwarf.
please reopen when it is fixed

zhouronghua:
sorry for I have not replied in holiday.

I didn't find way to reopen the PULL REQUEST, so I create another PR in https://github.com/mozilla/sccache/pull/2264

PR #2266, opened 2024-10-02
Simplify Xcode CI / Docs

scoopr:
I happened to stumble upon the `C_COMPILER_LAUNCHER` Xcode attribute,
which fits the bill perfectly, so use that instead.


scoopr:
Meh, just as we got the #2257 merged, some Xcode warning exposes that it supports `C_COMPILER_LAUNCHER` attribute, which is a lot simpler than the wrapper script

scoopr:
Aha, but actually now it enables the response file use, so that needs to be disabled, I was a bit hasty on this PR, it needs a bit more work to be correct.

scoopr:
Alright, when using `C_COMPILER_LAUNCHER`, instead of wrapper script for `CC`, it seems to be passing a bit more flags (as it knows the compiler better), So I added it to passthrough the `-ivfstatcache` path. Also noticed a single `-x objective-c++-header` from my project, so went ahead and added that as supported too.

I think it is much better like this.

sylvestre:
nope, thanks for the ping :)

PR #2269, opened 2024-10-07
Add support for caching target json specs for Rust compilation.

### LINKED ISSUE #2174: Caching custom Rust targets
I have a very large Rust project that uses a custom target `.json`, and I would like to add support for this to `sccache`. Is this possible? If so, could anyone give me advice on the implementation as I'm not familiar with the `sccache` codebase.

ColinFinck:
This has been implemented in a backward-compatible way:
- `--target` arguments specifying target names are still passed as arguments just like before.
- `--target` arguments specifying a json file path have previously not been cached at all. The json file is now hashed like other external file dependencies. 

Fixes #2174

### DIFF
--- a/docs/Rust.md
+++ b/docs/Rust.md
@@ -7,6 +7,5 @@ sccache includes support for caching Rust compilation. This includes many caveat
 * Compilation from stdin is not supported, a source file must be provided.
 * Values from `env!` require Rust >= 1.46 to be tracked in caching.
 * Procedural macros that read files from the filesystem may not be cached properly
-* Target specs aren't hashed (e.g. custom target specs)
 
 If you are using Rust 1.18 or later, you can ask cargo to wrap all compilation with sccache by setting `RUSTC_WRAPPER=sccache` in your build environment.
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -3736,4 +3736,47 @@ proc_macro false
 
         assert_eq!(h.profile, Some("foo-a1b6419f8321841f.profraw".into()));
     }
+
+    #[test]
+    fn test_parse_target() {
+        // Parse a --target argument that is a string (not a path to a .json file).
+        let h = parses!(
+            "--crate-name",
+            "foo",
+            "--crate-type",
+            "lib",
+            "./src/lib.rs",
+            "--emit=dep-info,link",
+            "--out-dir",
+            "/out",
+            "--target",
+            "string"
+        );
+        assert!(h.arguments.contains(&Argument::WithValue(
+            "--target",
+            ArgData::Target(ArgTarget::Name("string".to_owned())),
+            ArgDisposition::Separated
+        )));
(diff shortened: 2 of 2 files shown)


sylvestre:
Could you please add a test here to make sure we don't regress?
https://github.com/mozilla/sccache/blob/main/.github/workflows/integration-tests.yml

ColinFinck:
Hey @sylvestre, thanks a lot for your quick response! I've checked the integration-tests.yml file you gave me, but it only contains a single test for each platform/compiler, not for individual flags of the Rust compiler. I'm willing to add a test for the newly supported flag, but I need a good example on how sccache currently does per-flag tests.

ColinFinck:
@sylvestre I've rebased to latest `main` and added a test in 616dd8ce2bff26cb658c93824ccb119497fe42b0

ColinFinck:
Bump

sylvestre:
sorry, i missed your comment. CI started!

ColinFinck:
@sylvestre Thanks for getting back to this PR. However, I'm still seeing "2 workflows awaiting approval" in GitHub.

sylvestre:
done
could you please add a test in 
https://github.com/mozilla/sccache/blob/main/tests/system.rs
too ?

ColinFinck:
> could you please add a test in
> https://github.com/mozilla/sccache/blob/main/tests/system.rs
> too ?

@sylvestre There is currently not a single test in that file which exercises the Rust compiler. This is only done in `src/compiler/rust.rs`. Which is why I added my test there.
Multiple C compilers are tested in `tests/system.rs`, but not `rustc`. Adding a test framework for the Rust compiler to that file exceeds the scope of this PR and goes beyond my limited knowledge of sccache.

I hope we're able to merge the few lines of this PR without requiring an entirely new test framework for the Rust compiler first.

sylvestre:
> @sylvestre There is currently not a single test in that file which exercises the Rust compiler. 

@ColinFinck we have proper Rustc testing:

See:
https://github.com/mozilla/sccache/blob/main/tests/sccache_cargo.rs
and
https://github.com/mozilla/sccache/blob/main/tests/sccache_rustc.rs

ColinFinck:
> > @sylvestre There is currently not a single test in that file which exercises the Rust compiler.
> 
> @ColinFinck we have proper Rustc testing:
> 
> See: https://github.com/mozilla/sccache/blob/main/tests/sccache_cargo.rs and https://github.com/mozilla/sccache/blob/main/tests/sccache_rustc.rs

@sylvestre These are different files. I already had a look at them, but the first one just contains very high-level tests of running `cargo` and checking for cache hits and misses. The second file tests against a mock `rustc` and not the real one. None of these files are prepared for adding a low-level test of validating a single parameter (`--target`) passed to a real `rustc` compiler.

Which is why I added a test for `--target` to `src/compiler/rust.rs`. This is as much as I can do given the current state of the sccache testing framework.

PR #2271, opened 2024-10-09
Fix: fix sccache bug for dwo file generate

zhouronghua:
When output dir changed, and -g -gsplit-dwarf is enabled, 
because the object file have link to the dwo file (include directories),
the new object file should be gererated, but not old cached object.
In this patch, we add a new precompile macro, for example, "-D_gsplit_dwarf_path=foo.dwo", to force sccache to generate new object file depend on the output directories.

we can check if it work by this case:
case:
```bash
echo "int test(){}" > test.cc
mkdir o1 o2
sccache g++ -c -g -gsplit-dwarf test.cc -o o1/test.o
sccache g++ -c -g -gsplit-dwarf test.cc -o o2/test.o
strings o2/test.o |grep o2/test.dwo
```
The new o2/test.o should contains new link to o2/test.dwo, but not cached o1/test.dwo

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -575,6 +575,9 @@ where
     };
     if split_dwarf {
         let dwo = output.with_extension("dwo");
+        common_args.push(OsString::from(
+            "-D_gsplit_dwarf_path=".to_owned() + dwo.to_str().unwrap(),
+        ));
         // -gsplit-dwarf doesn't guarantee .dwo file if no -g is specified
         outputs.insert(
             "dwo",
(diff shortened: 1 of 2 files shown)


sylvestre:
Could you please add a test to make sure we don't regress in the future? thanks

zhouronghua:
"test_parse_arguments_split_dwarf" this case has been modified to make it work all the time

sylvestre:
this isn't enough. I would like to see a test reproducing:
```
echo "int test(){}" > test.cc
mkdir o1 o2
sccache g++ -c -g -gsplit-dwarf test.cc -o o1/test.o
sccache g++ -c -g -gsplit-dwarf test.cc -o o2/test.o
strings o2/test.o |grep o2/test.dwo
```

also, you didn't updated the clang.rs file, is that expected ?

zhouronghua:
ok, I will add another test for it, but it need time.

parse_arguments in clang.rs will call parse_arguments in gcc.rs, so no need to change them all.

sylvestre:
see https://github.com/mozilla/sccache/blob/main/tests/system.rs for example

zhouronghua:
thanks, that's really help.

sylvestre:
Needs rustfmt

zhouronghua:
done

sylvestre:
Thanks

PR #2282, opened 2024-11-06
feat: Use profile-generate to replace outdated -Zprofile options

Xuanwo:
Related to https://github.com/mozilla/grcov/issues/1240

I start this PR to fix the main branch's CI failure, but find out that we need to use profile-generate to replace outdated -Zprofile options.

Refer https://doc.rust-lang.org/rustc/profile-guided-optimization.html for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,7 +239,7 @@ jobs:
         env:
           CARGO_INCREMENTAL: "0"
           RUSTC_WRAPPER: ""
-          RUSTFLAGS: "-Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off"
+          RUSTFLAGS: "-Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Cprofile-generate=target/debug"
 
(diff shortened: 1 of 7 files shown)


Xuanwo:
cc @sylvestre would you like to take a look? This PR will address the failing CI on main.

Xuanwo:
Snap workflow has been disable since it never works.

sylvestre:
thanks !

glandium:
I think we should revert this change because:
- It broke support for command lines that contain any other -Z flag
- It broke support for -Zprofile, which may still be used with older version of rustc
- It tracks a profraw file that is irrelevant: it would be the equivalent to gcda, not gcno. And actually it never exists, because even when the program is run and a profraw file is generated, its name is different.

Xuanwo:
Hi, I reviewed all the documentation for [Instrumentation-based Code Coverage](https://doc.rust-lang.org/rustc/instrument-coverage.html#instrumentation-based-code-coverage) and [Profile-guided Optimization](https://doc.rust-lang.org/rustc/profile-guided-optimization.html#profile-guided-optimization) and now understand the mistake pointed out by @glandium. (Thank you for that!) I’m currently working on it and will restore everything back to normal.

PR #2284, opened 2024-11-07
feat: set user agent in webdav requests

lukevmorris:
I have a WebDAV server that I use as a backend remote cache for multiple build tools. Many of these build tools (e.g. `bazel`, `gradle`) differentiate themselves in their user agent header, whereas `sccache` does not. This PR configures the WebdavCache service with an http client preconfigured with a user agent header like so: `sccache/0.8.2`. These changes are nearly identical to those in https://github.com/mozilla/sccache/pull/2137. Happy to make whatever changes you feel may be necessary. Thanks!

### DIFF
--- a/docs/Azure.md
+++ b/docs/Azure.md
@@ -6,6 +6,4 @@ the container for you - you'll need to do that yourself.
 
 You can also define a prefix that will be prepended to the keys of all cache objects created and read within the container, effectively creating a scope. To do that use the `SCCACHE_AZURE_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
 
-Requests sent to Azure Blob Storage will have a user agent header indicating the current sccache version, e.g. `sccache/0.8.2`.
-
 **Important:** The environment variables are only taken into account when the server starts, i.e. only on the first run.
--- a/docs/Azure.md
+++ b/docs/Azure.md
@@ -1,7 +1,7 @@
 # Azure
 
-To use Azure Blob Storage, you'll need your Azure connection string and an _existing_ Blob Storage container name. Set the `SCCACHE_AZURE_CONNECTION_STRING`
-environment variable to your connection string, and `SCCACHE_AZURE_BLOB_CONTAINER` to the name of the container to use. Note that sccache will not create
+To use Azure Blob Storage, you'll need your Azure connection string and an _existing_ Blob Storage container name.  Set the `SCCACHE_AZURE_CONNECTION_STRING`
+environment variable to your connection string, and `SCCACHE_AZURE_BLOB_CONTAINER` to the name of the container to use.  Note that sccache will not create
 the container for you - you'll need to do that yourself.
 
 You can also define a prefix that will be prepended to the keys of all cache objects created and read within the container, effectively creating a scope. To do that use the `SCCACHE_AZURE_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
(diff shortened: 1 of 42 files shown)


src/cache/webdav.rs
```diff
@@ -32,11 +34,19 @@ impl WebdavCache {
...
 
         let op = Operator::new(builder)?
             .layer(LoggingLayer::default())
             .finish();
         Ok(op)
     }
 }
+
+/// Set the user agent (helps with monitoring on the server side)
+fn set_user_agent() -> HttpClient {
```
Xuanwo, line 48: I recall that we already have similar code that builds an HTTP client. Would you like to check the codebase again?

lukevmorris: You're right! I've removed the copied code and referenced `set_user_agent` from `src/cache/s3.rs`.

Xuanwo (commented):
Others look good to me. Thank you @lukevmorris for working for this.

src/cache/s3.rs
```diff
@@ -68,7 +68,7 @@ impl S3Cache {
 }
 
 /// Set the user agent (helps with monitoring on the server side)
-fn set_user_agent() -> HttpClient {
+pub fn set_user_agent() -> HttpClient {
```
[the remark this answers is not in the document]
lukevmorris: Done -- apologies if there are cleaner ways to do this, I'm not very familiar with Rust.

Xuanwo (approved):
Thank you @lukevmorris for working on this!

sylvestre:
Could you please document this ?

https://github.com/mozilla/sccache/?tab=readme-ov-file
just a small section like
"sccache queries are sent with user agent XXX"
etc

sylvestre:
thanks.
Sorry for the dumb question. I am wondering why it should not be done for all backend? Why limiting only s3 & webdav ?

(your patch is conflicting)

lukevmorris:
@sylvestre not a dumb question at all. I had considered that but wasn't sure what scope of change would be appreciated. I can try and make that change if you'd like?

sylvestre:
yeah, i think it makes sense :)

lukevmorris:
Ok I ended up moving `set_user_agent()` into its own file that's included for only specific features, to try and clean up all my `cfg` blocks. Every opendal service backed by an http client is configured to use the same user agent. Redis and Memcached don't accept an http_client, I assume because they have their own specific communication protocol, so I've omitted those from this changeset. I've also added a line of documentation to readmes for each of the affected services.

Let me know what you think!

lukevmorris:
*facepalm* I didn't check in the new file. Hopefully that addresses the test failures.

docs/Azure.md
```diff
@@ -1,9 +1,11 @@
...
 
-To use Azure Blob Storage, you'll need your Azure connection string and an _existing_ Blob Storage container name.  Set the `SCCACHE_AZURE_CONNECTION_STRING`
-environment variable to your connection string, and `SCCACHE_AZURE_BLOB_CONTAINER` to the name of the container to use.  Note that sccache will not create
+To use Azure Blob Storage, you'll need your Azure connection string and an _existing_ Blob Storage container name. Set the `SCCACHE_AZURE_CONNECTION_STRING`
+environment variable to your connection string, and `SCCACHE_AZURE_BLOB_CONTAINER` to the name of the container to use. Note that sccache will not create
 the container for you - you'll need to do that yourself.
 
 You can also define a prefix that will be prepended to the keys of all cache objects created and read within the container, effectively creating a scope. To do that use the `SCCACHE_AZURE_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
 
+Requests sent to Azure Blob Storage will have a user agent header indicating the current sccache version, e.g. `sccache/0.8.2`.
```
sylvestre, line 9: i don't think we should duplicate it
just declare it once :)

lukevmorris: Ok sorry! I centralized it in one section at the bottom of the README. What do you think?

PR #2291, opened 2024-11-21
Pass correct argument to rustc -Z ls

### LINKED ISSUE #2290: Fix rustc -Z ls handling
I was investigating a high memory usage on sccache rust compilation part and I came accross a failure in the build log https://github.com/mozilla/sccache/blob/b0a6355d9ef18df3d9d26531254d8ab52ba306ca/src/compiler/rust.rs#L2388.

At first I assumed it was just a byproduct of configure step, but it looks like `-Z ls` changes in https://github.com/rust-lang/rust/commit/ff00763dd1c542abe9897ff62c3acff5baa6d2eb#diff-03c3ed2f6e27e58df63382e00323eea6fd1ce3a4f9546941b2b717fead25e520

One might need to update to `-Z ls=root`.

lissyx:
Behavior was changed back in rustc 1.74.

Fixes #2290

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -35,6 +35,8 @@ use fs_err as fs;
 use log::Level::Trace;
 use once_cell::sync::Lazy;
 #[cfg(feature = "dist-client")]
+use semver::Version;
+#[cfg(feature = "dist-client")]
 use std::borrow::Borrow;
 use std::borrow::Cow;
 #[cfg(feature = "dist-client")]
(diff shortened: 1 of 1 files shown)


src/compiler/rust.rs
```diff
@@ -2372,7 +2372,7 @@ impl RlibDepReader {
         trace!("Discovering dependencies of {}", rlib.display());
 
         let mut cmd = process::Command::new(&self.executable);
-        cmd.args(["-Z", "ls"])
+        cmd.args(["-Z", "ls=root"])
```
lissyx, line 2375: Do we want to care for old rustc ?

lissyx: @sylvestre the change is there since 1.74, so this is broken since. I'm not sure how much we want to care for versions of rustc older than 1.74 here?

PR #2296, opened 2024-12-19
Always use sccache's own jobserver

glandium:
and close the pipes to the inherited jobserver.


sylvestre:
if i understood correctly, it fixed https://bugzilla.mozilla.org/show_bug.cgi?id=1499382

PR #2298, opened 2024-12-21
Fix `test_rlib_dep_reader_call` failure edge case

trxcllnt:
This PR fixes a failure in the [`test_rlib_dep_reader_call`](https://github.com/mozilla/sccache/blob/6bc815456247292c5d770d9a54f0afbc99cf7326/src/compiler/rust.rs#L3168-L3177) test when `$CARGO_HOME` isn't in the user's `$HOME`.

I develop in a container based on `messense/rust-musl-cross:x86_64-musl`. The `$CARGO_HOME` dir is at `/root/.cargo`, and I create a non-root user and mount in the `sccache` source.

For some reason, `rustc` fails when run with an empty env fails (like in the test):
```shell
$ echo $HOME
/home/user
$ echo $CARGO_HOME 
/root/.cargo
$ env -i $CARGO_HOME/bin/rustc --crate-type=rlib -o /tmp/x.rlib - <<< ""
error: rustup could not choose a version of rustc to run, because one wasn't specified explicitly, and no default is configured.
help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain.
```

However, defining the `RUSTUP_HOME` envvar succeeds:
```shell
$ env -i RUSTUP_HOME=$RUSTUP_HOME $CARGO_HOME/bin/rustc --crate-type=rlib -o /tmp/x.rlib - <<< ""
```

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -2314,12 +2314,6 @@ impl RlibDepReader {
             stderr,
         } = cmd.output()?;
 
-        if !status.success() {
-            bail!(
-                "Failed to compile a minimal rlib with {}",
-                executable.display()
-            )
-        }
         if !stdout.is_empty() {
             bail!(
                 "rustc stdout non-empty when compiling a minimal rlib: {:?}",
@@ -2330,6 +2324,12 @@ impl RlibDepReader {
             bail!(
                 "rustc stderr non-empty when compiling a minimal rlib: {:?}",
                 String::from_utf8_lossy(&stderr)
+            )
+        }
+        if !status.success() {
+            bail!(
+                "Failed to compile a minimal rlib with {}",
+                executable.display()
             )
         }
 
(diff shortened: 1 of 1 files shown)


sylvestre:
would it be possible to add a test here to make sure we don't regress?
thanks

trxcllnt:
@sylvestre I can't think of how to test this in sccache. It seems to occur when `rustc` is run in an empty environment and `$RUSTUP_HOME` isn't in the current user's `$HOME`, and I'm not sure how we'd be able to recreate this in a test.

The fact that the test isn't failing in CI or for other devs sounds like a possible bug in `rustc`, as if it's using `$HOME/.cargo` as the fallback when when `$RUSTUP_HOME` is unset? That would explain why this test fails for me, but not others.

trxcllnt:
I guess one way would be to run all `ci.yaml` tests as a non-root user in the `messense/rust-musl-cross` containers, but that seems heavy-handed...

sylvestre:
ok, thanks

Issue #2299, opened 2024-12-21
sccache incorrectly yields hits for code guarded by `#ifndef __CUDA_ARCH__`


trxcllnt:
First reported here: https://github.com/NVIDIA/cccl/issues/3103

sccache is incorrectly yielding cache hits on code guarded by `#ifndef __CUDA_ARCH__`, e.g.:

```c++
#ifndef __CUDA_ARCH__
const auto x = "host-only code";
#endif
```

If the code inside the `#ifndef` changes, sccache yields a cache hit when it should've been a cache miss, because [`nvcc -E` injects `-D__CUDA_ARCH__`](https://godbolt.org/z/aes867GGo) (and likely others like `-D__CUDA_ARCH_LIST__`) into the host preprocessor invocations.

This bug has existed from before I rewrote how sccache handles nvcc compilations.
Luckily, we can fix this because I rewrote how sccache handles nvcc compilations :slightly_smiling_face:.

PR #2300, opened 2024-12-21
Always decompose nvcc compilations

trxcllnt:
Never cache the outer CUDA compilation (because `nvcc -E` can't be trusted).

Always decompose via `nvcc --dryrun`, then cache and report the host compiler call as a CUDA compilation.

Fixes https://github.com/mozilla/sccache/issues/2299.

### DIFF
--- a/src/compiler/nvcc.rs
+++ b/src/compiler/nvcc.rs
@@ -461,7 +461,20 @@ pub fn generate_compile_commands(
         output_file_name: output.file_name().unwrap().to_owned(),
     };
 
-    Ok((command, None, Cacheable::Yes))
+    Ok((
+        command,
+        None,
+        // Never assume the outer `nvcc` call is cacheable. We must decompose the nvcc call into
+        // its constituent subcommands with `--dryrun` and only cache the final build product.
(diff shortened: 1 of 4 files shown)


src/compiler/nvcc.rs
```diff
@@ -461,7 +461,20 @@ pub fn generate_compile_commands(
...
+        // its constituent subcommands with `--dryrun` and only cache the final build product.
+        //
+        // Always decomposing `nvcc --dryrun` is the only way to ensure caching nvcc invocations
+        // is fully sound, because the `nvcc -E` preprocessor output is not sufficient to detect
+        // all source code changes.
+        //
+        // Specifically, `nvcc -E` always defines __CUDA_ARCH__, which means changes to host-only
+        // code guarded by an `#ifndef __CUDA_ARCH__` will _not_ be captured in `nvcc -E` output.
+        Cacheable::No,
+    ))
```
trxcllnt, line 477: We might now be able to get away with doing less in the `preprocess()` function, since we effectively don't care about most of what it does.

PR #2297, opened 2024-12-19
Revert "feat: Use profile-generate to replace outdated -Zprofile options"

sylvestre:
Reverts mozilla/sccache#2282

PR #2306, opened 2024-12-30
fix: bring "cargo build -Zprofile" support back

Xuanwo:
Close https://github.com/mozilla/sccache/pull/2297.

In https://github.com/mozilla/sccache/pull/2282, I mistakenly removed the `-Zprofile` support, which is still used by many older Rust nightly toolchains. This PR restores this feature.

Additionally, this PR updates the CI to ensure it will not be broken by similar changes in the future.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -805,14 +805,15 @@ jobs:
 
           ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
 
-  rust-test-coverage:
+  # The test cargo "cargo build -Zprofile"
+  rust-test-Z-profile:
     runs-on: ubuntu-latest
     needs: build
 
     env:
       RUSTC_WRAPPER: /home/runner/.cargo/bin/sccache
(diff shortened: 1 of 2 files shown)


Xuanwo:
cc @sylvestre and @glandium for a review, thanks!

.github/workflows/integration-tests.yml
```diff
@@ -825,21 +826,24 @@ jobs:
...
-        run: cargo clean && cargo +nightly test
+        run: cargo clean && cargo +nightly-2024-11-01 test
 
       - name: Output
         run: ${SCCACHE_PATH} --show-stats
 
       - name: "Coverage test #2"
         working-directory: ./coverage-test
-        run: cargo clean && cargo +nightly test
+        run: cargo clean && cargo +nightly-2024-11-01 test
```
sylvestre, line 846: maybe document why Nov 1st?

Xuanwo: Added before. Let's me extra it as an env.

Xuanwo:
`-Zprofile` works now:

image

Issue #974, opened 2021-03-10
Setting up Sccache on Ec2 instances.


rahulbansal16:
Hi,

I am trying to set up the sccache on the Ec2 instances and I am not able to get it up and running. I have tried few techniques to make it work but I have not got any success.

I thought of documenting all the steps that I took and will keep on posting the new ones.

**Intended Setup**
I want to put the scheduler server on a dedicated machine and connect the build servers to the scheduler. I want the scheduler to be accessible by the General Public with the required token.

This means that the build server should be running on the public IP instead of the private one.

**Approaches tried**

1. **Setting the public_addr value to the public IP in the build server** 

    The public IP needs to be associated with the OS for this approach to work. I did not have much idea about doing that so I thought of trying the next approach. Refer to the StackOverflow question for more details.

2. **Listening on the Private IP**

    I tried listening on the Private IP of the build server to at least get the sccache working for the subnet. I set the value of the **public_addr** value to 172.23.23.22:10501. This leads to an error of **invalid_bearer_token_mismatched_address**. I have set the scheduler auth type to DANGEROUSLY_INSECURE.

3. **Setting the public_addr to 127.0.0.1[:10501](http://localhost:10501) and doing a Port Forwarding from Scheduler and Client**

    In this, I thought of keeping the public_addr value to 127.0.0.1:10501 and doing a port forwarding from the client and scheduler to build server port 10501. This gives the error channel 3: open failed: connect failed: Connection refused 

4. **Hardcoding the ip addresses**

    I hardcoded the address in the line [https://github.com/mozilla/sccache/blob/915e977bc3d7339030119462db3b04b856cf08ee/src/dist/http.rs#L772](https://github.com/mozilla/sccache/blob/915e977bc3d7339030119462db3b04b856cf08ee/src/dist/http.rs#L772) with the public_addr "0.0.0.0:10501" to make it possible to expose the port outside VM and returning server_id in the line [https://github.com/mozilla/sccache/blob/915e977bc3d7339030119462db3b04b856cf08ee/src/dist/http.rs#L719](https://github.com/mozilla/sccache/blob/915e977bc3d7339030119462db3b04b856cf08ee/src/dist/http.rs#L719)
    
   **EDIT**
 I got it working via the 4th approach

luser:
>  I want to put the scheduler server on a dedicated machine and connect the build servers to the scheduler. I want the scheduler to be accessible by the General Public with the required token.

[...]

Issue #930, opened 2021-01-11
Support for build servers not behind `https`


daboross:
It seems like sccache hardcodes `https` for build servers:

https://github.com/mozilla/sccache/blob/46d51c1404684e5c051ed0cec8ff96410eff806f/src/dist/http.rs#L250

Could this be made configurable, or otherwise documented? The distributed setup guides mentions that it's recommended to put the scheduler behind a HTTPs server:

>  It's strongly recommended to listen on localhost and put a HTTPS server in front of it.

However, I can't find anything mentioning a hard requirement to put builders behind the same.

My use case for this is using sccache on a local LAN network. Right now I'm trying to set it up just using SSH tunnels - the scheduler and builder only to `localhost`, and I connect to server through an SSH tunnel. This setup fails with the following error, though:

```
 WARN 2021-01-11T04:00:37Z: sccache::compiler::compiler: [inflections]: Could not perform distributed compile, falling back to local:
Error 500: {"description":"assign job failed, job un-assigned from the server",
"cause":{"description":"POST to scheduler assign_job failed",
"cause":{"description":"https://[::1]:10501/api/v1/distserver/assign_job/43: error trying to connect: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1913: (Hostname mismatch)",
"cause":{"description":"error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1913: (Hostname mismatch)",
"cause":{"description":"error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1913:",
"cause":null}}}}}:
Error 500: {"description":"assign job failed, job un-assigned from the server",
"cause":{"description":"POST to scheduler assign_job failed",
"cause":{"description":"https://[::1]:10501/api/v1/distserver/assign_job/43: error trying to connect: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1913: (Hostname mismatch)",
"cause":{"description":"error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1913: (Hostname mismatch)",
"cause":{"description":"error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1913:",
"cause":null}}}}}
```

<details>

[...]

PR #2307, opened 2024-12-30
Add bind_address config field

samuel-jimenez:
Allows binding to local address, uses `public_addr` as default.
Fix for #930, #974.

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -225,6 +225,7 @@ fn run(command: Command) -> Result<i32> {
             builder,
             cache_dir,
             public_addr,
+            bind_address,
             scheduler_url,
             scheduler_auth,
             toolchain_cache_size,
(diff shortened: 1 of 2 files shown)


sylvestre:
Could you please document the change and add tests to make sure we don't regress?

Thanks

docs/DistributedConfiguration.md
```diff
@@ -0,0 +1,146 @@
+# Available Distributed Configuration Options
```
lissyx, line 1: Shouldn't that be part of existing `Distributed.md` under a `Configuration` title? As a user that's where I'd first go look, having it split looks weird

lissyx (approved):
LGTM with maybe the documentation move?

docs/DistributedConfiguration.md
samuel-jimenez: My rationale for creating a new file is having a canonical place to document configuration, as opposed to samples spread throughout `Distributed.md` and  `DistributedQuickstart.md`.
Keeping the file separate also allows a quick reference for distributed configuration options and mirrors `Configuration.md` file structure.

I agree a `Configuration` section would also accomplish my first point. However, accessing the information would become a bit more unwieldy, and would bloat `Distributed.md`, which is already the largest doc.
Would a link in `Distributed.md` allay your concerns?

samuel-jimenez:
I have added two commits for documentation comparison.

lissyx (approved):

lissyx:
Thanks!

PR #2308, opened 2025-01-03
Don't use CARGO_REGISTRIES_* configuration in hash keys

kornelski:
Private registry configuration in Cargo can be specified using `CARGO_REGISTRIES_$name_*` env vars. That it includes `CARGO_REGISTRIES_$name_TOKEN`, which is an auth secret. The tokens have high entropy and change often, so inclusion of such env vars in the hash key completely ruins caching in sccache.

I don't think any of the `CARGO_REGISTRIES_` env vars need to be tracked, because Cargo identifies dependencies by their package IDs, which already includes a normalized registry URL. Only the resolved package ID affects the result, not the mechanics of how Cargo discovered it.

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -1466,12 +1466,21 @@ where
             .collect();
         env_vars.sort();
         for (var, val) in env_vars.iter() {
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
Makes sense. Thank you @kornelski for this!

Issue #2313, opened 2025-01-08
sccache could be more zealous closing file descriptors when daemonizing


ntrrgc:
Since `sccache` will be invoked as a generic compiler wrapper, in some situations it might be invoked with pretty exotic file descriptors. For instance, nobody stops me from doing this in a Bash shell, even if it's somewhat questionable:

```
$ exec 4<> /tmp/a.c  # Open /tmp/a.c as file descriptor number 4 in Bash
$ sccache /usr/bin/gcc -o /tmp/executable -x c /dev/fd/4  # sccache will inherit fd 4, so this will work
$ exec 4>&-  # Close file descriptor 4 in Bash
```

One of the tasks of the compiler wrapper is to spawn a sccache server when one is not already running. When starting a background process we want as little baggage from the original process as possible, and extra file descriptors are an instance of such baggage that can lead to unexpected problems.

A notorious example of such problems is that **ninja will get stuck if executing the sccache wrapper in a tty**, because ninja uses a covert pipe to detect when its subprocesses finish execution when sharing a console (!): https://github.com/ninja-build/ninja/issues/2444

You can probably imagine other scenarios in which the daemon accidentally sharing some file descriptors may cause problems.

[daemonize](https://github.com/knsd/daemonize) (the crate used for most of the setup of spawning the server process) unfortunately does not handle this except for the standard discriptors (stdin, stdout, stderr), and it couldn't without breaking compatibility as a daemon may still need to share specific file descriptors in specific situations.

Nowadays you can use something like `close_range()` (FreeBSD.  Linux 5.9, glibc 2.34). Other approaches involve using `/proc/<PID>/fd` or similar mechanisms to figure out and close extraneous file descriptors.

sylvestre:
which version of sccache did you try ? 

We made progress here
https://github.com/mozilla/sccache/pull/2296
https://bugzilla.mozilla.org/show_bug.cgi?id=1499382

ntrrgc:
I tried this on the current top of tree (9ca8bebb3db47e15549c9b1296764debd50d365e).

It's reproducible very often with meson projects that end up invoking the compiler wrapper.

```
$ sccache --stop-server
$ touch meson.build && ninja -C builddir/
```

PR #2314, opened 2025-01-08
Close non-stdio file handles when daemonizing

ntrrgc:
Otherwise, when the compiler wrapper spawns the sccache server, the server may end up with unintended file descriptors, which can lead to unexpected problems.

This is particularly problematic if any of those file descriptors that accidentally end up in the server process is a pipe, as the pipe will only be closed when all the processes with that file descriptor close it or exit.

This was causing sccache to hang ninja, as ninja uses the EOF of a pipe given to the subprocess to detect when that subprocess has exited: https://github.com/ninja-build/ninja/issues/2444#issuecomment-2577999973

This patch adds a dependency on the
[close_fds](https://crates.io/crates/close_fds) crate, which automatically chooses an appropriate mechanism to close a range of file descriptors. On Linux 5.9+ that mechanism will be libc::close_range().

Fixes https://github.com/mozilla/sccache/issues/2313


ntrrgc:
The test `test_rust_cargo_cmd_readonly_preemtive_block` was failing. After investigation, I found the reason:

```rust
if env::var("SCCACHE_ERROR_LOG").is_ok() {
    let f = create_error_log()?;
    // Can't report failure here, we're already daemonized.
    daemonize()?;
    redirect_error_log(f)?;
```

That code was opening a log file, daemonizing, then swapping stderr with it. We could pass the file descriptor of `f` to the list of exceptions of close_open_fds(), but in this particular case there is no need, as swapping stderr with a file is already a built-in feature of Daemonize.

I'm updating the patch to make daemonize accept a log file as argument and do the file handle swap at that point, which also removes the need for the `redirect_stderr()` and `redirect_error_log()` functions in sccache.

ntrrgc:
Several CI jobs are consistently failing, so probably there are still file handles unaccounted. Will have to reproduce the failures locally and find the missing handles.

ntrrgc:
I checked the failures now and it seems it is just formatting nits and failures to build on Windows. If there was something more serious, I can't see it now...

Oh well, I'm updating the patch with the formatting nits applied and with the `close_fd` crate marked for use only in `cfg(unix)`.

ntrrgc:
Any thoughts on the patch? Can someone approve the CI workflows so they can run?

ntrrgc:
```
error[E0061]: this function takes 0 arguments but 1 argument was supplied
   --> src\commands.rs:645:17
    |
645 |                 daemonize(Some(f))?;
    |                 ^^^^^^^^^ ------- unexpected argument of type `std::option::Option<fs_err::File>`
    |
note: function defined here
   --> src\util.rs:942:8
    |
942 | pub fn daemonize() -> Result<()> {
    |        ^^^^^^^^^
help: remove the extra argument
```

This is very odd because `daemonize()` is modified by this patch to add that extra argument, and you can see that in the diff with those same line numbers:

```diff
diff --git a/src/util.rs b/src/util.rs
index 7174681..04d348c 100644
--- a/src/util.rs
+++ b/src/util.rs
@@ -841,7 +841,7 @@ impl<'a> Hasher for HashToDigest<'a> {
 
 /// Pipe `cmd`'s stdio to `/dev/null`, unless a specific env var is set.
 #[cfg(not(windows))]
-pub fn daemonize() -> Result<()> {
+pub fn daemonize(log_file: Option<File>) -> Result<()> {
```

It's as if the patch had been only partially applied by the CI server. The same error is happening in many of them.

ntrrgc:
Oooooh, because they're the Windows builds! I keep forgetting the `#[cfg(not(windows))]`. There has to be a Windows specific variant that I need to update.

ntrrgc:
`toml_format` was complaining about the packages not being in alphabetical order (it didn't explicitly say it, but I could confirm it locally).

```
error: function `redirect_stderr` is never used
   --> src\commands.rs:130:4
    |
130 | fn redirect_stderr(f: File) {
    |    ^^^^^^^^^^^^^^^
```

It should actually be used. `daemonize()` in Windows used to be a full no-op, but now that it takes the log file as argument, it should also set it as stderr in Windows, so the API is the same for all platforms. Fixed it by moving `redirect_stderr()` to `util.rs` and using it in the Windows version of `daemonize()`:

```rust
/// Daemonizing is a no-op on Windows, but we still must set the log file as
/// stderr.
#[cfg(windows)]
pub fn daemonize(log_file: Option<File>) -> Result<()> {
    if let Some(log_file) = log_file {
        redirect_stderr(log_file);
    }
    Ok(())
}
```

The failures of the `test_rust_cargo_cmd_readonly_preemtive_block` in Windows are just a consequence of that log redirection missing in Windows, so that should be fixed now too:

```rust
    let log_contents = fs::read_to_string(sccache_log)?;
    assert!(predicates::str::contains("server has setup with ReadOnly").eval(log_contents.as_str()));
```

Hopefully no more remaining failures!

AJIOB:
Hi @sylvestre,

Is it ok for you now?

ahartmetz:
What's the status here? In case it was stuck because of CI failing due to no fault of these changes, well, CI should be OK again.

PR #2315, opened 2025-01-10
Scheduler: oversubscribe cores just a little to make up for slack

ahartmetz:
...not by factor of 10 as it was done previously

### DIFF
--- a/src/bin/sccache-dist/main.rs
+++ b/src/bin/sccache-dist/main.rs
@@ -309,7 +309,8 @@ fn init_logging() {
     }
 }
 
-const MAX_PER_CORE_LOAD: f64 = 10f64;
+// Maximum number of jobs per core - only occurs for one core, usually less, see load_weight()
+const MAX_PER_CORE_LOAD: f64 = 2f64;
 const SERVER_REMEMBER_ERROR_TIMEOUT: Duration = Duration::from_secs(300);
 const UNCLAIMED_PENDING_TIMEOUT: Duration = Duration::from_secs(300);
 const UNCLAIMED_READY_TIMEOUT: Duration = Duration::from_secs(60);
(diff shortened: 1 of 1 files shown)


src/bin/sccache-dist/main.rs
```diff
@@ -309,7 +309,7 @@ fn init_logging() {
     }
 }
 
-const MAX_PER_CORE_LOAD: f64 = 10f64;
+const MAX_PER_CORE_LOAD: f64 = 2f64;
```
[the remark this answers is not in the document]
ahartmetz: That OK?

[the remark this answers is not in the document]
ahartmetz: (In the next version that I forced-pushed - sorry, I'm more used to how it works in GitLab)

ahartmetz:
We agree these CI failures are clearly unrelated to this change, right?

src/bin/sccache-dist/main.rs
```diff
@@ -399,6 +400,18 @@ impl Default for Scheduler {
     }
 }
 
+fn load_weight(job_count: usize, core_count: usize) -> f64 {
+    // Do not oversubscribe cores very much to avoid out of memory situations and cache thrashing -
+    // oversubscribe just a little to make up for network and other latency.
+    let cores_plus_slack = core_count + 1 + core_count / 8;
```
lissyx, line 406: Do you have a reasonning for the values here? especially `+1` and `core_count / 8`

ahartmetz: I don't have measurements or anything. It's just "send a few more jobs than the core count so that cores don't wait for the network (to receive jobs) or I/O (while compiling)". I guess it would be cool to model the network part explicitly and have the concept of "pre-loading" jobs which AFAIU Icecream has, but that'd be much more work. And once you get into the territory where jobs run faster than they can be transferred, it gets pretty involved to do it well, I think.

lissyx: maybe worth mentionning that in the comment so that people know where those values comes from :)

ahartmetz: OK, I'll add something

ahartmetz: Added

lissyx (approved):

PR #2319, opened 2025-01-18
add ubuntu-24.04-arm in the CI

sylvestre:


mati865:
There is an issue with ubuntu-24.04-arm runners making them very unstable: https://github.com/rust-lang/rust/issues/135867
I'd advise sticking with ubuntu-22.04-arm until it's fixed.

sylvestre:
Thanks for the information, i wasn't aware :)

mati865:
FYI, the issue with unstable AArch64 runners is fixed.

PR #2324, opened 2025-01-28
Fix and update sccache for CUDA 12.8

trxcllnt:
This PR updates sccache, tests, and CI for CUDA Toolkit 12.8.

This PR fixes a bug related to the `.module_id` file generated by `cudafe++`. This file is now unique (in CTK 12.8) across `cudafe++` invocations, is important for device symbol visibility, and must be consistent between `cudafe++` and all `cicc` calls.

The bug arises when building a `.cu.o` that includes cached PTX. When the `cudafe++` command is re-run it generates a new unique `.module_id` file. This version has a different id than the cached `.ptx` files, leading to a mismatch in the device symbol names in the PTX/cubins vs. the symbols used by the final host-compilation step.

Luckily the fix is straightforward -- cache `cudafe++` invocations. This is safe to do since the `cudafe++` input is the output from the host preprocessor, so any changes that affect device code will yield different hashes.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -8,7 +8,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Clone repository
         uses: actions/checkout@v4
@@ -28,7 +28,7 @@ jobs:
           path: ./target/debug/sccache
 
   redis-deprecated:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: build
 
     services:
@@ -76,7 +76,7 @@ jobs:
           ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
 
   redis:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: build
 
     services:
@@ -124,7 +124,7 @@ jobs:
           ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
 
   s3_minio:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: build
 
     # Setup minio server
@@ -184,7 +184,7 @@ jobs:
           ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
 
   azblob_azurite:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: build
 
     # Setup azurite server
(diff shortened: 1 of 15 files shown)


robertmaynard (approved):

trxcllnt:
@sylvestre `ubuntu-latest` is 24.04 now, so the build jobs link sccache to glibc 2.38. The hip integration job uses 22.04 with an older glibc and [fails](https://github.com/mozilla/sccache/actions/runs/13018600869/job/36313970858#step:8:28).

I copied this fix over from another branch, and at the time `rocm/dev-ubuntu-24.04` wasn't published. I just checked again and it looks like it exists now, so I'll change it back.

tests/harness/Dockerfile.sccache-dist
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:latest
```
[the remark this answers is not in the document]
trxcllnt: This is `latest` to ensure the container uses glibc >= the version sccache was built with.

For example, I build and test on an Ubuntu 24.04 host. When the tests run this container with sccache built from my host, it fails to launch since Ubuntu 22.04 has an older glibc.

I could set it to 24.04, but that just defers the problem for 2 more years.

The only potential problem I see is that `apt install libcap2 bubblewrap` could break in some future release if those packages are renamed (or Ubuntu removes apt, heh), but those would have to be dealt with eventually anyway.

.github/workflows/integration-tests.yml
```diff
@@ -8,7 +8,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
sylvestre, line 11: it isn't clear to me why this change?

sylvestre:
Could you please update one of the readme to document a bit this ?
For example, cudafe could be defined and explain :)
besides these two comments, looks good to me
thanks

.github/workflows/integration-tests.yml
trxcllnt: When GitHub updates `ubuntu-latest` to 26.04 next year, the integration tests will start failing again in the same way as [here](https://github.com/mozilla/sccache/pull/2324#issuecomment-2625476046) until the HIP job is updated to `rocm/dev-ubuntu-26.04`.

It will stay broken until a new version is published (or we switch images). This lets us update on our schedule while keeping keep CI green. Since we have to update this file in both cases, I chose the path where jobs don't start mysteriously failing due to GitHub changes.

trxcllnt:
@sylvestre sure, but can we make it a follow-up PR? The current sccache release is currently poisoning caches when used with CUDA 12.8.

sylvestre:
sure

PR #2328, opened 2025-02-10
Don't use CARGO_BUILD_JOBS in hash keys

fsouza:
`CARGO_BUILD_JOBS` only affects Cargo's parallelism, not rustc's outputs, so it can be safely excluded from the cache key.

This enables better cache reutilization when building from environments with different number of CPUs.

I ran into this while experimenting with sccache on a containerized system where a library can be shared/compiled between different job sizes, and was surprised that I was getting cache misses for those libraries.

Given how small the change is, I figured opening the PR wouldn't hurt, but please let me know if you prefer me to move this to an issue for further discussion before actually having the PR reviewed.

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -1507,7 +1507,11 @@ where
             // CARGO_REGISTRIES_*_TOKEN contains non-cacheable secrets.
             // Registry override config doesn't need to be hashed, because deps' package IDs
             // already uniquely identify the relevant registries.
-            if var == "CARGO_MAKEFLAGS" || var.starts_with("CARGO_REGISTRIES_") {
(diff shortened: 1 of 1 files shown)


fsouza:
(Perhaps we can discuss what other variables should be excluded or whether sccache could have a knob that allows the user to select which environment variables to exclude from the key, though I understand that having a knob may complicate things unnecessarily)

drahnr (approved):

drahnr:
I think a long-term solution would be to have a meta-key, that allows to exclude arbitrary environment variables from being included. Out of scope for this PR though.

ahartmetz:
Someone who can press the merge button here? It's just sitting around in approved state.

PR #2329, opened 2025-02-11
Allow using S3 virtual host style endpoints

brndnmtthws:
sccache currently breaks on S3 endpoints that use virtual host style URLs. This is because [opendal strips the bucket name from URLs](https://opendal.apache.org/docs/rust/src/opendal/services/s3/backend.rs.html#504-505).

This PR provides a config option to force opendal to keep the bucket name in the URL, which unbreaks sccache for such buckets.


ADustyOldMuffin:
Fixed #2344, thanks you beat me to it! Any timeline on getting this merged and out?

Xuanwo (approved):
Thank you for working on this!

andylolu2:
Would be amazing if we can make a new release for virtual host style support! 🙏🏻

PR #2335, opened 2025-02-17
Clean up logging

samuel-jimenez:
Improve logging as discussed in point 1 of #2332.

Splits out heartbeats and raw HTTP as `sccache_heartbeat` and `sccache_http`, respectively.


sylvestre:
please run rustfmt :)

PR #2336, opened 2025-02-17
Remove Rust jobserver fd reference

### LINKED ISSUE #2333: Distributed Rust compilation fails
Attempting to build a crate with a `sccache-dist` setup:
```
cargo clean
SCCACHE_RECACHE=1 cargo build --release
```

Distributed compilation fails with the following message:
```
warning: failed to connect to jobserver from environment variable `CARGO_MAKEFLAGS="-j --jobserver-fds=8,9 --jobserver-auth=8,9"`: cannot open file descriptor 8 from the jobserver environment variable value: Bad file descriptor (os error
 9)
  |
  = note: the build environment is likely misconfigured
```

samuel-jimenez:
Strip `CARGO_MAKEFLAGS` environment variable. Under a distributed environment, the file descriptors referenced by the client are not present on a build server.

Resolves #2333.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -310,6 +310,7 @@ pub fn try_parse() -> Result<Command> {
                         && k != "MINICOM"
                         && k != "DESTDIR"
                         && k != "RPM_PACKAGE_VERSION"
+                        && k != "CARGO_MAKEFLAGS"
                 });
 
                 let cmd = matches


sylvestre:
would it be possible to add a test to make sure we don't regress? thanks

drahnr (approved):

drahnr:
@sylvestre sccache currently broken so a new patch release might make sense

sylvestre:
@drahnr broken how?

PR #2337, opened 2025-02-19
feat(GCS): Update `CustomTime` metadata field at cache entry hits

### LINKED ISSUE #2318: [GCS] Feature req: Use `CustomTime` attribute to record when cached entry was last hit (for eviction/cleanup rules)
Hey!

We are using a _Google Cloud Storage Bucket_ through _`sccache`_ to cache across CI compilations of [LLVM](http://github.com/llvm/llvm-project). Unfortunately, it looks like none of the cloud provider implementations allow for keeping the size of the cache in check — it is only possible for local-disk caching (which we can not do as the CI machines are ephemeral VMs).

In order to keep the costs of this shared cache reasonable, we implemented an _Age_ lifecycle rule in _GCS_: if a cached entry is too old, _GCS_ will deal with deleting it.
According to _GCS_, an _Age_ rule is matched as:

> Age is counted from when an object was uploaded to the current bucket.

This presents a problem with ["death-waves"](https://www.giantbomb.com/profile/gamer_152/blog/death-wave-cities-skylines-and-simulatory-failure/275492): Every $n$ days after an initial CI run had populated the cache with about $6\,000$ translation units, they **all** get deleted, and the subsequent CI run will essentially have to do an almost full rebuild. The only cache hits at $\ge n$ days will be the more recently modified content — which is often the smaller part of the files in the project.

It would be great if _`sccache`_ with a cloud bucket could behave in the same _LRU_ way as [_`ccache`_ locally does](https://ccache.dev/manual/4.10.2.html#_automatic_cleanup). Doing a full _LRU_ is problematic without handling size limits, **but** it seems there is another lifecycle property that can be used, at least in the case of _GCS_: [**CustomTime**](https://cloud.google.com/storage/docs/metadata#custom-time).

_LRU_-like behaviour could be simulated by setting the _Days since custom time_ lifecycle rule (instead of _Age_), making sure only files that had not been cache hit for $n$ days are evicted from the bucket.
However, _`sccache`_ does not populate this field at all:

_Details_ view of a file in an _`sccache`_ bucket, showing that **Custom time** is empty.

It seems from the documentation that this is a simple timestamp field that could be populated through the API, so all we need is a simple additional HTTP request that tells Google to update this field to "current time" once the caching logic hits a file successfully.

----

It seems like similar metadata options are available for other cloud providers, such as [S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html) or [Azure](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-properties-metadata) but I have no idea or experience whether these could be used for controlling the cache's lifecycle as well as it can be for _GCS_.

whisperity:
Resolves #2318.

This commit introduces the architectural support for implementing a cache-provider-specific routine that updates an `atime`-like (access time) timestamp every time a cache entry is successfully hit. Several cloud providers implement such semantics under different names and APIs, most commonly for the purpose of providing a better time-out and culling mechanism. The logic in _sccache_ only performs the bookkeeping of this information. With this feature, it is possible to give the necessary inputs to the cloud provider's lifecycle routines to allow keeping cloud costs and size low by culling cache objects that had not been hit for a set amount of time.

Added the support for the aforementioned logic specifically for the GCS cache implementation, where the custom access time medata is available under the `CustomTime` attribute, which can be bumped only ever forward in time. In case a bumping is unsuccessful (due to clocks getting out-of-synch), the error is silently discarded.

Screenshot of the Google Cloud Console inspecting the metadata of a cache object where _Custom Time_ is set.

### DIFF
--- a/src/test/mock_storage.rs
+++ b/src/test/mock_storage.rs
@@ -65,6 +65,9 @@ impl Storage for MockStorage {
             Duration::from_secs(0)
         })
     }
+    async fn timestamp_cache_hit(&self, _key: &str) -> Result<Duration> {
+        Ok(Duration::from_secs(0))
+    }
     fn location(&self) -> String {
         "Mock Storage".to_string()
     }
--- a/docs/Gcs.md
+++ b/docs/Gcs.md
@@ -65,6 +65,16 @@ export SCCACHE_GCS_KEY_PATH=secret-gcp-storage.json
 Cache location                  GCS, bucket: Bucket(name=<bucket name in GCP>), key_prefix: (none)
 ```
 
+## Lifecycle management
+
+Sccache updates the `CustomTime` metadata field of cache objects every time
+there was a cache hit.
+This can be used to implement automatic cleanup in GCS using the
+["Custom time before"](https://cloud.google.com/storage/docs/lifecycle#customtimebefore)
+or ["Days since custom time"](https://cloud.google.com/storage/docs/lifecycle#dayssincecustomtime)
+conditions on the bucket in order to remove cache entries which have not been
+actively used for a certain amount of time.
+
 ## Deprecation
 
 `SCCACHE_GCS_OAUTH_URL` have been deprecated and not supported, please use `SCCACHE_GCS_SERVICE_ACCOUNT` instead.
--- a/src/cache/cache.rs
+++ b/src/cache/cache.rs
@@ -16,7 +16,7 @@
 use crate::cache::azure::AzureBlobCache;
 use crate::cache::disk::DiskCache;
 #[cfg(feature = "gcs")]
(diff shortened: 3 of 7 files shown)


src/test/mock_storage.rs
```diff
@@ -65,6 +65,9 @@ impl Storage for MockStorage {
             Duration::from_secs(0)
         })
     }
+    async fn timestamp_cache_hit(&self, _key: &str) -> Result<Duration> {
```
whisperity, line 68:
```suggestion
    async fn timestamp_cache_hit(&self, _key: &str) -> Result<Option<Duration>> {
```

D'oh!

PR #2338, opened 2025-02-21
Handle -Xarch_<arch> <arg> for Clang

### LINKED ISSUE #2090: `-Xarch` Clang compiler option not supported
I'm not sure what the exact conditions are, but CMake (sometimes?) generates `-Xarch_<arch>` before architecture-specific PCH includes with (both LLVM and Apple) Clang on macOS. The option's syntax is kind of weird (`-Xarch_<arg1> <arg2>` [Clang CLI ref](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-Xarch_-arg1)) and it not being properly supported results in arguments getting reordered weirdly causing the following errors:
```
/usr/local/bin/sccache /usr/local/opt/llvm/bin/clang++ <...> -arch x86_64 <...> -Winvalid-pch -Xarch_x86_64 -include/Users/runner/work/geode/geode/build/loader/CMakeFiles/geode-loader.dir/cmake_pch_x86_64.hxx <...>
clang++: error: no such file or directory: 'x86_64'
clang++: error: invalid Xarch argument: '-Xarch_x86_64 -arch', options requiring arguments are unsupported
```
```
[2024-02-13T11:50:19Z DEBUG sccache::compiler::gcc] cmd after -arch rewrite: Some(<...> "/usr/local/opt/llvm/bin/clang++" <...> "-include" "/Users/runner/work/geode/geode/build/loader/CMakeFiles/geode-loader.dir/cmake_pch_x86_64.hxx" <...> "-Winvalid-pch" "-Xarch_x86_64" "-arch" "x86_64" <...>)
```
I managed to work around the issue for my specific use case in [this](https://github.com/cgytrus/sccache/commit/f284de9f9c2c6ec9623b127879f30cf6ec4d538b) commit, but obviously this isn't sustainable long-term, since it should actually parse 2 arguments from the option (`arg1` and `arg2`) which doesn't seem to be possible right now.

alcroito:
A new ArgDisposition is introduced that combines the Concatenated and Separated behaviors, but produces results similar to a plain Separated argument with the full original -Xarch_<arch> argument.

Fixes #2090


tests/system.rs
```diff
@@ -1720,6 +1725,39 @@ fn test_clang_cache_whitespace_normalization(
     }
 }
 
+// Checks that the -include$PWD/test.h flag does not get re-ordered and
+// placed somewhere other than directly after the -Xarch_<arch> flag.
+// If it gets re-ordered, the compilation will fail.
+fn test_clang_xarch_flags(compiler: Compiler, tempdir: &Path) {
```
alcroito, line 1731: I added an e2e test here, based on the original issue comment about reordered `-include` flag breaking compilation.

src/compiler/args.rs
```diff
@@ -456,6 +458,20 @@ impl<T: ArgumentValue> ArgInfo<T> {
                     a => a?,
                 }
             }
+            ArgInfo::TakeArg(s, create, ArgDisposition::ConcatenatedAndSeparated(_)) => {
+                let len = s.len();
+                debug_assert_eq!(&arg[..len], s);
+
+                if let Some(a) = get_next_arg() {
+                    Argument::WithValue(
+                        arg.to_string().leak(),
```
alcroito, line 467: There's no simple way to get rid of the leak here, due to the way `Argument::WithValue` is defined. 

It expects the argument to be a `&'static str`, e.g defined as a string literal in the source code, like most other arguments are defined via `take_arg!`. 

But the nature of the `-Xarch_<value1> <value2>` argument is that we can't know at source code time what `<value1>` will be, it's dynamic / can be any string literal, and that goes against the design of `Argument::WithValue`.

One approach could be to have a global static `HashSet<String>` and store newly encountered values they are encountered, but that's not exactly pretty due to the global nature of it.

Another could be to refactor `Argument::WithValue` to take an enum of either a `&'static str` or a `String`, but that might end up in some cloning of strings around the code base. And I'm not sure if that will work in all the relevant code paths.

src/compiler/clang.rs
```diff
@@ -210,6 +210,7 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     take_arg!("-MT", OsString, CanBeSeparated, DepTarget),
     flag!("-Wno-unknown-cuda-version", PassThroughFlag),
     flag!("-Wno-unused-parameter", PassThroughFlag),
+    take_arg!("-Xarch", OsString, ConcatenatedAndSeparated('_'), PassThrough),
```
alcroito, line 213: It's actually unclear to me so far what `PassThrough` does after looking through the code.

Does it ignore the value passed, or does it bust the cache for every different value?

src/compiler/args.rs
alcroito: The enum approach mentioned, would basically refactor a lot of the code to use:
```
use std::borrow::Cow;
type CowStaticStr = Cow<'static, str>;
```
and replace `&'static str` with that type alias.

It's quite a bit of lines touched though, in a local attempt i already had to touch 60 lines of code, and it seems it will be a lot more. I can try to finish it, if it's deemed ok.

drahnr: @alcroito I think that's the right approach, it'd allow the default to remain _almost_ as efficient as before, while not leaking mem in the presented case.

alcroito:
Closing for now due lack of time on the big refactoring, and unclear semantics of PassThrough.

Issue #2330, opened 2025-02-11
Time sensitive: GitHub Actions cache service integration


Link-:
The actions/cache backend service has been rewritten from the ground up for improved performance and reliability. The new service has been gradually rolling out since February 1st, 2025. The legacy service **will be sunset by March 1st**.

We have identified that this project is integrating with the legacy cache service without using the official and supported [package](https://github.com/actions/toolkit/tree/main/packages/cache). Unfortunately, that means that you have to introduce code changes in order to be compatible with the new service we're rolling out.

The new service uses an entirely new set of internal API endpoints. To help with your changes we have provided the proto definitions below to help generate compatible clients to speed up your migration. 

**These internal APIs were never intended for consumption the way your project is at the moment. Since this is not a paved path we endorse, it's possible there will be breaking changes in the future. We are reaching out as a courtesy because we do not wish to break the workflows dependent on this project.**

Please introduce the necessary changes ASAP before the end of February. Otherwise, storing and retrieving cache entries will start to fail. There will be no need to offer backward compatibility as the new service will be rolled out to all repositories by February 13th 2025. 

The cutover point in time will be the moment the new service is rolled out to a given repository. Then, users will no longer have access to cache entries from the legacy service and are expected to store / retrieve cache entries from the new service.

<details>
  <summary>Expand for proto definitions</summary>

#### cache.proto
```
syntax = "proto3";

import "cachemetadata.proto";

package v1;

service CacheService {
    // Generates a SAS URL with write permissions to upload a cache archive
    rpc CreateCacheEntry(CreateCacheEntryRequest) returns (CreateCacheEntryResponse);
    // Indicate the completion of a cache archive upload. Triggers post-upload processing
    rpc FinalizeCacheEntryUpload(FinalizeCacheEntryUploadRequest) returns (FinalizeCacheEntryUploadResponse);
    // Generates a SAS URL with read permissions to download a cache archive
    rpc GetCacheEntryDownloadURL(GetCacheEntryDownloadURLRequest) returns (GetCacheEntryDownloadURLResponse);
}

message CreateCacheEntryRequest {
    // Scope and other metadata for the cache entry
    results.entities.v1.CacheMetadata metadata = 1;
    // An explicit key for a cache entry 
    string key = 2;
    // Hash of the compression tool, runner OS and paths cached
    string version = 3;
}

message CreateCacheEntryResponse {
    bool ok = 1;
    // SAS URL to upload the cache archive
    string signed_upload_url = 2;
}

[...]

PR #2339, opened 2025-02-24
feat: Bump opendal to 0.52 to support ghac v2

Xuanwo:
Fix https://github.com/mozilla/sccache/issues/2330

---

Hi, @sylvestre, we will need a release to allow users to upgrade sccache before 3/1 (the sunset of old ghac service)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -754,12 +754,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


drahnr (approved):

sylvestre:
I will make a release today or tomorrow

PR #2343, opened 2025-02-26
 Apply log filtering to syslog with env_filter crate

samuel-jimenez:
Expands --syslog argument to accept a comma-separated list of directives to control logging in the same manner as SCCACHE_LOG.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,6 +109,8 @@ zip = { version = "0.6", default-features = false }
 zstd = "0.13"
 
 # dist-server only
+env_filter = "0.1.3"
+hostname = "0.4.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -109,6 +109,8 @@ zip = { version = "0.6", default-features = false }
 zstd = "0.13"
 
 # dist-server only
+env_filter = "0.1.3"
```
sylvestre, line 112: if it is only used for dist, we should make it optional no ?
 (or default-features = false

sylvestre:
could you please add a test to verify that it works correctly ? and won't regress in the future

sylvestre:
please reopen when ready

PR #2348, opened 2025-03-03
Update CI coverage: grcov/codecov

samuel-jimenez:
`-Zprofile` support has been [removed from Rust](https://github.com/rust-lang/rust/pull/131829).
Instead of attempting to work around it (#2306), this patch switches the CI to use LLVM instrumentation-based coverage,  
as recommended in [grcov #1240](https://github.com/mozilla/grcov/issues/1240).

Codecov syntax has been updated to v5. This action now [requires a token to be set](https://github.com/marketplace/actions/codecov#usage). This commit uses the `CODECOV_TOKEN` secret to hold this [token](https://docs.codecov.io/docs/frequently-asked-questions#section-where-is-the-repository-upload-token-found-). However, if desired,  token use can be [disabled per repository](https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token), or by configuring [OIDC](https://github.com/marketplace/actions/codecov#using-oidc).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -281,6 +281,11 @@ jobs:
     #            rustc: nightly
     env:
       RUST_BACKTRACE: 1
+      COVERAGE_REPORT_DIR: "target/debug"
+      COVERAGE_REPORT_FILE: "target/debug/lcov.info"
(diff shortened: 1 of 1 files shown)


sylvestre:
excellent, thanks!

PR #2350, opened 2025-03-05
Expand coverage to sccache-dist tests

samuel-jimenez:
Add testing matrix option for `dist-tests`.
Configures `LLVM_PROFILE_FILE` environment variable to collect coverage results.


sylvestre:
thanks!

PR #2362, opened 2025-03-14
Reword and correct the preprocessor cache mode documentation

ahartmetz:
The correction is that there is a *long* list of compiler options that disable preprocessor cache mode.

### DIFF
--- a/docs/Local.md
+++ b/docs/Local.md
@@ -8,13 +8,13 @@ The local storage only supports a single sccache server at a time. Multiple conc
 
 ## Preprocessor cache mode
 
-This is inspired by [ccache's direct mode](https://ccache.dev/manual/3.7.9.html#_the_direct_mode) and works roughly the same.
+This is inspired by [ccache's direct mode](https://ccache.dev/manual/3.7.9.html#_the_direct_mode) and works roughly the same. It allows to skip the preprocessor for cached results when compiling C/C++.
 
 [preprocessor cache] mode is controlled by a configuration option which is true by default, as well as additional conditions described below.
 
-In preprocessor cache mode, sccache caches the preprocessor step for C/C++ whenever possible. This can make the compilation much faster, since the preprocessor step takes a non-negligible amount of the entire compile time.
+In non-preprocessor cache mode, sccache's compilation pipeline for one C/C++ source file is inputs → preprocessing → cache lookup → (return outputs if cached / compilation otherwise). In preprocessor cache mode, the compilation pipeline is inputs → cache lookup → (return outputs if cached / preprocessing → compilation otherwise). This can make it much faster to return compilation outputs from cache.
 
-In order to cache the preprocessor step, [sccache] needs to remember, among other things, all files included by a given input file. To quote ccache's documentation:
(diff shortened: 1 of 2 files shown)


ahartmetz:
Draft because: The second commit needs review from someone who understands preprocessor cache mode! I think this explanation is much clearer, but I don't know if it is actually true. The previous text mentioned "caching the preprocessor result", but that doesn't seem to make any sense. If the preprocessor result is the same, compilation can be trivially skipped as well. So I think what is actually happening is skipping from input files to output files directly if there is a cached result, not skipping anything if there is a cache miss.

docs/Local.md
```diff
@@ -8,13 +8,13 @@ The local storage only supports a single sccache server at a time. Multiple conc
 
 ## Preprocessor cache mode
 
-This is inspired by [ccache's direct mode](https://ccache.dev/manual/3.7.9.html#_the_direct_mode) and works roughly the same.
+This is inspired by [ccache's direct mode](https://ccache.dev/manual/3.7.9.html#_the_direct_mode) and works roughly the same. It allows to skip the preprocessor for cached results when compiling C/C++.
 
 [preprocessor cache] mode is controlled by a configuration option which is true by default, as well as additional conditions described below.
 
-In preprocessor cache mode, sccache caches the preprocessor step for C/C++ whenever possible. This can make the compilation much faster, since the preprocessor step takes a non-negligible amount of the entire compile time.
+In non-preprocessor cache mode, sccache's compilation pipeline for one C/C++ source file is inputs → preprocessing → cache lookup → (return outputs if cached / compilation otherwise). In preprocessor cache mode, the compilation pipeline is inputs → cache lookup → (return outputs if cached / preprocessing → compilation otherwise). This can make it much faster to return compilation outputs from cache.
```
drahnr, line 15: This line would need some wrapping, as well as using backticks for the processing pipeline description if not a a triple-back-tick code block I find it hard to read inline as it is, harming the intent of the PR. It also requires a precise definition of "outputs" - is that the pre-processor generated C/C++ code or the compiled object file?

docs/Local.md
```diff
@@ -8,13 +8,13 @@ The local storage only supports a single sccache server at a time. Multiple conc
...
 
-In order to cache the preprocessor step, [sccache] needs to remember, among other things, all files included by a given input file. To quote ccache's documentation:
+To ensure that the cached outputs for a source file correspond to the un-preprocessed inputs, [sccache] needs to remember, among other things, all files included by that source file (i.e. the complete set of inputs). To quote ccache's documentation:
```
drahnr, line 17: I'd be explicit about the issue with `__TIME__` and `__DATE__` macros and the config flag that allows one to handle the trade-off.

drahnr (changes requested):
I think it needs one more pass to achieve its goal, the direction is good!

Thank you!

drahnr:
If you take a look into the code (older ref) https://github.com/mozilla/sccache/blob/cbe79c9800097c0cc69c76e1d9a1d93c348c3312/src/compiler/c.rs#L335 is where we do a lookup in the storage instance based on the preprocessor hashkey which is then used to do a lookup of the preprocessor output. And only then proceeeds to actual compilation `Box<dyn Compilation>` type utilizing the preprocessor output as input for the C compiler.

drahnr:
@ahartmetz is this ready for a review pass?

ahartmetz:
> @ahartmetz is this ready for a review pass?

It is now.

PR #2363, opened 2025-03-14
Configurable core count

ahartmetz:
Some nodes can run out of memory if all of their cores are used. There may be other reasons to limit the amount of cores used.

### DIFF
--- a/tests/harness/mod.rs
+++ b/tests/harness/mod.rs
@@ -216,6 +216,7 @@ fn sccache_server_cfg(
             token: DIST_SERVER_TOKEN.to_owned(),
         },
         toolchain_cache_size: TC_CACHE_SIZE,
+        core_count: None,
     }
 }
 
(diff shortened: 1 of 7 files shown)


tests/harness/mod.rs
```diff
@@ -430,6 +431,7 @@ impl DistSystem {
             Some(SocketAddr::from(([0, 0, 0, 0], server_addr.port()))),
             self.scheduler_url().to_url(),
             token,
+            4,
```
sylvestre, line 434: it would be great if the test could verify that, indeed, 4 core are being used

ahartmetz: I guess so, but that would add a lot more  work than the actual change - and you are seeing some of the first Rust code that I've written here, so I really don't know the tradeoffs etc to make a good decision about how to pull it off.
I have some ideas about scheduling improvements, though absolutely no guarantees that I'll get to them. That would need a corresponding test harness (maybe with a kind of mock compiler that takes a configurable amount of CPU time and memory and writes some kind of tracing output) with one of the more trivial checks being that the max number of jobs is not exceeded.

ahartmetz: I mean actually verify, not jusk asking the scheduler what it thinks its limit is, which seems a bit pointless because it "obviously works" (right?)

PR #2364, opened 2025-03-14
Improve the CARGO_INCREMENTAL checking

anholt:
Improve the error message to give the user a better hint of what to do in response, and make it so you can sccache --help without throwing the error.


Xuanwo (approved):
Thank you, makes sense to me.

Issue #2288, opened 2024-11-16
compiler wrapper trying to create temporary files in /dev


mischief:
hi,

without any additional setup or ceremony, i built HEAD with rustc/cargo 1.81.0 in gentoo and invoked sccache like the readme suggests: https://github.com/mozilla/sccache?tab=readme-ov-file#usage

```
mischief@beast:~/src/sccache $ cargo build --release
   Compiling libc v0.2.155
   Compiling proc-macro2 v1.0.75
   Compiling unicode-ident v1.0.12
...
    Finished `release` profile [optimized] target(s) in 3m 24s
```

i'm not a rust person, but i find it odd that sccache would attempt to make a temporary file in /dev/. a user program should never do this.

this occurred after more than 1 invocation of sccache with the same parameters. the first invocation had no error.

```
mischief@beast:~/src/sccache $ ./target/release/sccache cc -o /dev/null -c ~/code/c/fbufsize.c
sccache: encountered fatal error
sccache: error: Permission denied (os error 13) at path "/dev/.tmp48KCJX"
sccache: caused by: Permission denied (os error 13) at path "/dev/.tmp48KCJX"
```

fbufsize.c is nothing special, just a trivial program i had laying around.
```
#include <stdio.h>
#include <stdio_ext.h>

int main(void)
{
        setvbuf(stdout, NULL, _IONBF, 0);
        setvbuf(stderr, NULL, _IONBF, 0);
        printf("%ld\n", __fbufsize(stdout));
        printf("%ld\n", __fbufsize(stderr));
        return 0;
}
```

Xuanwo:
> ./target/release/sccache cc -o /dev/null -c ~/code/c/fbufsize.c

What will happen if we change `-o /dev/null`?

mischief:
why would sccache need to do this if the compilers it is wrapping doesn't? seems like an easy fix.

myzhang1029:
I understand their justification of an atomic fs move operation, but I feel like `rustc` accessing a half-written file itself is a bug in rustc/cargo instead of something `sccache` should solve.
This change originated in https://github.com/mozilla/sccache/commit/16ecd0ba3181a3afa25979a0ac917b5927b94b8d and was subsequently moved to `cache/cache.rs` in https://github.com/mozilla/sccache/commit/8a85b5fd74a028cc8b98757a6baa67e86cc24633.

PR #2369, opened 2025-03-20
Fall back to direct cache write if tempfile creation on the same fs fails

myzhang1029:
Please see #2288 for the discussion.

### DIFF
--- a/src/cache/cache_io.rs
+++ b/src/cache/cache_io.rs
@@ -143,6 +143,10 @@ impl CacheRead {
                 optional,
             } in objects
             {
+                if path == Path::new("/dev/null") {
+                    debug!("Skipping output to /dev/null");
+                    continue;
+                }
                 let dir = match path.parent() {
                     Some(d) => d,
                     None => bail!("Output file without a parent directory!"),
--- a/src/cache/cache_io.rs
+++ b/src/cache/cache_io.rs
@@ -143,14 +143,8 @@ impl CacheRead {
                 optional,
             } in objects
             {
-                #[cfg(unix)]
-                if path == Path::new("/dev/null") {
-                    debug!("Skipping output to /dev/null");
-                    continue;
-                }
-                #[cfg(windows)]
-                if path == Path::new("NUL") {
-                    debug!("Skipping output to NUL");
+                if path == null_path() {
+                    debug!("Skipping output to {}", path.display());
                     continue;
                 }
                 let dir = match path.parent() {
@@ -197,6 +191,15 @@ impl CacheRead {
         })
         .await?
     }
+}
+
+#[cfg(unix)]
+fn null_path() -> &'static Path {
+    Path::new("/dev/null")
+}
+#[cfg(windows)]
+fn null_path() -> &'static Path {
+    Path::new("NUL")
 }
 
 /// Data to be stored in the compiler cache.
--- a/src/cache/cache_io.rs
+++ b/src/cache/cache_io.rs
@@ -172,7 +172,7 @@ impl CacheRead {
                         // Fall back to writing directly to the final location
                         warn!("Failed to create temp file on the same file system: {e}");
                         let mut f = std::fs::File::create(&path)?;
-                        // `optional`` is false in this branch, so do not ignore errors
+                        // `optional` is false in this branch, so do not ignore errors
                         let mode = self.get_object(&key, &mut f)?;
                         if let Some(mode) = mode {
                             if let Err(e) = set_file_mode(path.as_path(), mode) {
--- a/src/cache/cache_io.rs
+++ b/src/cache/cache_io.rs
@@ -15,7 +15,7 @@ use crate::errors::*;
 use fs_err as fs;
 use std::fmt;
 use std::io::{Cursor, Read, Seek, Write};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use tempfile::NamedTempFile;
 use zip::write::FileOptions;
 use zip::{CompressionMethod, ZipArchive, ZipWriter};
@@ -154,15 +154,35 @@ impl CacheRead {
                 // Write the cache entry to a tempfile and then atomically
                 // move it to its final location so that other rustc invocations
                 // happening in parallel don't see a partially-written file.
-                let mut tmp = NamedTempFile::new_in(dir)?;
-                match (self.get_object(&key, &mut tmp), optional) {
-                    (Ok(mode), _) => {
-                        tmp.persist(&path)?;
-                        if let Some(mode) = mode {
-                            set_file_mode(&path, mode)?;
+                match (NamedTempFile::new_in(dir), optional) {
+                    (Ok(mut tmp), _) => {
+                        match (self.get_object(&key, &mut tmp), optional) {
+                            (Ok(mode), _) => {
+                                tmp.persist(&path)?;
+                                if let Some(mode) = mode {
+                                    set_file_mode(path.as_path(), mode)?;
+                                }
+                            }
+                            (Err(e), false) => return Err(e),
+                            // skip if no object found and it's optional
+                            (Err(_), true) => continue,
                         }
                     }
-                    (Err(e), false) => return Err(e),
+                    (Err(e), false) => {
+                        // Fall back to writing directly to the final location
+                        warn!("Failed to create temp file on the same file system: {e}");
+                        let mut f = std::fs::File::create(&path)?;
+                        // `optional`` is false in this branch, so do not ignore errors
+                        let mode = self.get_object(&key, &mut f)?;
+                        if let Some(mode) = mode {
+                            if let Err(e) = set_file_mode(path.as_path(), mode) {
+                                // Here we ignore errors from setting file mode because
+                                // if we could not create a temp file in the same directory,
+                                // we probably can't set the mode either (e.g. /dev/stuff)
+                                warn!("Failed to reset file mode: {e}");
+                            }
+                        }
+                    }
                     // skip if no object found and it's optional
                     (Err(_), true) => continue,
                 }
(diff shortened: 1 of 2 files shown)


src/cache/cache.rs
```diff
@@ -207,22 +208,41 @@ impl CacheRead {
...
+                            (Err(e), false) => return Err(e),
+                            // skip if no object found and it's optional
+                            (Err(_), true) => continue,
+                        }
+                    }
+                    (Err(e), false) => {
+                        // Fall back to writing directly to the final location
+                        warn!("Failed to create temp file on the same file system: {e}");
+                        let mut f = File::create(&path)?;
+                        // `optional`` is false in this branch, so do not ignore errors
```
AJIOB, line 240:
```suggestion
                        // `optional` is false in this branch, so do not ignore errors
```
nit: looks like a typo

sylvestre:
would it be possible to add a test to make sure we don't regress in the future?
thanks

myzhang1029:
Rebased on `main` and added two tests.

sylvestre:
could you please add high level tests in https://github.com/mozilla/sccache/blob/main/tests/system.rs too ? thanks

like
https://github.com/mozilla/sccache/blob/7025295445c607dac66843c572a2a8adc25e8136/tests/system.rs#L1825

myzhang1029:
@sylvestre Sorry for the delayed response. I am planning to test this by adding two functions that are similar to `test_basic_compile` in `system.rs`.

One of them should test `/dev/null` and the other for `/dev/stdout` (`AssertCmd::assert` seems to capture `stdout` already --- better approach maybe?) They would:
1. Compiles a file
2. Repeats the compilation command
3. Check that both commands succeed and that #miss and #hit are both 1.

myzhang1029:
On a second thought, I have a question:

I think I named the PR a bit too narrow. Maybe like 
> Fall back to direct cache write if parent directory is not writable

I was think of reframing the PR this way, because this way, we don't have to hardcode `/dev` anymore. We can instead:
1. run the first compilation command
2. `chmod u-w` the destination dir
3. run the second (cached) pass

src/cache/cache.rs
```diff
@@ -207,22 +208,41 @@ impl CacheRead {
                 optional,
             } in objects
             {
+                if path == Path::new("/dev/null") {
```
AJIOB: It should be `NUL` for Windows with MSVC, isn't it?

myzhang1029: Sorry I do not have a MSVC environment to test this so I wasn't sure. 

For `gcc`, it seems to be complicated by `gcc` appending `.exe` automatically:
- `-o NUL`: it gives me `NUL.exe`
- `-o NUL.`: `ld.exe: cannot open output file NUL.: Invalid argument`
- `-Wl,-o,NUL`: `ld.exe: final link failed: file truncated`
- `-c -o NUL`: nothing (good)
- `-c -o NUL.o`: `NUL.o` (as expected)

AJIOB: Looks like we should support `-c -o NUL` as a base one, isn't it?

P.S. I don't think we really support `-o NUL` variant with the compiling & linking together: @Xuanwo , do we?

myzhang1029: How do you think we should detect the OS? would a constant marked with `#[cfg(unix)]` and `#[cfg(windows)]` be enough?

myzhang1029:
@AJIOB I have pushed the change. Sorry for the delay.

src/cache/cache.rs
AJIOB: I prefer to keep /dev/null for windows too, because of cygwin and mingw as a possible shells

myzhang1029:
There is an issue with the tests: we use the same `INPUT`, so the cache stats depend on the previous tests. I feel like it would become a maintenance disaster to hardcode the expected stats into the tests in this case. Would it be acceptable to put a pair of `start_local_daemon` `stop_local_daemon` in each test?

myzhang1029:
CI should work now.

I have relaxed one more error handling: in the case that same-fs file creation fails, we also allow `chmod` to fail.

AJIOB:
Hi @myzhang1029 ,

I face with that error
As a result, the `OpenSSL` Perl incorrectly generates assembly file.

Do you have some plans to actualize the PR?

myzhang1029:
Hello @AJIOB 

Is my understanding correct that this patch would fix that error you mentioned?

Lots has happened in the repo and let me work on this today!

AJIOB:
I hope it helps.

In the trace mode I have:
```
[2026-03-11T12:18:02Z DEBUG sccache::compiler::compiler] Found gcc
[2026-03-11T12:18:02Z TRACE sccache::server] Inserting POSSIBLY PROXIED cache map info for "/usr/bin/gcc"
[2026-03-11T12:18:02Z DEBUG sccache::server] check_compiler: Supported compiler
[2026-03-11T12:18:02Z DEBUG sccache::server] parse_arguments: Ok: ["-Wa,-v", "-c", "-o", "/dev/null", "-x", "assembler", "/dev/null"]
[2026-03-11T12:18:02Z DEBUG sccache::compiler::compiler] [null]: get_cached_or_compile: ["-Wa,-v", "-c", "-o", "/dev/null", "-x", "assembler", "/dev/null"]
[2026-03-11T12:18:02Z TRACE sccache::util] Hashed 0 files in 0.000 s
[2026-03-11T12:18:02Z DEBUG sccache::compiler::c] generate_hash_key: Disabling preprocessor cache because assembler language doesn't need C preprocessing
[2026-03-11T12:18:02Z DEBUG sccache::compiler::compiler] [null]: generate_hash_key took 0.000 s
[2026-03-11T12:18:02Z DEBUG sccache::compiler::compiler] [null]: Hash key: 11b765162fadb5068b7bb35ddec6a07e17f458f37cef7c0d884955a66981f15c
[2026-03-11T12:18:02Z TRACE sccache::cache::disk] DiskCache::get(11b765162fadb5068b7bb35ddec6a07e17f458f37cef7c0d884955a66981f15c)
[2026-03-11T12:18:02Z DEBUG sccache::compiler::compiler] [null]: Cache hit in 0.014 s
[2026-03-11T12:18:02Z ERROR sccache::server] ["null"] fatal error: Permission denied (os error 13) at path "/dev/.tmp9gyxnV"
[2026-03-11T12:18:02Z ERROR sccache::server] ["null"] 	Permission denied (os error 13) at path "/dev/.tmp9gyxnV"
```

AJIOB:
P.S. Just an easy-to-reproduce test on the latest sccache 0.14.0:
```
$ gcc -Wa,-v -c -o /dev/null -x assembler /dev/null
GNU assembler version 2.42 (x86_64-linux-gnu) using BFD version (GNU Binutils for Ubuntu) 2.42
$ sccache gcc -Wa,-v -c -o /dev/null -x assembler /dev/null
sccache: error: Permission denied (os error 13) at path "/dev/.tmpNJtyLa"
sccache: caused by: Permission denied (os error 13) at path "/dev/.tmpNJtyLa"
```

AJIOB:
Hi @myzhang1029 ,

Do you have any updates?

myzhang1029:
Sorry for the delay as my area was hit by some severe weather.

This time is largely just rebasing the code and resolving conflicts, but this version should already fix the `openssl` building error you mentioned. What is left is to figure out the tests properly.

AJIOB:
Hi @sylvestre,

Is this PR OK to you?

myzhang1029:
Moved the unit tests to `cache_io.rs` since it uses a private interface.
I've moved the integration tests to branch `pr2369-tests` since I realized that the expected constants (counts) depend on the platform.

Edit: The integration tests have been merged back.

myzhang1029:
@AJIOB I believe everything that has been discussed so far is addressed now. Please kindly review!

AJIOB (approved):
All looks fine for me

AJIOB:
CC @sylvestre ,

Can we merge it, please?

myzhang1029:
Just gonna rebase onto `main` to see if the new tests still work :)

AJIOB:
Hi @drahnr ,

Is it OK for you?

myzhang1029:
another rebase onto main, mainly for checking compatibility with #2581

src/cache/cache_io.rs
```diff
@@ -143,22 +143,52 @@ impl CacheRead {
...
+                #[cfg(unix)]
+                if path == Path::new("/dev/null") {
+                    debug!("Skipping output to /dev/null");
+                    continue;
+                }
+                #[cfg(windows)]
+                if path == Path::new("NUL") {
+                    debug!("Skipping output to NUL");
+                    continue;
+                }
```
drahnr, line 157: Could we move the check into a function so the target specifics don't appear in scope?

tests/system.rs
```diff
@@ -265,6 +267,110 @@ fn test_basic_compile(compiler: Compiler, tempdir: &Path) {
...
+        assert!(info.stats.cache_misses.get("C/C++").is_none());
+        let adv_key = adv_key_kind("c", compiler.name);
+        assert_eq!(&2, info.stats.cache_hits.get_adv(&adv_key).unwrap());
+        assert!(info.stats.cache_misses.get_adv(&adv_key).is_none());
+    });
+}
+
+#[cfg(not(unix))]
+fn test_basic_compile_into_dev_stdout(_: Compiler, _: &Path) {
+    warn!("Not unix, skipping tests with /dev/stdout");
```
drahnr: Please downgrade to an info level log.

drahnr (approved):

src/cache/cache_io.rs
myzhang1029: Done

myzhang1029:
I got my hands on a windows machine and it turned out I simplified the NUL mess. [Now fixed](https://github.com/mozilla/sccache/pull/2369/changes#diff-cfa6a2b9469ebacdfb8512b0b7486cd85707ad354d82887c8b70fe20e136c859R210).

myzhang1029:
@drahnr The testsuites seem to work on my FreeBSD VM. Could you please have a look? Thanks!

myzhang1029:
@drahnr @sylvestre Hi Both,

How do you feel about merging this patch?

PR #2356, opened 2025-03-07
Support nvcc device debug flags

trxcllnt:
This PR adds support for `nvcc -G`/`nvcc --device-debug`. Presently `-G` is interpreted as a GCC argument that expects a numeric value. nvcc doesn't support that, so instead it should be intercepted and passed to the `nvcc --dryrun` commands.

While adding the tests for the new flag, I took the opportunity to refactor the CUDA tests (https://github.com/mozilla/sccache/commit/52bd3aa6d0c982e512896fd12c71eab9750f4c98). Rather than individually asserting various struct fields match exact values, it's easier to create a `ServerStats` instance, increment the values I expect to change, and compare against the real stats.

When I did this, I discovered there was an unexpected cache error related to a hash collision between one of the inner `ptxas` calls colliding with the later `nvcc --cubin` invocation. This is an edge case, but I'm glad the refactored tests caught it.

In addition, we can now run the tests with `SCCACHE_DEBUG=1` and optionally `SCCACHE_LOG=`/`SCCACHE_ERROR_LOG=` to instruct the test harness not to delete the tmpdirs, and to redirect sccache daemon logs to a file:
```shell
$ RUST_BACKTRACE=1 SCCACHE_DEBUG=1 SCCACHE_ERROR_LOG=$(pwd)/sccache.log \
  cargo test --no-fail-fast -F all,dist-server,vendored-openssl --package sccache test_cuda_sccache_command
```

PR #2382, opened 2025-04-15
Refactor system/dist/CUDA tests

trxcllnt:
This PR extracts the test changes from #2356 per @drahnr's request.

The system and dist tests now use an `SccacheClient` instance (https://github.com/mozilla/sccache/commit/a519fc30490dde92f5029cd834115689f81910a5), which is a struct that wraps an sccache client process. Each `SccacheClient` instance starts its sccache client on a distinct port, and shuts it down when `SccacheClient` dropped.

With this improvement, system (and dist) tests can now run in parallel. Since each test creates its own isolated `SccacheClient` instance, no need to carefully manage global `start_local_daemon()`, `stop_local_daemon()`, `zero_stats()` etc. calls.

The second commit (https://github.com/mozilla/sccache/commit/b307d9a269e0dc701e19b13b38512fe5a1532f81) refactors the CUDA tests to be easier to read and update. Rather than individually asserting various struct fields match exact values, it's easier to create a `ServerStats` instance, increment the values I expect to change, and compare against the real stats.


trxcllnt:
The failing nvcc tests on Windows will be fixed by bug fixes in the follow-up PR.

sylvestre:
i am not comfortable landing a change that would deliberately break the CI :/

trxcllnt:
@sylvestre https://github.com/mozilla/sccache/pull/2383 branches from this PR, I just split them up per a request from @drahnr.

#2383 has both sets of changes (diff of the two branches [here](https://github.com/trxcllnt/sccache/compare/fea/enhance-system-tests...trxcllnt:sccache:fix/consistent-nvcc-internal-file-names?expand=1)), so it's fine to ignore this one and merge #2383 instead.

I will also be adding `nvcc --device-debug` and nvcc binary compilation support in separate follow-up PRs, since those are very minor changes compared to these.

trxcllnt:
cc: @robertmaynard for review

drahnr:
My suggestions was to split out the structural test changes, I didn't say broken CI would be acceptable :) If there are specific tests that would be fixed later on, please defer them to the relevant other PR(s).

trxcllnt:
The test improvements exposed an underlying bug, because now they compare fields that weren't compared before. I apologize if that wasn't clear before.

The test changes are isolated to this PR to make them easier to review, but they are also included in the PR (https://github.com/mozilla/sccache/pull/2383) that includes the fix. The diff of the test changes and the fix is [here](https://github.com/trxcllnt/sccache/compare/fea/enhance-system-tests...trxcllnt:sccache:fix/consistent-nvcc-internal-file-names?expand=1).

The changes to the tests can be merged separately from the bug fix by merging this PR, then immediately merging #2383, or they can be merged together by closing this PR and only merging #2383.

Let me know if there's anything I can do to make reviewing these easier.

PR #2383, opened 2025-04-15
Use consistent names for internal `nvcc` files

trxcllnt:
This PR extracts the fixes changes from https://github.com/mozilla/sccache/pull/2356 per @drahnr's request.

This PR branched from/is a follow-up to https://github.com/mozilla/sccache/pull/2382. See the diff of the two branches [here](https://github.com/trxcllnt/sccache/compare/fea/enhance-system-tests...trxcllnt:sccache:fix/consistent-nvcc-internal-file-names?expand=1).

The names for the internal files depend on the compilation flag and device architectures. `nvcc` generates a different name for the `.cpp1.ii`, `.cudafe1.c`, `.cudafe1.stub.c`, `.cudafe1.gpu`, `.ptx`, and `.cubin` files when the compile flag is `-ptx`, `-cubin`, or `-c`, and also on whether there's one vs. many `-gencode` arguments. Additionally, it will either include or omit the `--gen_module_id_file` flag from the `cicc` invocation based on whether the compile flag is `-ptx`, `-cubin`, or `-c`.

Some examples:
```shell
# compile flag: -ptx, single arch
$ nvcc -x cu -ptx x.cu -o x.cu.o -gencode=arch=compute_60,code=[compute_60,sm_60] --dryrun --keep --keep-dir /x 2>&1 | grep -P '(cpp1\.ii|\.ptx|\.cubin)'
#$ gcc -E ... "x.cu" -o "/x/x.cpp1.ii" 
#$ "$CICC_PATH/cicc" ... --gen_module_id_file --module_id_file_name "/x/x.module_id"  "/x/x.cpp1.ii" -o "x.cu.o"

# compile flag: -cubin, single arch
$ nvcc -x cu -cubin x.cu -o x.cu.o -gencode=arch=compute_60,code=[sm_60] --dryrun --keep --keep-dir /x 2>&1 | grep -P '(cpp1\.ii|\.ptx|\.cubin)'
#$ gcc -E ... "x.cu" -o "/x/x.cpp1.ii" 
#$ "$CICC_PATH/cicc" ... --gen_module_id_file --module_id_file_name "/x/x.module_id" --gen_c_file_name "/x/x.cudafe1.c" --stub_file_name "/x/x.cudafe1.stub.c" --gen_device_file_name "/x/x.cudafe1.gpu"  "/x/x.cpp1.ii" -o "/x/x.ptx"
#$ ptxas -arch=sm_60 -m64  "/x/x.ptx"  -o "x.cu.o"

# compile flag: -c, single arch
$ nvcc -x cu -c x.cu -o x.cu.o -gencode=arch=compute_60,code=[compute_60,sm_60] --dryrun --keep --keep-dir /x 2>&1 | grep -P '(cpp1\.ii|\.ptx|\.cubin)'
#$ gcc -E ... "x.cu" -o "/x/x.cpp1.ii" 
#$ "$CICC_PATH/cicc" ... --module_id_file_name "/x/x.module_id" --gen_c_file_name "/x/x.cudafe1.c" --stub_file_name "/x/x.cudafe1.stub.c" --gen_device_file_name "/x/x.cudafe1.gpu"  "/x/x.cpp1.ii" -o "/x/x.ptx"
#$ ptxas -arch=sm_60 -m64  "/x/x.ptx"  -o "/x/x.sm_60.cubin" 

# compile flag: -c, multiple archs
$ nvcc -x cu -c x.cu -o x.cu.o -gencode=arch=compute_60,code=[sm_60] -gencode=arch=compute_70,code=[compute_70,sm_70] --dryrun --keep --keep-dir /x 2>&1 | grep -P '(cpp1\.ii|\.ptx|\.cubin)'
#$ gcc -E ... "x.cu" -o "/x/x.compute_60.cpp1.ii" 
#$ "$CICC_PATH/cicc" ... --module_id_file_name "/x/x.module_id" --gen_c_file_name "/x/x.compute_60.cudafe1.c" --stub_file_name "/x/x.compute_60.cudafe1.stub.c" --gen_device_file_name "/x/x.compute_60.cudafe1.gpu"  "/x/x.compute_60.cpp1.ii" -o "/x/x.compute_60.ptx"
#$ ptxas -arch=sm_60 -m64  "/x/x.compute_60.ptx"  -o "/x/x.compute_60.cubin" 
#$ gcc -E ... "x.cu" -o "/x/x.compute_70.cpp1.ii" 
#$ "$CICC_PATH/cicc" ... --module_id_file_name "/x/x.module_id" --gen_c_file_name "/x/x.compute_70.cudafe1.c" --stub_file_name "/x/x.compute_70.cudafe1.stub.c" --gen_device_file_name "/x/x.compute_70.cudafe1.gpu"  "/x/x.compute_70.cpp1.ii" -o "/x/x.compute_70.ptx"
#$ ptxas -arch=sm_70 -m64  "/x/x.compute_70.ptx"  -o "/x/x.compute_70.sm_70.cubin" 
```

From the above, we observe that:
* `.cpp1.ii`, `.cudafe1.c`, `.cudafe1.stub.c`, `.cudafe1.gpu`, and `.ptx` files are either:
  * `x.<suffix>`
  * `x.compute_XX.<suffix>`
* `.cubin` files are either:
  * `x.cubin`
  * `x.compute_XX.cubin`
  * `x.compute_XX.sm_XX.cubin`
* without `-c`, the `cicc` command includes `--gen_module_id_file`
* with `-c`, the `cicc` command omits `--gen_module_id_file`

Now that we're hashing all the `cudafe++`, `cicc`, and `ptxas` arguments, these differences yield cache misses when they should be hits. So for simplicity, I updated the renaming logic to rename to the longest form of each (i.e. `x.compute_XX.ptx`, `x.compute_XX.sm_XX.cubin`), and always add the `--gen_module_id_file` flag to `cicc` invocations.


trxcllnt:
Looks like integration tests are failing due to https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts

trxcllnt:
cc: @robertmaynard for review

robertmaynard (approved):

src/compiler/cicc.rs
```diff
@@ -329,23 +317,22 @@ pub fn generate_compile_commands(
 }
 
 ArgData! { pub
-    Output(PathBuf),
-    PassThrough(OsString),
-    UnhashedFlag,
+    ExtraOutput(PathBuf),
```
trxcllnt, line 320: I renamed this from `UnhashedOutput -> ExtraOutput` to reflect that these argument types are now hashed.

src/compiler/gcc.rs
```diff
@@ -905,10 +906,21 @@ where
...
+                        Language::C | Language::Cxx => language = Some("cpp-output".into()),
+                        Language::GenericHeader | Language::CHeader | Language::CxxHeader => {}
+                        _ => *(language.as_mut()?) = "none".into(),
+                    }
+                } else {
+                    match parsed_args.language {
+                        Language::C => language = Some("cpp-output".into()),
+                        Language::GenericHeader | Language::CHeader | Language::CxxHeader => {}
+                        _ => language.as_mut()?.push_str("-cpp-output"),
+                    }
```
trxcllnt, line 923: This change slipped in when I cherry-picked the changes into this branch. It isn't strictly related to `nvcc`, but it does fix the issue using `nvc++` as a host compiler with sccache-dist.

src/compiler/msvc.rs
```diff
@@ -1025,11 +1021,8 @@ where
                 writeln!(f, "{}:", dep)?;
             }
         }
-        Ok(process::Output {
-            status,
-            stdout,
-            stderr: stderr_bytes,
-        })
```
trxcllnt, line 1032: The L974-L1032 changes are also unrelated to the functionality, pulled in from another branch by accident. I'll revert them.

sylvestre:
i wish this kind of changes would be done in a separate PR 
I can't either:
* squash this PR into a single commit given some unrelated changes
* merge it as it has some commits like "[revert unrelated changes](https://github.com/mozilla/sccache/pull/2383/commits/caf89552f165a08a63c08b60758381bbf76a89d4)"

Maybe split this PR into several

Note: it is why it takes time to merge your PR, split them into smaller PR would make our life much easier

trxcllnt:
@sylvestre I am fine with either squashing or merging. If you'd prefer to squash, are there files you'd like me to revert? If merge, I can rebase out the follow-up commits.

drahnr:
> Maybe split this PR into several

would be my personal preference.

sylvestre:
same, smaller PR would be ideal :)

trxcllnt:
I rebased on `main` and squashed the changes in https://github.com/trxcllnt/sccache/commit/ac44e9a6a66fb2f857a37fdf3f4727714c01cdd9, https://github.com/trxcllnt/sccache/commit/3c08b18caa73f161fb60ef8148c71a6671cf0934, and https://github.com/trxcllnt/sccache/commit/caf89552f165a08a63c08b60758381bbf76a89d4 into a single commit.

I can make a separate PR with the test changes after this one. How does that sound?

src/compiler/nvcc.rs
```diff
@@ -401,25 +407,45 @@ pub fn generate_compile_commands(
         .unwrap()
         .path;
 
+    // Build nvcc's internal files in `$TMPDIR/$hash_key` so the paths are
+    // stable across compilations. This is important because this path ends
+    // up in the preprocessed output, so using random tmpdir paths leads to
+    // erroneous cache misses.
+    let out_dir = env::temp_dir().join("sccache_nvcc").join({
+        // Combine `hash_key` with the output path in case
```
drahnr, line 415: Can we rename `hash_key` to something less ambiguous? It's temp dir subdir stabilizer, `hash_key` is not encoding that, it also needs docs in the `trait` definition.

src/compiler/nvcc.rs
```diff
@@ -401,25 +407,45 @@ pub fn generate_compile_commands(
...
+        // the same file is concurrently built to separate
+        // output paths.
+        let mut m = crate::util::Digest::new();
+        m.update(hash_key.as_bytes());
+        m.update(cwd.join(output).as_os_str().as_encoded_bytes());
+        m.finish()
+    });
+    fs::create_dir_all(&out_dir).ok();
```
drahnr, line 423: `let _ = fs::create_dir_all(..)` would be preferable, although I don't understand how it'd be acceptable to fail creating the directory?

trxcllnt: That's fine with me

src/compiler/nvcc.rs
trxcllnt: I called it `hash_key` because it's the key produced by the call to `generate_hash_key()` for this `nvcc` compilation. I am fine calling it something else, it just seemed the most descriptive and consistent name.

trxcllnt:
I do plan to update the `system.rs` tests to account for the new numbers, it just takes a few hours in the current state and I've been busy with other things lately.

src/compiler/nvcc.rs
trxcllnt: Done

trxcllnt:
I updated the tests in `system.rs` so they're all now passing.

The rust v1.75.0 jobs are failing to `cargo install grcov`, but that looks to be happening in other PRs and is unrelated to the changes here. Is there a cargo flag or envvar we can set to allow unstable features when installing `grcov`?

drahnr:
CC @sylvestre re grcov

sylvestre:
it is change in the dep tree of grcov

src/compiler/nvcc.rs
drahnr: Please rename this variable, `hash_key` my preference would be `unique_comp_key`, that's a clear indicator of _what_ it is and what it could be used for.

src/compiler/nvcc.rs
```diff
@@ -743,19 +865,38 @@ where
...
-                    // If `nvcc` is invoked with `-c` (or any of its variants), remove the
-                    // `--gen_module_id_file` flag. In this mode, we instruct `cudafe++`
-                    // to generate this file, so cicc shouldn't generate it again.
-                    Some("-c") | Some("--compile") | Some("-dc") | Some("--device-c")
-                    | Some("-dw") | Some("--device-w") => {
-                        if let Some(idx) = args.iter().position(|x| x == &gen_module_id_file_flag) {
-                            args.splice(idx..(idx + 1), []);
-                        }
+                // Fix for CTK < 12.0:
+                // Remove `--gen_module_id_file` if cudafe++ already does it
```
drahnr, line 869: I am not sure we want to encode `cudafe++` specific behaviour. I pressume `cudafe++` is just one example, and the fix is adjusting the path.

If this is done though, then any custom logic for output dependencies _might_ depend on that. Could you elaborate if you see any further side effects of this?

src/compiler/nvcc.rs
```diff
@@ -743,19 +865,38 @@ where
...
+                if cudafe_has_gen_module_id_file_flag {
+                    if let Some(idx) = args.iter().position(|x| x == &gen_module_id_file_flag) {
```
drahnr:
```suggestion
                    if let Some(idx) = args.iter().position(|arg| arg == &gen_module_id_file_flag) {
```

src/compiler/nvcc.rs
```diff
@@ -743,19 +865,38 @@ where
...
+                // Add these flags if they're missing:
+                // * `--gen_c_file_name test_a.compute_XX.cudafe1.c`
+                // * `--stub_file_name test_a.compute_XX.cudafe1.stub.c`
+                // * `--gen_device_file_name test_a.compute_XX.cudafe1.gpu`
+                //
+                // This ensures the same `cicc` command is generated regardless
+                // of whether the compilation flag is `-c`, `-ptx`, or `-cubin`
+
+                // e.g. test_a.compute_XX.cpp1.ii
+                let mut nidx = args.len() - 3;
```
drahnr, line 884: Did we check anywhere this holds?

src/compiler/nvcc.rs
```diff
@@ -743,19 +865,38 @@ where
...
+                let name = args[nidx].clone();
+                // test_a.compute_XX.cpp1.ii -> test_a.compute_XX
+                let name = name.split(".cpp1.ii").next().unwrap();
```
drahnr, line 887:
```suggestion
                let name = name.rsplit_once().ok_or_else(|| { err })?.0;
```

drahnr: Q: `.cpp1.ii.cpp1.ii` should be split how?

src/compiler/nvcc.rs
```diff
@@ -743,19 +865,38 @@ where
...
+
+                for (flag, name) in [
+                    (&gen_c_file_name_flag, format!("{name}.cudafe1.c")),
```
drahnr, line 890: Could you elaborate the signficance of the `cudafe1.*` file extension? I pressume these are intermediate, generate C files specific to `cudafe++` and that's hardcoded in `cudafe++`?

src/compiler/nvcc.rs
```diff
@@ -743,19 +865,38 @@ where
...
-                    Some("-c") | Some("--compile") | Some("-dc") | Some("--device-c")
-                    | Some("-dw") | Some("--device-w") => {
-                        if let Some(idx) = args.iter().position(|x| x == &gen_module_id_file_flag) {
-                            args.splice(idx..(idx + 1), []);
-                        }
+                // Fix for CTK < 12.0:
+                // Remove `--gen_module_id_file` if cudafe++ already does it
+                if cudafe_has_gen_module_id_file_flag {
+                    if let Some(idx) = args.iter().position(|x| x == &gen_module_id_file_flag) {
+                        args.splice(idx..idx + 1, []);
```
drahnr, line 872: I am pretty sure there is a better way to the above.

src/compiler/nvcc.rs
```diff
@@ -671,50 +723,123 @@ where
...
+    // ```
+    //
+    // The second command yields a false-positive cache miss because the names are different.
+    //
+    // This matters because CI jobs will often compile .cu files in "many-arch" mode, but devs
+    // who have just one GPU locally prefer to only compile for their specific GPU arch. It is
+    // preferrable if they can reuse the PTX populated by "many-arch" CI jobs.
+    //
+    // So to avoid this, we detect these "single-arch" compilations and rewrite the names to
+    // match what nvcc generates for "many-arch" compilations.
```
drahnr, line 786: :+1:

src/compiler/nvcc.rs
```diff
@@ -671,50 +723,123 @@ where
...
+        .map(|(_, dir, exe, args)| (dir.to_owned(), exe.to_owned(), args.to_owned()))
+        .collect::<Vec<_>>();
+
+    // First pass over commands because in CTK < 12.0, `cudafe++` is at the end of the commands list,
+    // but we need to set `cudafe_has_gen_module_id_file_flag` in order to adjust the cicc commands.
+    for (dir, exe, args) in all_commands.iter_mut() {
+        if let Some("cudafe++") = exe.file_stem().and_then(|s| s.to_str()) {
+            // Fix for CTK < 12.0:
+            // Add `--gen_module_id_file` if the cudafe++ args include `--module_id_file_name`
+            if let Some(idx) = args.iter().position(|x| x == &module_id_file_name_flag) {
```
drahnr:
```suggestion
            if let Some(idx) = args.iter().position(|arg| arg == &module_id_file_name_flag) {
```

src/compiler/nvcc.rs
```diff
@@ -1114,104 +1237,262 @@ fn fold_env_vars_or_split_into_exe_and_args(
...
+                        };
+
+                        if arg_is_msvc_preprocessor_output {
+                            format!("-Fi{}", arg)
+                        } else {
+                            arg
+                        }
+                    })
+                    .collect::<Vec<_>>()
+                    .iter()
```
drahnr, line 1478: nit: do we need the intermediate `collect()`?

src/compiler/nvcc.rs
trxcllnt: Here the name is something like `x.compute_60.cpp1.ii`, so splitting on the suffix and taking the first value gives just the name `x.compute_60`.

drahnr: Is that intended how it's supposed to work?

src/compiler/nvcc.rs
trxcllnt: Unfortunately yes. Both `map` selectors borrow `nvcc_internal_files` as mutable, so the first `map` result must be collected before running the second `map` :slightly_frowning_face:.

src/compiler/nvcc.rs
trxcllnt: Yes, the last three arguments to cicc are always `input.cpp1.ii -o output.ptx`

src/compiler/nvcc.rs
trxcllnt: Yes, this is intended. This block is part of ensuring consistent `cicc` arguments, regardless of which `-gencode` flags are or aren't passed.

The input file name is something like `foo.compute75.cpp1.ii`, so the three flags we splice in are:
```shell
--gen_c_file_name foo.compute75.cudafe1.c \
--stub_file_name foo.compute75.cudafe1.stub.c \
--gen_device_file_name foo.compute75.cudafe1.gpu
```

So the logic here extracts the `foo.compute75` component from the input file name and uses it as the prefix for the file name for each of these three flags.

src/compiler/nvcc.rs
trxcllnt: This logic normalizes the command order and arguments due to differing behavior between CUDA <=11 and >=12.

CUDA <=11 nvcc generates the commands in an order like this:
```shell
cicc -arch=60 --module_id_file_name foo.module_id --gen_module_id_file ...
...
cicc -arch=70 --module_id_file_name foo.module_id ...
...
cudafe++ --module_id_file_name foo.module_id ...
```

In the above, the first `cicc` call creates `foo.module_id` on disk due to `--gen_module_id_file`, which is consumed by the later `cicc` and `cudafe++` calls. This is problematic, since we want to potentially run both `cicc` calls in parallel.

CUDA >= 12 nvcc generates the `cudafe++` command first, and puts the `--gen_module_id_file` flag on it.

So here we reorganize and splice in/out arguments so the commands are always of the CUDA >=12 form.

> any custom logic for output dependencies might depend on that

Since we're ensuring the command arguments are consistent, we should always get the same outputs for all our `cudafe++`, `cicc`, and `ptxas` invocations, even though nvcc's original commands are dependent on the `-gencode` flags.

src/compiler/nvcc.rs
trxcllnt: Done

src/compiler/nvcc.rs
trxcllnt: Done

src/compiler/nvcc.rs
trxcllnt: Done

src/compiler/nvcc.rs
```diff
@@ -438,28 +464,21 @@ pub fn generate_compile_commands(
...
-    arguments.push(
-        (if parsed_args.input.is_absolute() {
-            parsed_args.input.clone()
-        } else {
-            cwd.join(&parsed_args.input).canonicalize().unwrap()
-        })
-        .into(),
-    );
+    // inputs that should produce cache hits.
+    arguments.push(cwd.join(&parsed_args.input).canonicalize().unwrap().into());
```
drahnr, line 468: This does change behaviour, we now assume all relative paths?

src/compiler/nvcc.rs
```diff
@@ -516,58 +544,82 @@ impl CompileCommandImpl for NvccCompileCommand {
...
+                            let src = entry.path();
+                            let dst = keep_dir.join(entry.file_name());
+                            if dst == src || !src.exists() {
+                                continue;
+                            }
+                            trace!(
+                                "[{}]: maybe_keep_temps_then_clean move {src:?} -> {dst:?}",
+                                output_path.display()
+                            );
+                            fs::rename(src, dst)?;
```
drahnr, line 614: Can we ensure it's the same filesystem? Or have a fallback? Otherwise `rename` will fail.

src/compiler/nvcc.rs
trxcllnt: If `parsed_args.input` is absolute, `cwd.join()` is a no-op. This is a more succinct form of the code it replaces.

src/compiler/nvcc.rs
trxcllnt: How would we check if `src` and `dst` are on the same file system? Should I use `fs::copy()` and `fs::remove_file()` instead of `fs::rename()`?

trxcllnt:
Not sure what's going on with Windows2022, but it looks like `clang-cl` was updated recently?

~I can try restricting those jobs to CUDA 12.5 instead of 12.8.~ edit: this isn't the issue.

I see `Found clang++` in the logs of the failing tests, so it appears `clang-cl` isn't being detected as `msvc-clang` [here](https://github.com/mozilla/sccache/blob/5b1e93e5dbd8113bb66ac990148e8ba354dd3545/src/compiler/compiler.rs#L1502).

I ran CI on a branch to print out the version: `Found clang++ (version: "Clang 19.1.5")`

Changing the test to search for `clang-cl` on Windows [fails](https://github.com/trxcllnt/sccache/actions/runs/15936233757/job/44956573708#step:9:1819) with the same error. This feels like a bug in the latest VS release (or the recent GHA Windows runner image [update](https://github.com/actions/runner-images/pull/12459)).

src/compiler/nvcc.rs
drahnr: Revisiting, it's ok for the time being

drahnr (approved):
@sylvestre LGTM, you might want to give it a quick pass

trxcllnt:
Don't merge this yet, I need to make a few updates.

ahartmetz:
FYI, windows CI is fixed now, in case that was blocking you.

sylvestre:
needs some rebasing

PR #2385, opened 2025-04-15
github storage: force version 2

sylvestre:


trxcllnt:
Per the [OpenDAL source](https://github.com/apache/opendal/blob/b1d036523869f5d2e196e8d67234fb30e021a60f/core/src/services/ghac/core.rs#L351-L352), I think `ACTIONS_CACHE_URL` in this file needs to be `ACTIONS_RESULTS_URL`.

sylvestre:
right, thanks!

.github/workflows/integration-tests.yml
```diff
@@ -688,7 +689,7 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
-            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RESULTS_URL', process.env.ACTIONS_RESULTS_URL || '');
```
trxcllnt, line 692: Also need to update lines 258, 542, and 602.

sylvestre: i was sure i did a full replace :( 
Thanks!

trxcllnt (approved):

docs/Configuration.md
```diff
@@ -201,7 +201,7 @@ The full url appears then as `redis://user:passwd@1.2.3.4:6379/?db=1`.
 
 #### gha
 
-* `SCCACHE_GHA_CACHE_URL` / `ACTIONS_CACHE_URL` GitHub Actions cache API URL
+* `SCCACHE_GHA_CACHE_URL` / `ACTIONS_RESULTS_URL` GitHub Actions cache API URL
```
AJIOB, line 204:
```suggestion
* `SCCACHE_GHA_RESULTS_URL` / `ACTIONS_RESULTS_URL` GitHub Actions cache API URL
```

And in the implementation, please.

IMO, they should be in sync.

AJIOB, line 204: What about backward compatibility of the variable names? Your changes break it. Is it what you really want?

PR #2392, opened 2025-05-06
Fix preprocessor cache mode + (distributed compilation || -MD dependency scanning)

ahartmetz:


sylvestre:
nice
a few jobs are failing btw

ahartmetz:
This seems really annoying to auto-test, and to be honest, I haven't even tested it manually yet ;)
I will obviously need to do at least that.

ahartmetz:
So I did do some practical testing and ran into another blocker for direct mode, this time that it didn't work with -MD (which is used by at least Make and Ninja backends of CMake, so very widely I'd say).
@glandium, do the last three commits look reasonable to you?

ahartmetz:
Well, that cherry-pick didn't fix the tests.

sylvestre:
nice, green builds :)

Could you please add tests ?
Especially tests that would fail without your changes and would pass now
thanks

src/compiler/rust.rs
```diff
@@ -1321,35 +1321,11 @@ where
...
-                    target_json,
-                    ..
-                },
-        } = *self;
-        trace!("[{}]: generate_hash_key", crate_name);
+        trace!("[{}]: generate_hash_key", self.parsed_args.crate_name);
         // TODO: this doesn't produce correct arguments if they should be concatenated - should use iter_os_strings
-        let os_string_arguments: Vec<(OsString, Option<OsString>)> = arguments
+        let os_string_arguments: Vec<(OsString, Option<OsString>)> = self
+            .parsed_args
```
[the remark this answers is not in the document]
ahartmetz: Is that actually considered a problem and worth solving?
From how I know the similar "self.variable" usage in Python, the self. part is more considered part of the name and self-documentation rather than useless boilerplate.

Anything else that needs doing?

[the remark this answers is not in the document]
sylvestre: nope, just ignore it :)

ahartmetz:
> nice, green builds :)
> 
> Could you please add tests ? Especially tests that would fail without your changes and would pass now thanks

For dependency files (-MD etc) with preprocessor cache mode, I have it easy: when preprocessor cache mode was disabled in case of -MD etc detected: c0a0b6ebd8a0b03aa76cf658d3d44b282aa14f01, there were tests added which I kept, and they failed in builds without the "dist" feature, which was required to make the new logic work before https://github.com/mozilla/sccache/pull/2392/commits/60b826bc554e309089dffefdcd2f7c7d91c71fde. 
(Edit: nope, that's wrong - different failure mode: these tests were for non-preprocessor cache mode.)

For distributed compilation with preprocessor cache mode, I will actually need to write the tests.

ahartmetz:
> For dependency files (-MD etc) with preprocessor cache mode, I have it easy: when preprocessor cache mode was disabled in case of -MD etc detected: [c0a0b6e](https://github.com/mozilla/sccache/commit/c0a0b6ebd8a0b03aa76cf658d3d44b282aa14f01), there were tests added which I kept, and they failed in builds without the "dist" feature, which was required to make the new logic work before [60b826b](https://github.com/mozilla/sccache/commit/60b826bc554e309089dffefdcd2f7c7d91c71fde). (Edit: nope, that's wrong - different failure mode: these tests were for non-preprocessor cache mode.)
> 
Ugh. I always have to reload a lot of context when getting back to this. In fact, there are existing tests covering that part, quoting from my updated commit message:
This is covered by the existing test test_sccache_command(preprocessor_cache_mode = true) -> run_sccache_command_tests() -> test_gcc_clang_depfile().


> For distributed compilation with preprocessor cache mode, I will actually need to write the tests.

And I did!

ahartmetz:
The ordering of commits follows a red-green pattern now: first break the tests by removing workarounds, then fix them again by fixing the bugs "properly".

ahartmetz:
The test failures are probably unrelated again. Other PRs changing different things have the same failures.

src/compiler/compiler.rs
```diff
@@ -498,15 +498,35 @@ where
...
+                    // In this mode, cache entries are exclusively distinguished by their preprocessed
+                    // source contents. But two files may differ in their names and / or the names of
+                    // included files while still producing the same preprocessed output, so they get the
+                    // same cache entry. That entry will have wrong (file names) dependency informaton in
+                    // the dependency file except for the compilation unit that originally produced it.
+                    // Since we did local preprocessing, that should already have produced the dependency
+                    // file - just leave that one alone and don't overwrite it from the cache.
+                    outputs
+                        .iter()
+                        .filter(|fobj_source| fobj_source.key != "d")
```
sylvestre: please add a comment at the end of the line to remind what is "d" here

src/compiler/compiler.rs
```diff
@@ -434,7 +434,7 @@ where
             .generate_hash_key(
                 &creator,
                 cwd.clone(),
-                env_vars,
+                env_vars.clone(),
```
sylvestre, line 437: the clone is now mandatory ?

src/compiler/compiler.rs
ahartmetz: done

src/compiler/compiler.rs
ahartmetz: Yes, it's somewhat explained in commit 920293468d1b49559bec1e5da3647e0ec0870188. Just now I tried some minimally invasive ways to to avoid the clone but didn't find a good way.

sylvestre:
thanks and sorry it took forever

PR #2395, opened 2025-05-14
chore: Remove not working mozilla code

Xuanwo:
The endpoint used by sccache is not working anymore. Let's just remove them.


sylvestre:
it was for office usage of sccache, i didn't know we still have it

PR #2403, opened 2025-06-08
fix: in stats, Compare values AND keys to have a fully deterministic order

iTrooz:
When using `watch sccache -s`, the order of "Successful distributed compiles" can change randomly if two keys have the same value
This PR fixes this

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1818,10 +1818,19 @@ impl ServerStats {
                 stat_width = stat_width + suffix_len
             ));
         }
+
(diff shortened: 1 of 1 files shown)


sylvestre:
Could you please add a quick test to make sure we don't regress in the future? thanks

iTrooz:
Done ! Note that the test is not foolproof, because of the [randomness in hashmap looping](https://doc.rust-lang.org/std/collections/struct.HashMap.html)

iTrooz:
@sylvestre Hello, any news ?

PR #2404, opened 2025-06-10
ci/grcov: build with --locked

sylvestre:


sylvestre:
doesn't fix the issue

PR #2405, opened 2025-06-14
feat: handle human size prefixes

iTrooz:
this PR allows you to specify human sizes in the configuration. For example:
```
[dist]
toolchain_cache_size = "5G"

[cache.disk]
size = "20G"
```

As a bonus, I also added for lowercase sizes, e.g. "2k"


ahartmetz:
Any open problems here?

PR #2408, opened 2025-06-17
fix: change the way to decide whether there's a `RUSTC_WORKSPACE_WRAPPER`

BugenZhao:
#1280 added support for locating the `rustc` binary when there's a workspace wrapper set via `RUSTC_WORKSPACE_WRAPPER`. A check on the existence of env var `CARGO` was included as a guardrail against false positives.

However, this doesn't seem to always work. In my reproduction at https://github.com/bugenzhao/sccache-wrapper-reproduce, it seems that the environment variable `CARGO` is not set when `cargo` attempts to detect the compiler as its initial step. As a result, we fail to decide the `maybe_rustc_executable` and eventually run into the testing of C compiler:

```
sccache: caused by: Compiler not supported: "/risingwave/ci/scripts/rustc-workspace-wrapper.sh: line 16: exec: -E: invalid option\nexec: usage: exec [-cl] [-a name] [command [argument ...]] [redirection ...]\n"
```

This PR attempts to fix this by directly checking whether `RUSTC_WORKSPACE_WRAPPER` is set.


sylvestre:
please add tests to make sure we don't regress in the future 
maybe in
https://github.com/mozilla/sccache/blob/main/tests/sccache_cargo.rs
or
https://github.com/mozilla/sccache/blob/main/tests/sccache_rustc.rs

PR #2409, opened 2025-06-17
Non hashable args like fdebug-prefix-map & related

borjamunozf:
Hello!

I realized that compilers arguments/flags for gcc like the mentioned above are being used in the hash_key function. This means that strategies for handling the absolute paths to relatives regarding debug info or in similar fashion that involves passing a _dynamic value_ into the argument depending of the builder origin will create a different hash key.

From my point of view, these kind of arguments should not affect the computer hash key at all:
- -_fdebug-prefix-map=${PWD}=.
- -fmacro-prefix-map
- ffile-prefix-map
- fdebug-compilation-dir

A naive change could just include a non-hashable list of args that should be skipped or not considered into the hash key.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -25,8 +25,8 @@ use crate::dist;
 use crate::dist::pkg;
 use crate::mock_command::CommandCreatorSync;
 use crate::util::{
-    Digest, HashToDigest, MetadataCtimeExt, TimeMacroFinder, Timestamp, decode_path, encode_path,
-    hash_all,
+    decode_path, encode_path, hash_all, Digest, HashToDigest, MetadataCtimeExt, TimeMacroFinder,
+    Timestamp,
 };
 use async_trait::async_trait;
 use fs_err as fs;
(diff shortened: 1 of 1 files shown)


necto:
> * -fmacro-prefix-map
> * ffile-prefix-map

Note that these arguments do affect preprocessor output, so they should still be passed to the preprocessor. Using Clang compiler, I had success with the following patch:
```diff
modified   src/compiler/clang.rs
@@ -220,6 +220,7 @@ counted_array!(pub static ARGS: [ArgInfo<gcc::ArgData>; _] = [
     flag!("-fcxx-modules", TooHardFlag),
     take_arg!("-fdebug-compilation-dir", OsString, Separated, PassThrough),
     take_arg!("-fembed-offload-object", PathBuf, Concatenated(b'='), ExtraHashFile),
+    take_arg!("-ffile-prefix-map", OsString, Concatenated(b'='), PreprocessorArgument),
     flag!("-fmodules", TooHardFlag),
     flag!("-fno-color-diagnostics", NoDiagnosticsColorFlag),
     flag!("-fno-pch-timestamp", PassThroughFlag),
```

PR #2411, opened 2025-06-27
feat: Add timestamp to .sccache_check name (#2132)

magreenblatt:
Use a unique name for the `.sccache_check` object so that we don't need to grant overwrite permissions on the bucket.


magreenblatt:
@Xuanwo  can you review? Thanks!

sylvestre:
could you please add tests to make sure we don't regress in the future? thanks

Issue #2415, opened 2025-06-30
Improve read/write handling on "Invalid Credentials" error


magreenblatt:
Access tokens might be invalidated while a build is running, resulting in "Invalid Credentials" errors for read and write operations. When a read operation fails with "Invalid Credentials" sccache still attempts to write that same object after local compilation. Additionally, sccache reports these failures as "Cache write errors" instead of as "Cache read errors".

I'd therefore like to propose the following improvements to sccache behavior in the case of "Invalid Credentials" read errors:

1. Attempt to refresh the access token (if a refresh attempt hasn't already been made)
2. If read still fails with "Invalid Credentials" the report the failure as "Cache read errors"
3. Do not attempt to write the same object after read fails with "Invalid Credentials"

Example debug logging for the read/write failure:

```
[0m[38;5;8m[[0m2025-06-28T17:36:18Z [0m[1m[31mERROR[0m opendal::services[0m[38;5;8m][0m service=gcs name=sccache-bucket path=7/9/9/799ca2d5ea0c4120cdb00e301153ff6b619a34c95d097fad836c15dd428b8acd: read failed Unexpected (permanent) at read => Invalid Credentials
    
    Context:
       uri: https://storage.googleapis.com/storage/v1/b/sccache-bucket/o/7%2F9%2F9%2F799ca2d5ea0c4120cdb00e301153ff6b619a34c95d097fad836c15dd428b8acd?alt=media
       response: Parts { status: 401, version: HTTP/1.1, headers: {"content-type": "text/html; charset=UTF-8", "x-guploader-uploadid": "...", "date": "Sat, 28 Jun 2025 17:36:18 GMT", "vary": "Origin", "vary": "X-Origin", "server": "UploadServer", "expires": "Sat, 28 Jun 2025 17:36:18 GMT", "cache-control": "private, max-age=0", "content-length": "19"} }
       service: gcs
       path: 7/9/9/799ca2d5ea0c4120cdb00e301153ff6b619a34c95d097fad836c15dd428b8acd
       range: 0-

[...]

PR #2416, opened 2025-06-30
feat: Parse GCS token expire time (#2415)

magreenblatt:


magreenblatt:
@Xuanwo Can you review? Thanks!

sylvestre:
could you please update the doc to document this? thanks

sylvestre:
could you please add tests to make sure we don't regress in the future? thanks

sylvestre:

PR #2419, opened 2025-07-02
make sccache -s always return current cache size

iTrooz:
sccache server disk cache is usually initiated by the first compilation request. If `sccache -s` is issued before then, sccache will not return the currently occupied cache size. This PR fixes that by querying the underlying LruDiskCache in `current_size()`. I don't think they are performance implications of this, but I am not sure.

I also renamed LazyDiskCache to LazyLruDiskCache, because I think the name is more clear. Let me know if you want me to revert that change

### DIFF
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -173,7 +173,7 @@ impl Storage for DiskCache {
     }
 
     async fn current_size(&self) -> Result<Option<u64>> {
-        Ok(self.lru.lock().unwrap().get().map(|l| l.size()))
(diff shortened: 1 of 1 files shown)


iTrooz:
Hello, do you need me to do something on this PR ?

iTrooz:
@sylvestre Hello ! I keep seeing pushes made to this PR. Do you need me to do something on it, or are you just testing it ?

sylvestre:
I was just running the benchmarks again

iTrooz:
Ok ! Lmk if anything prevents this from being merged

Issue #993, opened 2021-04-30
Symbolic link doesn't seem to work on Linux


lackhoa:
Using `sccache-v0.2.15-x86_64-unknown-linux-musl`, after symlinking:
```
$ ln -s sccache /usr/local/bin/gcc
```

I get this behavior
```
$ gcc
sccache: No command specified
sccache 0.2.15

USAGE:
    sccache [FLAGS] [OPTIONS] [cmd]...

FLAGS:
        --dist-auth       authenticate for distributed compilation
        --dist-status     show status of the distributed client
    -h, --help            Prints help information
    -s, --show-stats      show cache statistics
        --start-server    start background server
        --stop-server     stop background server
    -V, --version         Prints version information
    -z, --zero-stats      zero statistics counters

OPTIONS:
        --package-toolchain <executable> <out>    package toolchain for distributed compilation
        --stats-format <stats-format>
            set output format of statistics [default: text]  [possible values: text, json]


ARGS:
    <cmd>...    

Enabled features:
    S3:        true
    Redis:     true
    Memcached: true
    GCS:       true
    Azure:     true
```
sccache still thinks it is being called as `sccache`, and not as `gcc`.

The "wrap-around" behavior works fine before linking
```
$ sccache gcc
gcc: fatal error: no input files
compilation terminated.
```

But on my Mac, the same sccache `v0.2.15` works fine with gcc being symlinked to sccache:
```
$ gcc
clang: error: no input files
```

lackhoa:
Oh, I see that gcc was a symlink as well
```
$ ls -l /usr/bin/gcc
lrwxrwxrwx 1 root root 21 Dec 18 13:46 /usr/bin/gcc -> /etc/alternatives/gcc
```

lackhoa:
If someone wants to reproduce this, I can offer this dockerfile
```
FROM ubuntu:bionic

# GCC 8
RUN apt-get install -y gcc-8 g++-8

# Use GCC 8 by default (https://askubuntu.com/questions/1028601/)
RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 \
                        --slave   /usr/bin/g++ g++ /usr/bin/g++-8

# sccache
RUN export SCCACHE_VER=sccache-v0.2.15-x86_64-unknown-linux-musl && \
    wget https://github.com/mozilla/sccache/releases/download/v0.2.15/${SCCACHE_VER}.tar.gz && \
[...]

PR #2420, opened 2025-07-05
Fix symbolic links to sccache on linux

sertonix:
Whether or not `env::current_exe` follows symlinks is platform dependant. On linux it follows them which made it impossible to do things like `ln -s sccache /usr/bin/cc`. By reading `env::args()` directly we don't have this issue.

Fixes https://github.com/mozilla/sccache/issues/993
Based on https://github.com/rivy/rs.coreutils/commit/ad3852a72d2235ae501a40b4e1f36d406cc53b1b

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -197,7 +197,10 @@ pub fn try_parse() -> Result<Command> {
     let mut args: Vec<_> = env::args_os().collect();
 
     if !internal_start_server {
-        if let Ok(exe) = env::current_exe() {
+        if let Ok(exe) = match env::args().next() {
+            Some(s) if !s.is_empty() => Ok(PathBuf::from(s)),
+            _ => env::current_exe(),
+        } {
             match exe
                 .file_stem()
                 .and_then(|s| s.to_str())
@@ -205,7 +208,7 @@ pub fn try_parse() -> Result<Command> {
             {
                 // If the executable has its standard name, do nothing.
                 Some(ref e) if e == env!("CARGO_PKG_NAME") => {}
-                // Otherwise, if it was copied/hardlinked under a different $name, act
+                // Otherwise, if it was copied/linked under a different $name, act
                 // as if it were invoked with `sccache $name`, but avoid $name resolving
                 // to ourselves again if it's in the PATH.
                 _ => {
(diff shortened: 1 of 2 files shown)


sertonix:
@sylvestre Could you take a look at this?

sylvestre:
@sertonix could you please add a test to make sure we don't regress? thanks

src/cmdline.rs
```diff
@@ -197,15 +197,18 @@ pub fn try_parse() -> Result<Command> {
     let mut args: Vec<_> = env::args_os().collect();
 
     if !internal_start_server {
-        if let Ok(exe) = env::current_exe() {
+        if let Ok(exe) = match env::args().next() {
```
ahartmetz, line 200: I think it would make sense to mention the platform-dependent symlink following issue here in a comment instead of only in the commit message. This code looks a bit mysterious otherwise. Much more convenient not to have to use git blame or so to find that important info.

I hope this gets merged btw, this issue is a strange papercut that gives a bad/unfinished first impression.

sertonix:
> could you please add a test to make sure we don't regress?

I would like to but I am uncertain how to do this.

ahartmetz:
> > could you please add a test to make sure we don't regress?
> 
> I would like to but I am uncertain how to do this.

I've had the same problem, but it's often not as hard as it looks.
- What is not supposed to happen?
- How do you make it happen? (Probably with Rust filesystem API to create the symlink and existing sccache test infrastructure to create a temp dir, run a test compile... have a look around)
- How do you check if it happened? (Probably success or failure of compiling a file)
- Run tests with and without the fix to (more or less) verify that they both work

ahartmetz:
@sertonix If you're motivated to finish it, you could come to the sccache Matrix channel and ask for help. I'm sometimes on there as well. https://chat.mozilla.org/#/room/#sccache:mozilla.org (or translate it to the usual Matrix "coordinates" in your client if you have one)

ahartmetz:
@sertonix If you don't mind, I can have a whack at writing a test for this

sertonix:
It might take a while before I get to writing tests so feel free to give it a try. I pushed a small fix that I made while using this patch

ahartmetz:
I think you can close this one.

sertonix:
Right

PR #2421, opened 2025-07-07
Disable reqwest default features to not depend on openssl

### LINKED ISSUE #841: sccache always links to OpenSSL
When building the current master version of sccache with `cargo build --release` or even `cargo build --release --features=dist-client` the resulting sccache binary always links against OpenSSL. According to the Readme this should not be the case, only when using GCS it should be required.

I can reproduce this with the following steps:
* Use this Docker image: `rust@sha256:e158d68ef891c1f41edd6f1bf2adafdcdb360ca98302865d6d35a8ed44236bdd`. This is the official Rust image for 32bit environment
* `docker run -it rust@sha256:e158d68ef891c1f41edd6f1bf2adafdcdb360ca98302865d6d35a8ed44236bdd /bin/bash`
* `git clone https://github.com/mozilla/sccache.git`
* `cd sccache`
* `cargo build --release`

However I can't see the same issue when using the 64bit Rust image (`docker run -it rust /bin/bash`). Though I'm not sure if this has to do with something else.

tylerhawkes:
Fixes #841

I also did a cargo update to get the latest dependencies. All the tests were able to build, and I did get a failure locally, but probably due to my setup.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aes"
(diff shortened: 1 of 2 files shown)


rtyler (approved):

sylvestre:

tylerhawkes:
> sorry but it needs to be rebased

@sylvestre done

sylvestre:
fails with:
```

  Downloaded base64ct v1.8.0
error: failed to download replaced source registry `crates-io`

Caused by:
  failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/base64ct-1.8.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.75.0 (1d8b05cdd 2023-11-20)).
```

tylerhawkes:
@sylvestre I think I have everything downgraded that is incompatible with 1.75

tylerhawkes:
Turns out updating with an older version requirement is just too hard, especially when I can't test out the windows stuff, so I'm downgrading this to just the one change to remove the default dependencies for reqwest.

tylerhawkes:
@sylvestre I just updated this again. Maybe the tests will pass this time after all the other crate updates.

valkum:
Cargo install currently fails due to github.com/rust-native-tls/rust-native-tls/issues/367. It might be nice to get this merged to avoid such issues in the future (allowing users to switch between native-tls and rustls as they like).

PR #2424, opened 2025-07-24
Ignore `--diagnostic-width` argument when computing hash

### LINKED ISSUE #2418: [feat] Ignore `--diagnostic-width` when computing hash
`--diagnostic-width=` depends on the terminal settings, and it shouldn't affect compilation. However, `sccache` includes this flag in hash computation. It means these two commands don't share cache:

```bash
# 1. include `--diagnostic-width=` for TTY terminals
cargo build

# 2. exclude `--diagnostic-width=`. No cache hits!
cargo build > _.log 2>&1
```

Could we have an `sccache` option to remove the `--diagnostic-width=` parameter from hash computation?

shadaj:
Fixes #2418

We ran into this in https://github.com/hydro-project/hydro, where we programmatically invoke Cargo to compile generated Rust sources, so without a TTY the hash of arguments would differ causing cache misses. This ignores that argument, and I validated on Hydro that it enables cache hits in the non-TTY builds.


shadaj:
@sylvestre I think you'll need to approve the GitHub Actions workflow; the linting errors after the original merge should be fixed!

sylvestre:
done

shadaj:
Looks like the failures are from `main` 😢

shadaj:
Friendly bump @sylvestre! The changes should be good to go.

sylvestre:
Could you please add tests to verify that we aren't regressing in the future? thanks

sylvestre:
Very sorry it took that long to get to this

shadaj:
@sylvestre added the test! And please don't apologize, really appreciate all the work you do as a maintainer :)

sylvestre:
thanks

PR #2430, opened 2025-08-05
Fix mtime for reproducable toolchains

nigredon1991:
Hello.
We are running sccache-dist servers in docker and mtime in toolchains is changing, because of this new toolchains appear on the servers.
This request fixes this problem.

### DIFF
--- a/tests/harness/mod.rs
+++ b/tests/harness/mod.rs
@@ -3,8 +3,8 @@ use fs_err as fs;
 use sccache::config::HTTPUrl;
 use sccache::dist::{self, SchedulerStatusResult, ServerId};
 use sccache::server::ServerInfo;
-use std::fs::remove_dir_all;
 use std::env;
+use std::fs::remove_dir_all;
 use std::io::Write;
 use std::net::{self, IpAddr, SocketAddr};
 use std::path::{Path, PathBuf};
@@ -495,10 +495,7 @@ impl DistSystem {
                 check_output(&output);
                 println!("{output:?}");
                 let str_output = String::from_utf8_lossy(&output.stdout);
-                return str_output
-                    .lines()
-                    .filter(|&part| !part.is_empty())
-                    .count();
+                return str_output.lines().filter(|&part| !part.is_empty()).count();
             }
             ServerHandle::Process { pid: _, url: _ } => {
                 panic!("restart not yet implemented for pids")


sylvestre:
could you please add a test to make sure we don't regress ?

nigredon1991:
Yes, of course, I will.

nigredon1991:
Added a test, changed the solution to a more correct one during the process

tests/harness/mod.rs
```diff
@@ -72,6 +73,19 @@ pub fn stop_local_daemon() -> bool {
         .is_ok_and(|status| status.success())
 }
 
+pub fn clear_cache_local_daemon(tmpdir: &Path) -> bool {
+    trace!("clear local cache daemon");
+    let client_build_dir = tmpdir.join(DIST_CACHE_RELPATH);
+    assert!(client_build_dir.exists());
+    Command::new("rm")
```
sylvestre: Using 'rm -rf' command directly poses security risks - please use std::fs::remove_dir_all

tests/dist.rs
```diff
@@ -114,7 +114,7 @@ fn test_dist_basic() {
 
     let mut system = harness::DistSystem::new(&sccache_dist, tmpdir);
     system.add_scheduler();
-    system.add_server();
+    let server_handle = system.add_server();
```
sylvestre: seems that you don't use it ?

tests/harness/mod.rs
```diff
@@ -469,6 +483,33 @@ impl DistSystem {
...
+                        "-1",
+                        &format!("{BUILD_DIR_CONTAINER_PATH}/toolchains"),
+                    ])
+                    .output()
+                    .unwrap();
+                check_output(&output);
+                println!("{output:?}");
+                let str_output = String::from_utf8_lossy(&output.stdout);
+                return str_output
+                    .split("\n")
```
sylvestre: please use lines()

tests/harness/mod.rs
nigredon1991: Fixed

tests/dist.rs
nigredon1991: Fixed

tests/harness/mod.rs
nigredon1991: Fixed

nigredon1991:
Thanks for the edits. I fixed everything.

tests/harness/mod.rs
```diff
@@ -76,9 +76,14 @@ pub fn stop_local_daemon() -> bool {
 pub fn clear_cache_local_daemon(tmpdir: &Path) -> bool {
     trace!("clear local cache daemon");
     let client_build_dir = tmpdir.join(DIST_CACHE_RELPATH);
-    assert_eq!(true , client_build_dir.exists());
+    assert!(client_build_dir.exists());
     Command::new("rm")
```
sylvestre: can you use the Rust api to do it instead of a command ?

sylvestre: ignore this comment, i was testing something :)

PR #2443, opened 2025-09-17
Downgrade CUDA to 12.5 for Windows 2022 CI to fix integration

ahartmetz:
Based on the following warning and then error, which look related: clang++: warning: CUDA version is newer than the latest partially supported version 12.5 [-Wunknown-cuda-version]
corecrt_wstdio.h:486:24: error: non-const lvalue reference to type '__builtin_va_list' cannot bind to a value of unrelated type 'va_list' (aka 'char *')

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,21 +118,21 @@ jobs:
             rustc: beta
             extra_desc: cuda11.8
           - os: windows-2022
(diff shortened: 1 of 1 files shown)


ahartmetz:
Doesn't work. For VS2022, there exists no CUDA version that works with both the MSVC version and the Clang version included with it.

PR #2444, opened 2025-09-17
Don't run tests using Clang as CUDA compiler on Windows

ahartmetz:
Visual Studio 2022 actually ships Clang these days. The included Clang v19 is too old to work with CUDA versions new enough to work with the MVSC version 14 also included with VS2022.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,16 +122,19 @@ jobs:
             # Oldest supported version, keep in sync with README.md
             rustc: "1.75.0"
             extra_args: --no-fail-fast
+            notest_cuda_compilers: clang++
             extra_desc: cuda12.8
           - os: windows-2022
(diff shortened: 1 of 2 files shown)


ahartmetz:
Another option would be to disable the CUDA / Clang test if an environment variable is set, and set it in appropriate CI configurations. It might work again with the next Visual Studio release. Or install current Clang in the CI, obviously, but that seems like a chore and would also increase CI runtime. At least (I didn't know and it was kind of hard to find) LLVM does seem to provide binary downloads for Windows, e.g.: https://github.com/llvm/llvm-project/releases/tag/llvmorg-21.1.0

sylvestre:
@trxcllnt ok with you ? :)

ahartmetz:
That string list filtering code that I've written seems rather verbose btw, any tips for doing better?

PR #2448, opened 2025-09-22
Remove or replace "Windows 2019" CI config

ahartmetz:
Remove where there is a similar config with Windows 2022, raise the version where there isn't one.


sylvestre:
excellent, thanks

PR #2452, opened 2025-10-02
Write stats to cache_stats_file after compile

zmc:
Fixes: #33 

This would be my first contribution to this project, and I am relatively inexperienced with Rust, so I'd love feedback on this.


sylvestre:
did you evaluate the perf impact of this? thanks

ahartmetz:
What's the use case for this? It seems pretty exotic. I don't really mind since it's opt-in, anyway...

sylvestre:
rustfmt is needed btw

PR #2460, opened 2025-10-13
fix: use single-threaded tokio runtime in sccache client

venkyt-arista:
The sccache client was using Runtime::new() which creates a multi-threaded tokio runtime with worker threads equal to the number of CPU cores. On high-core-count servers running many parallel builds, this created an excessive number of threads.

For example:
- 96 vCPU server
- 10 concurrent make invocations
- Each make using -j16
- Result: 96 × 10 × 16 = 15,360 threads created by sccache wrappers

While these threads are short-lived, with continuous build pipelines this constant thread creation/destruction causes unnecessary overhead. The sccache client only performs simple I/O operations (connecting to server, sending requests, receiving responses) and doesn't need worker threads.

This change replaces all client-side Runtime::new() calls with Builder::new_current_thread().enable_all().build() to use a single-threaded runtime, reducing thread count from num_cpus to 1 per client invocation.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -92,7 +92,9 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     trace!("run_server_process");
     let tempdir = tempfile::Builder::new().prefix("sccache").tempdir()?;
     let socket_path = tempdir.path().join("sock");
-    let runtime = Runtime::new()?;
+    let runtime = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
     let exe_path = env::current_exe()?;
     let workdir = exe_path.parent().expect("executable path has no parent?!");
 
@@ -190,7 +192,9 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     trace!("run_server_process");
 
     // Create a mini event loop and register our named pipe server
-    let runtime = Runtime::new()?;
+    let runtime = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
     let pipe_name = &format!(r"\\.\pipe\{}", Uuid::new_v4().as_simple());
 
     // Spawn a server which should come back and connect to us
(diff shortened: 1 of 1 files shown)


sylvestre:
could you please add a test to make sure we don't regress in the future on this? thanks

venkyt-arista:
> could you please add a test to make sure we don't regress in the future on this? thanks

Will do. Doing a few more tests manually right now. Will add a test before taking this PR out of draft state.
Thanks!

sylvestre:
ping ?

venkyt-arista:
Sorry, got caught up with a bunch of things. This is still on my radar. Hope to get to this in the next couple of weeks.

ahartmetz:
See also: https://github.com/mozilla/sccache/pull/2474
Can you coordinate with the author of that one?

zsrkmyn:
Oh, I didn't notice this PR before. It's so interesting that the issue exists for years but we both sent the PR within the same month LOL. I'm fine to drop my PR if this is a more complete fix for #2473 .

src/commands.rs
```diff
@@ -92,7 +92,9 @@ fn run_server_process(startup_timeout: Option<Duration>) -> Result<ServerStartup
     trace!("run_server_process");
     let tempdir = tempfile::Builder::new().prefix("sccache").tempdir()?;
     let socket_path = tempdir.path().join("sock");
-    let runtime = Runtime::new()?;
+    let runtime = tokio::runtime::Builder::new_current_thread()
```
AJIOB, line 95: nit: should we allow running the server on more than one core? Only one server process should be used for all client connections.

P.S. Maybe the server concurrency should be limited to 4 or 8 threads and not executed on the all cores. And we should be ready for less than 4-core systems too (just use min of constant and hardware concurrency)

AJIOB:
Hi @venkyt-arista,

Do you plan to finalize this PR?

sylvestre:
still draft, please reopen when it is ready

src/commands.rs
AJIOB: The server Tokio runtime starts in https://github.com/mozilla/sccache/blob/c0db872c953f20fde2d45ab68ddc74d5100febbc/src/server.rs#L443 , and it's still multi-threaded

PR #2464, opened 2025-10-16
feat: Add axum impl for dist server

### LINKED ISSUE #2234: Replace `rouille` with `axum`
Since we are heavily using `tokio` and `reqwest`, it would be more sensible to use an asynchronous web framework rather than a blocking one.

Migrating to `axum` will have the following benefits too:

- It's easier to maintain the API than the macro.
- Remove some old future-incompatibile deps:

```shell
warning: the following packages contain code that will be rejected by a future version of Rust: buf_redux v0.8.4, multipart v0.18.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

Xuanwo:
Close https://github.com/mozilla/sccache/issues/2234

This PR adds an Axum implementation for the dist server. We introduced this as a new feature called `dist-server-axum` with no functional changes to the existing server. This allows our community to test the new server as an experimental option.

All behavior remains identical, and all configurations are unchanged. You can run this server exactly as before by simply replacing the binary.

We welcome your feedback and trials. This is just the first step toward modernizing sccache, with the goal of making it a fully async, modular, and easier-to-maintain project.

---

**This PR was primarily authored with Amp using Claude-Sonnet-4.5 and then hand-reviewed by me. I AM responsible for every change made in this PR. I aimed to keep it aligned with our goals, though I may have missed minor issues. Please flag anything that feels off, I'll fix it quickly.**


sylvestre:
did you evaluate any of the wins here ? thanks

Xuanwo:
> did you evaluate any of the wins here ? thanks

The ultimate goal is to remove the dying `rouille` and turn sccache into a true async server.

After the whole refactor been done, we can expected:

- better performance (async io / async http server)
- easier maintenance (use up-to-date crates and remove old ones)
- prevent CVEs (avoid depending on unmaintained crate)

We haven't discussed the full plan yet, but I expect the axum impl will become the default after several versions and finally we'll remove the rouille impl.

sylvestre:
ok, thanks
did you run some benchmarks ?

Xuanwo:
> did you run some benchmarks ?

Sorry, I didn’t. I’m working on macOS now and don’t have access to a Linux server. I wish we could get some feedback from the community.

sylvestre:
it needs to be rebased

ahartmetz:
By the way, there's a similar PR, but with Warp instead of Axum. This isn't an argument for one or the other, but it's maybe good to know and make a conscious decision. (AFAICS, both would be fine: they both work and sccache's needs are relatively simple)
https://github.com/mozilla/sccache/pull/2241

trxcllnt:
Hi @Xuanwo, I've been working on something similar here at NVIDIA. Would you be open to collaborating? The work is in our fork [here](https://github.com/rapidsai/sccache), and I'm available to discuss over email, slack, video call, etc. I also have [diagrams](https://gist.github.com/trxcllnt/bae20d8196f2a2d65e8f625e1d9d7681), an example [usage script](https://gist.github.com/trxcllnt/eaab5d814dd3069ec2103a7cccabf5d1#file-sccache-setup-sh-L138-L139), and I can share slides.

We do up to 1M-2M compiles/hour, 1500-2000 compile requests/second, 25GiB/minute. We had to re-architect a few things in `sccache-dist` to support this scale (and higher), but considering the numbers we're seeing, I think it's worth it. We have seen uncached build times drop 50-90% in most cases, and are already seeing significant savings in CI costs.

Ultimately the goal is to make an `sccache-dist` service available company-wide, so all our teams can build as if they had an infinite number of cores, i.e. `make -j $(ulimit -Hn)` :fire:.

We're starting to think about how to upstream everything, including other various security, performance, resiliency, scaling, and correctness fixes.

Massively parallel distributed CUDA builds are a long-standing ask from our customers. We're huge sccache fans and would love to contribute back.

@sylvestre We would also love to discuss with you if you're free.

sylvestre:
@trxcllnt with pleasure - let's coordinate by email s@mozilla.com

PR #2469, opened 2025-10-23
Share cache entries even if `-Cprefer-dynamic` and `-Cextra-filename` differ

shadaj:
Makes two changes that improve the cache hit rate if switching between compilation with `-Cprefer-dynamic` and not. First, because sccache only caches `rlib` / `staticlib` crates, `-Cprefer-dynamic` has no effect and can be ignored (it is often set via global `RUSTFLAGS`).

Similarly, the Cargo-generated fingerprint is easily invalidated by custom `RUSTFLAGS`, but since sccache has fine-grained knowledge of the `rustc` invocation it should ignore this fingerprint. We do this by ignoring this flag in the cache key, and also stripping the `-Cextra-filename` (which Cargo uses to append the fingerprint to outputs) from the keys of the output files.

Most of this behavior is (to my understanding) covered by the existing tests, but I adjusted a hashing test to ensure these two flags are correctly ignored.

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -158,6 +158,8 @@ pub struct ParsedArguments {
     crate_name: String,
     /// The crate types that will be generated
     crate_types: CrateTypes,
+    /// The hash pased to -Cextra-filename
+    extra_filename: Option<String>,
     /// If dependency info is being emitted, the name of the dep info file.
     dep_info: Option<PathBuf>,
     /// If profile info is being emitted, the path of the profile.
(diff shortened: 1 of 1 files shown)


src/compiler/rust.rs
```diff
@@ -1456,15 +1459,25 @@ where
...
                     !(arg == "--extern"
                         || arg == "-L"
                         || arg == "--out-dir"
-                        || arg == "--diagnostic-width")
+                        || arg == "--diagnostic-width"
+                        || (arg == "-C" && value == &Some(OsString::from("prefer-dynamic")))
+                        || (arg == "-C"
+                            && value
+                                .as_ref()
+                                .is_some_and(|v| v.starts_with("extra-filename"))))
```
MingweiSamuel, line 1476: Maybe
```rust
(arg == "C" && (value == &Some(..) || value.as_ref().is_some_and(..))
```
is more readable?

MingweiSamuel (approved):

PR #2474, opened 2025-10-27
Limit client thread to 1 for Command::Compile

### LINKED ISSUE #2473: sccache makes too many `clone` syscall
I have a 112-core Intel CPU, and each time a `sccache` is invoked, it makes 112 `clone` syscall; and when I try to parallel build with `-j112`, there are 112*112=12544 clones in total, which may cause `EAGAIN` error (especially in a container environment).

Reproducer:

```
$ nproc
112
$ sccache --start-server
sccache: Listenning on address ...
$ strace -e clone sccache gcc -c a.c |& wc -l
114
```

Version:
```
$ sccache --version
sccache 0.12.0
```

zsrkmyn:
This tries to fix #2473.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -42,6 +42,6 @@ jobs:
         env:
           CODSPEED_LOG: debug
         with:
-          mode: instrumentation
+          mode: simulation
           run: cargo codspeed run > /dev/null
           token: ${{ secrets.CODSPEED_TOKEN }}
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -556,7 +556,7 @@ jobs:
       - name: Chmod for binary
         run: chmod +x ${SCCACHE_PATH}
 
-      - name: Test
+      - name: Test C/C++
         run: |
           export CXX="${SCCACHE_PATH} clang++"
           $CXX -c `pwd`/tests/test_clang_multicall.c
@@ -575,6 +575,16 @@ jobs:
           ${SCCACHE_PATH} --show-stats
 
           ${SCCACHE_PATH} --show-stats | grep -e "Cache hits\s*[1-9]"
+
+      - name: Test ASM
+        run: |
+          export ASM="${SCCACHE_PATH} clang"
+          $ASM -c `pwd`/tests/test_intel_asm.s
+
+      - name: Test ASM with preprocessor
+        run: |
+          export ASM="${SCCACHE_PATH} clang"
+          $ASM -c `pwd`/tests/test_intel_asm_to_preproc.S
 
   hip:
     # Probably wouldn't matter anyway since we run in a container, but staying
@@ -699,7 +709,7 @@ jobs:
       - name: Chmod for binary
         run: chmod +x ${SCCACHE_PATH}
 
-      - name: Test
+      - name: Test C/C++
         run: |
           export CXX="${SCCACHE_PATH} g++"
           $CXX -c `pwd`/tests/test_clang_multicall.c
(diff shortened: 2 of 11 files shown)


zsrkmyn:
Ooops. I just found that rustc supports parallel compilation natively. As I'm a C/C++ user, I'm not sure if the fix is correct for rust compilation.

ahartmetz:
> Ooops. I just found that rustc supports parallel compilation natively. As I'm a C/C++ user, I'm not sure if the fix is correct for rust compilation.

I don't see how this change would make things worse. This is a change in the async runtime, which I don't think in any way propagates to the Rust compiler, a presumably exec'd child process. The only thing that might be different for Rust is that you want a lower amount of compile jobs so that the number of compiler threads doesn't greatly exceed the number of cores (that isn't currently done AFAIK). Apart from that, for Rust compilation in sccache, the question seems to be the same as for C++: what is the ideal thread count to avoid slowing down in sccache while also not wasting CPU time and memory on unused thread construction, stacks etc? Given that sccache doesn't use a lot of CPU itself, the answer might well be one, or maybe two or so.

zsrkmyn:
@ahartmetz Thanks for the reply! Then I believe the fix is correct. IIUC, the newly created threads are used to run the compilation command only (see line 812), and sccache itself runs on the main thread; so creating 1 new thread for is enough, right?

ahartmetz:
> @ahartmetz Thanks for the reply! Then I believe the fix is correct. IIUC, the newly created threads are used to run the compilation command only (see line 812), and sccache itself runs on the main thread; so creating 1 new thread for is enough, right?

I'm only reasonably sure about the (non-)difference between C/C++ and Rust regarding this patch. I'll leave commenting about more general issues to others who know the codebase (and Rust and Rust async...) better than me.

ahartmetz:
By the way, I think you might misunderstand something: The compiler is always run in a subprocess, which can spawn as many threads as it wants, but will nevertheless only have one communication channel to its parent process sccache, so one thread or less, if I/O multiplexing is used (which I think it might be via async), required to deal with it.

A separate issue is that one *might* want to spawn fewer multi-threaded compilers to avoid oversubscribing cores too much, but AFAIK even the Rust compiler often uses only a single thread, so this is a difficult trade-off to make, and just running as many compiler processes as there are hardware threads (as it's currently done) seems like a reasonable choice.

zsrkmyn:
It's sound to spawn as many compiler processes as the number of hw threads. The issue here is that we start 1 sccache server, and 112 (the thread number) sccache clients, and each **client** again starts 112 threads but only 1 is used to communicate with the compiler instantance (as you mentioned). 

The patch here only fixes the thread number clients start, but not those in server side.

ahartmetz:
> I'm only reasonably sure about the (non-)difference between C/C++ and Rust regarding this patch. I'll leave commenting about more general issues to others who know the codebase (and Rust and Rust async...) better than me.

Yeah, I'm not so sure anymore! Seems like rustc participates in the jobserver protocol in order to share available cores sensibly. Not sure what that means for this patch.
https://doc.rust-lang.org/rustc/jobserver.html

b-lerohellec:
I think you can tune the configuration of the [runtime](https://github.com/mozilla/sccache/blob/31b993da874c8bd656d461d8865d4612bfb13567/src/commands.rs#L810) used for local compiling.
A `new_current_thread` with `enable_io()` must be enough to simply run a `block_on` call.

```rust
let jobserver = Client::new_num(1);
let conn = connect_or_start_server(&get_addr(), startup_timeout)?;
// Used only for a unique block_on(…) in case compilation has to be performed locally
let mut runtime = tokio::runtime::Builder::new_current_thread().enable_io().build()?;
let res = do_compile(…)
```

zsrkmyn:
> I think you can tune the configuration of the [runtime](https://github.com/mozilla/sccache/blob/31b993da874c8bd656d461d8865d4612bfb13567/src/commands.rs#L810) used for local compiling. A `new_current_thread` with `enable_io()` must be enough to simply run a `block_on` call.
> 
> ```rust
> let jobserver = Client::new_num(1);
> let conn = connect_or_start_server(&get_addr(), startup_timeout)?;
> // Used only for a unique block_on(…) in case compilation has to be performed locally
> let mut runtime = tokio::runtime::Builder::new_current_thread().enable_io().build()?;
> let res = do_compile(…)
> ```

I believe #2460 already did that :-)

zsrkmyn:
Close this as #2460 may be a more appropriate fix.

PR #2479, opened 2025-11-05
Use generator in `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT` in README

Andrej730:
As it is probably the recommended way to do integrate cache in cmake, without introducing unexpected debug symbols to Release / MinSizeRel builds. 

Also consistent with the usage described in the official docs - https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html


Andrej730:
@sylvestre can you please take a look at that one too? 😅

Issue #1844, opened 2023-07-25
MSVC /Fo parameter not handled correctly


mpoeter:
According to the [MSVC documentation](https://learn.microsoft.com/en-us/cpp/build/reference/fo-object-file-name?view=msvc-170) `/Fo` "Specifies an object (.obj) file name or directory to be used instead of the default."

And further:
> To use the /Fo option to set an output directory for all object files created by the CL command, specify the directory as the pathname argument. A directory is indicated by a trailing slash or backslash in the pathname argument. Use an escaped backslash (a double backslash), if you're using a quoted path. The directory path can be absolute, or relative to the source directory. The specified directory must exist, or the compiler reports error D8003. The directory isn't created automatically.

At the moment sccache always treats `/Fo` as a filename. Unfortunately by default Visual Studio sets this to `$(IntDir)` ("Path to the directory specified for intermediate files."). So as a result instead of a `main.obj` I get a `Release.obj`.

```
MSBuild.exe .\CppTest.sln /t:Build /p:Configuration=Release "/verbosity:normal" "/p:CLToolExe=cl.exe" "/p:CLToolPath=C:\dev\sccache"
MSBuild version 17.3.1+2badb37d1 for .NET Framework
Building the projects in this solution one at a time. To enable parallel build, please add the "-m" switch.
Build started 7/25/2023 2:04:36 PM.
Project "C:\dev\CppTest\CppTest.sln" on node 1 (Build target(s)).
ValidateSolutionConfiguration:
  Building solution configuration "Release|x64".
Project "C:\dev\CppTest\CppTest.sln" (1) is building "C:\dev\CppTest\CppTest.vcxproj" (2) on node 1 (default targets).
PrepareForBuild:
  Creating directory "x64\Release\".
  Creating directory "x64\Release\CppTest.tlog\".
InitializeBuildStatus:
  Creating "x64\Release\CppTest.tlog\unsuccessfulbuild" because "AlwaysCreate" was specified.
ClCompile:
  C:\dev\sccache\cl.exe /c /Z7 /nologo /W3 /WX- /diagnostics:column /sdl /O2 /Oi /GL /D NDEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /MD /GS /Gy /fp:precise /Zc:wchar_t / 
[...]

PR #2482, opened 2025-11-06
msvc: handle '/FoRelease\' command-line argument

### LINKED ISSUE #1687: sccache places object file at different location compared to cl.exe and causes link failure when @rsp file is used

jimis:
MSBuild may provide as output file (/Fo) plus a folder ending with a trailing backslash, for example /FoRelease\.

msvc detects this case and chooses {folder}/{input_file_name}.obj as output file. For example, for /cfoo.c /FoRelease\, the expected output path (by linker and msbuild) is Release/foo.obj.

Fixes #1687 #1844.

Edit: This pull request is basically #1836 rebased and changed minimally to pass all tests. The changes are:
  * Enable the test only on Windows
  * Fix style

The "author" header in the commit preserves the author of the original pull request.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -750,7 +750,21 @@ pub fn parse_arguments(
             );
         }
         Some(o) => {
-            if o.extension().is_none() && compilation {
+            if o.as_os_str()
+                .to_string_lossy()
+                .ends_with(std::path::MAIN_SEPARATOR)
+            {
+                match Path::new(&input).file_name() {
+                    Some(i) => outputs.insert(
+                        "obj",
+                        ArtifactDescriptor {
+                            path: o.join(Path::new(i)).with_extension("obj"),
+                            optional: false,
+                        },
+                    ),
+                    None => cannot_cache!("invalid input file"),
+                };
+            } else if o.extension().is_none() {
                 outputs.insert(
                     "obj",
                     ArtifactDescriptor {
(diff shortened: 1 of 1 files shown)


jimis:
I think codecov is confused. The lines it's reporting as uncovered, are lines inside the test.

exoosh:
I tested this and it works great. Thank you.

PR #2483, opened 2025-11-07
msvc: fix detect_showincludes_prefix when VS_UNICODE_OUTPUT is set

exoosh:
This reintroduces #1835 (FYI: @AudranDoublet). Copying the details over here:

-----

`VS_UNICODE_OUTPUT` is set by Visual Studio to a value that tells MS tools running from within the IDE where to send their output. Thus, `cl.exe` invocation used to retrieve the `/showIncludes` prefix returns an empty output when this variable is set to a valid value.

The proposed fix is to use `/E` option that tells `cl.exe` to output file content in stdout and retrieved headers in stderr. When set, `cl.exe` ignores the value of `VS_UNICODE_OUTPUT`.

Another fix could be to explicitly unset `VS_UNICODE_OUTPUT` in the command invocation.

#1830 #909 mentioning these issues as it may fix them.

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -29,7 +29,7 @@ use std::collections::{HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 use std::io::{self, BufWriter, Read, Write};
 use std::path::{Path, PathBuf};
-use std::process::{self, Stdio};
+use std::process;
 
 use crate::errors::*;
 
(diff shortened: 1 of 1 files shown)


exoosh:
Trying to tend to the failed checks. It also turns out that this fix does not address #1830 for me on VS2022 with cl.exe 19.43.34810, @AudranDoublet.

But your finding certainly is an excellent starting point for me to investigate further.

exoosh:
Will rebase against `main` before this comes out of draft. Currently I rebased #2482 against `main` and added this change on top. The result seems to be a usable `sccache.exe` from the VS IDE.

sylvestre:
most of the jobs are failing

AJIOB:
Hi,  @exoosh 

Do we need this PR? Looks like #2544 resolved the problem

PR #2491, opened 2025-11-12
Don't hash --check-cfg

kornelski:
It's an assertion which doesn't affect successfully built results.


sylvestre:
please add a test for this, thanks!

etaloof:
I've added the test in #2507. Please let me know there is something missing!

Issue #2494, opened 2025-11-14
Should CARGO_ env vars be hashed at all?


kornelski:
sccache treats `CARGO_` env vars specially, and includes them in the hash. However, there are variable that are irrelevant for the output, but can break caching:

https://github.com/mozilla/sccache/blob/d4ae480184c050f29a41dac119dded829d2d5d85/src/compiler/rust.rs#L1517-L1530

and there's lot more to exclude:

 * `CARGO_BUILD_JOBS` – concurrency does not affect results
 * `CARGO_REGISTRY_*` – same as `CARGO_REGISTRIES_*` (random tokens, redundant with package IDs)
 * `CARGO_TARGET_DIR` – already included in the file paths
 * `CARGO_BUILD_BUILD_DIR`
 * `CARGO_BUILD_TARGET_DIR`
 * `CARGO_HTTP_*` – no effect on the outputs
 * `CARGO_NET_*`
 * `CARGO_TERM_*`
 * `CARGO_ALIAS_*`
 * `CARGO_CACHE_AUTO_CLEAN_FREQUENCY"`

and I'm wondering what even is needed to be included?

The problem is that `CARGO_` env vars that modify config are only one of multiple ways of modifying Cargo's config. So either `sccache` is not properly capturing the config set via `config.toml` or CLI flags, or the config overrides via env vars are redundant. I think they are redundant, as long as the rlib metadata hash is included, and possibly extra-filename set by Cargo. If Cargo's hash put in the rlib filenames is good enough for Cargo, it should be good enough for sccache?

Xuanwo:
Yep, I think it's fair to remove unrelated cargo envs.

kornelski:
What about removing *all* cargo envs?

For example, hashing `CARGO_HOME` tells you where it is, but doesn't tell you what's in `$CARGO_HOME/config.toml`. So the env var won't change if the config changes, and OTOH the env var may differ between builds, but the builds can end up using the same config anyway.

Builds with `CARGO_INCREMENTAL=0` set hash to a different hash than builds with `CARGO_PROFILE_RELEASE_INCREMENTAL=0`, even though that's the same thing when building with `--release`. And they all hash to something else when using `[profile.release] incremental = 0` to set it, and not env vars. Or when running `cargo b --release --config profile.release.incremental=0` which doesn't set the env vars, but does change the config without modifying any files on disk.

weihanglo:
[...]

PR #2495, opened 2025-11-15
Avoid hashing CARGO_ variables that are redundant/unrelated

kornelski:
Excludes more irrelevant/redundant `CARGO_` variables based on a blocklist

(Another option could be to remove hashing of the `CARGO_` variables completely — #2494)

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -1518,15 +1518,19 @@ where
                 continue;
             }
 
-            // CARGO_MAKEFLAGS will have jobserver info which is extremely non-cacheable.
-            // CARGO_REGISTRIES_*_TOKEN contains non-cacheable secrets.
-            // Registry override config doesn't need to be hashed, because deps' package IDs
-            // already uniquely identify the relevant registries.
-            // CARGO_BUILD_JOBS only affects Cargo's parallelism, not rustc output.
-            if var == "CARGO_MAKEFLAGS"
-                || var.starts_with("CARGO_REGISTRIES_")
-                || var == "CARGO_BUILD_JOBS"
-            {
+            let may_skip = var == "CARGO_MAKEFLAGS" // will have jobserver info which is extremely non-cacheable.
+                || var == "CARGO_BUILD_JOBS" // does not affect results
+                || var.starts_with("CARGO_REGISTRIES_") // CARGO_REGISTRIES_*_TOKEN contains non-cacheable secrets, and package IDs already uniquely identify the relevant registries.
+                || var.starts_with("CARGO_REGISTRY_") // same as CARGO_REGISTRIES_*
+                || var == "CARGO_TARGET_DIR" // already included in the file paths
+                || var == "CARGO_BUILD_BUILD_DIR"
+                || var == "CARGO_BUILD_TARGET_DIR"
+                || var.starts_with("CARGO_HTTP_") // not relevant
+                || var.starts_with("CARGO_NET_")
+                || var.starts_with("CARGO_TERM_")
+                || var.starts_with("CARGO_ALIAS_")
+                || var == "CARGO_CACHE_AUTO_CLEAN_FREQUENCY";
+            if may_skip {
                 continue;
             }
 
@@ -3506,6 +3510,7 @@ proc_macro false
                         OsString::from("CARGO_BUILD_JOBS"),
                         OsString::from("ignored"),
                     ),
+                    (OsString::from("CARGO_TERM_COLOR"), OsString::from("never")),
                 ]
                 .to_vec(),
                 false,


Xuanwo (approved):
Thank you for this change!

sylvestre:
would it be possible to have a test?

src/compiler/rust.rs
```diff
@@ -1518,15 +1518,19 @@ where
...
+                || var.starts_with("CARGO_REGISTRY_") // same as CARGO_REGISTRIES_*
+                || var == "CARGO_TARGET_DIR" // already included in the file paths
+                || var == "CARGO_BUILD_BUILD_DIR"
+                || var == "CARGO_BUILD_TARGET_DIR"
+                || var.starts_with("CARGO_HTTP_") // not relevant
+                || var.starts_with("CARGO_NET_")
+                || var.starts_with("CARGO_TERM_")
+                || var.starts_with("CARGO_ALIAS_")
+                || var == "CARGO_CACHE_AUTO_CLEAN_FREQUENCY";
+            if may_skip {
```
sylvestre, line 1533: maybe something like:
```suggestion
            // Skip environment variables that shouldn't affect caching
            const CARGO_SKIP_VARS: &[&str] = &[
                "CARGO_MAKEFLAGS",  // jobserver info which is extremely non-cacheable
                "CARGO_BUILD_JOBS", // does not affect results
                "CARGO_TARGET_DIR", // already included in the file paths
                "CARGO_BUILD_BUILD_DIR",
                "CARGO_BUILD_TARGET_DIR",
                "CARGO_CACHE_AUTO_CLEAN_FREQUENCY",
            ];
            const CARGO_SKIP_PREFIXES: &[&str] = &[
                "CARGO_REGISTRIES_", // contains non-cacheable secrets, package IDs already identify registries
                "CARGO_REGISTRY_",   // same as CARGO_REGISTRIES_*
                "CARGO_HTTP_",       // not relevant
                "CARGO_NET_",
                "CARGO_TERM_",
                "CARGO_ALIAS_",
            ];

            let may_skip = CARGO_SKIP_VARS.contains(&var.as_str())
                || CARGO_SKIP_PREFIXES
                    .iter()
                    .any(|prefix| var.starts_with(prefix));
            if may_skip {
```

sylvestre:
ping ?

sylvestre:
@copilot resolve the merge conflicts in this pull request

src/compiler/rust.rs
```diff
@@ -1518,15 +1518,19 @@ where
...
-            // CARGO_MAKEFLAGS will have jobserver info which is extremely non-cacheable.
-            // CARGO_REGISTRIES_*_TOKEN contains non-cacheable secrets.
-            // Registry override config doesn't need to be hashed, because deps' package IDs
-            // already uniquely identify the relevant registries.
-            // CARGO_BUILD_JOBS only affects Cargo's parallelism, not rustc output.
-            if var == "CARGO_MAKEFLAGS"
-                || var.starts_with("CARGO_REGISTRIES_")
-                || var == "CARGO_BUILD_JOBS"
-            {
+            let may_skip = var == "CARGO_MAKEFLAGS" // will have jobserver info which is extremely non-cacheable.
```
sylvestre, line 1521: could this be a list of exact names plus a list of prefixes instead of a 12-branch boolean? easier to extend

src/compiler/rust.rs
```diff
@@ -1518,15 +1518,19 @@ where
...
+                || var == "CARGO_BUILD_JOBS" // does not affect results
+                || var.starts_with("CARGO_REGISTRIES_") // CARGO_REGISTRIES_*_TOKEN contains non-cacheable secrets, and package IDs already uniquely identify the relevant registries.
```
sylvestre, line 1523: that trailing comment is too long, please move it above the block

src/compiler/rust.rs
```diff
@@ -1518,15 +1518,19 @@ where
...
+                || var.starts_with("CARGO_REGISTRY_") // same as CARGO_REGISTRIES_*
+                || var == "CARGO_TARGET_DIR" // already included in the file paths
```
sylvestre, line 1525: are we sure? cargo can pass a relative --out-dir, so the same relative path under two different target dirs would collide. could you please explain the why here

src/compiler/rust.rs
sylvestre, line 1533: `may_skip` reads odd since we always skip when it's true, just `skip`? or inline it in the if

src/compiler/rust.rs
```diff
@@ -3506,6 +3510,7 @@ proc_macro false
                         OsString::from("CARGO_BUILD_JOBS"),
                         OsString::from("ignored"),
                     ),
+                    (OsString::from("CARGO_TERM_COLOR"), OsString::from("never")),
```
sylvestre, line 3513: only CARGO_TERM_COLOR is covered, please add the other new ones to the test too (CARGO_TARGET_DIR, CARGO_HTTP_*, CARGO_REGISTRY_*)

PR #2502, opened 2025-11-18
chore: update to gzp 2

tottoto:
Updates to `gzp` 2.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1237,7 +1237,7 @@ where
     ) -> SccacheResponse {
         match compiler {
             Err(e) => {
-                debug!("check_compiler: Unsupported compiler: {}", e.to_string());
+                debug!("check_compiler: Unsupported compiler: {e}");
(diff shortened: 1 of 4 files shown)


sylvestre:
@dependabot rebase

src/server.rs
```diff
@@ -1237,7 +1237,7 @@ where
     ) -> SccacheResponse {
         match compiler {
             Err(e) => {
-                debug!("check_compiler: Unsupported compiler: {}", e.to_string());
+                debug!("check_compiler: Unsupported compiler: {e}");
```
sylvestre, line 1240: unrelated change

sylvestre:
did you run some benchmark here?

src/server.rs
tottoto: This seems to be detected after this change.

https://github.com/mozilla/sccache/actions/runs/19536092765/job/55930464405

PR #2511, opened 2025-12-14
Print hostname info for remote jobs that ran, but failed

ahartmetz:


sylvestre:
would it be possible to have a test to verify the output?

sylvestre:
please run rustfmt

src/compiler/compiler.rs
```diff
@@ -922,6 +922,14 @@ where
                 .handle_outputs(&path_transformer, &output_paths, &extra_inputs)
                 .with_context(|| "failed to rewrite outputs from compile")
         );
+
+        if jc.output.code != 0 {
+            // Add server info to help diagnose host-specific failures, e.g. due to flaky hardware.
+            // Failed builds are not cached so this tampering should not cause too much trouble.
+            let server_info = format!("sccache: Job failed on server {}:\n", server_id.addr());
```
sylvestre, line 929: is there a place where we could test that? thanks

ahartmetz: As mentioned on the Matrix channel, I just used it (as a patch on top of v0.10 even, to minimize differences to "production") and it saved a lot of time. But auto-testing seems pretty difficult in relation to a feature that is rather unlikely to break without major carelessness. Manual testing is easy because this code runs on any job that did run but without success. You can compile a C source file containing just "#error This does not compile".

ahartmetz:
> please run rustfmt

Done

src/compiler/compiler.rs
ahartmetz: Actually no, that simple example doesn't work. With #error, it already fails in the preprocessor, and the compiler needs to run without the linking step for sccache to accept it as a compile job. This works:
`echo "void foo() { return 1; }" > fail.c ; gcc -c fail.c`

PR #2512, opened 2025-12-15
feat: Add riscv64 support

ffgan:
Hi there.

Recently, riscv64 (rv64) has gained widespread adoption, and we might consider adding rv64 support to sccache. Our build image `messense/rust-musl-cross` has excellent support for rv64, so we can easily add rv64 support to sccache, but there are two points that need special mention.

One of our dependencies is openssl. The version we currently use fails to compile on rv64, so we need to upgrade the version.

I chose to upgrade to the latest version, `300.5.4+3.5.4,` which compiles successfully for non-s390x architectures. For s390x, there's a compilation issue: `Error: Unrecognized opcode: 'cijne'`. After reviewing related issues, I inserted a compilation flag specifically for s390x to enable it to compile successfully. For the specific code, please refer to the content in the PR. 

The patch comes from two sources: [1](https://github.com/openssl/openssl/issues/27323), [2](https://github.com/pytogo/portforward/pull/49/changes).

No other modifications are needed to complete the compilation. I ran CI in my fork, and the results show everything is good. Please check the links below for details:

1. [ci](https://github.com/ffgan/sccache/actions/runs/20231390879)
2. [integration-tests](https://github.com/ffgan/sccache/actions/runs/20231390833)

Regarding the `ci`, there are some failed jobs that I need to briefly explain. Some failures are due to running in my fork and lacking `token`, which causes `Upload coverage results` to fail—this doesn't affect this PR. Others are due to `The macOS-13 based runner images are now retired`, which also doesn't affect this PR.

Therefore, we can add rv64 support very well. 

If you have any questions about the above content, please @ me directly. I'm happy to answer all questions.

---
**Other Info**
Co-authored by: nijincheng@iscas.ac.cn;

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,6 +277,14 @@ jobs:
           toolchain: ${{ matrix.target == 'aarch64-apple-darwin' && 'beta' || 'stable' }}
           target: ${{ matrix.target }}
         if: ${{ !matrix.container }}
+      
+      - name: fix build openssl error on s390x
+        if: ${{  matrix.target  == 's390x-unknown-linux-musl' }}
+        run: |
+          # see https://github.com/openssl/openssl/issues/27323
+          echo CFLAGS="-march=z10 -mzarch" >> $GITHUB_ENV
+          echo CXXFLAGS="$CFLAGS" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


ffgan:
Hi @sylvestre . Thank you for merging this PR. Could we please consider releasing a new version so that we can directly use sccache for riscv64?

PR #2518, opened 2025-12-18
add a nix flake

TroyKomodo:
Not sure if wanted, can close if not.

Adds nix flake so that nix shells can inherit the required tooling to build / develop the project.

### developer env usage

```bash
# direnv + nix
direnv allow

# nix only
nix develop .
```

### package building

The nix flake also adds the package building process used by end users (separate component can be removed).

```
nix build .#sccache
```

#### module integration

```nix
# Add flake input
sccache.url = "github:mozilla/sccache";

# Add overlay
nixpkgs.overlays = [ sccache.overlays.default ];
environment.systemPackages = [ pkgs.sccache ]; # nixos
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,6 +1,6 @@
 if has nix; then
-  if ! has nix_direnv_version || ! nix_direnv_version 3.0.7; then
-    source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.7/direnvrc" "sha256-bn8WANE5a91RusFmRI7kS751ApelG02nMcwRekC/qzc="
+  if ! has nix_direnv_version || ! nix_direnv_version 3.1.0; then
+    source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.1.0/direnvrc" "sha256-yMJ2OVMzrFaDPn7q8nCBZFRYpL/f0RcHzhmw/i6btJM="
   fi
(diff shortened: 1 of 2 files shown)


TroyKomodo:
> I am not familiar with the `.envrc` files, and I am not fond of pulling online resources when doing local development. However this file might be slightly outdated (3.0.7 -> 3.1.0) compared to https://github.com/nix-community/nix-direnv/blob/master/templates/flake/.envrc

direnv basically initializes the shell when entering the directory. Also plays nice with vscode using the https://marketplace.visualstudio.com/items?itemName=mkhl.direnv extension which causes all of the other vscode extensions to inherit the nix dev shell. (such as rust-analyzer) 

Hmm, i see the latest release is https://github.com/nix-community/nix-direnv/releases/tag/3.0.7, but i see the 3.1.0 tag so i think they just forgot to create a git release for it. I'll update.

> I would not have included the NixOS / HomeManager modules only for registering the flake output as part of the system packages, unless there was some added value such as registering a systemd jobs for starting the sccache server.

I think thats out of scope for this. I'll remove them.

> I will also note that the lack of overlay for the definition of packages imply that this flake creates an additional instances of Nixpkgs (line 54: `import nixpkgs`) when using any of the modules, which is a bad practice.

I think I fixed it?

Issue #1020, opened 2021-07-13
[feature] Multiple cache levels


browdus:
Hi everyone! Is it possible, or has it been considered, to allow multiple levels of cache? For instance, using local disk as an L1 cache, and s3/azure/gcs as L2.

I envision it working something like the following:

```
Check disk cache according to current rules
if (found)
    return item

check s3/azure/gcs according to current rules
if (found)
    return item

compile item
populate disk cache
populate s3/azure/gcs cache

return item
```

This lets us use nvme as a hot cache, while still being able to hit the network cache if needed. This (theoretically) avoids a lot of network traffic and associated costs, and reduces build times on average. It also gives us a transparent way to seed new build hosts when they spin up, which makes automatic scaling of the build fleet more effective. Since each host has it's own hot cache, the load on s3/azure/gcs should scale better than linearly with the number of build hosts (on average; linear being the worst case).

Generically, it could be useful to mix/match all the cache types. For example, local disk as L1, on-prem memcached/redis as L2, and s3/azure/gcs as L3. Such a setup could be used to share build cache with all the devs in an office without hitting the internet as much, while still being able to share that build cache with other offices, etc.

Thanks!

browdus:
related: https://github.com/mozilla/sccache/pull/241

browdus:
For anyone wanting to do something similar, you can hack something together by combining `ccache` and `sccache`. The result is not optimal (multiple compression steps, for example), but it works reasonably well. This effectively gives you an L1 disk cache, and an L2 of anything `sccache` can handle.

```bash
export CCACHE_PREFIX="sccache"
ccache gcc ...
```

If you are using cmake, make sure you don't have something like the following, as `sccache` will fail to handle linking (as designed).

```cmake
set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
```

Be-ing:
> set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)

Use CMAKE_C_COMPILER_LAUNCHER / CMAKE_CXX_COMPILER_LAUNCHER rather than RULE_LAUNCH_COMPILE:
https://github.com/mozilla/sccache/issues/947

PR #2521, opened 2025-12-18
Add `SCCACHE_BASEDIRS` support

### LINKED ISSUE #35: Implement an equivalent to `CCACHE_BASEDIR`

Felixoid:
This is an attempt to fix #35.

The ENV variable `SCCACHE_BASEDIRS` and configuration parameter `basedirs` are added.

As well as new tests to validate the behavior.

### DIFF
--- a/tests/integration/scripts/test-basedirs.sh
+++ b/tests/integration/scripts/test-basedirs.sh
@@ -67,10 +67,12 @@ test_backend() {
 
     echo "Cache hits for $backend_name: $CACHE_HITS"
 
-    if [ "$CACHE_HITS" -gt 0 ]; then
+    SECOND_MISSES=$(echo "$STATS_JSON" | python3 -c "import sys, json; stats = json.load(sys.stdin).get('stats', {}); print(stats.get('cache_misses', {}).get('counts', {}).get('C/C++', 0))")
+
+    if [ "$FIRST_MISSES" == "$SECOND_MISSES" ]; then
         echo "✓ PASS: $backend_name - Basedir test successful"
     else
-        echo "✗ FAIL: $backend_name - No cache hits detected with basedirs"
+        echo "✗ FAIL: $backend_name - Basedir test failed, cache misses did not remain the same: $FIRST_MISSES != $SECOND_MISSES"
         echo "$STATS_JSON" | python3 -m json.tool
         exit 1
     fi


src/config.rs
```diff
@@ -946,7 +970,19 @@ fn config_from_env() -> Result<EnvConfig> {
         oss,
     };
 
-    Ok(EnvConfig { cache })
+    // ======= Base directory =======
+    // Support multiple paths separated by '|' (a character forbidden in paths)
+    let basedir = env::var_os("SCCACHE_BASEDIR")
```
Felixoid: Probably, after the rewrite to `basedirs`, this parameter should be called `SCCACHE_BASEDIRS` to reflect the current state?

src/config.rs
```diff
@@ -978,6 +1014,10 @@ pub struct Config {
     pub fallback_cache: DiskCacheConfig,
     pub dist: DistConfig,
     pub server_startup_timeout: Option<std::time::Duration>,
+    /// Base directory (or directories) to strip from paths for cache key computation.
+    /// Similar to ccache's CCACHE_BASEDIR.
+    /// Currently only the first directory is used if multiple are specified.
```
Felixoid: Outdated comment

src/util.rs
```diff
@@ -1015,6 +1015,74 @@ pub fn num_cpus() -> usize {
...
+    if basedirs.is_empty() || preprocessor_output.is_empty() {
+        return preprocessor_output.to_vec();
+    }
+
+    // Prepare normalized basedirs sorted by length (longest first) to match longest prefix first
+    let mut basedirs_data: Vec<_> = basedirs
+        .iter()
+        .map(|basedir| {
+            let normalized = basedir.to_string_lossy();
+            let trimmed = normalized.trim_end_matches('/').trim_end_matches('\\');
```
AJIOB: Does the `c:/test` and `c:\test` both recognize as a prefix if only `c:/test` was passed as a prefix one?

Windows build may mix slashes in the paths

Felixoid: No, and I'm not sure how it should be addressed. 

In any case, it's possible to put both cases info the basedirs parameter. Or you rather mean, that both slashes should be considered as the directory separator?

Felixoid:
I got an idea, that basedirs should be added to the stats command as well

src/util.rs
AJIOB: Yes, both slashes may be a directory separator.

For this example we have only one slash in the path, but if we have a `c:/very/long/base/directory/path/with/slashes`, we need to pass 2^7=128 variants, that is not good

Felixoid: Sad story, never knew MS started supporting forward slashes as the directory separators. And the case where they're mixed sounds cursed to me.

But the point is valid; normalization should be done automatically.

mathstuf: Windows and macOS have always had case folding anyways. There are also now per-directory flags for Linux to be case-insensitive (e.g., ext4), but has also always had the issue (e.g., FAT). I believe `inode` comparisons can also fail with subvolume mounts.

Anyways, case and separator confusion are the only issues that I think need to be considered here. Actually using case-insensitive or confusable subvolume file system setups for `sccache` on Linux use cases seems cursed.

AJIOB: Windows case-insensitivity is also a good case. Linux (and possibly macOS) are case-insensitive by default. I suggest to implement these cases too.

P.S. NTFS can be made also case-sensitive, but it's rare case.
P.P.S. Please, add some notes about case-sensitivity and slashes to the docs. It should be known by the end user IMO

Felixoid: As far as I can tell, you suggest making `basedirs` elements case-insensitive for all cases for simplicity.

Yes, it looks like a straightforward approach. I'll make sure to implement it.

mathstuf: I don't think you can make it case-insensitive on all platforms. I think it just needs a normalization pass applied to it. I know Windows has an API to get the actual case of a path. macOS probably has some mechanism (ideally better than `readdir`).

Note that on Windows, it is not necessarily just a case thing. You have DOS 8.3 replacements (`c:\Docume~1` for `C:\Documents and Settings`) that might be reported/used in some places.

Can a trace-level log be added for any attempted path comparisons (as well as results) so that these corner cases can be detected by users in the wild?

Felixoid: I prefer a case-sensitive approach for Unix OSs and a case-insensitive approach for Windows. The macOS behavior was one of the most confusing I have ever seen, with inconsistent case sensitivity. So, I partially reverted the latest commit with global case insensitivity.

<details><summary>MacOS FS behavior</summary>

As far as I remember, a several-year-old experiment did show the following apple's FS behavior:

- If there was only `a` file, `ls A` would show it
- If there were `a` and `A` files, `ls A` would show the correct `A` file
- `touch a` and `touch A` would create files properly

I think the path should match exactly in this case, without complex logic for searching files.

</details>

I can't argue with the Win approach; I don't use it anywhere, and I've forgotten every edge case it could have.

Can I implement a case-sensitive match on Unix and a case-insensitive match on Windows, so that the latter can be improved later if necessary?

I added `trace!` logging, which showed that not everything was covered at the moment. It should be better now.


BTW, thanks a lot for a fast-paced review! :+1:

mathstuf: macOS does Unicode normalization too (NFC IIRC, but it doesn't really matter). macOS is indeed weird and is in a case-sensitive transition of some kind, but I lack an Apple fortuneteller to know any plans or schedules.

macOS is far better at case preservation (i.e., if you refer to `a` as `A`, most everything will remember that). Windows tools will do what they do and internal normalization is common (slash flipping, all-upper, all-lower, 8.3 transforms, UNC syntax, you name it). I'm fine with the Windows is case insensitive and everything else is case sensitive plan as mixups mostly happen on Windows. Having the ability to get a log of issues will definitely help people use consistent cases where they need to.

AJIOB:
> I got an idea, that basedirs should be added to the stats command as well

I think we can also provide the stat about base dir usage translation:
* Number of base dir applied requests
* Number of base dir skipped requests 

In this case we can see, do we need to provide more/better base dirs or not

src/util.rs
whisperity: > As far as I remember, a several-year-old experiment did show the following apple's FS behavior:
> 
>   * If there was only `a` file, `ls A` would show it
> 
>   * If there were `a` and `A` files, `ls A` would show the correct `A` file
> 
>   * `touch a` and `touch A` would create files properly

Depends on your ~~partition~~ volume settings. By default, a stock MacOS installation will have a case **in**sensitive volume, so `touch a` and `touch A` will refer to the same file. As for the POSIX API, when you obtain a file descriptor, the file descriptor will remember exactly the path on how you obtained it (AFAIK, it does not get normalised), but when the actual file accesses translate through the kernel, the case will be lost. The FS, however, remembers the "first" casing of the file on how you created it.

What's interesting is that despite the underlying FS being case insensitive, the Shell itself retains case sensitivity when globbing… (This might be tuneable depending on the shell!)

See:

```bash
❯ echo "hello world" > hello.txt
❯ echo "HELLO WORLD" > HELLO.txt
❯ ls -alh *.txt
Permissions Size User       Date Modified Name
.rw-r--r--    12 whisperity 24 Dec 12:01  hello.txt
❯ cat hello.txt; cat HELLO.txt
HELLO WORLD
HELLO WORLD

❯ echo "goodbye" > BYE.TXT
❯ echo "GOODBYE" > bye.txt
❯ ls -alh *.txt
Permissions Size User       Date Modified Name
.rw-r--r--    12 whisperity 24 Dec 12:01  hello.txt
❯ ls -alh *.TXT
Permissions Size User       Date Modified Name
.rw-r--r--     8 whisperity 24 Dec 12:02  BYE.TXT
❯ cat bye.txt BYE.TXT
GOODBYE
GOODBYE
```

Many tools, such as Git or Subversion can run into issues when in the repository people change files' cases or commit multiple files that only differ in a case. Recent versions of Git handle this gracefully, notifying you that the checkout failed because the case-insensitivity overwrote an existing file with a different content. Subversion is much worse in this regard, it just shows a content conflict. All these are resolvable issues (if you have commit rights), but takes some VCS-specific maintainer spelunking to untangle.

You can create a case-**sensitive** volume, and mount it somewhere, at which point everything will behave as if you were on Linux. 🙂 In fact, if you ever reach a point that you need to self-compile GCC on Mac, this is something you will have to suffer.

```bash
❯ cd /Volumes/Case-Sensitive      # Just a random volume I have created and mounted long ago thanks to GCC…
❯ echo "hello world" > hello.txt
❯ echo "HELLO WORLD" > HELLO.txt
❯ echo "goodbye" > BYE.TXT
❯ echo "GOODBYE" > bye.txt
❯ cat *.txt
HELLO WORLD
GOODBYE
hello world
❯ cat *.TXT
goodbye
❯ ls -alhi
 inode Permissions Size User       Date Modified Name
937334 .rw-r--r--     8 whisperity 24 Dec 12:06  bye.txt
937333 .rw-r--r--     8 whisperity 24 Dec 12:06  BYE.TXT
937335 .rw-r--r--    12 whisperity 24 Dec 12:07  HELLO.txt
937332 .rw-r--r--    12 whisperity 24 Dec 12:06  hello.txt
```

MacOS 15.5

docs/Configuration.md
```diff
@@ -6,6 +6,19 @@
...
+# different absolute paths when compiling the same source code.
+# Can be an array of paths. When multiple paths are provided,
+# the longest matching prefix is used.
+# Path matching is case-insensitive on Windows and case-sensitive on other OSes.
+# For example, if basedir is "/home/user/project", then paths like
+# "/home/user/project/src/main.c" will be normalized to "./src/main.c"
+# for caching purposes.
+basedirs = ["/home/user/project"]
+# Or multiple directories:
+# basedirs = ["/home/user/project", "/home/user/workspace"]
```
whisperity: Bit of a rephrasing suggestion:


```suggestion
# Base directories to strip from source paths during cache key
# computation.
#
# Similar to ccache's CCACHE_BASEDIR, but supports multiple paths.
#
# 'basedirs' enables cache hits across different absolute root
# paths when compiling the same source code, such as between
# parallel checkouts of the same project, Git worktrees, or different
# users in a shared environment.
# When multiple paths are provided, the longest matching prefix
# is applied.
# 
# Path matching is case-insensitive on Windows and case-sensitive on other OSes.
#
# Example:
#   basedir = ["/home/user/project"] results in the path prefix rewrite:
#   "/home/user/project/src/main.c" -> "./src/main.c"
basedirs = ["/home/user/project"]
# basedirs = ["/home/user/project", "/home/user/workspace"]
```

I'd not say "can be an array of paths" if the "one path" example already is an array. What happens if you do `basedirs = "/home/foo/bar"`? (I hope it's not the dreaded Python-like behaviour that it starts chewing away an array of characters one by one…)

docs/Configuration.md
```diff
@@ -134,6 +147,7 @@ Note that some env variables may need sccache server restart to take effect.
 
 * `SCCACHE_ALLOW_CORE_DUMPS` to enable core dumps by the server
 * `SCCACHE_CONF` configuration file path
+* `SCCACHE_BASEDIRS` base directory (or directories) to strip from paths for cache key computation. This is similar to ccache's `CCACHE_BASEDIR` and enables cache hits across different absolute paths when compiling the same source code. Multiple directories can be separated by `|` (pipe character). When multiple directories are specified, the longest matching prefix is used. Path matching is **case-insensitive** on Windows and **case-sensitive** on other operating systems. Environment variable takes precedence over file configuration. Only absolute paths are supported; relative paths will be ignored with a warning.
```
whisperity: Usually when you provide multiple directories or file paths in environment variables (such as `PATH`, `LD_LIBRARY_PATH`, `LD_PRELOAD`, etc.), the convention is to use `;` as a separator. Why was `|` chosen here instead?

src/cache/cache.rs
```diff
@@ -736,12 +740,35 @@ pub fn storage_from_config(
...
+        .basedirs
+        .iter()
+        .filter_map(|p| {
+            if p.is_absolute() {
+                Some(p.clone())
+            } else {
+                warn!(
+                    "Ignoring relative basedir path: {:?}. Only absolute paths are supported.",
+                    p
+                );
```
whisperity: If a relative basedir is an invalid configuration, wouldn't it make more sense to error here and get the dev/server admin to fix their config?

src/config.rs
```diff
@@ -584,6 +584,9 @@ pub struct FileConfig {
     pub cache: CacheConfigs,
     pub dist: DistConfig,
     pub server_startup_timeout_ms: Option<u64>,
+    /// Base directory (or directories) to strip from paths for cache key computation.
+    /// Can be a single path or an array of paths.
+    pub basedirs: Vec<PathBuf>,
```
whisperity:
```suggestion
    /// Base directories to strip from paths for cache key computation.
    pub basedirs: Vec<PathBuf>,
```

`Vec` already captures the notion that it can be multiple elements.

src/util.rs
whisperity: Also, a higher-level question. **Is there no Rust library that could handle these OS-specifics for paths instead of us having to juggle around strings in a potentially clunky or edge-sharp way?** In Python, you'd have `pathlib` standard library giving you the platform-specific right behaviour. On GNU Linux, there's also the `realpath` and `readlink` (GNU realpath can even do `--relative-to`!) binaries and POSIX `realpath()` call which do normalisations and such.

src/util.rs
```diff
@@ -1015,6 +1015,120 @@ pub fn num_cpus() -> usize {
...
+                } else {
+                    false
+                };
+
+                if exact_match || normalized_match {
+                    // Check if this is actually a path boundary (preceded by whitespace, quote, or start)
+                    let is_boundary = i == 0
+                        || preprocessor_output[i - 1].is_ascii_whitespace()
+                        || preprocessor_output[i - 1] == b'"'
+                        || preprocessor_output[i - 1] == b'<';
```
whisperity: This might be ultra cursed even beyond the case sensitivity issue here, but on Linux and MacOS, whitespace and `"` are perfectly legit members of a path element's name, as long as they are escaped.

`/home/Whisperity/Projects/The\ LLVM\ Compiler\ Infrastructure\ Project/.Worktrees/Research:\ \"The Big Lebowski\"` is a legit path to have as a basedir. Granted, you can simplify all these escapes if you wrap things in quotes, but there is a countably infinite depth of escape hell you can get into here.

```
❯ gstat "./Projects/The LLVM Compiler Infrastructure Project/.Worktrees/Research: \"The Big Lebowski\""
  File: ./Projects/The LLVM Compiler Infrastructure Project/.Worktrees/Research: "The Big Lebowski"
  Size: 64              Blocks: 0          IO Block: 4096   directory
Device: 1,17    Inode: 117446528   Links: 2
Access: (0755/drwxr-xr-x)  Uid: (  501/whisperity)   Gid: (   20/   staff)
```

AJIOB: IMO, the quotes are rare case, so I can suggest to add it's support additionally

docs/Configuration.md
AJIOB: nit: Linux way is to use `:` as a PATH delimiter, but the Windows uses `:` for drive letter providing

whisperity: > nit: Linux way is to use `:` as a PATH delimiter

Ah, yes. Of course I mixed it up with CMake, which generally uses `;` as an array separator…

Felixoid: I chosen `|` because any of `,` and `:` could be a part of the valid path. Didn't think of other options, but testing it blown my mind:

```
> ls -ld test*dir
drwxr-xr-x 2 felixoid felixoid 6 Dec 24 14:13  test:dir
drwxr-xr-x 2 felixoid felixoid 6 Dec 24 14:12 'test;dir'
drwxr-xr-x 2 felixoid felixoid 6 Dec 24 14:13 'test|dir'
```

The colon doesn't look like an option to me because of the `c:/` format. Coma is too familiar.

I am open for suggestions. Maybe, `;` looks like a decent compromise.

src/cache/cache.rs
Felixoid: By "error," you mean "fail to start"? It could be an option as well, I didn't think of it as a misconfiguration.

Looks OK to me to hard fail here.

src/config.rs
Felixoid: Right, it's a leftover from the first implementation.

src/util.rs
Felixoid: > `/home/Whisperity/Projects/The\ LLVM\ Compiler\ Infrastructure\ Project/.Worktrees/Research:\ \"The Big Lebowski\"`

I think, it's fine. The file will still be processed. The `preprocessor_output[i - 1] == b'"'` checks the opening border of the path.

To get stroke by some issue, I could imagine the following:

`basedir = '/home/user/project'` and `preprocessor_output = '"/home/user2/test-dir/tragic\ mistake\ \"/home/user/project/real-dir/file.c'`

But I'd consider it the system's current limit. The main issue here, `preprocessor_output` is the plain text. It can contain a lot of things =\

src/util.rs
Felixoid: > **Is there no Rust library that could handle these OS-specifics for paths instead of us having to juggle around strings in a potentially clunky or edge-sharp way?**

The same issue as described nearby. The `preprocessor_output` is not a path; it's a long plain text. To cut a path from there to check a prefix/relative path, we need to identify its boundaries.

While thinking about it, I realized the function is suboptimal anyway. Instead of iterating byte by byte, it should search for every basedir, and then take the lowest found position.

Back to your point, how do you think it could be done to normalize paths in the `preprocessed_output`?

Felixoid:
> * Number of base dir applied requests
> * Number of base dir skipped requests

It's a tricky one. After taking a look, the number of successful substitutions is relatively easy to implement, although the counter should be threaded to the ServerStats.

But how to count the number of skipped directories? What is it? If a `base_directory` didn't match any of the `output` at all?

docs/Configuration.md
mathstuf: Please use `;` on Windows and `:` elsewhere. There's no "safe" character on POSIX as any non-NUL, non-`/` byte is valid in filenames and everything uses `:` for path list separators. If `:`-escaping is needed, someone will come and make noise about it. It can still be considered whether it is a valid use case at that time.

AJIOB: IMO, we can select `;` as a cross-platform delimiter, similar to CMake.

If we have a platform-dependent selector, what should be used for Mingw and Cygwin? Or it will depend of the linker is used (GNU LD or MSVC link.exe)? Some not-so-easy to know questions

AJIOB:
> > * Number of base dir applied requests
> > * Number of base dir skipped requests
> 
> It's a tricky one. After taking a look, the number of successful substitutions is relatively easy to implement, although the counter should be threaded to the ServerStats.
> 
> But how to count the number of skipped directories? What is it? If a `base_directory` didn't match any of the `output` at all?

No, just to check what number of cache requests has been converted using any of base dirs and what number was kept as an absolute one

AJIOB:
About paths: we can try to normalize paths via https://docs.rs/normpath/latest/normpath/ or something similar instead of doing one more implementation.

docs/Configuration.md
mathstuf: Cygwin is a POSIX, so that is `:`. MinGW is an alternate toolchain/runtime for Windows, so that is `;`.

AJIOB: JFYI, both of Mingw and Cygwin provides bash as their shell and use Linux-like paths.

I suggest to unify the delimiter to the single cross-platform symbol or make it configurable too (but it, may be an almost unused option).

I agreed with any of variants, because the feature looks nice

Felixoid: I feel the OS dependent delimiter would be very confusing. I can imagine how much cognitive load it would add. And if somebody uses infrastructure-as-a-code for different environments, it will be one more unnecessary "if os ==.." branch.

I'd go with cross-platform CMAKE-like approach, as it works fine. And then, for any other lists, it could be used too (e.g. there's a feature request for caching levels, where levels should be set as a list as well).

mathstuf: IMO, having a tool use some foreign path list separator is more confusing.

Even CMake uses `:` for environment path list separators on non-Windows. The `;` separators used everywhere are in the CMake language itself.

Felixoid: https://github.com/mozilla/sccache/pull/2521#discussion_r2645676588

Both are valid symbols in the path, so I didn't follow why PATH env separator should be inherited in sccache. The semantic doesn't look the same to me.

mathstuf: The pattern used by `PATH` is typically used for any environment variable which needs to specify multiple paths in a string.

Felixoid: Given that: 
- Colon it's a valid path symbol, despite it works for legacy PATH semantics
- It doesn't work across all platforms, that sccache sports
- It should be a list/array separator, not a path separator (see my point regarding cache levels)
Do you still insist on the different symbols for this case? 

I think mostly of the letter case. IMO, if now one will decide on this delimiter, it must remain as the os dependent list separator. So, the separator in the theoretical following SCCACHE_LEVELS will be as well different on different platforms. 

I'd say, I don't care about it on the highest level. But I can imagine the frustration about either semantics (across variables) or os inconsistency. I'd be very unhappy about it. 

I put my voice to choose the symbol for an array elements delimiter, consistent for most of the cases. Sticking to a very specific case is a stair to the dark wood.

mathstuf: I'm unfamiliar with `SCCACHE_LEVELS` and didn't see an issue about it. Do you have a link?

Given that I'm not familiar with the feature, I will speculate some on what it might mean. I'm still OK with path-list envvars using an OS-dependent separator and list-as-string (e.g., `backend1,backend2`) envvars using an OS-independent separator (that we know is not valid as a backend name).

For the first point, the *only* symbol available that is not a valid path symbol is NUL (`\0`) (`/` cannot be in a filename, but given the must-be-absolute rule, it will appear in every entry anyways) which cannot be in an envvar value anyways as it is also the terminator for it, so something needs to be sacrificed. History has decided that it is `:` and I'd encourage following that here. If support for escaping it is necessary, that can be visited when someone trips over it. Windows does have symbols to "sacrifice" as they cannot appear in paths otherwise and `;` is the symbol that history has chosen there.

Felixoid: > I'm unfamiliar with `SCCACHE_LEVELS` and didn't see an issue about it. Do you have a link?

It's about https://github.com/mozilla/sccache/issues/1020, I started to think about it recently as well. For multiple caching levels, when it would be more than `local -> remote`, it requires an array-like environment variable, which is why I thought of the global elements separators rather than a situative one.

> I'm still OK with path-list envvars using an OS-dependent separator and list-as-string (e.g., `backend1,backend2`) envvars using an OS-independent separator (that we know is not valid as a backend name).

Ok, fair enough. It differs from my opinion, but it's more important to agree on something here. Let's finish the feature and address an annoying issue :+1: 

I'll add an OS-dependent separator to the parse and update the documentation tomorrow, when I'll continue working on the feature.

Felixoid:
> About paths: we can try to normalize paths via https://docs.rs/normpath/latest/normpath/ or something similar instead of doing one more implementation.

It looks like https://doc.rust-lang.org/stable/std/path/fn.absolute.html is what I should use for both cases. It normalizes slashes, but keeps the cases as is.

Although see my [comment](https://github.com/mozilla/sccache/pull/2521#discussion_r2645871121) regarding the `preprocessor_output`. `normalize_path` is used not only to needles from `basedirs`, but to a haystack as well. Both https://doc.rust-lang.org/stable/std/path/fn.absolute.html and https://docs.rs/normpath/latest/normpath/trait.PathExt.html#tymethod.normalize_virtually work on Path-like objects. They don't fit the purpose of the plaintext normalization, as far as I can tell.

docs/Configuration.md
Felixoid:
```
---- config::test_deserialize_basedirs_missing stdout ----

thread 'config::test_deserialize_basedirs_missing' panicked at src/config.rs:1481:51:
called `Result::unwrap()` on an `Err` value: Error { inner: Error { inner: TomlError { message: "invalid type: string \"/home/user/project\", expected a sequence", original: Some("\n        basedirs = \"/home/user/project\"\n        [cache.disk]\n        dir = \"/tmp/cache\"\n        size = 1073741824\n\n        [dist]\n    "), keys: ["basedirs"], span: Some(20..40) } } }
```

Toml takes care of it

src/cache/cache.rs
Felixoid:
```
> SCCACHE_BASEDIRS='/tmp|tmp' SCCACHE_DIR=~/.cache/sccache SCCACHE_NO_DAEMON=1 ./target/debug/sccache --stop-server
sccache: error: Basedir path must be absolute: "tmp"
```

This is fixed

Felixoid:
> No, just to check what number of cache requests has been converted using any of base dirs and what number was kept as an absolute one

I am afraid it causes changes in too many places, including the signatures of `hash_key` and other significant components. I want to cut this task from the PR, if you don't mind; it's already quite huge.

src/util.rs
Felixoid: I think, with `std::path::absolute` + trimming leading slashes, it's good enough to go

src/util.rs
Felixoid: This looks addressed to me as well

src/config.rs
```diff
@@ -989,32 +1012,57 @@ impl Config {
...
+        // Environment variable takes precedence over file config
+        let basedirs_raw = if !env_basedirs.is_empty() {
+            env_basedirs
+        } else {
+            file_basedirs
+        };
+
+        // Validate that all basedirs are absolute paths
+        let mut basedirs = Vec::new();
+        for p in basedirs_raw {
```
AJIOB:
```suggestion
        basedirs.reserve(basedirs_raw.size());
        for p in basedirs_raw {
```

nit: less reallocations

src/util.rs
```diff
@@ -1015,10 +1015,121 @@ pub fn num_cpus() -> usize {
...
+    // Find all potential matches for each basedir using fast substring search
+    // Store as (position, length) sorted by position
+    let mut matches: Vec<(usize, usize)> = Vec::new();
+    #[cfg(target_os = "windows")]
+    let preprocessor_output = &normalize_path(preprocessor_output);
+
+    for basedir_path in basedirs.iter() {
+        let basedir_str = basedir_path.to_string_lossy();
+        let basedir = basedir_str
+            .trim_end_matches('/')
```
AJIOB: nit: why we do this trims every time, not while loading it from config?

src/util.rs
```diff
@@ -1015,10 +1015,121 @@ pub fn num_cpus() -> usize {
...
+            .trim_end_matches('\\')
```
AJIOB:
```suggestion
            .trim_end_matches(|c| c == '/' || c == '\\')
```

Or the `c:\abs/\` will be translated to `c:\abs\` instead of `c:\abs`

src/util.rs
```diff
@@ -1015,10 +1015,121 @@ pub fn num_cpus() -> usize {
...
+            .as_bytes();
+        #[cfg(target_os = "windows")]
+        // Case insensitive
+        let basedir = normalize_path(basedir);
```
AJIOB: nit: this can be done once too - while loading

src/config.rs
Felixoid: Maybe, the single `Vec::with_capacity(basedirs_raw.size())` then?

src/util.rs
```diff
@@ -1015,10 +1015,121 @@ pub fn num_cpus() -> usize {
...
+///             lowercases ASCII characters for consistency.
+/// This function used for two purposes:
+///     - basedir_path: already normalized by std::path::absolute,
+///       but still can mismatch the case of the actual file name
+///     - preprocessor_output: plain text
+#[cfg(target_os = "windows")]
+fn normalize_path(path: &[u8]) -> Vec<u8> {
+    path.iter()
+        .map(|&b| match b {
+            b'A'..=b'Z' => b + 32,
```
AJIOB:
```suggestion
            b'A'..=b'Z' => b + 'a' - 'A',
```

nit: IMO it's more readable, that magic constant

src/util.rs
Felixoid: On one hand, it's a good point.

On the other hand, I feel it's more reliable than to hope, that the directories are cleaned in the `config` generation.

Probably, we can move this logic there AND test that `strip_basedirs` works on the output from Config.

docs/Configuration.md
```diff
@@ -134,6 +154,7 @@ Note that some env variables may need sccache server restart to take effect.
 
 * `SCCACHE_ALLOW_CORE_DUMPS` to enable core dumps by the server
 * `SCCACHE_CONF` configuration file path
+* `SCCACHE_BASEDIRS` base directory (or directories) to strip from paths for cache key computation. This is similar to ccache's `CCACHE_BASEDIR` and enables cache hits across different absolute paths when compiling the same source code. Multiple directories can be separated by `;` on Windows hosts and by `:` on any other. When multiple directories are specified, the longest matching prefix is used. Path matching is **case-insensitive** on Windows and **case-sensitive** on other operating systems. Environment variable takes precedence over file configuration. Only absolute paths are supported; relative paths will cause an error and prevent the server from start.
```
mathstuf:
```suggestion
* `SCCACHE_BASEDIRS` base directory (or directories) to strip from paths for cache key computation. This is similar to ccache's `CCACHE_BASEDIR` and enables cache hits across different absolute paths when compiling the same source code. Multiple directories can be separated by `;` on Windows hosts and by `:` on any other operating system. When multiple directories are specified, the longest matching prefix is used. Path matching is **case-insensitive** on Windows and **case-sensitive** on other operating systems. Environment variable takes precedence over file configuration. Only absolute paths are supported; relative paths will cause an error and prevent the server from start.
```

README.md
```diff
@@ -278,12 +278,49 @@ This is most useful when using sccache for Rust compilation, as rustc supports u
...
+Normalizing Paths with `SCCACHE_BASEDIRS`
+-----------------------------------------
+
+By default, sccache requires absolute paths to match for cache hits. To enable cache sharing across different build directories, you can set `SCCACHE_BASEDIRS` to strip a base directory from paths before hashing:
+
+```bash
+export SCCACHE_BASEDIRS=/home/user/project
+```
+
+You can also specify multiple base directories by separating them by `;` on Windows hosts and by `:` on any other. When multiple directories are provided, the longest matching prefix is used:
```
mathstuf:
```suggestion
You can also specify multiple base directories by separating them by `;` on Windows hosts and by `:` on any other operating system. When multiple directories are provided, the longest matching prefix is used:
```

docs/Configuration.md
```diff
@@ -6,6 +6,26 @@
...
+# computation.
+#
+# Similar to ccache's CCACHE_BASEDIR, but supports multiple paths.
+#
+# 'basedirs' enables cache hits across different absolute root
+# paths when compiling the same source code, such as between
+# parallel checkouts of the same project, Git worktrees, or different
+# users in a shared environment.
+# When multiple paths are provided, the longest matching prefix
+# is applied.
```
mathstuf: @copilot 
```suggestion
# When multiple matching paths are provided, the longest prefix
# is used.
```

src/cache/readonly.rs
```diff
@@ -121,6 +127,40 @@ mod test {
...
+    #[test]
+    fn readonly_storage_forwards_basedirs() {
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .worker_threads(1)
+            .build()
+            .unwrap();
+
+        let tempdir = tempfile::Builder::new()
+            .prefix("sccache_test_readonly_basedirs")
```
mathstuf:
```suggestion
            .prefix("readonly_storage_forwards_basedirs")
```

?

src/compiler/c.rs
```diff
@@ -1623,10 +1786,143 @@ mod test {
...
+        let digest = "abcd";
+
+        // Test 1: Same hash with different absolute paths when basedir is used
+        let preprocessed1 = b"# 1 \"/home/user1/project/src/main.c\"\nint main() { return 0; }";
+        let preprocessed2 = b"# 1 \"/home/user2/project/src/main.c\"\nint main() { return 0; }";
+
+        let basedirs = [
+            PathBuf::from("/home/user1/project"),
+            PathBuf::from("/home/user2/project"),
+        ];
```
mathstuf, line 1709: Can a case be added where the basedirs given to the two `hash_key` calls differ? I want to make sure that `basedirs` itself is not included in the hash.

src/compiler/preprocessor_cache.rs
```diff
@@ -634,4 +642,92 @@ mod test {
...
+        let hash2_with_basedirs = preprocessor_cache_entry_hash_key(
+            "test_digest",
+            Language::C,
+            &[],
+            &[],
+            &[],
+            &file2_path,
+            false,
+            config,
+            &[dir1.path().to_path_buf(), dir2.path().to_path_buf()],
```
mathstuf: Same here; file1 with dir1 (only) and file2 with dir2 (only) should also match.

src/server.rs
```diff
@@ -1932,18 +1933,25 @@ impl ServerInfo {
...
             cache_location = storage.location();
             use_preprocessor_cache_mode = storage
                 .preprocessor_cache_mode_config()
                 .use_preprocessor_cache_mode;
             (cache_size, max_cache_size) =
                 futures::try_join!(storage.current_size(), storage.max_size())?;
+            basedirs = storage
+                .basedirs()
+                .iter()
+                .map(|p| p.to_string_lossy().to_string())
```
mathstuf: Instead of using strings, use `PathBuf` and its `.display()` method when printing it out. Alternatively, detect when lossy conversions have happened and decorate them with something like "(mangled through Unicode)" to non-UTF-8 paths.

And while we're at it, can a non-Unicode path test case be added to make sure it plumbs through properly?

src/config.rs
```diff
@@ -989,32 +1012,57 @@ impl Config {
...
 
-        let EnvConfig { cache } = env_conf;
+        let EnvConfig {
+            cache,
+            basedirs: env_basedirs,
+        } = env_conf;
         conf_caches.merge(cache);
 
+        // Environment variable takes precedence over file config
+        let basedirs_raw = if !env_basedirs.is_empty() {
```
mathstuf: Shouldn't it be possible to have `SCCACHE_BASEDIRS=` to ignore the configuration file basedirs?

src/config.rs
```diff
@@ -1335,10 +1385,361 @@ fn config_overrides() {
...
+
+    let file_conf = FileConfig {
+        cache: Default::default(),
+        dist: Default::default(),
+        server_startup_timeout_ms: None,
+        basedirs: vec![],
+    };
+
+    let config = Config::from_env_and_file_configs(env_conf, file_conf).unwrap();
+    assert_eq!(config.basedirs, Vec::<PathBuf>::new());
```
mathstuf:
```suggestion
    assert!(config.basedirs.is_empty());
```

Seems cleaner to me. Same below

src/config.rs
```diff
@@ -1335,10 +1385,361 @@ fn config_overrides() {
...
+
+        [dist]
+    "#;
+
+    let config: FileConfig = toml::from_str(toml).unwrap();
+    assert_eq!(config.basedirs, Vec::<PathBuf>::new());
+}
+
+#[test]
+#[serial]
```
mathstuf: Can the reason be added here so that they're not overly-serialized? https://docs.rs/serial_test/latest/serial_test/attr.serial.html

src/config.rs
```diff
@@ -1335,10 +1385,361 @@ fn config_overrides() {
...
+#[cfg(not(target_os = "windows"))]
+fn test_env_basedirs_with_spaces() {
+    use std::path::PathBuf;
+
+    // Test that spaces around paths are trimmed
+    unsafe {
+        std::env::set_var(
+            "SCCACHE_BASEDIRS",
+            " /home/user/project : /home/user/workspace ",
```
mathstuf, line 1736: Trimming spaces is novel to me (for these path-list environment variables). Is there precedence for it you can cite?

Note that on non-Windows, paths can legitimately end with a space (but it is odd).

src/util.rs
```diff
@@ -1015,10 +1015,120 @@ pub fn num_cpus() -> usize {
...
+/// Path matching is case-insensitive to handle various filesystem behaviors and build system
+/// configurations uniformly across all operating systems. On Windows, this function also handles
+/// paths with mixed forward and backward slashes, which can occur when different build tools
+/// produce preprocessor output.
+///
+/// Only paths that start with one of the basedirs are modified. The paths are expected to be
+/// in the format found in preprocessor output (e.g., `# 1 "/path/to/file"`).
+pub fn strip_basedirs(preprocessor_output: &[u8], basedirs: &[PathBuf]) -> Vec<u8> {
+    if basedirs.is_empty() || preprocessor_output.is_empty() {
+        return preprocessor_output.to_vec();
```
mathstuf: Why re-allocate instead of returning a `Cow` if the contents are unchanged?

src/util.rs
```diff
@@ -1015,10 +1015,120 @@ pub fn num_cpus() -> usize {
...
+    // Find all potential matches for each basedir using fast substring search
+    // Store as (position, length) sorted by position
+    let mut matches: Vec<(usize, usize)> = Vec::new();
+    #[cfg(target_os = "windows")]
+    let preprocessor_output = &normalize_path(preprocessor_output);
+
+    for basedir_path in basedirs.iter() {
+        let basedir_str = basedir_path.to_string_lossy();
+        let basedir = basedir_str
+            .trim_end_matches(|c| c == '/' || c == '\\')
```
mathstuf: `\` should not be trimmed on non-Windows.

src/util.rs
```diff
@@ -1015,10 +1015,120 @@ pub fn num_cpus() -> usize {
...
+        let basedir = normalize_path(basedir);
+        // Use memchr's fast substring search
+        let finder = memchr::memmem::find_iter(preprocessor_output, &basedir);
+
+        for pos in finder {
+            // Check if this is a valid boundary (start, whitespace, quote, or '<')
+            let is_boundary = pos == 0
+                || preprocessor_output[pos - 1].is_ascii_whitespace()
+                || preprocessor_output[pos - 1] == b'"'
+                || preprocessor_output[pos - 1] == b'<';
```
mathstuf: I think this captures `#line` mentions for sure. `__FILE__` is the other concern that I don't see tests for. That can have:

```c
printf("%s\n",__FILE__); // `,` before it
;__FILE__; // `;` before it
{__FILE__} // `{` before it
const char* fname=__FILE__; // `=` before it
log(__FILE__); // `(` before it
```

Also, perhaps do `is_word_boundary(preprocessor_output.get(pos - 1))` to avoid re-indexing all the time. The optimizer *should* be able to factor it out, but it is definitely cleaner to see `prev ==` instead IMO.

src/util.rs
```diff
@@ -1015,10 +1015,120 @@ pub fn num_cpus() -> usize {
...
+
+    // Build the result in a single pass
+    let mut result = Vec::with_capacity(preprocessor_output.len());
+    let mut current_pos = 0;
+
+    for (match_pos, match_len) in filtered_matches {
+        // Copy everything before the match
+        result.extend_from_slice(&preprocessor_output[current_pos..match_pos]);
+        // Replace the basedir with "."
+        result.push(b'.');
```
mathstuf: Perhaps use `__SCCACHE_BASEDIR__` so that it doesn't collide with a reference to a relative path that also lives under a basedir?

Imagine:

```
#line "./relative_path.h"
#line "/a/basedir/match/relative_path.h"
```

I worry that this can cause false matches (though the contents are likely to be very different, it might be `__FILE__` that is more likely to collide in practice).

src/util.rs
```diff
@@ -1015,10 +1015,120 @@ pub fn num_cpus() -> usize {
...
+///             lowercases ASCII characters for consistency.
+/// This function used for two purposes:
+///     - basedir_path: already normalized by std::path::absolute,
+///       but still can mismatch the case of the actual file name
+///     - preprocessor_output: plain text
+#[cfg(target_os = "windows")]
+fn normalize_path(path: &[u8]) -> Vec<u8> {
+    path.iter()
+        .map(|&b| match b {
+            b'A'..=b'Z' => b + 'a' - 'A',
```
mathstuf:
```suggestion
            c @ b'A'..=b'Z' => c.to_ascii_lowercase(),
```

src/util.rs
Felixoid:
```
# 0 "test.c"                           ||| # 0 "/tmp/test.c"
# 0 "<built-in>"                       ||| # 0 "<built-in>"
# 0 "<command-line>"                   ||| # 0 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4 ||| # 1 "/usr/include/stdc-predef.h" 1 3 4
# 0 "<command-line>" 2                 ||| # 0 "<command-line>" 2
# 1 "test.c"                           ||| # 1 "/tmp/test.c"
# 1 "/usr/include/stdio.h" 1 3 4       ||| # 1 "/usr/include/stdio.h" 1 3 4
# 28 "/usr/include/stdio.h" 3 4        ||| # 28 "/usr/include/stdio.h" 3 4
```

I like the idea. But instead, I'd like to delete the match along with the following slash. And for that, we should actually normalize the basedirs by _adding_ a trailing slash. And it's sound OK to have a current dir `./` there by default.

src/server.rs
Felixoid: I have already almost finished refactoring by putting all the `strip_basedirs` logic into `Config::from_env_and_file_configs`, so the `basedirs` became `Vec<u8>` and won't be changed anywhere

src/config.rs
Felixoid: All of them manipulate the same variable, so `#[serial(SCCACHE_BASEDIRS)]` should work pretty well

src/util.rs
AJIOB: Maybe without any condition? Just do it for every char, not for every byte.

We can have non-english locale too

src/config.rs
Felixoid:
```
> ls -ld test\ space\ and\ ending\
drwxr-xr-x 2 felixoid felixoid 6 Dec 30 18:31 'test space and ending '
```

pretty valid, right. I think, I won't bother to trim spaces, so it will remain misconfiguration with failing to start

src/util.rs
Felixoid: https://github.com/mozilla/sccache/pull/2521#discussion_r2653493910

so, stripping `\|/` on windows, `/` on other OSes, and then adding the trailing `/`

src/util.rs
Felixoid: The question is how to make it universal and safe.

For example, if we are going to Turkish, where ı and i are different letters, that have İ and I as capital, it's already bad enough.

Cyrillic would work, I think, but it should be UTF, right?

src/util.rs
Felixoid:
```
> cat test.c
#include <stdio.h>

int main() {
    printf("Hello from %s\n",__FILE__);
    return 0;
}
> /usr/bin/gcc -x c -E ./test.c | grep './test.c'
# 0 "./test.c"
# 1 "./test.c"
# 2 "./test.c" 2
# 3 "./test.c"
    printf("Hello from %s\n","./test.c");
```

`__FILE__` looks covered

src/config.rs
Felixoid: Good question, didn't think of it.

I think so, it should clean out the cofigured value

src/util.rs
mathstuf: At that point, you need to ask the platform for its APIs to do it. The codepage in use might not be Unicode at all.

src/util.rs
mathstuf: Ah right, all expansions are going to be preceded by `"`. All good!

src/config.rs
```diff
@@ -989,32 +1014,83 @@ impl Config {
...
+                {
+                    normalize_win_path(&bytes)
+                }
+
+                #[cfg(not(target_os = "windows"))]
+                {
+                    bytes
+                }
+            };
+            basedirs.push(normalized);
```
Felixoid: The tests for trailing slashes must be implemented here for checking how it works with `strip_basedirs`

AJIOB:
Hi all.

Is this PR ready? Or we're waiting for something?

IMO, this feature is very useful to have a better cache matching

Felixoid:
I went on holiday, and I still haven't had time to address some review points. In particular:

- [x] - Clean out configured value by an empty ENV https://github.com/mozilla/sccache/pull/2521#discussion_r2653284178
- [x] - Add tests to `src/config.rs` for trailing slashes https://github.com/mozilla/sccache/pull/2521#discussion_r2654143989
- [x] - Don't trim spaces from the dir name https://github.com/mozilla/sccache/pull/2521#discussion_r2653304566
- [x] - Use Cow https://github.com/mozilla/sccache/pull/2521#discussion_r2653305982
- [x] - Proper trimming https://github.com/mozilla/sccache/pull/2521#discussion_r2653306699
- [x] - Change the stripping to just remove `basedir` from the path, because replacing it by `./` causes funny special effects https://github.com/mozilla/sccache/pull/2521#discussion_r2653493910
- [x] - And the worst discussion is lowercase https://github.com/mozilla/sccache/pull/2521#discussion_r2653324247. The suggested approach didn't work out of the box, and I don't know how to fix it. *upd*: I mean, generally fix the approach. Something will still remain broken, I am afraid.

src/util.rs
AJIOB: Looks like we can use https://docs.rs/typed-path/latest/typed_path/ trait to do almost all magic we need

src/server.rs
Felixoid: This was addressed

src/config.rs
Felixoid: Adressed in https://github.com/mozilla/sccache/pull/2521/commits/56272a15b6a7b9dca054a619791f3c30bfd800b2 -> https://github.com/mozilla/sccache/pull/2521/commits/e5b74ad4a38f8436156123fdeafbc9add67f4ecf

src/compiler/c.rs
Felixoid: this and https://github.com/mozilla/sccache/pull/2521#discussion_r2653277183 are adressed in https://github.com/mozilla/sccache/pull/2521/commits/8f0b7eb81aa8d6a17fdba4cde0972450d463a2a8

src/util.rs
Felixoid: https://github.com/mozilla/sccache/pull/2521/commits/a0437695535f4e2529829e94331ac90e98375685

src/util.rs
Felixoid: I'll use https://docs.rs/typed-path/latest/typed_path/ here.

Then, the trailing `/` will be added explicitly and completely stripped from the `preprocessor_output` instead of being replaced by `.`.

The last question is whether `./` should be treated as a special case. Because `./test.c` and `test.c` won't match

Felixoid: In https://github.com/mozilla/sccache/pull/2521/commits/87151ba8660208544080b55d5b687071bd1eea3f, I implemented this logic. The basedir is stripped completely to match `/dir/test.c` and `test.c` in the cache.

The `./test.c` is still not covered by the code.

src/util.rs
Felixoid: The typed_path is used in https://github.com/mozilla/sccache/pull/2521/commits/87151ba8660208544080b55d5b687071bd1eea3f

In https://github.com/mozilla/sccache/pull/2521/commits/9687112d92877d1a02dbe7b91eaa541aee50080d, the ascii code was expanded to proceed UTF-8 as well

src/config.rs
Felixoid: Implemented in https://github.com/mozilla/sccache/pull/2521/commits/87151ba8660208544080b55d5b687071bd1eea3f

Felixoid:
Finally, all the points are addressed. Hopefully, it's good enough.

ahartmetz:
Ah, you even thought of the preprocessor cache 👍
Slightly annoying that it requires munging the whole file (not for the first time that that happens) due to how input files end up in the output, but what can you do.

src/config.rs
```diff
@@ -989,32 +1013,84 @@ impl Config {
...
+            // normalize windows paths
+            let normalized = {
+                #[cfg(target_os = "windows")]
+                {
+                    normalize_win_path(&bytes)
+                }
+
+                #[cfg(not(target_os = "windows"))]
+                {
+                    bytes
```
AJIOB, line 1103: What if the dir already has the `/` in the end, before you adding it? Should we normalize the dir on non-win platforms too?

src/config.rs
```diff
@@ -989,32 +1013,84 @@ impl Config {
...
+                }
+            };
+            basedirs.push(normalized);
+        }
+
+        if !basedirs.is_empty() {
+            let basedirs_str: Vec<String> = basedirs
+                .iter()
+                .map(|b| String::from_utf8_lossy(b).into_owned())
+                .collect();
```
AJIOB, line 1116: nit: Do we really need to build the vector? Maybe we can just print the results of iterating?

src/util.rs
```diff
@@ -1015,6 +1016,164 @@ pub fn num_cpus() -> usize {
...
+
+    // Find all potential matches for each basedir using fast substring search
+    // Store as (position, length, basedir_idx) sorted by position
+    let mut matches: Vec<(usize, usize, usize)> = Vec::new();
+    // We must return the original preprocessor output on all platforms,
+    // so we only normalize a copy for searching.
+    #[cfg(not(target_os = "windows"))]
+    let normalized_output = preprocessor_output;
+    #[cfg(target_os = "windows")]
+    let normalized_output = &normalize_win_path(preprocessor_output);
```
AJIOB, line 1050: We may change the program output, if it contains the `__FILE__` value printing to stout. Please, add the note about that.

P.S. Also, if we have the hardcover value - it can be also replaced (such as searching hardcoded `/usr/bin` prefix) - we can break the program behavior. Does the classic ccache have the same problems?

src/util.rs
```diff
@@ -1015,6 +1016,164 @@ pub fn num_cpus() -> usize {
...
+    // Build the result in a single pass
+    let mut result = Vec::with_capacity(preprocessor_output.len());
+    let mut current_pos = 0;
+
+    for (match_pos, match_len) in filtered_matches {
+        // Copy everything before the match
+        result.extend_from_slice(&preprocessor_output[current_pos..match_pos]);
+        // Replace the basedir is removed completely, including trailing slash (it is expected, see
+        // Config::basedir)
+        current_pos = match_pos + match_len;
```
AJIOB, line 1103: nit: in docs you tell that the prefix is replaced by `./`. Looks like it's not valid anymore

src/config.rs
Felixoid: It's done in `let p_norm = p.normalize();` a few lines above, `normalize_win_path` changes slash direction and cases to lowercase

src/config.rs
Felixoid: Thought about it. Yes, it could be in `for d in basedirs_raw`, but I thought the single `debug` line is just a little bit nicer. Could be both `¯\_(ツ)_/¯`

TBH, the overhead is negligible; it's done only once.

I'll add the check for debug level

src/util.rs
Felixoid: It looks to me from the code, that the output is used only by the digest calculation in `hash_key` and `preprocessor_cache_entry_hash_key`.

For the compilation, should the compilers themselves be used?

src/util.rs
Felixoid: Thanks, I'll take another look

Felixoid: Fixed

src/config.rs
Felixoid: The logging level condition was added

src/config.rs
Felixoid: added a comment

src/util.rs
AJIOB: It's may be not true: 
https://github.com/mozilla/sccache/blob/d9d2eb6ffb71f3d1a8a4174848941bfc67b25eb0/src/compiler/c.rs#L671-L672

Felixoid: The `preprocessor_output` remains the same there.

It's defined in line
https://github.com/mozilla/sccache/blob/b0290f335ca1ca39e211927139ea9f57e64cdf6c/src/compiler/c.rs#L461

Then it's passed to `hash_key`
https://github.com/mozilla/sccache/blob/b0290f335ca1ca39e211927139ea9f57e64cdf6c/src/compiler/c.rs#L627-L638

In `hash_key` it's processed to create Cow
https://github.com/mozilla/sccache/blob/b0290f335ca1ca39e211927139ea9f57e64cdf6c/src/compiler/c.rs#L1507-L1509

https://github.com/mozilla/sccache/blob/b0290f335ca1ca39e211927139ea9f57e64cdf6c/src/util.rs#L1109

It looks like the flow is safe to me

AJIOB (approved):

Felixoid:
[Rebased](https://github.com/mozilla/sccache/compare/1d0156e9719dcf8ee7d117792f091abe49fe1f8b..31f9f6bf6ad3177245b03e7960ad9106203357d9) it on top of the current master

src/compiler/c.rs
```diff
@@ -1467,6 +1469,11 @@ static CACHED_ENV_VARS: LazyLock<HashSet<&'static OsStr>> = LazyLock::new(|| {
 });
 
 /// Compute the hash key of `compiler` compiling `preprocessor_output` with `args`.
+///
+/// If `basedirs` are provided, paths in the preprocessor output will be normalized by
+/// stripping the longest matching basedir prefix. This enables cache hits across different
+/// absolute paths (similar to ccache's CCACHE_BASEDIR).
+#[allow(clippy::too_many_arguments)]
```
mathstuf: Something for a new PR, but a `Builder` pattern to deal with these arguments looks like a nice enhancement. The (test) call sites being a zoo of `&[]` and unlabeled `bool` literals isn't that nice IMO.

README.md
```diff
@@ -320,12 +320,49 @@ This is most useful when using sccache for Rust compilation, as rustc supports u
...
+
+Path matching is **case-insensitive** on Windows and **case-sensitive** on other operating systems.
+
+This is similar to ccache's `CCACHE_BASEDIR` and helps when:
+* Building the same project from different directories
+* Sharing cache between CI jobs with different checkout paths
+* Multiple developers working with different username paths
+* Working with multiple project checkouts simultaneously
+
+**Note:** Only absolute paths are supported. Relative paths will prevent server from start.
```
mathstuf:
```suggestion
**Note:** Only absolute paths are supported. Relative paths will prevent server from starting.
```

mathstuf (approved):
Looks good to me. Just one nit and a possible followup.

Thanks!

README.md
Felixoid: I thought of a special case for `.` directory.

As I said in https://github.com/mozilla/sccache/pull/2521#discussion_r2677698365, `./test.c` and `test.c` are different cases. And with a current approach, it's literally impossible to handle this `./`

WDYT, should this special case be added to https://github.com/mozilla/sccache/pull/2521/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR1051-R1053 ?

My concern is that it can be TOO broad.

Felixoid: Checked, the `Utf8TypedPathBuf::from("./").normalize()` returns empty string. (any of `${empty_string}`, `.`, or `./src/..`)

Checked a strange pattern like `./src/../test.c`, and it's returned as is:

```
> gcc -E ./src/../test.c | tail
# 973 "/usr/include/stdio.h" 3 4

# 2 "./src/../test.c" 2


# 3 "./src/../test.c"
int main() {
    printf("Hello from %s\n","./src/../test.c");
    return 0;
}
```

Let's behavior remain unchanged: garbage in -> garbage out. Somebody could expect `./src/..` to clean out, but in the best effort, we could convert it to `./`.

I'll leave it as room for future improvements.

src/compiler/c.rs
Felixoid: I'll take a look, maybe it's worth spending some time in the current PR to get the tests nicer.

README.md
mathstuf: I think this was in response to the wrong comment?

Felixoid: It's related to `Only absolute paths are supported` partially.

First, I used the opportunity to discuss a special case for `.`. But after some experiments, I found it's a bad idea.

src/compiler/c.rs
Felixoid: Constructor + optional fields fit it slightly better, imho. `compiler_digest, language, arguments, preprocessor_output` are used everywhere in tests, with `extra_hashes, env_vars, plusplus, basedirs` as optional. Builder looks like an overhead in this case.

Felixoid:
@sylvestre, it's ready for merge, I think

Felixoid:
Another rebase:
- The COS backend was added
- Added integration tests for basedirs for each already tested backend

Felixoid:
I am pretty sure I didn't touch anything related to the [TimeMacroFinder::find_time_macros](https://codspeed.io/mozilla/sccache/branches/Felixoid%3Aadd-basedir-configuration?uri=benches%2Fsccache_bench.rs%3A%3Atime_macro_finder_chunked&runnerMode=Simulation&utm_source=github&utm_medium=check&utm_content=benchmark&page=1) in this PR

upd: yes, it's the same in another [PR](https://codspeed.io/mozilla/sccache/branches/avikivity%3Adisable-ccache?uri=benches%2Fsccache_bench.rs%3A%3Atime_macro_finder_chunked&runnerMode=Instrumentation&utm_source=github&utm_medium=comment&utm_content=benchmark&page=1)

sylvestre:
bravo for the hard work!

PR #2524, opened 2025-12-20
Avoid double-caching when ccache is installed in PATH

### LINKED ISSUE #2519: Unwanted double-caching when ccache is also installed
If ccache is installed and has /usr/lib64/ccache in $PATH, then `sccache gcc ...` will actually call /usr/lib64/ccache/gcc, which is in fact ccache, resulting in double-caching.

It would be easier for users migrating from ccache (or having some projects using ccache and some projects using sccache) if sccache automatically set the CCACHE_DISABLE environment variable before invoking the compiler.

avikivity:
On Linux, ccache is typically installed in $PATH as /usr/lib64/ccache/g++
or similar. If we keep it there, all compilations will be cached by both
ccache and sccache.

While the user could easily disable ccache with CCACHE_DISABLE, it's
reasonable to assume many will forget and will have their disk space
doubly consumed by both caches. Better to recognize this and disable
ccache under sccache.

This patch does this resolving the input compiler name to an absolute
path, with these rules:
 - if a directory in PATH contains a component `ccache`, it is skipped
 - if the compiler exists in a directory in PATH, but is a symlink
   to a target `ccache`, it is skipped

These two tests will catch most ccache masquerade directories.

Fixes https://github.com/mozilla/sccache/issues/2519

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -35,7 +35,7 @@ use crate::dist::pkg;
 use crate::lru_disk_cache;
 use crate::mock_command::{CommandChild, CommandCreatorSync, RunCommand, exit_status};
 use crate::server;
-use crate::util::{fmt_duration_as_secs, run_input_output};
+use crate::util::{fmt_duration_as_secs, resolve_compiler_avoiding_ccache, run_input_output};
 use crate::{counted_array, dist};
 use async_trait::async_trait;
 use filetime::FileTime;
@@ -194,7 +194,9 @@ impl CompileCommandImpl for SingleCompileCommand {
             env_vars,
             cwd,
         } = self;
-        let mut cmd = creator.clone().new_command_sync(executable);
+        // Resolve compiler avoiding ccache wrappers to prevent double-caching.
+        let resolved_executable = resolve_compiler_avoiding_ccache(executable, env_vars);
+        let mut cmd = creator.clone().new_command_sync(&resolved_executable);
         cmd.args(arguments)
             .env_clear()
             .envs(env_vars.clone())
@@ -1692,8 +1694,11 @@ compiler_version=__VERSION__
     .to_vec();
     let (tempdir, src) = write_temp_file(&pool, "testfile.c".as_ref(), test).await?;
 
+    // Resolve compiler avoiding ccache wrappers to prevent double-caching.
     let executable = executable.as_ref();
-    let mut cmd = creator.clone().new_command_sync(executable);
+    let resolved_executable = resolve_compiler_avoiding_ccache(executable, &env);
+
+    let mut cmd = creator.clone().new_command_sync(&resolved_executable);
     cmd.stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .envs(env.iter().map(|s| (&s.0, &s.1)));
@@ -1734,7 +1739,7 @@ compiler_version=__VERSION__
         }
     });
     if let Some(kind) = lines.next() {
-        let executable = executable.to_owned();
+        let executable = resolved_executable;
         let version = lines
             .next()
             // In case the compiler didn't expand the macro.
--- a/src/util.rs
+++ b/src/util.rs
@@ -1174,10 +1174,181 @@ pub fn normalize_win_path(path: &[u8]) -> Vec<u8> {
     result
 }
 
+/// Resolve the compiler executable, avoiding ccache wrappers.
+///
+/// This function handles scenarios where ccache might interfere:
+/// 1. PATH contains directories like /usr/lib/ccache or /usr/lib64/ccache
+///    which contain wrapper scripts that would call ccache
+/// 2. The compiler executable itself is a symlink to ccache
+///
+/// Searches PATH for the compiler name, skipping any candidate whose path
+/// contains "ccache" or that resolves (via symlink) to ccache.
+/// Returns the absolute path to the first valid match, or the original
+/// executable if no valid match is found.
+///
+/// If the executable is already an absolute path, it is returned as-is
+/// without searching PATH.
+pub fn resolve_compiler_avoiding_ccache(
+    executable: &Path,
+    env_vars: &[(OsString, OsString)],
+) -> PathBuf {
+    // If the path is absolute, return it as-is without searching PATH
+    if executable.is_absolute() {
+        return executable.to_path_buf();
+    }
+
+    const CCACHE_DIR_COMPONENT: &str = "ccache";
+
+    // Helper to check if a path points to ccache (via symlink resolution)
+    let resolves_to_ccache = |path: &Path| -> bool {
+        std::fs::canonicalize(path)
+            .ok()
+            .and_then(|canonical| canonical.file_name().map(|n| n.to_os_string()))
+            .map(|name| {
+                let name_lower = name.to_string_lossy().to_lowercase();
+                name_lower == "ccache" || name_lower.starts_with("ccache.")
+            })
+            .unwrap_or(false)
+    };
+
+    // Helper to check if a path contains a ccache directory component
+    let path_contains_ccache = |path: &Path| -> bool {
+        path.components()
+            .any(|c| c.as_os_str() == CCACHE_DIR_COMPONENT)
+    };
+
+    // Get the compiler name to search for
+    let compiler_name = match executable.file_name() {
+        Some(name) => name,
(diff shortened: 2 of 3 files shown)


avikivity:
update: fix rustfmt error

avikivity:
@sylvestre please review

avikivity:
@sylvestre can I help in any way to push this along?

src/util.rs
```diff
@@ -1015,11 +1015,106 @@ pub fn num_cpus() -> usize {
...
+/// Filter out ccache directories from PATH environment variable.
+///
+/// This prevents double-caching when ccache is installed and has
+/// /usr/lib/ccache or /usr/lib64/ccache in PATH, since those directories
+/// contain wrapper scripts that would call ccache.
+pub fn filter_ccache_from_path(env_vars: Vec<(OsString, OsString)>) -> Vec<(OsString, OsString)> {
+    use std::env;
+    use std::path::Path;
+
+    const CCACHE_DIRS: &[&str] = &["/usr/lib/ccache", "/usr/lib64/ccache"];
```
sylvestre: The CCACHE_DIRS list only contains Linux paths. Consider adding macOS Homebrew paths (/usr/local/opt/ccache/libexec, /opt/homebrew/opt/ccache/libexec) or adding a comment explaining why they're omitted

src/util.rs
```diff
@@ -1015,11 +1015,106 @@ pub fn num_cpus() -> usize {
     std::thread::available_parallelism().map_or(1, std::num::NonZeroUsize::get)
 }
 
+/// Filter out ccache directories from PATH environment variable.
+///
+/// This prevents double-caching when ccache is installed and has
+/// /usr/lib/ccache or /usr/lib64/ccache in PATH, since those directories
+/// contain wrapper scripts that would call ccache.
+pub fn filter_ccache_from_path(env_vars: Vec<(OsString, OsString)>) -> Vec<(OsString, OsString)> {
```
sylvestre: Function allocates a new Vec even when no ccache dirs are present in PATH. Consider short-circuiting to return the input Vec unchanged when filtering is not needed

src/util.rs
sylvestre: also, it would be nice to be able to configure things somewhere

src/compiler/compiler.rs
```diff
@@ -194,10 +194,13 @@ impl CompileCommandImpl for SingleCompileCommand {
             env_vars,
             cwd,
         } = self;
+        // Filter out ccache directories from PATH to avoid double-caching
+        // when ccache is also installed on the system.
+        let env_vars = filter_ccache_from_path(env_vars.to_vec());
```
sylvestre: Filter is only applied to SingleCompileCommand but NvccCompileCommand (in nvcc.rs) also implements CompileCommandImpl and may need the same treatment it is intentional ?

sylvestre:
sorry for the latency

src/util.rs
avikivity: We can generalize it to any path that contains a component ccache.

src/util.rs
avikivity: Sure

src/compiler/compiler.rs
avikivity: Not intentional, I'll take a look.

src/util.rs
avikivity: About configuration, is it still needed if we just filter out directories with ccache components?

avikivity:
v2:
 - filter out any directory with a ccache component, so it also works on MacOS
 - avoid allocations if no ccache directories are present
 - apply to detect_c_compiler
 - apply to nvcc

avikivity:
v3:
 - apply rustfmt

avikivity:
v4:
 - use slice instead of Vec

avikivity:
I believe the performance regression report is spurious, the code doesn't touch the macro finder code.

ahartmetz:
So I'm not the maintainer here and this is merely a suggestion. I'm a bit uncomfortable with the brittleness of this approach to recognizing ccache. An additional, somewhat - perhaps less - brittle heuristic that could be added on POSIX platforms is to recognize when gcc, g++, clang, clang++ [...] is a symlink to a binary called "ccache". Assuming that ccache works like icecream and sccache (which btw supports symlinks compiler -> sccache now).

avikivity:
> So I'm not the maintainer here and this is merely a suggestion. I'm a bit uncomfortable with the brittleness of this approach to recognizing ccache. An additional, somewhat - perhaps less - brittle heuristic that could be added on POSIX platforms is to recognize when gcc, g++, clang, clang++ [...] is a symlink to a binary called "ccache". Assuming that ccache works like icecream and sccache (which btw supports symlinks compiler -> sccache now).

I believe it's unnecessary - it's unlikely people will have directories in path named ccache that host the real compilers instead of ccache masquerades.

I'm happy to make that change if the maintainers think it is necessary though.

ahartmetz:
> I believe it's unnecessary - it's unlikely people will have directories in path named ccache that host the real compilers instead of ccache masquerades.
> 
> I'm happy to make that change if the maintainers think it is necessary though.

I don't mean ANDing the two conditions, I mean ORing. The false positive rate of a compiler binary linking to a binary called ccache, which is actually not ccache, should be negligible.

avikivity:
> > > So I'm not the maintainer here and this is merely a suggestion. I'm a bit uncomfortable with the brittleness of this approach to recognizing ccache. An additional, somewhat - perhaps less - brittle heuristic that could be added on POSIX platforms is to recognize when gcc, g++, clang, clang++ [...] is a symlink to a binary called "ccache". Assuming that ccache works like icecream and sccache (which btw supports symlinks compiler -> sccache now).
> > 
> > 
> > I believe it's unnecessary - it's unlikely people will have directories in path named ccache that host the real compilers instead of ccache masquerades.
> > I'm happy to make that change if the maintainers think it is necessary though.
> 
> I don't mean ANDing the two conditions, I mean ORing

To clarify: IF the directory contains a path component ccache, OR if the looked-up compiler is a symlink to something named ccache, THEN skip this compiler executable?

Makes sense.

ahartmetz:
> > I don't mean ANDing the two conditions, I mean ORing
> 
> To clarify: IF the directory contains a path component ccache, OR if the looked-up compiler is a symlink to something named ccache, THEN skip this compiler executable?
> 
> Makes sense.

That's what I meant, yes

ahartmetz:
(Background: I MAY have had a distributed compiler, which I compiled myself, installed in [obscure]/local/bin at some point, but I have never renamed the distributed compiler binary to something else than its usual name - which would probably break something because it avoids calling itself by checking its binary name)
Edit: on second thought, I think such tools usually just eliminate their own binary name / path / inode, whatever it may be, from the compiler search - but still, the ccache / icecream / sccache binary is usually called exactly that, with the exception of sccache, which required hardlinks until recently.

On third thought, what the hell was I thinking, if it's not called gcc or g++ etc, it won't be found as a compiler at all, so you'd need to resolve the symlink *to* ccache, which is... doable but where this maybe gets a bit "too much".

avikivity:
update:
 - rebased
 - changed approach - the utility function resolves relative compiler names to absolute paths, skipping over directories with `ccache` components or files that as symlinks that resolve to `ccache`.

avikivity:
update:
 - applied rustfmt

src/util.rs
```diff
@@ -1174,10 +1174,181 @@ pub fn normalize_win_path(path: &[u8]) -> Vec<u8> {
...
+) -> PathBuf {
+    // If the path is absolute, return it as-is without searching PATH
+    if executable.is_absolute() {
+        return executable.to_path_buf();
+    }
+
+    const CCACHE_DIR_COMPONENT: &str = "ccache";
+
+    // Helper to check if a path points to ccache (via symlink resolution)
+    let resolves_to_ccache = |path: &Path| -> bool {
```
ahartmetz, line 1203: Thanks, nice job!

avikivity:
@sylvestre the gentlest of pings

avikivity:
@sylvestre how can we move this forward?

sylvestre:
Sorry

iTrooz:
> To clarify: IF the directory contains a path component ccache, OR if the looked-up compiler is a symlink to something named ccache, THEN skip this compiler executable?

Is there really a case when checking what the symlink resolves to wouldn't be enough ? I can't think of any such cases

Issue #2303, opened 2024-12-25
`test_rlib_dep_reader_call` asserts `env::var("CARGO_HOME").is_ok()` but it's not necessarily true


nc7s:
.. not for Debian shipped cargo, apparently: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1090040

Remotely related to https://github.com/mozilla/sccache/pull/2298.

A patch using an unqualified `Path::from("rustc")` works ™, though is probably not _good_.

PR #2527, opened 2025-12-23
Fix failing test_rlib_dep_reader_call for omit CARGO_HOME

Felixoid:
Fixes https://github.com/mozilla/sccache/issues/2303

If the `CARGO_HOME` is empty, but `rustc` is installed globally, it will be used in the `compiler::rust::test::test_rlib_dep_reader_call`.

Now, the test fails as:

```
---- compiler::rust::test::test_rlib_dep_reader_call stdout ----

thread 'compiler::rust::test::test_rlib_dep_reader_call' panicked at src/compiler/rust.rs:3321:33:
called `Result::unwrap()` on an `Err` value: No such file or directory (os error 2)
```

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -3308,18 +3308,15 @@ proc_macro false
             env_vars.push(("RUSTUP_HOME".into(), rustup_home));
         }
 
-        // Try to find rustc from CARGO_HOME first, then fall back to `which`
-        let rustc_path = if let Ok(cargo_home) = std::env::var("CARGO_HOME") {
-            let mut path = PathBuf::from(cargo_home);
-            path.push("bin");
-            path.push("rustc");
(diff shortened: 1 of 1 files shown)


sylvestre:
is it possible to add a test to make sure it doesn't regress? thanks

Felixoid:
I don't know how to do it. The test is environment-dependent, so what should the regression test do?

Neither of `rustc` in `$PATH` nor `$CARGO_home/bin/rustc` looks mandatory to me. It's unclear how the regression can be caught.

Felixoid:
I added the comment on why the code looks like this. It's the best I can think of.

Checking if `rustc` exists when `CARGO_HOME` is not set would fail for environments where there's no globally installed rust. And vise-versa as well, that's why this fallback is actually necessary.

PR #2528, opened 2025-12-24
Ignore -parallel-jobs Clang flags used in ROCm/HIP

GZGavinZhao:
This is similar to NVCC's [`--threads`](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#threads-number-t) flag which only controls how many threads is used to compile each GPU architecture in parallel. It doesn't affect the resulting build artifact.


GZGavinZhao:
Wrong base branch, oops.

Issue #2537, opened 2026-01-03
Add support for asm sources support


AJIOB:
Similar to the ccache, I suggest to add the assembly language support.

Sometimes assembly language supports C preprocessing too.

See https://ccache.dev/platform-compiler-language-support.html, **Source code languages** part

PR #2538, opened 2026-01-04
Add support for C preprocessor output

AJIOB:


src/compiler/c.rs
```diff
@@ -431,6 +433,14 @@ where
                 );
             }
 
+            if !support_preprocessing && use_preprocessor_cache_mode {
+                debug!(
+                    "parse_arguments: Disabling preprocessor cache because language {} does not support preprocessing",
```
sylvestre: debug message says 'parse_arguments' but this code is in generate_hash_key(), is that expected?

AJIOB: Yes, just to keep consistent with the few rows before.

I can change all debug entries with invalid prefix to be consistent, but externally it uses via `parse_arguments` API

ahartmetz (changes requested):
The commit message should either say "preprocessor output" or "preprocessed input". Preprocessor output is sccache input.

ahartmetz:
By the way, what's the use case for this? Would be nice to mention it somewhere. When I see this, I think "Huh, I guess this could make sense, but I've compiled a ton of software and never had a need for it".
AFAICS, not having this feature would at worst cause a minor waste of time and disk space in the rare case that you are compiling already preprocessed code.

AJIOB:
> By the way, what's the use case for this? Would be nice to mention it somewhere. When I see this, I think "Huh, I guess this could make sense, but I've compiled a ton of software and never had a need for it". AFAICS, not having this feature would at worst cause a minor waste of time and disk space in the rare case that you are compiling already preprocessed code.

I have 2 reasons to do it:
1. Support asm files (that clang/gcc do natively, see #2537) - some configure scripts uses the same binaries for C and asm. And some asm files (.s) MUSTN'T be preprocessed
2. Fix TODOs

The C preprocessed files can be easier verified,  that the arch-dependent asm files.

Are these reasons enough?

ahartmetz:
> Are these reasons enough?

Yes, thanks. I see the TODOs now. Would it be possible to add .s (and maybe .S, I know that one does need preprocessing) support in the same branch then, or is that a larger change that you don't want to make yet?

AJIOB:
I want to do it as a dedicated PR: need to test it additionally

src/compiler/compiler.rs
```diff
@@ -277,19 +285,33 @@ impl Language {
             Language::Hip => "hip",
         }
     }
+
+    pub fn support_preprocessing(self) -> bool {
```
ahartmetz: should be called supports_preprocessing (but see below)

src/compiler/compiler.rs
```diff
@@ -277,19 +285,33 @@ impl Language {
...
+        !matches!(
```
ahartmetz, line 290: It seems to me that, technically, at least Ptx and Cubin also don't support preprocessing, and Rust maybe neither in the sense that matters here. That information isn't used, but still... Which makes me think: is this function answering the right question? Maybe the right question is "Should this kind of source file run through the C preprocessor"? - so needs_c_preprocessing() or something...

AJIOB: I don't know about PTX and Cubin - keep them as is - but I agree with renaming and excluding Rust from supported list

AJIOB: P.S. as I can see in docs, PTX supports cpp: https://docs.nvidia.com/cuda/parallel-thread-execution/#source-format

P.P.S. Technically,  PTX is also an assembler

ahartmetz: > I don't know about PTX and Cubin - keep them as is - but I agree with renaming and excluding Rust from supported list

Thanks, that should be fine.

src/compiler/c.rs
```diff
@@ -448,156 +456,167 @@ where
...
+                preprocessor_result.stdout.len()
+            );
 
-        trace!(
-            "[{}]: Preprocessor output is {} bytes",
-            self.parsed_args.output_pretty(),
-            preprocessor_result.stdout.len()
-        );
+            (preprocessor_result.stdout, include_files)
+        } else {
```
ahartmetz, line 613: You could move this to the beginning (followed by else if) and prevent a lot of "chaff" whitespace changes and a level of indentation. Yes it's not the common case, but OTOH, having a trivial implementation first is helpful to see what the full version is essentially doing.

ahartmetz:
This looks OK to me, but I feel too noob at Rust (or sccache for that matter) to approve it - at least for now. I'll change my mind if nobody else steps up for a several weeks.

src/compiler/c.rs
AJIOB: Do you suggest adding next parts duplication? Or just swap if & else parts between each other?

P.S. You can see the difference without spaces on the github UI

sylvestre:
lgtm 
thanks!

src/compiler/c.rs
ahartmetz: Yeah, I could only understand what's going on well enough to suggest this after I found the option. (And git diff and git blame have it, too, TIL)
So anyway...

ahartmetz: What I meant was:
```
let (preprocessor_output, include_files) = if !needs_preprocessing {
    //...
} else if let Some(preprocessor_key) = &preprocessor_key {
// ... everything else as before
```
Not sure if that syntax even works

PR #2539, opened 2026-01-05
Add GCC pipe flag ignoring support

### LINKED ISSUE #2102: Add support for GCC's `-pipe` option
As GCC said in the [official docs](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#index-pipe), it's just a note for the compiler to internally use piping instead of temporary files.

I think we should just ignore it from caching for better cache entries checking.

AJIOB:
Fixes #2102

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -172,7 +172,7 @@ use self::ArgData::*;
 
 const ARCH_FLAG: &str = "-arch";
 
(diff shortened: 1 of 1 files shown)


sylvestre:
could you please add a test ? thanks

AJIOB:
> could you please add a test ? thanks

Done, thanks. The changes are inspired by #2529

PR #2544, opened 2026-01-08
 msvc: fix detect_showincludes_prefix with MSBuild

AJIOB:
The problem usually occurs when VS_UNICODE_OUTPUT is set. This is a @AudranDoublet #1835 and @exoosh #2483 fixes with working tests: thank you!

Should fix issues #1830 and #909

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -514,6 +514,19 @@ jobs:
           $SCCACHE_EXE "$(where cl.exe)" -c "@args.rsp"
           $SCCACHE_EXE --show-stats
 
+      - name: Prepare - MSBuild support
+        shell: bash
+        working-directory: ./tests/msvc-msbuild
+        # Force stop server to drop compiler cache - need to test showInclude prefix detection
+        run: |
+          $SCCACHE_EXE --stop-server
+          mkdir -p compiler
+          cp $SCCACHE_EXE compiler/cl.exe
+
+      - name: Compile - MSBuild support
+        working-directory: ./tests/msvc-msbuild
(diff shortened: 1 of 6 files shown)


sylvestre:
could you please add a test to make sure we don't regress? thanks

AJIOB:
> could you please add a test to make sure we don't regress? thanks

@sylvestre , done

sylvestre:
> > could you please add a test to make sure we don't regress? thanks
> 
> @sylvestre , done

to make sure i understand, what you added in the CI was previously failing to build, right ?
thanks

AJIOB:
Yes, it previously fails.

src/compiler/msvc.rs
```diff
@@ -216,16 +212,20 @@ where
     let output = run_input_output(cmd, None).await?;
 
     if !output.status.success() {
-        bail!("Failed to detect showIncludes prefix")
+        warn!(
+            "Failed to detect showIncludes prefix (status: {:?})",
+            output.status.code().unwrap_or(-1)
+        )
```
sylvestre:
```suggestion
        );
```

src/compiler/msvc.rs
```diff
@@ -216,16 +212,20 @@ where
...
 
     let process::Output {
-        stdout: stdout_bytes,
+        stderr: stderr_bytes,
         ..
     } = output;
-    let stdout = from_local_codepage(&stdout_bytes)
-        .context("Failed to convert compiler stdout while detecting showIncludes prefix")?;
-    for line in stdout.lines() {
+    let output = from_local_codepage(&stderr_bytes)
```
sylvestre: can we avoid the variable shadowing here?

AJIOB: Fixed, thanks

src/compiler/msvc.rs
AJIOB: Fixed, thanks

PR #2545, opened 2026-01-09
Assembly language support

### LINKED ISSUE #2537: Add support for asm sources support

AJIOB:
Fixes #2537

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,47 @@
+name: Benchmarks
+
+# spell-checker:ignore codspeed dtolnay Swatinem
+
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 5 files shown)


sylvestre:
silly question but is there really a need for this ?

AJIOB:
> silly question but is there really a need for this ?

Yes. OpenSSL, for example, uses the same executable for building both asm and C sources. So, if we want to use the same wrapper for OpenSSL - we need to have support assembly too.

P.S. ccache also supports asm, see https://ccache.dev/platform-compiler-language-support.html

PR #2548, opened 2026-01-11
codspeed: evaluate the memory benchmarking

sylvestre:


sylvestre:
doesn't impact the product itself, automerging

PR #2557, opened 2026-01-15
Distributed compilation support for asm & preprocessor outputs

### LINKED ISSUE #2556: assembly support broken
The assembly support added in https://github.com/mozilla/sccache/pull/2545 seems to be broken, when executing distributed compilation, the `-x` argument generated for the compiler is invalid (file contents irrelevant):

```c
$ sccache cc -c test.S
sccache: Job failed on server 192.168.1.3:10600:
cc: error: language assembler-with-cpp-cpp-output not recognized
cc: error: language assembler-with-cpp-cpp-output not recognized
sccache: Compiler killed by signal 1
```

Seems to be due to this condition: https://github.com/mozilla/sccache/blob/d9d2eb6ffb71f3d1a8a4174848941bfc67b25eb0/src/compiler/gcc.rs#L915

cc @AJIOB 

TIA for looking into this!

AJIOB:
Fixes #2556

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -212,6 +212,11 @@ pub enum CompilerKind {
     Rust,
 }
 
+// Used for tests
+// Need to be consistent with `Language`
+#[allow(dead_code)]
+const EXPECTED_LANGUAGE_COUNT: usize = 21;
+
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum Language {
     AssemblerToPreprocess,
@@ -303,6 +308,28 @@ impl Language {
                 | Language::ObjectiveCxxPreprocessed
                 | Language::Rust
         )
+    }
+
+    pub fn is_c_like_header(self) -> bool {
+        matches!(
+            self,
+            Language::CHeader
+                | Language::CxxHeader
+                | Language::ObjectiveCHeader
+                | Language::ObjectiveCxxHeader
+                | Language::GenericHeader
+        )
+    }
+
+    pub fn to_c_preprocessed_language(self) -> Option<Language> {
+        match self {
+            Language::AssemblerToPreprocess => Some(Language::Assembler),
+            Language::C => Some(Language::CPreprocessed),
+            Language::Cxx => Some(Language::CxxPreprocessed),
+            Language::ObjectiveC => Some(Language::ObjectiveCPreprocessed),
+            Language::ObjectiveCxx => Some(Language::ObjectiveCxxPreprocessed),
+            _ => None,
+        }
     }
 
     /// Common implementation for GCC and Clang language argument mapping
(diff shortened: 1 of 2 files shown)


src/compiler/gcc.rs
```diff
@@ -909,10 +909,17 @@ where
...
                 language_to_arg(parsed_args.language).map(|lang| lang.into());
             if !rewrite_includes_only {
-                match parsed_args.language {
-                    Language::C => language = Some("cpp-output".into()),
-                    Language::GenericHeader | Language::CHeader | Language::CxxHeader => {}
-                    _ => language.as_mut()?.push_str("-cpp-output"),
+                if matches!(
+                    parsed_args.language,
+                    Language::GenericHeader | Language::CHeader | Language::CxxHeader
+                ) {
```
sylvestre: The if-else chain in gcc.rs:915-926 could be simplified by inverting the header check condition to reduce nesting

src/compiler/compiler.rs
```diff
@@ -1856,13 +1868,39 @@ mod test {
...
 
+    #[test]
+    fn test_c_preprocessing_consistent() {
+        for lang in all::<Language>() {
+            let maybe_processed = lang.to_c_preprocessed_language();
+
+            if let Some(processed) = maybe_processed {
+                assert!(
+                    !processed.needs_c_preprocessing(),
+                    "{:?} should not need preprocessing - it is a reasult of {:?} processing",
```
sylvestre:
```suggestion
                    "{:?} should not need preprocessing - it is a result of {:?} processing",
```

ahartmetz:
Please, like last time: "preprocessor output" or "preprocessed input". Be nice to the people who are going to read what you write.

AJIOB:
> Please, like last time: "preprocessor output" or "preprocessed input". Be nice to the people who are going to read what you write.

Oops, thanks again

AJIOB:
Hi @sylvestre ,

Is it OK for you?

Looks like I've finalized it

src/compiler/compiler.rs
```diff
@@ -1859,13 +1871,39 @@ mod test {
...
+                    "{:?} should not need preprocessing - it is a result of {:?} processing",
+                    processed,
+                    lang
+                );
+            }
+
+            if !lang.needs_c_preprocessing() {
+                assert!(
+                    maybe_processed.is_none(),
+                    "{:?} should not be processed, but it is produce the {:?} as a C preprocessing result",
```
sylvestre:
```suggestion
                    "{:?} should not be processed, but it produces the {:?} as a C preprocessing result",
```

Cargo.toml
```diff
@@ -42,6 +42,7 @@ chrono = "0.4"
 clap = { version = "4.5.13", features = ["derive", "env", "wrap_help"] }
 directories = "6.0"
 encoding_rs = "0.8"
+enum-iterator = "2.3.0"
```
sylvestre: is it really necessary to add this dep just for this ?

src/compiler/gcc.rs
```diff
@@ -911,12 +911,17 @@ where
...
-                    Language::GenericHeader | Language::CHeader | Language::CxxHeader => {}
-                    _ => language.as_mut()?.push_str("-cpp-output"),
-                }
+            if rewrite_includes_only
+                || matches!(
+                    parsed_args.language,
+                    Language::GenericHeader | Language::CHeader | Language::CxxHeader
+                )
+            {
+                // Nothing to do
```
sylvestre: please update the code to avoid the nothing to do

Cargo.toml
AJIOB: I can't find another way to do foreach all enum elements.

So the answer is yes, it's necessary

AJIOB:
@sylvestre , is it OK?

Or you prefer to manually list all possible enum values twice?

AJIOB:
@sylvestre , check,  please,  is this PR OK for you.

src/compiler/compiler.rs
```diff
@@ -1866,6 +1888,64 @@ mod test {
...
+                    lang
+                );
+            }
+
+            if !lang.needs_c_preprocessing() {
+                assert!(
+                    maybe_processed.is_none(),
+                    "{:?} should not be processed, but it produces the {:?} as a C preprocessing result",
+                    lang,
+                    maybe_processed.unwrap()
```
sylvestre: can we avoid the unwrap here?
Should be removed since Debug format handles Option directly.

src/compiler/compiler.rs
```diff
@@ -1866,6 +1888,64 @@ mod test {
     use test_case::test_case;
     use tokio::runtime::Runtime;
 
+    #[test]
+    fn test_c_preprocessing_consistent() {
+        // Must contain all possible values from `Language`.
+        // Cannot use `enum_iterator` because it adds macro to the real enum,
+        // and we don't want to add additional dependencies.
+        let languages = vec![
```
sylvestre, line 1901: Add a `const EXPECTED_LANGUAGE_COUNT: usize = 21` and assert against it to catch when languages are added/removed.

src/compiler/gcc.rs
```diff
@@ -911,11 +911,11 @@ where
             // https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Overall-Options.html
             let mut language: Option<String> =
                 language_to_arg(parsed_args.language).map(|lang| lang.into());
-            if !rewrite_includes_only {
-                match parsed_args.language {
-                    Language::C => language = Some("cpp-output".into()),
-                    Language::GenericHeader | Language::CHeader | Language::CxxHeader => {}
-                    _ => language.as_mut()?.push_str("-cpp-output"),
+            if !rewrite_includes_only && !parsed_args.language.is_c_like_header() {
```
sylvestre: the nested if can be simplified by inverting the condition for better readability.

sylvestre:
clippy isn't happy

AJIOB:
Looks like all is good now

PR #2559, opened 2026-01-16
Add Objective C Header for consistency

AJIOB:
We have C, C++ and Objective C++ headers, but not the Objective C ones.

Add it for consistency


AJIOB:
Hi @sylvestre ,

Looks like we have broken s390x build

PR #2564, opened 2026-01-18
Integration tests

Felixoid:
Attempt to rearrange integration tests to make them run locally.

Now, they can run as `make -C tests/integration test` from the repo root.

GHA definition was reduced significantly.

`tests/test_crate` is used for cargo tests; locally, all tests are finished in around one minute in the case of the second run.

### DIFF
--- a/tests/integration/scripts/test-backend.sh
+++ b/tests/integration/scripts/test-backend.sh
@@ -19,7 +19,7 @@ TEST_ENV_VAR="test_value_$(date +%s)" && export TEST_ENV_VAR
 cargo clean
 cargo build
 
-# Use pythong to parse JSON output, installing jq each time is slow and unreliable
+# Use python to parse JSON output, installing jq each time is slow and unreliable
 # The difference python vs jq is 55 seconds vs 90 seconds
 echo "Checking stats after first build..."
 STATS_JSON=$("$SCCACHE" --show-stats --stats-format=json)


sylvestre:
cool stuff
would it be possible to split it into different PR ?

Felixoid:
It was wrong HEAD on the first push. But now it's already a different PR.

Only integration tests related changes are here

sylvestre:
sure but https://github.com/mozilla/sccache/pull/2564/changes/dacf2a6b9dd3e3133ac80f49af4ae9c8718b080e
could be done a sep PR, no ?

Felixoid:
Yeah, ok. I feel it's a part of preparation, but it could be done in a dedicated PR

#2565

sylvestre:
it makes reviews easier and then faster :)

tests/integration/Makefile
```diff
@@ -0,0 +1,116 @@
...
+	@echo "Building sccache in Docker..."
+	@docker run --rm \
+		-v $(PROJECT_ROOT):/workspace \
+		-v integration-cargo-cache:/usr/local/cargo/registry \
+		-w /workspace \
+		-e OPENSSL_NO_VENDOR=1 \
+		rust:latest \
+		bash -c "apt-get update -qq && apt-get install -y -qq libssl-dev pkg-config && \
+		         cargo build --all-features && \
+		         chown -R $$(id -u):$$(id -g) target"
```
AJIOB: nit: use `UID` & `GID`, defined in the top

tests/integration/scripts/test-gcc.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+set -euo pipefail
+
+SCCACHE="${SCCACHE_PATH:-/sccache/target/debug/sccache}"
+TEST_FILE="/workspace/tests/test_clang_multicall.c"
```
AJIOB: nit: is it OK to hardcode the `/workspace` prefix? Maybe it should be set/overridden from env or arg

Felixoid: Probably, it should be `/sccache`.

Then, the mounting scheme can be simplified a little bit

Felixoid: I don't see a reason to change the hardcoded value to an ENV, though. The scripts work only in docker-compose enviromnent anyway. And all paths are hardcoded there as well.

tests/integration/scripts/test-backend.sh
```diff
@@ -0,0 +1,53 @@
...
+cp -r /sccache/tests/test-crate /build/
+
+cd /build/test-crate
+
+echo "Build 1: Cache miss expected"
+TEST_ENV_VAR="test_value_$(date +%s)" && export TEST_ENV_VAR
+cargo clean
+cargo build
+
+# Use pythong to parse JSON output, installing jq each time is slow and unreliable
```
sylvestre:
```suggestion
# Use python to parse JSON output, installing jq each time is slow and unreliable
```

sylvestre:
thanks

PR #2565, opened 2026-01-19
Move integration tests related stuff to subdir

Felixoid:
A dedicated PR, preparation for #2564


Felixoid:
https://github.com/rust-cross/rust-musl-cross/issues/212 is related to the musl build error

PR #2570, opened 2026-01-21
snap: update to core24

soumyaDghosh:


soumyaDghosh:
@sylvestre Could it be possible for you to trigger a new build of sccache for the snap store?

Issue #2481, opened 2025-11-05
Split the compiler cache and preprocessed cache


xis19:
In Linux, both the compiler cache and preprocessed cache are stored in `/home/user/.cache/sccache`. The compiler cache and preprocessed cache are different, as the former only writes the file "atomically," while the latter requires random access to the data, also creating numerous temporary files.

The motivation is to store the compiled objects in S3. Since S3 does not support modifying file content (I am using [mountpoint-s3](https://github.com/awslabs/mountpoint-s3)), the preprocessor mode cannot work properly. If the preprocessed cache stays locally while the shared cache is in S3, the preprocessor mode and the shared storage can work together better.

ahartmetz:
I'm not sure if the preprocessor cache really needs to work like this, or just happens to.
In any case, you might lose some of its benefit (it only makes sense if it's much faster than preprocessing, which isn't that slow) due to relatively long round-trip times to S3, so I guess ideally, the preprocessor cache would both support a separate cache location and write out every file in one go.

xis19:
@ahartmetz  Not sure but I have read the source code, it seems isolating the preprocessor cache is not that hard, let me try a PR sometime,

xis19:
@ahartmetz  -- do you mind have a look at this change?

PR #2571, opened 2026-01-22
Extract the preprocessor cache out from disk cache

xis19:
With this patch, the compiler's output can be shared among the team while still benefiting from hashed preprocessed source files. The details are documented as commit message in 0a87c6121ff87911790a05faae89990afb837254, [Split Storage trait into Storage and PreprocessorCacheStorage traits](https://github.com/mozilla/sccache/commit/0a87c6121ff87911790a05faae89990afb837254).

This PR is to address #2481 .


sylvestre:
please run rustfmt

src/cache/cache.rs
```diff
@@ -224,6 +156,10 @@ impl Storage for opendal::Operator {
 #[allow(clippy::cognitive_complexity)] // TODO simplify!
 pub fn storage_from_config(
     config: &Config,
+fn get_preprocessor_cache_storage(config: &Config) -> Result<Arc<dyn PreprocessorCacheStorage>> {
```
ahartmetz: Commit splitting accident?

src/cache/cache.rs
```diff
@@ -406,11 +342,20 @@ pub fn storage_from_config(
         preprocessor_cache_mode_config,
         rw_mode,
     )))
+pub fn get_storage_from_config(
```
ahartmetz, line 236: That also doesn't look right

ahartmetz (changes requested):
It is good that you explain the change in the commit message, but it needs to go into the documentation and the examples there, too.

src/test/mock_preprocessor_cache.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+#[async_trait]
+impl PreprocessorCacheStorage for MockPreprocessorCacheStorage {
+    fn get_config(&self) -> &PreprocessorCacheModeConfig {
+        &self.config
+    }
+
+    // TODO Implement get_preprocessor_cache_entry and put_preprocessor_cache_entry
```
ahartmetz, line 41: Ah?

ahartmetz:
I really like most of the refactorings here - especially factoring out the preprocessor cache and moving its configuration under configuration. I'm just a bit ambivalent about the cache_io and cache_util stuff. But that's mostly because "a CacheRead" seems a bit strange as a stateful object to me, and that existed before, so OK.

Do I understand correctly that you need to manually configure the preprocessor cache dir for it to work? That would be additional mandatory boilerplate in the configuration, that's not good.

About the R/W mode, any particular reason for not using the same mode as the main cache?

src/cache/cache.rs
xis19: Yeah. That is surprising. I have re-pushed the stuff to address the issue.

xis19:
> please run rustfmt

Done

xis19:
> Do I understand correctly that you need to manually configure the preprocessor cache dir for it to work? That would be additional mandatory boilerplate in the configuration, that's not good.

I have added the code to address this. Now by default it would use the directory `${diskcache::dir}/preprocessor`
 
> About the R/W mode, any particular reason for not using the same mode as the main cache?

I would think there could be a particular situation, e.g., the main cache is shared and mounted read-only for a new user. The person still needs to generate and store hash data locally to see if the main cache has the object file already.

src/test/mock_preprocessor_cache.rs
xis19: In the test infrastructure, the `MockPreprocessorCache` is just a placeholder. The test code expects it return "cache miss" all the time, and the code is in the trait. I would expect in the future the mock code does more, but at this stage I suppose the cache miss result is sufficient?

sylvestre:
This is a huge patch, could you please split it into different PR? thanks

docs/Caching.md
```diff
@@ -48,12 +48,63 @@ See https://github.com/mozilla/sccache/blob/8567bbe2ba493153e76177c1f9a6f98cc7ba
...
+to remember, among other things, all files included by the source file. sccache also needs to recognize
+when "external factors" may change the results, such as system time if the `__TIME__` macro is used
+in a source file. How conservative sccache is about some of these external factors is configurable, see below.
+
+Preprocessor cache mode will be disabled in any of the following cases:
+
+- Not compiling C or C++
+- The configuration option is false
+- Not using GCC or Clang
+- Not using local storage for the cache
```
ahartmetz, line 78: That's not true anymore, is it? Other information in this PR indicates that maybe it'll work with caches other than local disk as long as they support random seeks - but in any case, now it always is local disk? (That's somewhat contradictory, so please make it consistent)

I am also still wondering if requiring random seek capability is very important or if it just saved three lines of code somewhere and the requirement could be lifted. Well, that's an optional improvement.

xis19: Based on my experience, random write is definitely required. To create a shared cache for our team, I tried to use mountpoint-s3 to mount a S3 bucket to $HOME/.cache/sccache. What I have observed is that, if preprocessor cache is enabled, mountpoint-s3 would fail because it does not support random write. This implies that somewhere in the sccache code, random write is being used. This is the motivation I decided to extract preprocessor cache out from disk cache.

ahartmetz: My question is not if random writes are used, but if the random writes are fundamentally necessary or a sort of frivolous requirement that could be easily lifted. You don't want to write hundreds of lines of code to work around a tiny problem that you could just fix instead.

ahartmetz: Note, making the preprocessor cache compatible with sequential-only devices wouldn't make it useless to have a feature to split the caches: the preprocessor cache is smaller than the main cache and especially benefits from fast storage, which would be reasons to keep it on a local SSD.

xis19: TBH I did not read the lru code before. Digging into the details makes feel it does not need random write feature. However, it was observed that mountpoint-s3 complains it does not support random write. Not sure if I want to dig into that project at this time though. I do not understand why this happens. My best guess at this stage is that, the LRU cache does some kind of random write (together with the tempfile), and it is done somewhere outside the current code, maybe inside some Rust internal implementation details.

sylvestre:
sorry, a lot of conflicts with a recent merge

docs/Caching.md
ahartmetz: OK, that just leaves adapting the documentation to the new (potentially) split caches reality as a thing to do.

src/test/mock_preprocessor_cache.rs
ahartmetz: Stupid question: is the thoroughness of testing the preprocess or cache higher, lower or equal to before in this branch? In my checkout of main, I don't see anything explicit in the tests, so I guess it at best kind of runs along with other cache tests?

xis19: I am assuming that the default tests would work since it includes the check of hit/miss rate.

PR #2574, opened 2026-01-26
ci: show diff for toml_format

xtqqczze:
- The `--diff` parameter isn't available via npx
- Also runs faster
- Use [`taiki-e/install-action`](https://github.com/taiki-e/install-action) to install from a binary


sylvestre:
thanks

PR #2577, opened 2026-01-28
remove unused declaration

sylvestre:


sylvestre:
found by https://github.com/Boshen/cargo-shear

PR #2581, opened 2026-01-29
Implement multiple tiers caching with fallback and backfilling

### LINKED ISSUE #30: Two-tier caching

### LINKED ISSUE #1020: [feature] Multiple cache levels

### LINKED ISSUE #2493: Combined/layered/fallback caches
I would like ability to read from more than one cache in parallel, and store new objects only to one of them.

I have a setup where I can't trust arbitrary build jobs to write to a shared cache (similar to the #845), but I can provide some baseline read-only pre-populated cache. Currently `sccache` supports a read-only mode, but having only a read-only cache is suboptimal, because objects missing from the read-only cache will remain cache misses forever. Layering another writeable cache on top of it would make it possible to have complete caches for individual build jobs, which effectively behave like pre-populated from the shared cache (without actually copying it).

It would also be useful for primitive CI cache setups that can only download and upload whole tarballs. Currently using sccache's disk cache this way pays a high cost of re-uploading the entire cache instead of only appending new entries. With layered caches, a previous cache could be only downloaded to one directory (read-only), and any newly written objects written to a different directory that will get uploaded, and an occasional job that will periodically merge the new cache into the old read-only cache.

I think it would be relatively easy to implement this in sccache by providing additional Storage implementations that wrap other storage back-ends (like https://rclone.org/combine/)

Wrapping one Storage on top of another may also make it possible to implement some transformations like #1762

I'm happy to implement this. Let me know if you're interested.

### LINKED ISSUE #2566: Time sensitive: GitHub Actions Cache rate limits (from GitHub staff)
To preserve the health of our services and reliably serve all GitHub Actions Cache users, the following rate limit will take effect immediately: 

[200 cache entry uploads per minute per repository](https://github.blog/changelog/2026-01-16-rate-limiting-for-actions-cache-entries/).

If you exceed this limit, you will receive a 429 response and an error message indicating that you have exceeded the rate limit. 

You should not retry your request until after the time specified in the `Retry-After` header has passed. If your request continues to fail due to this rate limit, implement exponential backoff between retries, and throw an error after a specific number of retries. 

This limit helps prevent abuse and denial-of-service attacks and ensures that the service remains available for all users. 

> [!IMPORTANT]
> Introduce the needed changes to your library so that the rate limit errors are retried as per the recommended schedule. 

### Thrashing

It has also come to our attention that a substantial number of repositories adopting this library (or action) are exhibiting consistent cache thrashing behavior. 

Cache thrashing occurs when cache entries are frequently evicted and re-created in a continuous cycle, preventing effective cache utilization. This happens when your workflow generates more cache data than can be stored within your available cache storage quota. 

By default, GitHub Actions provides *10 GB of cache storage per repository*. When this limit is reached, GitHub automatically evicts the least recently used cache entries to make room for new ones. If your workflows consistently produce cache data that exceeds this limit, the cache is constantly being written and evicted—resulting in thrashing. 

This leads to: 

- Wasted compute time spent creating caches that are immediately evicted 
- Increased network bandwidth consumption 
- Little to no performance benefit from caching 

To avoid thrashing, consider increasing your cache storage limit. For more information, see ["Usage limits and eviction policy."](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy)

Felixoid:
This PR addresses multiple requests to have tiered caching.

Closes #30
Closes #1020
Closes #2493
Closes #2566 

What is added:

- multilevel configuration for the caching chain
- config parameter for cache writing tolerance: ignore errors, fail on write to L0, fail on a write to any level
- added stats for each caching level
- documentation
- tests
- integration tests


Felixoid:
I've done quite a few features today, like control for write error handling, and added server stats for multilevel caching.

sylvestre:
this is a huge patch, can you split a bit more? thanks

Felixoid:
Probably I could take something out of `cache.rs`, but the rest is actually `multilevel.rs`, and 2/3 of 2k LoC are tests.

Felixoid:
Rebased to the current `main`.

This PR now contains the multi-level cache feature exclusively. Most of the changes are the tests and documentation, which are around 70%.

trxcllnt:
I have an implementation of this in our NVIDIA fork [here](https://github.com/rapidsai/sccache/blob/93ba7c295a6fa3fcc2787a1daad895a0c672fe12/src/cache/tiered.rs). It's a bit smaller because it uses composition to implement the cache hierarchy ([example](https://github.com/rapidsai/sccache/blob/93ba7c295a6fa3fcc2787a1daad895a0c672fe12/src/cache/tiered.rs#L188-L193)).

I made `Cache::put()` and `Cache::get()` accept/return a `bytes::Bytes`, which was necessary for performance -- the DiskCache can use mmap and return `Bytes::from_owner(Mmap::map(&file))`, and the OpenDAL caches can serialize their buffers `to_bytes()`. That way the `Bytes` instance can be used either by reference or by zero-copy slice.

Is it worth discussing the trade-offs to implementation here?

Felixoid:
Hey-hey @trxcllnt, yeah, that's awesome. Thank you for the insight!

The improvement for the `Storage` trait looks very good from a performance perspective. While preparing this PR, I didn't want to change the architect so much and didn't consider the zero-copy approach.

The main question now is whether it is worth a) implementing in this PR, b) postponing until the next one, or rather c) implementing as a foundation for the changes, checking it works, and then moving to a dedicated PR.

**update**: I revisit it, and, at least for now, the copying looks negligible. The cold-cache fillback usually should be more or less quick compared to other IO (disk write/network). We can improve it in follow-up PRs.

Felixoid:
Dear @AJIOB and @mathstuf, I understand this PR looks huge, but most of the changes are docs and tests.

I'd very much appreciate this PR moving forward. It should significantly improve our local ClickHouse builds.

mathstuf (commented):
Docs look good, but I would like to request one thing (feel free to punt for now though). I don't like that precedence is restricted to service *types*. For example, I may have:

- local disk
- NFS share ("disk")
- node-local Redis
- data center Redis
- local S3
- global S3

and want to interleave them in priority order. Basically: allow *naming* storages and then use the names in the chain configuration. Pure envvar configuration is probably lost, but maybe that's fine?

Overall, I love the idea, just asking for "one more thing" before things get baked in here. Again, feel free to ignore; it can probably be tacked on later too (by reserving all impl names or requiring a prefix on custom names).

src/cache/disk.rs
```diff
@@ -130,6 +159,31 @@ impl Storage for DiskCache {
             .await?
     }
 
+    async fn put_raw(&self, key: &str, data: Vec<u8>) -> Result<Duration> {
+        trace!("DiskCache::put_raw({}, {} bytes)", key, data.len());
```
AJIOB, line 143: nit: why not just call the `put` internally? IMO, they looks the same.

I prefer not to duplicate code if there's no significant reason to do it.

P.S. And for another backends too, please
P.P.S. And similar for `get` calls. Maybe we can call `get_raw`/`put_raw` from the `get`/`put` - it's also OK.

src/cache/multilevel.rs
```diff
@@ -0,0 +1,2217 @@
...
+                    feature = "gha",
+                    feature = "memcached",
+                    feature = "redis",
+                    feature = "s3",
+                    feature = "webdav",
+                    feature = "oss",
+                    feature = "cos"
+                )))]
+                {
+                    return Err(anyhow!(
```
AJIOB, line 487: Should we generate an error if we pass disk as a single multi-level backend? Looks like we can generate the config in this case

src/cache/multilevel.rs
```diff
@@ -0,0 +1,2217 @@
...
+
+    async fn put_preprocessor_cache_entry(
+        &self,
+        key: &str,
+        preprocessor_cache_entry: PreprocessorCacheEntry,
+    ) -> Result<()> {
+        for (idx, level) in self.levels.iter().enumerate() {
+            if let Err(e) = level
+                .put_preprocessor_cache_entry(key, preprocessor_cache_entry.clone())
+                .await
```
AJIOB: Should we do that sequentially? Why not in the async mode with next wait for all of them?

If I correctly understand the docs, it should be done independently

AJIOB (approved):

AJIOB:
In general, I like that changes.

There's awesome work, thanks, @Felixoid!

JFYI, you have a failed CUDA test. Please, recheck this point too.

src/cache/disk.rs
Felixoid: It's the other way around, actually. The `DiskCache` cache has raw methods to reflect the `Storage` trait implemented for `opendal::Operator`.

The `get` returns `to_bytes`, which decompresses when a compression level is used.

`get_raw`/`put_raw` are needed for a correct backfill without additional allocations.

`CacheRead` doesn't allow operating with raw bytes of the underlying `zip`:

```rust
/// Data stored in the compiler cache.
pub struct CacheRead {
    zip: ZipArchive<Box<dyn ReadSeek>>,
}
```

Using `to_vec()` guarantees byte-identical results between cache layers. We can't just get rid of them, unfortunately.

But `put` is another story; it fits well there.

src/cache/multilevel.rs
Felixoid: Oh yeah, it was a rough rebasing. I'll fix it

src/cache/multilevel.rs
Felixoid: Yes, and we do:

```
> SCCACHE_DIR=/tmp/sccache SCCACHE_MULTILEVEL_CHAIN=disk ./target/debug/sccache --show-stats
Compile requests                      0
Compile requests executed             0
Cache misses                          0
Cache hits rate                       -
Compilations                          0
Average compiler                  0.000 s
Multi-level cache levels              1
  L0 (disk) hits                      0
  L0 (disk) misses                    0
  L0 (disk) hit rate                  -
  L0 (disk) writes                    0
  L0 (disk) write failures            0
  L0 (disk) backfills from            0
  L0 (disk) backfills to              0
  L0 (disk) avg cache write       0.000 s
  L0 (disk) avg cache read hit    0.000 s
Cache location                  Multi-level (1 levels): L0: Local disk: "/tmp/sccache"
Base directories                (none)
Version (client)                0.14.0
Max cache size                       10 GiB
```

It's a feature-gate for the case, when non-`disk` storage is used.

AJIOB: IMO, we can just rely on line 443 output. Maybe with some tunes. No need for additional checks.

Or it generates some warnings about unreachable code?

src/cache/disk.rs
AJIOB: nit: add a comment about this. It's not easily understandable for me while reading the code

src/cache/multilevel.rs
Felixoid: It's another issue. Now, each name is actually checked, for example:

```rust
                        #[cfg(feature = "redis")]
		
                        "redis" => config.cache_configs.redis.clone().map(CacheType::Redis),
```

And if the feature is not built in, there's an error: (*update* I see, it's not true in any case. Yeah, I'll remove this block)

`Cache level '{}' requires a backend feature to be enabled`

If the block with `return Err(anyhow!("Unknown cache level: '{}'", level_name));` is reachable for `--no-default-features` build, there's either an unreachable code or a wrong error message.

Do you think, it's fine to raise `Unknown cache level: 'redis'` when the feature is not included?

Felixoid: Like this?

```diff
@@ -404,17 +404,6 @@ impl MultiLevelStorage {
                 trace!("Added disk cache level");
             } else {
                 // Build remote cache - get the appropriate CacheType
-                #[cfg(any(
-                    feature = "azure",
-                    feature = "gcs",
-                    feature = "gha",
-                    feature = "memcached",
-                    feature = "redis",
-                    feature = "s3",
-                    feature = "webdav",
-                    feature = "oss",
-                    feature = "cos"
-                ))]
                 {
                     let cache_type = match level_name.to_lowercase().as_str() {
                         #[cfg(feature = "s3")]
@@ -458,23 +447,6 @@ impl MultiLevelStorage {
                         ));
                     }
                 }
-                #[cfg(not(any(
-                    feature = "azure",
-                    feature = "gcs",
-                    feature = "gha",
-                    feature = "memcached",
-                    feature = "redis",
-                    feature = "s3",
-                    feature = "webdav",
-                    feature = "oss",
-                    feature = "cos"
-                )))]
-                {
-                    return Err(anyhow!(
-                        "Cache level '{}' requires a backend feature to be enabled (e.g., --features redis,s3)",
-                        level_name
-                    ));
-                }
             }
         }
```

Felixoid: ah no, it's there exactly for the `unreachable code` issue:

```
> cargo test --no-default-features
   Compiling sccache v0.14.0 (/home/felixoid/Space/Felixoid/github/mozilla/sccache)
warning: unused imports: `CacheType` and `self`
  --> src/cache/cache.rs:50:21
   |
50 | use crate::config::{self, CacheType, PreprocessorCacheModeConfig};
   |                     ^^^^  ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default

warning: unused import: `std::io`
  --> src/cache/cache.rs:53:5
   |
53 | use std::io;
   |     ^^^^^^^

warning: unreachable expression
   --> src/cache/multilevel.rs:436:21
    |
408 |                       let cache_type = match level_name.to_lowercase().as_str() {
    |  ______________________________________-
409 | |                         #[cfg(feature = "s3")]
410 | |                         "s3" => config.cache_configs.s3.clone().map(CacheType::S3),
411 | |                         #[cfg(feature = "redis")]
...   |
434 | |                     };
    | |_____________________- any code following this `match` expression is unreachable, as all arms diverge
435 |
436 | /                     if let Some(cache_type) = cache_type {
437 | |                         let storage = build_single_cache(&cache_type, &config.basedirs, pool)
438 | |                             .with_context(|| {
439 | |                                 format!("Failed to build cache for level '{}'", level_name)
...   |
447 | |                         ));
448 | |                     }
    | |_____________________^ unreachable expression
    |
    = note: `#[warn(unreachable_code)]` (part of `#[warn(unused)]`) on by default

error[E0425]: cannot find function `build_single_cache` in this scope
   --> src/cache/multilevel.rs:437:39
    |
437 |                         let storage = build_single_cache(&cache_type, &config.basedirs, pool)
    |                                       ^^^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
warning: `sccache` (lib) generated 3 warnings
error: could not compile `sccache` (lib) due to 1 previous error; 3 warnings emitted
warning: build failed, waiting for other jobs to finish...
warning: `sccache` (lib test) generated 3 warnings (3 duplicates)
error: could not compile `sccache` (lib test) due to 1 previous error; 3 warnings emitted
```

Felixoid:
Thanks a lot for the review!

@AJIOB: it looks like the CUDA tests were temporarily broken for an unknown reason. It was `Segmentation fault` in `sudo apt-get update`. Looks working fine now.

@mathstuf: as much as I like the idea, it looks like a next iteration for now. Probably, I'd like to work on it too. But it's for sure will be a design change for the configuration, and should be discussed in advance on how to proceed.

Now, the approach is simple and logical for the current configuration.

src/cache/disk.rs
Felixoid: Added it to `RemoteStorage` implementation

AJIOB (approved):

src/cache/multilevel.rs
AJIOB: Yes, it is. So, yes, just keep the current one, you're right

Felixoid:
@sylvestre do you think we can proceed?

src/cache/cache.rs
```diff
@@ -104,6 +119,11 @@ pub trait Storage: Send + Sync {
     /// Get the maximum storage size, if applicable.
     async fn max_size(&self) -> Result<Option<u64>>;
 
+    /// Get multi-level cache statistics, if this is a multi-level storage.
+    fn multilevel_stats(&self) -> Option<crate::cache::multilevel::MultiLevelStats> {
```
drahnr: nit: can we avoid the full path?

src/cache/multilevel.rs
```diff
@@ -0,0 +1,2232 @@
...
+    pub write_duration: Duration,
+}
+
+/// Statistics for multi-level cache operation.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct MultiLevelStats {
+    /// Per-level statistics
+    pub levels: Vec<LevelStats>,
+    /// Number of times a hit occurred before checking all levels
+    pub early_hits: u64,
```
drahnr: nit: isn't this already given in the sum of all of them but the last?

src/cache/multilevel.rs
```diff
@@ -0,0 +1,2232 @@
...
+            .collect();
+
+        // Wait for all writes to complete (errors are logged, not propagated)
+        futures::future::join_all(futures).await;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
drahnr, line 904: nit: Can we split the tests into a separate file?

docs/MultiLevel.md
```diff
@@ -0,0 +1,249 @@
...
+└───────────┘    └─────────┘     └─────────┘
+```
+
+### Read Path (Cache Hit at Level 2)
+
+```
+1. Check L0 (disk)    → Miss (5ms)
+2. Check L1 (redis)   → Miss (10ms)
+3. Check L2 (s3)      → Hit! (200ms)
+4. Return to compiler (Total: 215ms)
```
drahnr: nit: I think this should say _return to compiler wrapper / sccache_ iiuc?

drahnr (commented):
Bar the shell scripts, the core implementation does look sound. Also took a look @trxcllnt which is a more minimal approach with 2 cache layers (or using nesting).
Since this PR already got reviews I suggest to move it along.

@sylvestre if you find the time, take a look, otherwise I'll do sometime next week and we should land it imho.

Felixoid:
Thanks for the review!

I am on vacation RN, will try to find time to address the points somewhere around the following week.

src/cache/multilevel.rs
Felixoid: Looks like it's not only redundant but also has a logical error in how it is populated. I'll remove it, it's derivable from per-level stats.

Felixoid:
@drahnr I addressed all points. Sorry that it took so long. I tried not to touch the laptop for work

Felixoid:
In case #2644 is merged, this PR should benefit too; there are around 50 lines of boilerplate.

Felixoid:
Any update on the process, please?

AJIOB:
Hi @drahnr and @sylvestre ,

Is it OK for you? Can we merge these changes?

sunnyguan:
Thanks a lot for the work here, just curious if SCCACHE_BASEDIRS is handled properly when using the multi-tier setup?

sylvestre:
i am not ignoring you. just higher priorities lately, sorry

Felixoid:
> Thanks a lot for the work here, just curious if SCCACHE_BASEDIRS is handled properly when using the multi-tier setup?

I don't have any specific tests for it, but I don't see how it could break.

I'll try to add it too, just in case

Felixoid:
@sunnyguan thanks a lot for the spotlight, I indeed forgot to implement the `basedirs` treat for the `MultiLevelStorage` :facepalm:

Felixoid:
I believe the `cmake-modules` failure is unrelated; the fix of test is in https://github.com/mozilla/sccache/pull/2649

The dedicated issue for cmake 4.0 failure https://github.com/mozilla/sccache/issues/2650

Felixoid:
The merged https://github.com/mozilla/sccache/pull/2649 caused a conflict. Resolved it by rebasing

mmastrac:
I pulled this PR into a fork of sccache we've been testing in the next.js build and I'm happy to say it works as expected. I've tested with a local cache and network-addressable store (in our case, vercel artifacts which isn't fully in upstream opendal yet).

The only suggestion I might offer is a nit and a visual one at that:

```
Cache location                             Multi-level (2 levels): L0: Local disk: "/root/.sccache", L1: vercel-artifacts, name: , prefix: 
```

... would be nicer to read as:

```
Cache location                             Multi-level (2 levels)
  L0: Local disk: "/root/.sccache"
  L1: vercel-artifacts, name: , prefix: 
```

Felixoid:
I'll try to find it it's possible, but currently the value of a structure is a string. It can affect the json parsers

mmastrac:
We've been testing this in our CI for a few weeks now and I'm happy to say that it seems perfectly stable. Note that we're using a not-yet-upstreamed remote backend, but I believe that this PR would definitely be valuable to land.

<details>
<summary>a fairly detailed sample job log</summary>

```
Post job cleanup.
=== sccache stats ===
Compile requests                                999
Compile requests executed                       894
Cache hits                                      862
Cache hits (Assembler)                           20
Cache hits (C/C++)                               56
Cache hits (Rust)                               786
Cache misses                                     32
Cache misses (C/C++)                              1
Cache misses (Rust)                              31
Cache hits rate                               96.42 %
Cache hits rate (Assembler)                  100.00 %
Cache hits rate (C/C++)                       98.25 %
Cache hits rate (Rust)                        96.21 %
Cache timeouts                                    0
Cache read errors                                 0
Forced recaches                                   0
Cache write errors                                0
Cache errors                                      0
Compilations                                     32
Compilation failures                              0
Non-cacheable compilations                        0
Non-cacheable calls                             102
Non-compilation calls                             3
Unsupported compiler calls                        0
Average cache write                           0.092 s
Average compiler                              4.075 s
Average cache read hit                        0.002 s
Failed distributed compilations                   0
Multi-level cache levels                          2
  L0 (disk) hits                                855
  L0 (disk) misses                               39
  L0 (disk) hit rate                          95.64 %
  L0 (disk) writes                               32
  L0 (disk) write failures                        0
  L0 (disk) backfills from                        0
  L0 (disk) backfills to                          7
  L0 (disk) avg cache write                   0.031 s
  L0 (disk) avg cache read hit                0.000 s
  L1 (vercel-artifacts) hits                      7
  L1 (vercel-artifacts) misses                   32
  L1 (vercel-artifacts) hit rate              17.95 %
  L1 (vercel-artifacts) writes                   32
  L1 (vercel-artifacts) write failures            0
  L1 (vercel-artifacts) backfills from            7
  L1 (vercel-artifacts) backfills to              0
  L1 (vercel-artifacts) avg cache write       0.211 s
  L1 (vercel-artifacts) avg cache read hit    0.175 s

Non-cacheable reasons:
argument parse                                   67
-o                                               24
-                                                 8
missing input                                     3

Cache location                             Multi-level (2 levels): L0: Local disk: "/root/.sccache", L1: vercel-artifacts, name: , prefix: 
Base directories                           /root/actions-runner/_work/next.js/next.js/
Version (client)                           0.14.0
Cache size                                        7 GiB
Max cache size                                   10 GiB
Stopping sccache server...
Compile requests                                999
Compile requests executed                       894
Cache hits                                      862
Cache hits (Assembler)                           20
Cache hits (C/C++)                               56
Cache hits (Rust)                               786
Cache misses                                     32
Cache misses (C/C++)                              1
Cache misses (Rust)                              31
Cache hits rate                               96.42 %
Cache hits rate (Assembler)                  100.00 %
Cache hits rate (C/C++)                       98.25 %
Cache hits rate (Rust)                        96.21 %
Cache timeouts                                    0
Cache read errors                                 0
Forced recaches                                   0
Cache write errors                                0
Cache errors                                      0
Compilations                                     32
Compilation failures                              0
Non-cacheable compilations                        0
Non-cacheable calls                             102
Non-compilation calls                             3
Unsupported compiler calls                        0
Average cache write                           0.092 s
Average compiler                              4.075 s
Average cache read hit                        0.002 s
Failed distributed compilations                   0
Multi-level cache levels                          2
  L0 (disk) hits                                855
  L0 (disk) misses                               39
  L0 (disk) hit rate                          95.64 %
  L0 (disk) writes                               32
  L0 (disk) write failures                        0
  L0 (disk) backfills from                        0
  L0 (disk) backfills to                          7
  L0 (disk) avg cache write                   0.031 s
  L0 (disk) avg cache read hit                0.000 s
  L1 (vercel-artifacts) hits                      7
  L1 (vercel-artifacts) misses                   32
  L1 (vercel-artifacts) hit rate              17.95 %
  L1 (vercel-artifacts) writes                   32
  L1 (vercel-artifacts) write failures            0
  L1 (vercel-artifacts) backfills from            7
  L1 (vercel-artifacts) backfills to              0
  L1 (vercel-artifacts) avg cache write       0.211 s
  L1 (vercel-artifacts) avg cache read hit    0.175 s

Non-cacheable reasons:
argument parse                                   67
-o                                               24
-                                                 8
missing input                                     3

Cache location                             Multi-level (2 levels): L0: Local disk: "/root/.sccache", L1: vercel-artifacts, name: , prefix: 
Base directories                           /root/actions-runner/_work/next.js/next.js/
Version (client)                           0.14.0
Cache size                                        7 GiB
Max cache size                                   10 GiB
=== sccache error log (last 30 lines) ===
```
</details>

Felixoid:
@mmastrac looking at your output, and stepping back to the request in https://github.com/mozilla/sccache/pull/2581#issuecomment-4178940238

I feel it makes sense to shorten the `cache_location` to a simple `Multi-level (2 levels)`, and then add `stats.multi_level[]` would have a location to parse.

Besides, `stats.multi_level.levels[]` looks redundant. I'll get rid of the levels for good

Felixoid:
That's how it looks for the most recent changes:

<details>

```
> SCCACHE_DIR=/tmp/cache/sccache SCCACHE_MULTILEVEL_CHAIN=disk ./target/debug/sccache --show-stats --stats-format json | jq
{
  "stats": {
    "compile_requests": 0,
    "requests_unsupported_compiler": 0,
    "requests_not_compile": 0,
    "requests_not_cacheable": 0,
    "requests_executed": 0,
    "cache_errors": {
      "counts": {},
      "adv_counts": {}
    },
    "cache_hits": {
      "counts": {},
      "adv_counts": {}
    },
    "cache_misses": {
      "counts": {},
      "adv_counts": {}
    },
    "cache_timeouts": 0,
    "cache_read_errors": 0,
    "non_cacheable_compilations": 0,
    "forced_recaches": 0,
    "cache_write_errors": 0,
    "cache_writes": 0,
    "cache_write_duration": {
      "secs": 0,
      "nanos": 0
    },
    "cache_read_hit_duration": {
      "secs": 0,
      "nanos": 0
    },
    "compilations": 0,
    "compiler_write_duration": {
      "secs": 0,
      "nanos": 0
    },
    "compile_fails": 0,
    "not_cached": {},
    "dist_compiles": {},
    "dist_errors": 0,
    "multi_level": [
      {
        "name": "L0 (disk)",
        "location": "Local disk: \"/tmp/cache/sccache\"",
        "hits": 0,
        "misses": 0,
        "writes": 0,
        "write_failures": 0,
        "backfills_from": 0,
        "backfills_to": 0,
        "hit_duration": {
          "secs": 0,
          "nanos": 0
        },
        "write_duration": {
          "secs": 0,
          "nanos": 0
        }
      }
    ]
  },
  "cache_location": "Multi-level (1 levels)",
  "cache_size": null,
  "max_cache_size": 10737418240,
  "use_preprocessor_cache_mode": true,
  "version": "0.14.0",
  "basedirs": []
}
> SCCACHE_DIR=/tmp/cache/sccache SCCACHE_MULTILEVEL_CHAIN=disk ./target/debug/sccache --show-stats
Compile requests                      0
Compile requests executed             0
Cache misses                          0
Cache hits rate                       -
Compilations                          0
Average compiler                  0.000 s
Multi-level cache levels              1
  L0 (disk) hits                      0
  L0 (disk) misses                    0
  L0 (disk) hit rate                  -
  L0 (disk) writes                    0
  L0 (disk) write failures            0
  L0 (disk) backfills from            0
  L0 (disk) backfills to              0
  L0 (disk) avg cache write       0.000 s
  L0 (disk) avg cache read hit    0.000 s
Cache location                  Multi-level (1 levels)
  L0 (disk)                     Local disk: "/tmp/cache/sccache"
Base directories                (none)
Version (client)                0.14.0
Max cache size                       10 GiB
```

</details>

Exactly what you asked, but now `stats.multi_level[].location` belongs to each particular level :+1:

Felixoid:
@drahnr another round of the review, maybe? Or even merging it? ^_^

mmastrac:
Happy to vouch for this PR with successful build runs as well!

src/cache/cache.rs
```diff
@@ -279,6 +296,51 @@ impl Storage for RemoteStorage {
...
         &self.basedirs
     }
+
+    /// Get raw bytes from remote storage without any transformations.
+    ///
+    /// Uses `to_vec()` instead of `to_bytes()` to preserve raw data unchanged.
+    /// This is critical for multi-level caching: when backfilling from remote to local,
+    /// we need the exact bytes without OpenDAL layer transformations (e.g., decompression).
+    /// If compression layers are configured, `to_bytes()` would decompress the data,
+    /// which would corrupt the cache entry when written to another level.
```
trxcllnt: Can you explain this comment? The `Operator::read()` method returns an opendal Buffer, and it looks like the only difference between [`to_bytes()`](https://github.com/apache/opendal/blob/main/core/core/src/types/buffer.rs#L283-L307) and [`to_vec()`](https://github.com/apache/opendal/blob/main/core/core/src/types/buffer.rs#L314-L318) are that `to_vec()` always copies, but `to_bytes()` will only copy if the internal storage isn't contiguous.

Also I don't think sccache defines any opendal layer transformations, so I'm confused what this method achieves.

drahnr:
> @drahnr another round of the review, maybe? Or even merging it? ^_^

Sorry for the delay, I'll get to it tomorrow.

src/cache/cache.rs
Felixoid: Hey hey, I described it in https://github.com/mozilla/sccache/pull/2581#discussion_r2816227064

trxcllnt: I don't think that addresses my question about the comment on why `to_vec()` vs. `to_bytes()`. Regardless whether any opendal layer transformations happen, `to_vec()` and `to_bytes()` will still return identical data. I'd suggest rewriting this comment so it doesn't mislead future maintainers.

As for vec vs. bytes, you can construct `CacheRead` from `bytes::Bytes` instances via `CacheRead::from(Cursor::new(bytes))`. When I changed the NVIDIA fork to return `opendal::Buffer` instead of `Vec<u8>`, the `CacheRead` constructor was the only legacy callsite that could potentially introduce a copy (due to needing to materialize to `bytes::Bytes`). Even though it's probably out of scope for this PR, it was a relatively minor change that's worth revisiting in the future.

Felixoid: Oh wait, it was 2 months ago, so I mixed it up with `CacheRead` wrapped around `ZipArchive<Box<dyn ReadSeek>>`

Yes, for the `Operator::read` `to_bytes` is much better, no doubts.

I'll replace the `to_vec` by `to_bytes` and update the comment. It's indeed misleading now.

**update**: for some reason, GH stopped updating the page for new comments. I didn't see your new one while answering

docs/MultiLevel.md
```diff
@@ -0,0 +1,249 @@
...
+### 3. Developer Workstation with Cloud Backup
+
+**Problem**: Local disk fills up, don't want to lose cache history.
+
+**Solution**: Disk + cloud storage
+```bash
+SCCACHE_MULTILEVEL_CHAIN="disk,s3"
+SCCACHE_DIR="$HOME/.cache/sccache"
+SCCACHE_BUCKET="my-personal-sccache"
+SCCACHE_CACHE_SIZE="5G"  # Keep disk small
```
drahnr, line 114: I find this ambiguous. I'd prefer longer names where size limits are clearly associated to a specific backend. I.e. _CACHE_SIZE doesn't tell me to _which_ cache level it refers to.

Felixoid: I didn't touch the `SCCACHE_CACHE_SIZE` parameter.

https://github.com/mozilla/sccache/blob/main/docs/Configuration.md#disk-local

It's a local disk cache size limit.

```
[cache.disk]
dir = "/tmp/.cache/sccache"
size = 7516192768 # 7 GiBytes
```

I can change the comment to `# Keep local disk cache small` though.

Other backends don't have a way to limit the cache size.

docs/MultiLevel.md
```diff
@@ -0,0 +1,249 @@
...
+
+#### Read-Only Levels
+
+Any level configured as read-only (e.g., `SCCACHE_LOCAL_RW_MODE=READ_ONLY`) is automatically skipped during writes, regardless of write policy:
+
+```bash
+export SCCACHE_MULTILEVEL_CHAIN="disk,redis"
+export SCCACHE_MULTILEVEL_WRITE_POLICY="all"
+export SCCACHE_LOCAL_RW_MODE="READ_ONLY"  # Disk is read-only
+# Compilation succeeds if Redis write succeeds (disk is skipped)
```
drahnr, line 175: I find the configuration to be rather inconsistent and I cannot reason from the name of the var to its impact, which should be a goal for improved usability

docs/MultiLevel.md
```diff
@@ -0,0 +1,249 @@
...
+export SCCACHE_S3_USE_SSL="true"
+```
+
+### Via Configuration File
+
+```toml
+# ~/.config/sccache/config
+[cache.multilevel]
+chain = ["disk", "redis", "s3"]
+write_policy = "l0"  # Optional: ignore, l0 (default), or all
```
drahnr: do we also support `l1` implying write to `l1` only or to `l0+l1` but not `l2`?

src/cache/multilevel.rs
```diff
@@ -0,0 +1,915 @@
...
+            format!("  {} backfills to", self.name),
+            self.backfills_to.to_string(),
+            0,
+        ));
+
+        // 4. Timing stats
+        let avg_write_duration = if self.writes > 0 {
+            self.write_duration / self.writes as u32
+        } else {
+            Duration::default()
```
drahnr, line 259: nit: we might want to ensure this never shows in output, but is rendered as a `-`

src/cache/multilevel.rs
```diff
@@ -0,0 +1,915 @@
...
+                    continue;
+                }
+            }
+        }
+        debug!("Cache miss at all levels");
+
+        // Mark final miss for all checked levels
+        for idx in 0..self.levels.len() {
+            if let Some(stats) = self.atomic_stats.get(idx) {
+                stats.misses.fetch_add(1, Ordering::Relaxed);
```
drahnr: nit: adding a helper for the stats might be worth it

src/config.rs
```diff
@@ -37,6 +37,56 @@ use typed_path::Utf8TypedPathBuf;
 
 use crate::errors::*;
 
+/// Defines how the multi-level cache handles write failures.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum WritePolicy {
```
drahnr: Shouldn't this be `enum WriteErrorPolicy`?

drahnr (approved):
The logic is sound, however, I think another pass avoiding code duplication and making the naming easier to discover _without_ needing to read all of the docs, might be a worthy approach. This can also be done as a follow-up PR, although I am not keen to break public "config API".

Thank you!

src/config.rs
Felixoid: Yes, it reflects the purpose much better!

This renaming addresses the https://github.com/mozilla/sccache/pull/2581#discussion_r3087545006 and https://github.com/mozilla/sccache/pull/2581#pullrequestreview-4114773291 comments as well.

src/cache/multilevel.rs
Felixoid: ~Is it better to hide it completely on no writes? To keep the json consistent and typed.~

Nevermind, it's only for the `text` output format. But another thing, it's the same for the global stats:

```
> SCCACHE_DIR=/tmp/cache/sccache SCCACHE_MULTILEVEL_CHAIN=disk ./target/debug/sccache --show-stats
Compile requests                      0
Compile requests executed             0
Cache misses                          0
Cache hits rate                       -
Compilations                          0
Average compiler                  0.000 s
Multi-level cache levels              1
  L0 (disk) hits                      0
  L0 (disk) misses                    0
  L0 (disk) hit rate                  -
  L0 (disk) writes                    0
  L0 (disk) write failures            0
  L0 (disk) backfills from            0
  L0 (disk) backfills to              0
  L0 (disk) avg cache write       0.000 s
  L0 (disk) avg cache read hit    0.000 s
Cache location                  Multi-level (1 levels)
  L0 (disk)                     Local disk: "/tmp/cache/sccache"
Base directories                (none)
Version (client)                0.14.0
Max cache size                       10 GiB
```

I think it should stay consistent

Felixoid:
Thanks for the review!

I think all points addressed so far. :+1:

Felixoid:
Looks like an infrastructure error https://github.com/mozilla/sccache/actions/runs/24502299252/job/71611929011

I'll try adding retries here later in another PR. A simple rerun of `docker build` could help

docs/MultiLevel.md
drahnr: Out of scope, I'll deal with this eventually.

docs/MultiLevel.md
```diff
@@ -0,0 +1,249 @@
...
+export SCCACHE_DIR="/var/cache/sccache"
+export SCCACHE_CACHE_SIZE="10G"
+
+# Level 1: Redis cache
+export SCCACHE_REDIS_ENDPOINT="redis://localhost:6379"
+export SCCACHE_REDIS_EXPIRATION="86400"  # 24 hours
+
+# Level 2: S3 cache
+export SCCACHE_BUCKET="my-sccache-bucket"
+export SCCACHE_REGION="us-east-1"
```
drahnr, line 195: I'd have expected a `.._S3_..` component in the path?

src/cache/multilevel.rs
drahnr: Let's defer

src/config.rs
```diff
@@ -484,6 +537,62 @@ impl CacheConfigs {
...
+                    })?,
+                    "azure" => self.azure.clone().map(CacheType::Azure).ok_or_else(|| {
+                        anyhow!("Azure cache not configured but specified in levels")
+                    })?,
+                    "webdav" => self.webdav.clone().map(CacheType::Webdav).ok_or_else(|| {
+                        anyhow!("Webdav cache not configured but specified in levels")
+                    })?,
+                    "oss" => self.oss.clone().map(CacheType::OSS).ok_or_else(|| {
+                        anyhow!("OSS cache not configured but specified in levels")
+                    })?,
```
drahnr, line 578: nit: Can we avoid the 8x duplication here

drahnr (approved):
Just nits, let's merge and move on

docs/MultiLevel.md
Felixoid: Same as other configs. Most probably worth changing, but I didn't touch the existing configuration.

src/config.rs
Felixoid: Probably. We can benefit from having a map `&str, CacheType`. It's used in `multilevel.rs` in a similar way, but with feature gates.

Felixoid: I tried to implement it in https://github.com/mozilla/sccache/pull/2644/changes/827571378236922b8d9eb7cf4fe9899ec92e1cb8

PR #2583, opened 2026-01-30
docs: add comments about auth token requirements

jullanggit:
Adds comments to the documentation for distributed compilation explaining the requirement of the token being a valid HTTP header value.
Hopefully this saves the next person running into this a bit of time :)


sylvestre:
thanks

Issue #2278, opened 2024-10-28
Using bazel-remote as cache backend is not working


simait:
Hello everyone,

I've tried to setup sccache with bazel-remote, but it doesn't seem to work. First you get an error that it expects a blob-path, e.g. something like `/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`. So quickly just switching `.sccache_check` to `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` (empty blob) yields another error, i.e. `PROPFIND` is not supported (HTTP 405).

As `PROPFIND` sems to be an integral part of WebDAV (at least according to a quick google) I don't see any way of supporting bazel-remote without bazel-remote also supporting `PROPFIND`, but I'm not an expert in either area and I have little to no experience with WebDAV or opendal in particular so maybe I'm wrong.

Anyway, as bazel-remote is currently not supported and unless there is a way to support it in the short term I would ask that you update the documentation accordingly.

Best regards,
Simon Aittamaa

AJIOB:
Hi @simait 

We're using https://github.com/apache/opendal for remote backends.

Please, open the issue in that repo & link this one

Xuanwo:
Thank you @simait for the report, I will check this out.

weihanglo:
apache/opendal#7177 we exposed `disable_create_dir` option for WebDAV backend which I think sccache can propogate that, so bazel-remote cache usage can configure accordingly.

The other part of this issue, the `.sccache_check` one, I am thinking of having a `skip_health_check` config option, so again for bazel-remote it can set it false. An alternative is `health_check_path` and set it to `cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` (the above approach OP mentioned), so it is still health-checked. Though that makes little sense to a content-addressed storage I feel like.

weihanglo:
So besides these issues

* `.sccache_check` is not a valid SHA256 key for bazel-remote cache
* `PROPFIND` is not supported for bazel-remote cache

There are also

* Key normalization what bazel-remote doesn't do that: https://github.com/mozilla/sccache/blob/9f6cba4e86cf7f9a18ca9332370c980a9c27a9fa/src/cache/cache.rs?plain=1#L601-L604
[...]

PR #2591, opened 2026-02-03
feat(webdav): skip_health_check and disable_create_dir

weihanglo:
### ‼️ This can't be merged until opendal cuts a new release including the feature we want.

---

On startup, sccache verifies storage connectivity by reading/writing
a `.sccache_check` file. Some WebDAV servers reject this path format.
For example, bazel-remote requires SHA256 hash paths and returns 400
for arbitrary paths like `.sccache_check`.

Additionally, some servers don't support PROPFIND/MKCOL methods used
during directory creation. For example, bazel-remote returns 405.

This adds two independent options:

- `skip_health_check`: bypasses the probe and assumes read-write access
- `disable_create_dir`: skips PROPFIND/MKCOL calls during writes

Users can enable these via config or environment variables:

- `SCCACHE_WEBDAV_SKIP_HEALTH_CHECK=true`
- `SCCACHE_WEBDAV_DISABLE_CREATE_DIR=true`

Fixes: https://github.com/mozilla/sccache/issues/2278

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -126,6 +126,10 @@ username = "alice"
 password = "secret12"
 # Mutually exclusive with username & password. Bearer token value
 token = "token123"
+# Skip health check probe (for servers that reject .sccache_check path)
+skip_health_check = false
+# Disable PROPFIND/MKCOL calls (for servers that don't support them)
+disable_create_dir = false
 
 [cache.oss]
 bucket = "name"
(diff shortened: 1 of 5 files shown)


weihanglo:
@Xuanwo sorry for bothering you again. This is a way to fix bazel-remote support. With this, people can easily spin up a bazel-remote or nginx with minimal configuration to use as sccache backend. See <https://bazel.build/remote/caching>.

This would attract more usage of sccache and OpenDAL, since WebDAV is more a lightweight backend than others.

weihanglo:
Found more blocker in bazel-remote as webdav backend. Will post in the original issue. Closing as this won't go any farther

ChriFo:
Hi @weihanglo, thanks for this PR!

Even though it was closed because of the remaining bazel-remote blockers, I wanted to share that the `disable_create_dir` change works for another backend: **Sonatype Nexus with a RAW repository** (hosted, via WebDAV).

Without the option, writes fail because Nexus RAW repositories don't support the PROPFIND/MKCOL calls used for directory creation. With `SCCACHE_WEBDAV_DISABLE_CREATE_DIR=true`, uploading and reading cache entries works as expected.

The original blocker is also gone now: `disable_create_dir` shipped in OpenDAL 0.56.0 (apache/opendal#7177), and sccache main is already on OpenDAL 0.58.1 (#2715). So this should only need a rebase without the git dependency.

Would you consider reopening this (or splitting out `disable_create_dir` into its own PR)? If you don't have time, I'd be happy to pick it up. Happy to help test either way.

Thanks!

weihanglo:
feel free to pick it up

PR #2594, opened 2026-02-06
actions: add security audit workflow

xtqqczze:
Action will be executed periodically at midnight of each day and check if any new advisories appear for crate dependencies.

For each new advisory (including informal) an issue will be created.


xtqqczze:
@sylvestre

```log
The action rustsec/audit-check@v2 is not allowed in mozilla/sccache 
```

sylvestre:
oh right, let's revert it then

PR #2597, opened 2026-02-08
chore: Simplify `storage_from_config`

Felixoid:
It's another part of preparation for #2581

The diff is tiny, when `Hide whitespace` is used


Felixoid:
image

sylvestre:
> The diff is tiny, when Hide whitespace is used

TIL, thanks!

PR #2610, opened 2026-02-12
Add support for anonymous Azure blob access

agocke:
The current support for Azure blob storage can't deal with the situation where the blob store needs to be accessed anonymously. This PR adds the opendal http backend and an environment variable that can be used to enable Azure anonymous access. If the environment variable is set, sccache falls back to direct http access instead of the opendal authentication flow.


sylvestre:
some jobs are failing

agocke:
Thanks, investigating.

agocke:
I believe this is ready for re review/testing

sylvestre:
The toml_format job is failing

agocke:
Fixed! Simple formatting issue that I didn't understand at first :)

PR #2613, opened 2026-02-12
chore: switch thirtyfour_sync to thirtyfour

tottoto:
Switches unmaintained `thirtyfour_sync` to `thirtyfour`.

### DIFF
--- a/tests/oauth.rs
+++ b/tests/oauth.rs
@@ -6,9 +6,8 @@ use std::io::{self, Read, Write};
 use std::net::TcpStream;
 use std::path::Path;
 use std::process::{Command, Output, Stdio};
-use std::thread;
 use std::time::{Duration, Instant};
-use thirtyfour_sync::prelude::*;
+use thirtyfour::prelude::*;
 
 const LOCAL_AUTH_BASE_URL: &str = "http://localhost:12731/";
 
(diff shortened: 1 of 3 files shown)


sylvestre:
Please update the commit to document the why

tests/oauth.rs
```diff
@@ -74,54 +73,60 @@ fn retry<F: FnMut() -> Option<T>, T>(interval: Duration, until: Duration, mut f:
         if let Some(res) = f() {
             return Some(res);
         } else {
-            thread::sleep(interval);
+            std::thread::sleep(interval);
```
sylvestre, line 76: why this change ? seems unrelated

tottoto: Since this is now the only place where `std::thread` is used, importing it solely for this felt unnecessary.

sylvestre: please avoid unrelated changes :)

tottoto: Do you think this is not related?

sylvestre: yes

tottoto: I proposed including it in this pull request because it arose as a result of this change, and I thought it would be appropriate to include it.

PR #2614, opened 2026-02-12
clippy: fix from_iter_instead_of_collect lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#from_iter_instead_of_collect


xtqqczze:
> nor do I think the additional need for `Iteratools` is a net positive

We already use `itertools` here.

sylvestre:
needs to be rebased

PR #2618, opened 2026-02-16
feat(redis): add Redis Sentinel support for HA cache discovery

ex0a:
Adds Redis Sentinel support so sccache can discover the current master from a sentinel cluster instead of hardcoding a single Redis endpoint.

When `endpoint` starts with `redis-sentinel://`, sccache queries each sentinel node with `SENTINEL GET-MASTER-ADDR-BY-NAME`, connects to the discovered master, and builds the opendal operator against it.

URL format: `redis-sentinel://[:password@]host1[:port1][,host2[:port2],...]/master_name[/db]`

Example config:
```toml
[cache.redis]
endpoint = "redis-sentinel://sentinel1:26379,sentinel2:26379/mymaster"
```

`username`, `password`, and `db` config fields are ignored for sentinel URLs (embed credentials in the URL). A warning is logged if they're set.

URL parsing is extracted into `parse_sentinel_url()` with unit tests covering multi-node, password with `@`, db selection, and error cases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,13 +119,13 @@ nix = { version = "0.30.0", optional = true, features = [
   "signal",
   "process",
 ] }
+redis = { version = "0.32", features = ["sentinel"], optional = true }
 rouille = { version = "3.6", optional = true, default-features = false, features = [
   "ssl",
 ] }
 syslog = { version = "7", optional = true }
-version-compare = { version = "0.1.1", optional = true }
 tracing = "0.1.44"
-redis = { version = "0.32", features = ["sentinel"], optional = true }
+version-compare = { version = "0.1.1", optional = true }
 
 [dev-dependencies]
 assert_cmd = "2.0.13"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,12 +119,11 @@ nix = { version = "0.30.0", optional = true, features = [
   "signal",
   "process",
 ] }
-redis = { version = "0.32", features = ["sentinel"], optional = true }
+redis = { version = "0.32", optional = true }
 rouille = { version = "3.6", optional = true, default-features = false, features = [
   "ssl",
 ] }
 syslog = { version = "7", optional = true }
-tracing = "0.1.44"
 version-compare = { version = "0.1.1", optional = true }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,6 +124,8 @@ rouille = { version = "3.6", optional = true, default-features = false, features
 ] }
 syslog = { version = "7", optional = true }
 version-compare = { version = "0.1.1", optional = true }
+tracing = "0.1.44"
+redis = { version = "0.32", features = ["sentinel"], optional = true }
 
 [dev-dependencies]
 assert_cmd = "2.0.13"
@@ -176,7 +178,7 @@ gha = ["opendal/services-ghac", "reqwest"]
 memcached = ["opendal/services-memcached"]
 native-zlib = []
 oss = ["opendal/services-oss", "reqsign", "reqwest"]
-redis = ["url", "opendal/services-redis"]
+redis = ["dep:redis", "url", "opendal/services-redis"]
 s3 = ["opendal/services-s3", "reqsign", "reqwest"]
 webdav = ["opendal/services-webdav", "reqwest"]
 # Enable features that will build a vendored version of openssl and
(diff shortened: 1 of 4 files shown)


sylvestre:
sorry, it conflicts

ex0a:
What would you suggest? I'm using this personally but thought it might help someone else

AJIOB:
Hi @ex0a ,

You should rebase your changes to the latest `master` branch changes or just merge them to your branch - and resolve the conflicts.

Then you need to push your changes to your branch.

That's should be fine!

Cargo.toml
```diff
@@ -157,7 +159,7 @@ version = "0.52"
 [features]
 all = [
   "dist-client",
-  "redis",
```
AJIOB, line 161: Should we remove the redis feature here?

src/cache/cache.rs
```diff
@@ -281,235 +598,212 @@ impl Storage for RemoteStorage {
...
+                db,
+                url,
+                ttl,
+                key_prefix,
+            }) => {
+                let storage = match (endpoint, cluster_endpoints, url) {
+                    (Some(url_str), None, None) => {
+                        if url_str.starts_with("redis-sentinel://") {
+                             debug!("Init redis sentinel cache with url {url_str}");
+                             RedisCache::build_from_url(url_str, key_prefix, *ttl)
```
AJIOB: Do we ignore the username, password & database for the sentinel mode?

If yes, we should warn about it if the username, password or database has been provided

AJIOB:
CC @Xuanwo 

IMO, we should implement the Redis Sentinel support as a part of the OpenDAL. And just use it here.

Possibly, it's already done.

ex0a:
Thanks for the review! I've pushed updates addressing both points:
Feature gating: redis has been restored to the all feature list and the redis crate dependency is now properly optional via dep:redis, so it won't be compiled in unless the redis feature is enabled. I've also removed the hard tracing dependency I accidentally added.
Credentials warning: Added a warn!() when username, password, or db are provided alongside a redis-sentinel:// URL, since those fields are ignored in sentinel mode (credentials should be embedded in the URL itself).
Regarding OpenDAL: I considered this approach, but OpenDAL's Redis service doesn't currently support Sentinel natively. My implementation handles discovery at the sccache layer — querying sentinels to resolve the current master address, then building a normal single-node OpenDAL connection to it. This keeps the OpenDAL layer unchanged and avoids requiring upstream OpenDAL changes. Happy to revisit if OpenDAL adds Sentinel support in the future.

AJIOB (approved):

sylvestre:
please fix the jobs

ex0a:
Hi @sylvestre — CodSpeed results came in: +3.56% improvement on `normalize_win_path_utf8` (3.4 µs → 3.3 µs), no regressions. clippy, rustfmt, and taplo are all passing, and @AJIOB has approved. Let me know if there's anything else needed before merge.

ex0a:
Rebased onto latest main and addressed all four Copilot findings:

- Fixed the sentinel code path to actually call `build_sentinel` instead of `build_from_url`
- Switched to `rsplit_once('@')` for password parsing so passwords containing `@` don't break the URL split
- Dropped the unused `sentinel` feature flag from the redis crate dep
- Removed the stray `tracing` dependency

clippy, rustfmt, and taplo all pass clean.

src/cache/redis.rs
```diff
@@ -53,6 +54,153 @@ impl RedisCache {
...
+                        }
+                    }
+                    Err(e) => {
+                        last_error = Some(format!("Connection failed: {}", e));
+                        debug!("{}", last_error.as_ref().unwrap());
+                    }
+                },
+                Err(e) => {
+                    last_error = Some(format!("Client creation failed: {}", e));
+                    debug!("{}", last_error.as_ref().unwrap());
```
AJIOB: Nit: no need to unwrap

ex0a: Fixed — also extracted URL parsing into a separate function with unit tests while I was in there.

ex0a:
Friendly ping @sylvestre — this has been open since Feb 16 with AJIOB's approval, all CI passing (clippy/rustfmt/taplo/codspeed), and the Copilot review findings addressed. The remaining Codecov gap is the `build_sentinel()` function body which requires a live Sentinel cluster to exercise — the URL parsing logic is fully unit-tested.

Happy to add a Docker-based integration test if that's blocking, but wanted to check if there's anything else needed first.

ex0a:
Just a friendly ping, @sylvestre. Would you mind looking at this again? Thank you for your time.

PR #2621, opened 2026-02-17
msvc: Append the default .pdb extension for the /Fd argument

### LINKED ISSUE #2620: sccache fails to find the PDB file given with /Fd when the .pdb file extension is omitted
Sccache encounters fatal error when the pdb file is provided without the file extension ".pdb" via /Fd flag (eg. `/Fd"MyProject"` instead of `/Fd"MyProject.pdb"`), complaining it cannot find the file. According to [MSDN docs](https://learn.microsoft.com/en-us/cpp/build/reference/fd-program-database-file-name?view=msvc-170#:~:text=If%20you%20specify%20a%20file%20name%20that%20does%20not%20include%20an%20extension%2C%20the%20compiler%20uses%20%2Epdb%20as%20the%20extension), omitting the ".pdb" extension is a valid use case and the compiler should append it automatically if no file extension was provided.

ikolomiko:
MSVC appends the .pdb extension for the file path given for the /Fd argument, if the extension is omitted. This commit adds this functionality for feature parity.

Fixes #2620

### DIFF
--- a/src/compiler/msvc.rs
+++ b/src/compiler/msvc.rs
@@ -2011,6 +2011,166 @@ mod test {
     }
 
     #[test]
+    fn test_parse_arguments_pdb_no_extension() {
+        // Test that .pdb extension is appended when /Fd argument lacks an extension
+        let args = ovec!["-c", "foo.c", "-Zi", "-Fdfoo", "-Fofoo.obj"];
+        let ParsedArguments {
+            input,
+            language,
+            outputs,
+            preprocessor_args,
+            msvc_show_includes,
+            common_args,
+            ..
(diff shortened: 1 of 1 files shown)


AJIOB:
@ikolomiko, please, recheck the format: looks like it's failed

ikolomiko:
Oopsies, fixed it

AJIOB:
Nice! Can you add the test for that, please?

It will help with unexpected regression

ikolomiko:
@AJIOB sure, here it is!

AJIOB (approved):

ikolomiko:
Looks like the ci pipeline is borked :d

PR #2628, opened 2026-02-27
fix: create output directory before writing cached objects

decofe:
`extract_objects` calls `NamedTempFile::new_in(dir)` without ensuring the output directory exists first. This causes a fatal error during parallel compilation in large workspaces:

```
sccache: error: No such file or directory (os error 2) at path ".../target/debug/deps/.tmpUCVapr"
```

Adds `fs::create_dir_all(dir)` before the temp file creation. No-op if the dir already exists.


decofe:
Closing — this is likely not the actual root cause of the issue we were seeing. The output directory should already be created by cargo before invoking the rustc wrapper.

PR #2637, opened 2026-03-10
Fix GCS initialization

msuozzo:
GCS usage currently fails as the initializtion path happens outside of a
Tokio runtime context.


msuozzo:
I have a separate implementation which actually colors the callers as async but it's necessarily a lot bigger a change so I figured I'd start here.

Skylion007 (approved):

Issue #1174, opened 2022-05-16
Disk cache and clean rebuild speed


Meinersbur:
I noticed that sccache is a lot slower that ccache on a clean rebuild (`make clean && make`). I think this is due to the following effects:

1. In addition to hashing the preprocessor output, ccache also collects all preprocessor dependencies (either from parsing the preprocessor's line directive or the output of `-MD`), then hashes those files and writes them into the cache as a "manifest" file. Next time ccache gets invoked on the file, it looks whether any of those file have changed and if not, doesn't need to invoke the preprocessor. Ideally, it doesn't even need to rehash but consider file metadata whether a file has changed.

2. ccache has an option that stores the object files uncompressed. On a cache hit, it just creates a hard link (or clone on CoW file systems) to that file. This is also friendler for SSDs for not wasting write cycles. Note the some build systems may modify files inplace (eg. when executing `strip`), so ccache removes write permission to hardlinked files in the cache.

Could those two features also be implemented in sccache?

Issue #1053, opened 2021-10-12
Support reflinks & hardlinks


Logarithmus:
Having something like https://ccache.dev/manual/4.4.2.html#config_file_clone & https://ccache.dev/manual/4.4.2.html#config_hard_link would be super useful.

igxactly:
this would help storage use efficienty a lot in terms of both bandwidth and capacity!

PR #2640, opened 2026-03-15
feat: add file_clone option for reflink-based disk cache

quake:
### Summary

Add a `file_clone` option for the disk cache that stores cache entries as **uncompressed files** and restores them using filesystem reflinks (e.g. `clonefile()` on APFS or `FICLONE` on Linux).

When supported by the filesystem, restored artifacts share underlying storage blocks with the cache entry, avoiding both decompression and physical data duplication on disk.

This mode is particularly useful for workflows with frequent `cargo clean` + rebuild cycles (for example when switching branches), where build artifacts are repeatedly restored from cache.

Key points:

- New config option `file_clone = true` in `[cache.disk]`
  - env: `SCCACHE_FILE_CLONE=true`
  - default: `false`
- Cache entries stored as uncompressed directory structures
- Cache restores use reflink (via `reflink-copy`)
- Automatic fallback to the existing compressed ZIP+zstd format when reflinks are not supported
- Mixed-format reading: both compressed and uncompressed cache entries are supported

---

### Benchmark Results

Benchmarks were run on macOS (APFS) using three Rust projects:

- ripgrep
- fd
- bat

#### Build Times

| Project | Cold (Compressed) | Cold (File Clone) | Warm (Compressed) |
|--------|-------------------|-------------------|-------------------|
| fd | 1m 06s | 1m 04s | 46.70s |
| bat | 2m 03s | 2m 01s | 1m 02s |
| ripgrep | 28.34s | 28.42s | 9.84s |

Build times are effectively identical across both modes.

This indicates that `file_clone` does not materially change compile performance for typical Rust workloads.

---

#### Cache Storage

| Metric | Compressed | File Clone |
|------|-------------|------------|
| Cache size | 184 MiB | 523 MiB |
| Avg cache write | 0.001s | 0.008s |
| Avg cache read (hit) | 0.001s | <0.001s |

As expected, storing entries uncompressed increases cache size (~2.8×).

Read and write latencies remain sub-millisecond in both modes.

---

#### Disk Usage on Cache Restore

During warm builds, ~650 MiB of artifacts were restored from the cache.

On APFS, `df` reported **no additional disk usage**, confirming that restored files share underlying storage blocks with the cache via reflink.

For my local development workflow on a large project where I frequently switch branches and rebuild, this feature can reduce disk usage by **tens of gigabytes**. Since restored artifacts are reflinked from the cache instead of being fully copied, repeated cargo clean + rebuild cycles no longer multiply disk usage.

---

### Tradeoffs

|  | Compressed (default) | File Clone |
|---|---|---|
| Cache size | Smaller | Larger (~2.8×) |
| Disk usage on cache hit | Full copy of artifacts | Shared blocks (reflink) |
| Filesystem requirement | Any | APFS / XFS / Btrfs |
| Build time | Baseline | Essentially unchanged |

---

### Notes

This feature is conceptually similar to the `file_clone` option in **ccache**, which also leverages filesystem reflinks to avoid unnecessary data duplication.

Related issues:
https://github.com/mozilla/sccache/issues/1174
https://github.com/mozilla/sccache/issues/1053

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1332,7 +1332,7 @@ dependencies = [
  "iana-time-zone-haiku",
  "js-sys",
  "wasm-bindgen",
- "windows-core",
+ "windows-core 0.52.0",
 ]
 
 [[package]]
@@ -2452,6 +2452,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "reflink-copy"
+version = "0.1.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9efd944f26afa2406cbbabff39fac533c9bc24b13d7f1f12e14ae3e7bdc66cdb"
+dependencies = [
+ "cfg-if 1.0.0",
+ "libc",
+ "rustix",
+ "windows",
+]
+
+[[package]]
 name = "regex"
 version = "1.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2918,6 +2930,7 @@ dependencies = [
  "openssl",
  "predicates",
  "rand 0.8.5",
+ "reflink-copy",
  "regex",
  "reqsign 0.18.0",
  "reqwest",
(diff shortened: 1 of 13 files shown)


quake:
@sylvestre I force pushed to resolve a windows unit test error, could you help to trigger workflow running again, thanks.

PR #2641, opened 2026-03-15
wip: do most of the processing on the client side - instead of the server 

sylvestre:
see https://github.com/sylvestre/sccache/blob/97483f6178a4253990364025705b26d74c84eaf9/docs/Architecture.md for the description of the change


sylvestre:
moved here https://github.com/mozilla/sccache/pull/2642

PR #2649, opened 2026-03-23
Move cmake-modules to integration tests

Felixoid:
A few improvements:

- Use disk cache for the tests, avoid issues with GHA
- Add a dedicated failing cmake 4.X test ([example](https://github.com/mozilla/sccache/actions/runs/23459356336/job/68257326892?pr=2581))
- Move it to `make test-tools` to launch locally
- Move files to `tests/integration` directory
- Group logs in CI for easier reading

### DIFF
--- a/tests/integration/Makefile
+++ b/tests/integration/Makefile
@@ -22,6 +22,15 @@ else
 export SCCACHE_PATH := /sccache/target/debug/sccache
 endif
 
+# CI grouping helpers (GitHub Actions folding)
+ifdef CI
+group = @echo '::group::$(1)'
+endgroup = @echo '::endgroup::'
+else
+group =
+endgroup =
+endif
+
 BACKENDS := redis redis-deprecated memcached memcached-deprecated s3 azblob webdav basedirs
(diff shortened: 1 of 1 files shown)


Felixoid:
cc  @TroyKomodo

TroyKomodo:
This looks like a much better way to do the integration tests. More scalable!

Personally I would reach for snapshotting here for the json stats output since it shouldn't be changing run-to-run. Right now we just check that at least one file was cached.

This might be something for a broader change across all the integration tests.

Cc @sylvestre

Felixoid:
By snapshotting, I guess you mean the number of misses/hits. Because the full output looks redundant and can change at any time. Although the whole integration test checks whether the values change, not the exact numbers. Not sure if it's worth diverging now, or rather a small refactoring in the future.

Additionally, it was suspicious that tests work in docker and fail in https://github.com/mozilla/sccache/actions/runs/23459356336/job/68257326892?pr=2581#step:10:38. I added cmake 4.0 test with `pip install`, and now it catches the bug. I'll keep it here and create another issue.

PR #2665, opened 2026-04-03
Don't wait depfiles for gcc/clang preprocessed inputs

AJIOB:
Fixes #2664

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -699,23 +699,27 @@ where
         );
         profile_generate = true;
     }
-    if need_explicit_dep_target {
-        dependency_args.push(dep_flag);
-        dependency_args.push(dep_target.unwrap_or_else(|| output.clone().into_os_string()));
-    }
-    if let DepArgumentRequirePath::Missing = need_explicit_dep_argument_path {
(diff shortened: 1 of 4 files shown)


AJIOB:
Hi @sylvestre ,

Friendly ping for this PR (and anothers too, please).

I hope you have some time for it.

src/compiler/gcc.rs
```diff
@@ -699,23 +699,27 @@ where
...
-    if let Some(path) = dep_path {
-        outputs.insert(
-            "d",
-            ArtifactDescriptor {
-                path: path.clone(),
-                optional: false,
-            },
-        );
+    // If the language doesn't need preprocessing, it doesn't generate a dependency file. See issue #2664
+    if language.needs_c_preprocessing() {
```
AJIOB, line 704: Cannot do it without explicit external `if`, because clippy is not happy with `if let ... && ...` and with `if ... && let ...`

See https://github.com/rust-lang/rust/issues/53667 for more details

sylvestre:
please add integration tests for both gcc & clang
thanks

AJIOB:
Looks like freebsd error is not related to my changes.

So, the PR is ready.

@sylvestre , I've added the integration tests too

PR #2673, opened 2026-04-13
Normalize `-f*-prefix-map` args via `SCCACHE_BASEDIRS`

larseggert:
These compound flags embed absolute paths that `strip_basedirs` cannot reach via preprocessor-output normalization, causing cache misses when the same sources are built from different directories. Extract the `OLD` path component and strip basedir prefixes from it before hashing.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -27,7 +27,7 @@ use crate::dist::pkg;
 use crate::mock_command::CommandCreatorSync;
 use crate::util::{
     Digest, HashToDigest, MetadataCtimeExt, TimeMacroFinder, Timestamp, decode_path, encode_path,
-    hash_all, strip_basedirs,
+    hash_all, normalize_prefix_map_arg, strip_basedirs,
 };
 use async_trait::async_trait;
 use fs_err as fs;
@@ -1566,7 +1566,7 @@ impl<'a> HashKeyParams<'a> {
         m.update(CACHE_VERSION);
         m.update(self.language.as_str().as_bytes());
         for arg in self.arguments {
-            arg.hash(&mut HashToDigest { digest: &mut m });
+            normalize_prefix_map_arg(arg, self.basedirs).hash(&mut HashToDigest { digest: &mut m });
         }
         for hash in self.extra_hashes {
             m.update(hash.as_bytes());
--- a/src/util.rs
+++ b/src/util.rs
@@ -1109,6 +1109,45 @@ pub fn strip_basedirs<'a>(preprocessor_output: &'a [u8], basedirs: &[Vec<u8>]) -
     Cow::Owned(result)
 }
 
+const PREFIX_MAP_FLAGS: &[&[u8]] = &[
+    b"-fdebug-prefix-map=",
+    b"-fmacro-prefix-map=",
+    b"-ffile-prefix-map=",
+];
+
+/// Normalize a `-f{debug,macro,file}-prefix-map=OLD=NEW` compiler argument for cache key
+/// computation by stripping basedir prefixes from the OLD path component.
+///
+/// These flags embed absolute source or object directory paths. Because they are compound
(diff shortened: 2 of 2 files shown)


sylvestre:
@copilot resolve the merge conflicts in this pull request

experimenting!

src/compiler/c.rs
```diff
@@ -1566,7 +1566,7 @@ impl<'a> HashKeyParams<'a> {
         m.update(CACHE_VERSION);
         m.update(self.language.as_str().as_bytes());
         for arg in self.arguments {
-            arg.hash(&mut HashToDigest { digest: &mut m });
+            normalize_prefix_map_arg(arg, self.basedirs).hash(&mut HashToDigest { digest: &mut m });
```
sylvestre, line 1569: preprocessor_cache_entry_hash_key() still hashes the raw args, so in preprocessor cache mode we keep missing. please handle both (and dedup).

src/util.rs
```diff
@@ -1154,6 +1154,45 @@ pub fn strip_basedirs<'a>(preprocessor_output: &'a [u8], basedirs: &[Vec<u8>]) -
     Cow::Owned(result)
 }
 
+const PREFIX_MAP_FLAGS: &[&[u8]] = &[
```
sylvestre, line 1157: what about -fcoverage-prefix-map= and -fprofile-prefix-map=?

src/util.rs
```diff
@@ -1154,6 +1154,45 @@ pub fn strip_basedirs<'a>(preprocessor_output: &'a [u8], basedirs: &[Vec<u8>]) -
...
+    b"-fdebug-prefix-map=",
+    b"-fmacro-prefix-map=",
+    b"-ffile-prefix-map=",
+];
+
+/// Normalize a `-f{debug,macro,file}-prefix-map=OLD=NEW` compiler argument for cache key
```
sylvestre, line 1163: nobody will read a 10 lines doc comment :) 2 lines is enough

src/util.rs
```diff
@@ -1154,6 +1154,45 @@ pub fn strip_basedirs<'a>(preprocessor_output: &'a [u8], basedirs: &[Vec<u8>]) -
...
+///
+/// Returns `Cow::Owned` with the normalized argument if a basedir was stripped, or
+/// `Cow::Borrowed` of the original if the argument was not a prefix-map flag or no basedir matched.
+pub fn normalize_prefix_map_arg<'a>(arg: &'a OsStr, basedirs: &[Vec<u8>]) -> Cow<'a, OsStr> {
+    fn try_normalize(bytes: &[u8], basedirs: &[Vec<u8>]) -> Option<Vec<u8>> {
+        let (flag, rest) = PREFIX_MAP_FLAGS
+            .iter()
+            .find_map(|&flag| Some((flag, bytes.strip_prefix(flag)?)))?;
+        // Format is OLD=NEW; split on the first '='
+        let sep = rest.iter().position(|&b| b == b'=')?;
```
sylvestre, line 1179: what if OLD contains a '='? gcc splits on the last one iirc

src/util.rs
```diff
@@ -1154,6 +1154,45 @@ pub fn strip_basedirs<'a>(preprocessor_output: &'a [u8], basedirs: &[Vec<u8>]) -
...
+        let old = &rest[..sep];
+        // Strip the longest matching basedir prefix from OLD
+        let stripped_old = basedirs
+            .iter()
+            .filter(|dir| old.starts_with(dir.as_slice()))
```
sylvestre, line 1184: on windows basedirs are lowercased and slash-normalized (config.rs) while the arg isn't, so this never matches - strip_basedirs normalizes the haystack first, could you do the same here?

src/util.rs
```diff
@@ -1670,6 +1710,41 @@ mod tests {
         );
```
sylvestre, line 1710: please add a test in c.rs that two identical builds from different dirs give the same hash key — that's the actual bug here.

PR #2675, opened 2026-04-14
client: implement client-side compilation mode

sylvestre:
Instead of running the compilation in the server, it is done on the client level.
I had 2 digits per improvements on Firefox

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,8 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v5
 
+      # TODO: restore `cargo test` once per-feature tests are compatible with
+      # both server-side and client-side compilation modes.
       - name: Check feature ${{ matrix.feature }}
         run: cargo check --no-default-features --features ${{ matrix.feature }}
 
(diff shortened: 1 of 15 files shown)


sylvestre:
see
https://github.com/sylvestre/sccache/blob/client3/docs/Architecture.md#client-side-compilation-new-mode

vs
https://github.com/sylvestre/sccache/blob/client3/docs/Architecture.md#server-side-compilation-legacy-mode

glandium:
> see https://github.com/sylvestre/sccache/blob/client3/docs/Architecture.md#client-side-compilation-new-mode
> 
> vs https://github.com/sylvestre/sccache/blob/client3/docs/Architecture.md#server-side-compilation-legacy-mode

> Why this is fast: preprocessing in client-side mode is cheap — it only concatenates source files rather than running the full C/C++ preprocessor. This avoids the expensive #include expansion and macro evaluation that dominates traditional preprocessing time, making it practical to move this work to the client without a performance penalty.

Huh, I haven't even looked at the implementation, but this is a huge red flag. This shouldn't be changing anything about preprocessing. The preprocessing part is a very important part of how sccache works and it can't be skipped except in direct mode, but that wouldn't be a difference between the server mode and the client mode anyways. Direct mode, by the way, does not appear in neither graph.

glandium:
Superseded by #2735

PR #2676, opened 2026-04-17
fix: handle directories in dep-info source file hashing

mmastrac:
This patch has been in testing in the vercel/sccache fork for a number of weeks.

When `proc_macro::tracked::path()` registers a directory as a dependency, rustc's `dep-info` output includes the directory path. sccache previously crashed with "Is a directory" when trying to hash these paths.

Now recursively hashes all files within the directory (sorted for determinism), using relative paths as delimiters. This correctly captures directory dependencies so cache invalidation works when any file in the tracked directory changes.

Note that `proc_macro_tracked_path` is a nightly-only feature, but this nightly feature is being used by an `include_dirs` fork used by the next.js team. The hashing falls back to the same file-specific hash as before in all other cases.

Fixes: https://github.com/mozilla/sccache/issues/2653


sylvestre:
thanks

PR #2678, opened 2026-04-17
feat: strip basedirs from Rust hash key for cross-machine cache hits

### LINKED ISSUE #2652: Wire SCCACHE_BASEDIRS into Rust hash key
### Problem

`SCCACHE_BASEDIRS` was done by <https://github.com/mozilla/sccache/pull/2521>. Pretty impressive!

However, that only deals with C compiler. Rust compiler wasn't covered.

In Rust, the most popular build tool Cargo by default using absolute paths everywhere. Such as

* Absolute source paths of registry dependencies and local crates are included in the `--extern` and `-L` flags.
* If you are moving `CARGO_HOME` around, the path will change. See <https://github.com/rust-lang/cargo/issues/10915>
* There is not `RUSTLINKFLAGS`, while one of the most common absolute path pattern in rustflags are linker flags. See <https://github.com/rust-lang/cargo/issues/4349>.
* `-Ztrim-paths` might be a way to partially fix this (as part of fixing the artifaict content hash), but that would also introduce more `--remap-path-prefix=…`, which will contain more aboslute paths. (See <https://github.com/rust-lang/cargo/issues/12137>)

I think at least we need to implement a basic support to eliminate some of the most common absolute paths from the cache key, like `--remap-path-prefix=/abs/path`, `-Clinker=/abs/path`, `-Clin-arge=something,/abs/path`, and others.

### Drawback

`SCCACHE_BASEDIRS` only normalize paths for cache keys not the real compiler invocations. That means embedded path in binaries are shared across different build root.

This might be a user problem, as they can fix it via prefix-map flag, like `-fdebug-prefix-map`, `-ffile-prefix-map`, `-fmacro-prefix-map` in C compiler and `--remap-path-prefix` in rustc.

mmastrac:
`SCCACHE_BASEDIRS` now normalizes cwd, `CARGO_MANIFEST_DIR`, `CARGO_WORKSPACE_DIR`, `CARGO_TARGET_TMPDIR`, `CARGO_MANIFEST_PATH`, `CARGO_BIN_EXE_*`, `dep-info` env var values, and the concatenated argument string in the Rust compiler's hash key computation. This enables cache hits when the same crate is compiled from different absolute paths on different machines (e.g., CI runners with different checkout roots).

This handles arguments that contain absolute paths at non-terminal positions, such as `-Clink-arg=blah,/path/to/something` or (the nightly-only?) `--remap-path-prefix=/full/path=/new`.

Added a number of unit tests around the functionality, as well as an integration test where we compile a crate from two different directories.

This fixes https://github.com/mozilla/sccache/issues/2652

### DIFF
--- a/src/cache/readonly.rs
+++ b/src/cache/readonly.rs
@@ -110,7 +110,7 @@ mod test {
 
     #[test]
     fn readonly_storage_is_readonly() {
-        let storage = ReadOnlyStorage(Arc::new(MockStorage::new(None, false)));
+        let storage = ReadOnlyStorage(Arc::new(MockStorage::default()));
         assert_eq!(
             storage.check().now_or_never().unwrap().unwrap(),
             CacheMode::ReadOnly
@@ -119,8 +119,7 @@ mod test {
 
     #[test]
     fn readonly_storage_forwards_preprocessor_cache_mode_config() {
-        let storage_no_preprocessor_cache =
-            ReadOnlyStorage(Arc::new(MockStorage::new(None, false)));
+        let storage_no_preprocessor_cache = ReadOnlyStorage(Arc::new(MockStorage::default()));
         assert!(
             !storage_no_preprocessor_cache
                 .preprocessor_cache_mode_config()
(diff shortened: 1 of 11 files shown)


cdbattags:
Hi folks, any reason this can't go in yet? Needing the same myself!

sylvestre:
because most of the jobs fails

cdbattags:
@sylvestre - Oops, roger that, I'll start on a fork and try to clean up the PR!

@mmastrac - Any more updates to the Vercel fork since April that should ride along?

PR #2679, opened 2026-04-17
fix: replace panicking expect() with map_or() for edge-case file paths

srpatcha:
# Fix panic on edge-case file paths in LRU disk cache init

## Problem

`LruDiskCache::init()` in `src/lru_disk_cache/mod.rs` calls `.expect("Bad path?")` on `file.file_name()`, which panics if the path ends in `..` or is otherwise unusual. A build cache tool should handle unexpected file system states gracefully rather than crashing.

## Root Cause

`Path::file_name()` returns `None` when the path terminates in `..` or consists solely of a root or prefix. The `.expect()` call converts this into a panic. While uncommon, symlinks or filesystem corruption could produce such paths in the cache directory.

## Fix

Replaced `.expect("Bad path?").starts_with(TEMPFILE_PREFIX)` with `.map_or(false, |name| name.starts_with(TEMPFILE_PREFIX))`. Paths without a valid file name component are now treated as non-temporary files (skipping the cleanup branch) rather than crashing.

## Testing

- Create a cache directory containing a path component ending in `..`.
- Previously: sccache panics during cache init. Now: the entry is handled gracefully.
- Normal cache operation should be unaffected.

## Impact

Affects sccache users whose cache directories may contain unusual file paths. A panic in cache init prevents the entire build cache from working.


sylvestre:
you pr is doing too many different things  (like fix the panic and change files permissions)
please split it into several prs

srpatcha:
Hi @sylvestre, thank you for the feedback! You're right — I'll split this into separate PRs:

1. Fix the panicking expect() with map_or()
2. File permission changes (separate concern)

I'll close this PR and submit them individually. Thanks for the review!

srpatcha:
Closing this PR as requested by @sylvestre. Split into focused PRs:

1. #2688 - fix: replace panicking expect() with map_or() for file_name() check
2. #2689 - fix: rollback pending state on tempfile failure in prepare_add

The cache statistics module will be submitted as a separate PR. Thank you for the feedback!

PR #2688, opened 2026-04-28
fix: replace panicking expect() with map_or() for file_name() check

srpatcha:
## Summary

\Path::file_name()\ returns \None\ for paths ending in \..\ which caused a panic via \.expect()\. Replace with \.map_or()\ to handle unusual paths gracefully.

This is split from #2679 per reviewer feedback to keep changes focused.

## Changes

- \src/lru_disk_cache/mod.rs\: Replace \.expect()\ with \.map_or(false, ...)\ in the file_name check

Signed-off-by: Srikanth Patchava <spatchava@meta.com>

### DIFF
--- a/src/lru_disk_cache/mod.rs
+++ b/src/lru_disk_cache/mod.rs
@@ -188,8 +188,7 @@ impl LruDiskCache {
         for (file, size) in get_all_files(&self.root) {
             if file
                 .file_name()
-                .expect("Bad path?")
-                .starts_with(TEMPFILE_PREFIX)
+                .is_some_and(|name| name.starts_with(TEMPFILE_PREFIX))
             {
                 fs::remove_file(&file).unwrap_or_else(|e| {
                     error!("Error removing temporary file `{}`: {}", file.display(), e);
--- a/src/lru_disk_cache/mod.rs
+++ b/src/lru_disk_cache/mod.rs
@@ -728,4 +728,54 @@ mod tests {
         assert!(!f.tmp().join("cache").join("file2").exists());
         assert!(!p4.exists());
     }
+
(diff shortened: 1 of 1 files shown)


srpatcha:
Hi, fixed the clippy `unnecessary_map_or` lint by switching to `is_some_and`:

\\\diff
- .map_or(false, |name| name.starts_with(TEMPFILE_PREFIX))
+ .is_some_and(|name| name.starts_with(TEMPFILE_PREFIX))
\\\

This is the canonical idiom and matches what clippy stable suggests. Force-pushed; CI should pass on ubuntu, macOS, and windows now.

sylvestre:
could you please add a test to make sure we don't regress in the future? thanks

srpatcha:
Hi @sylvestre, thank you for the suggestion! Added two regression tests in commit `a41fea6`:

1. **`test_init_with_leftover_tempfile_does_not_panic`** — Drops a `.sccachetmpXXX` file into the cache root, then opens the cache via `LruDiskCache::new()`. Before the fix, this could panic at `.expect("Bad path?")` when `file_name()` returned `None`. After the fix, `is_some_and()` handles it safely. Also asserts the stale tempfile is cleaned up.

2. **`test_init_mixed_files_no_panic`** — Cache root contains both regular cache files and tempfile-prefixed files. Verifies that `init()` removes only the tempfiles and preserves regular entries, exercising the full `file_name()` branch.

These guard against future regressions of the panic this PR fixes. Ready for re-review!

srpatcha:
Hi @sylvestre, the two regression tests have been added (commit a41fea6) and clippy's is_some_and suggestion applied. Ready for re-review. Thanks!

PR #2689, opened 2026-04-28
fix: rollback pending state on tempfile failure in prepare_add

srpatcha:
## Summary

\prepare_add\ updated pending size before \	empfile_in()\, which never rolled back on failure. This caused \make_space()\ to panic later when encountering negative available space.

Moved state update after successful tempfile creation and replaced \.expect()\ with \?\ operator.

This is split from #2679 per reviewer feedback to keep changes focused.

## Changes

- \src/lru_disk_cache/mod.rs\: Move pending size update after tempfile success
- \src/server.rs\: Replace \.expect()\ with \?\ operator

Signed-off-by: Srikanth Patchava <spatchava@meta.com>

### DIFF
--- a/src/lru_disk_cache/mod.rs
+++ b/src/lru_disk_cache/mod.rs
@@ -728,4 +728,42 @@ mod tests {
         assert!(!f.tmp().join("cache").join("file2").exists());
         assert!(!p4.exists());
     }
+
+    #[test]
+    fn test_prepare_add_rollback_on_tempfile_failure() {
+        // Verify that if tempfile creation fails inside prepare_add, the
+        // cache's `pending` and `pending_size` state is NOT incremented.
+        // Before the fix, prepare_add updated state BEFORE creating the
+        // tempfile, so a tempfile failure left pending_size > 0 forever,
+        // which later caused make_space() to panic with negative space.
+        let f = TestFixture::new();
+        let cache_dir = f.tmp().join("cache");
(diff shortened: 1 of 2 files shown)


sylvestre:
would it be possible to have a unit test to cover this change? thanks

srpatcha:
Hi @sylvestre, thank you for the suggestion! Added a regression test in commit `0a5cfd8`:

`test_prepare_add_rollback_on_tempfile_failure` exercises the failure path by:

1. Creating an LruDiskCache normally
2. Removing the cache root directory underneath it
3. Calling prepare_add(), which now fails because tempfile_in() can't create a file in a missing dir
4. Asserting that `c.size() == baseline_size` (no pending state leaked)

Before the fix, the test would fail with `size() == 10` because the old code incremented `pending_size` before the tempfile creation. The test serves as a regression guard for #2689.

sylvestre:
please run rustfmt on the change

srpatcha:
Hi @sylvestre, ran `rustfmt 1.9.0` on the test (commit `7687b70`). The only formatting change was wrapping the multi-arg `assert!()` macro across 4 lines per the standard rustfmt style. Force-pushed. Ready for re-review!

srpatcha:
Hi @sylvestre, rustfmt has been applied (commit 7687b70) and the regression test is in place. Ready for re-review whenever you get a chance. Thanks!

PR #2690, opened 2026-04-29
prepare release 0.15.0

sylvestre:


weihanglo:
@sylvestre thanks for the release 🎉 

BTW, do we have plan publishing this to crates.io?

PR #2691, opened 2026-04-29
feat: Make REQUEST_TIMEOUT_SECS and CONNECT configurable

### LINKED ISSUE #2276: make "REQUEST_TIMEOUT_SECS" configurable
5 minutes is too short to compile some complex kernels, so I bumped to 10 minutes here. It would be good to make this configurable.

_Originally posted by @trxcllnt in https://github.com/mozilla/sccache/pull/2247#discussion_r1729737863_

nunokaeru:
Some kernels can take upwards of 5 minutes to compile but 10 minutes might be too high of a timeout in the most common use cases.

Add a runtime configuration option for both the Request and the Connect timeout.

Closes: https://github.com/mozilla/sccache/issues/2276

### DIFF
--- a/src/dist/http.rs
+++ b/src/dist/http.rs
@@ -1081,9 +1081,10 @@ mod client {
     use futures::TryFutureExt;
     use reqwest::Body;
     use std::collections::HashMap;
+    use std::env;
     use std::io::Write;
     use std::path::{Path, PathBuf};
-    use std::sync::{Arc, Mutex};
+    use std::sync::{Arc, LazyLock, Mutex};
     use std::time::Duration;
 
     use super::common::{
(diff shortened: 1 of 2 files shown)


src/dist/http.rs
```diff
@@ -1159,7 +1165,11 @@ mod client {
                 );
             }
             // Finish the client
-            let timeout = Duration::new(REQUEST_TIMEOUT_SECS, 0);
+            let timeout_env: u64 = env::var("SCCACHE_REQUEST_TIMEOUT_SECS")
```
AJIOB: nit: move this to the dedicated function, please. Or keep the default value as a global define - it should be configurable in one place

nunokaeru: Opted to use LazyCell<T> might be overkill but then they are both const and only one place. Should I add the timeout to the new async client with certificates as well?

AJIOB: IMO, yes

AJIOB (approved):

sylvestre:
please add a test to make sure we won't regress

nunokaeru:
> please add a test to make sure we won't regress

Will do and fix clippy issues too, might take a while as I’m lacking free time

nunokaeru:
> please add a test to make sure we won't regress

Thank you for the wait. Fixed the clippy issues and added a test for each variable

AJIOB (approved):

PR #2697, opened 2026-05-06
Avoid double-caching on Solus and AerynOS

ReillyBrogan:
Following up from https://github.com/mozilla/sccache/pull/2524 add the ccache path that Solus uses (`/usr/lib64/ccache/bin`) and the one that AerynOS uses (`/usr/lib/ccache/bin`).

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1304,7 +1304,9 @@ mod tests {
         // Create a PATH with ccache directories
         let path_with_ccache = env::join_paths([
             "/usr/lib64/ccache",
+            "/usr/lib64/ccache/bin",
             "/usr/lib/ccache",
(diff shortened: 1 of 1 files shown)


ahartmetz:
If I understand this correctly, you don't understand this correctly: This "only" adds the compiler wrapper paths from your distro to the tests to to make sure that they are recognized as compiler wrappers. That is fine, I guess, but it's not what makes the avoidance of double-caching work. That logic recognizes (among other things) a "ccache" path component, which is present in the added test paths, so it should work.

I don't make the decision whether or not to merge this, but if you want to have a chance at getting it merged, you should adjust the commit message accordingly.

PR #2703, opened 2026-05-11
gcc: Force local compilation on dist-client when assembler has .incbin

lissyx:
Assembly files with .incbin would trigger a build failure when running over sccache-dist:
> 0:03.40 E /home/alex/firefox/gecko-cinnabar/config/external/icu/data/icu_data.S:28:9: error: Could not find incbin file '/home/alex/firefox/gecko-cinnabar/config/external/icu/data/icudt78l.dat'

Force disable distributed compilation when hitting such a file.

### DIFF
--- a/src/compiler/gcc.rs
+++ b/src/compiler/gcc.rs
@@ -986,6 +986,24 @@ where
     let has_verbose_flag = arguments.contains(&OsString::from("-v"))
         || arguments.contains(&OsString::from("--verbose"));
 
+    #[cfg(feature = "dist-client")]
+    let assembler_has_incbin = match Language::from_file_name(parsed_args.input.clone().as_path()) {
+        Some(Language::AssemblerToPreprocess) => {
+            let mut contents = String::new();
+            let res = File::open(parsed_args.input.clone())
+                .and_then(|mut f| f.read_to_string(&mut contents));
+            if res.is_err() {
+                false
+            } else {
+                contents.contains(".incbin")
+            }
+        }
+        _ => false,
+    };
+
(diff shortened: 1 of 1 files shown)


src/compiler/gcc.rs
```diff
@@ -986,6 +986,24 @@ where
     let has_verbose_flag = arguments.contains(&OsString::from("-v"))
         || arguments.contains(&OsString::from("--verbose"));
 
+    #[cfg(feature = "dist-client")]
+    let assembler_has_incbin = match Language::from_file_name(parsed_args.input.clone().as_path()) {
+        Some(Language::AssemblerToPreprocess) => {
```
AJIOB, line 991: This should be checked for any assembly or C language, because the last one may contain inline ASM with that directive and the same problem

src/compiler/gcc.rs
```diff
@@ -986,6 +986,24 @@ where
...
+            let mut contents = String::new();
+            let res = File::open(parsed_args.input.clone())
```
AJIOB, line 993: And we cannot just use the source file - we must check the preprocessor output - the `.incbin` directive may come from the included header file

src/compiler/gcc.rs
lissyx: I'm afraid it's going to be a lot of useless IOs reading those content then?

AJIOB:
One more question about your original issue: can you use relative path as an `.incbin` input? Does it solve your problem?

P.S. The problem looks like the same we have with an `#include "/some/absolute/path"` directive

PR #2704, opened 2026-05-11
fix: use single-threaded tokio runtime in sccache (dist-)client

AJIOB:
It's the #2460 follow-up with `dist-client` similar fixes.

Thanks, @venkyt-arista, for the original fix

### DIFF
--- a/src/dist/client_auth.rs
+++ b/src/dist/client_auth.rs
@@ -536,11 +536,10 @@ pub fn get_token_oauth2_code_grant_pkce(
     mut auth_url: Url,
     token_url: &str,
 ) -> Result<String> {
-    // We don't need to create multi-threaded runtime here:
-    // every client will create it's own NUM_THREADS,
-    // and with 96-core server we will have 96 x 96 = 9216 threads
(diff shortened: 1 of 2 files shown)


AJIOB:
Hi @sylvestre ,

Is it ok to you?

I don't think we can add some useful testing for that changes.

src/dist/client_auth.rs
```diff
@@ -537,7 +536,9 @@ pub fn get_token_oauth2_code_grant_pkce(
     mut auth_url: Url,
     token_url: &str,
 ) -> Result<String> {
-    let runtime = Runtime::new()?;
+    let runtime = tokio::runtime::Builder::new_current_thread()
```
sylvestre, line 543: please document this line, it is far from obvious for non-tokio expert :)

sylvestre:
and please udpate the commit message to explain the why

AJIOB:
@sylvestre , done, thanks

AJIOB:
Hi @sylvestre ,

Graceful ping.

CI was green with yesterday evening latest changes

AJIOB:
P.S. CI is green

PR #2705, opened 2026-05-11
feat: all backends support making them as read-only

AJIOB:
For the local developers we want to provide the access to the remote cache server in read-only mode.

So, this PR allows read-only access to read-write storages via per-storage specific flag

### DIFF
--- a/docs/Azure.md
+++ b/docs/Azure.md
@@ -6,4 +6,6 @@ the container for you - you'll need to do that yourself.
 
 You can also define a prefix that will be prepended to the keys of all cache objects created and read within the container, effectively creating a scope. To do that use the `SCCACHE_AZURE_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
 
+The `SCCACHE_AZURE_RW_MODE` environment variable can be set to `READ_ONLY` to make sccache use this backend in read-only mode. The default is `READ_WRITE`.
+
 **Important:** The environment variables are only taken into account when the server starts, i.e. only on the first run.
--- a/src/config.rs
+++ b/src/config.rs
@@ -236,6 +236,8 @@ pub struct AzureCacheConfig {
     pub connection_string: String,
     pub container: String,
     pub key_prefix: String,
+    #[serde(default)]
+    pub rw_mode: CacheModeConfig,
 }
 
 /// Configuration switches for preprocessor cache mode.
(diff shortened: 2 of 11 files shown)


AJIOB:
Hi @sylvestre . Is it ok to you?

src/config.rs
```diff
@@ -974,20 +1017,10 @@ fn config_from_env() -> Result<EnvConfig> {
...
-                warn!("Invalid SCCACHE_GCS_RW_MODE -- defaulting to READ_ONLY.");
-                CacheModeConfig::ReadOnly
-            }
-            _ => {
-                warn!("No SCCACHE_GCS_RW_MODE specified -- defaulting to READ_ONLY.");
-                CacheModeConfig::ReadOnly
-            }
-        };
+        let rw_mode = cache_mode_from_env_var("SCCACHE_GCS_RW_MODE").unwrap_or_else(|| {
+            warn!("No SCCACHE_GCS_RW_MODE specified -- defaulting to READ_ONLY.");
```
sylvestre: If a user sets an invalid value (e.g.  CACHE_GCS_RW_MODE=readonly_typo), cache_mode_from_env_var first warns "... must be 'READ_ONLY' or 'READ_WRITE'", returns None, and then this prints "No SCCACHE_GCS_RW_MODE specified" - which is false; it was specified.

docs/Azure.md
```diff
@@ -6,4 +6,6 @@ the container for you - you'll need to do that yourself.
 
 You can also define a prefix that will be prepended to the keys of all cache objects created and read within the container, effectively creating a scope. To do that use the `SCCACHE_AZURE_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
 
+The `SCCACHE_AZURE_RW_MODE` environment variable can be set to `READ_ONLY` to make the server read-only. The default is `READ_WRITE`.
```
sylvestre:
```suggestion
The `SCCACHE_AZURE_RW_MODE` environment variable can be set to `READ_ONLY` to make sccache use this backend in read-only mode. The default is `READ_WRITE`.
```
no ?

sylvestre:
please add more tests:
- No test for the check() short-circuit returning ReadOnly when rw_mode is set
- No test for cache_mode_from_env_var parsing

AJIOB:
Thanks, will fix the PR.

Looks like the read-only short-circuit was tested via `test_operator_storage_redis_with_read_only` (was added in this PR)

AJIOB:
I've added 2 tests in that PR:
* `test_operator_storage_redis_with_read_only` to check read-only short-curcuit via remote storage access.. The local storage logic wasn't changed, so not tested
* `test_cache_mode_from_env_var` to check the logic of new function `cache_mode_from_env_var`

AJIOB:
@sylvestre , I've fixed all you suggested, thanks. Is it OK now?

AJIOB:
Hi @sylvestre,

Graceful ping about the PR

PR #2708, opened 2026-05-13
feat(nvcc): support argument: `--dependency-output`

ZRHann:
This adds `--dependency-output` to nvcc::ARGS, which is equivalent to gcc's `-MF` flag. Both flags map to the same `DepArgumentPath` variant and follow identical code paths in gcc::parse_arguments().

Without this fix, the file path following `--dependency-output` is treated as a `Raw` input file. When the actual `.cu` source is then seen as a second input, the parser returns `CannotCache("multiple input files")`, leading to `UnhandledCompile` and a cache miss.

docs for this flag: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#dependency-output-file-mf


ZRHann:
@sylvestre Could you take a look when you have time?

ZRHann:
@sylvestre Could you take a look when you have time?

sylvestre:
sorry, i missed your initial ping

CI started!

ZRHann:
@codspeedbot explain why this is faster

PR #2709, opened 2026-05-14
feat: add option to not check build server source IP

iTrooz:
Hello !

This PR adds an option to be able to disable the scheduler checking the IP of the build server when connecting.
While this option is useful in some cases, it can be annoying for more complex setups. For example, I am currently using sccache on servers behind a NAT that I do not control, and use port redirection services (like ngrok) to talk to the build servers.
This means that a build server does not have a single "public IP", but use a different IP when initiating or receiving connections.

Currently, I'm getting around that by adding a reverse-proxy in front of the scheduler, adding a fake X-Real-IP header, to force the scheduler to accept build server connections from any source IP. But I'd like to simplify my setup, hence this option

The default behaviour is unchanged

### DIFF
--- a/docs/Distributed.md
+++ b/docs/Distributed.md
@@ -278,6 +278,8 @@ of the scheduler and instruct clients to configure their `scheduler_url` with th
 appropriate `https://` address. The scheduler will verify the server's IP in this
 configuration by inspecting the `X-Real-IP` header's value, if present. The webserver
 used in this case should be configured to set this header to the appropriate value.
+This verification is enabled by default and can be disabled by setting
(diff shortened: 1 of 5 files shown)


sylvestre:
please update the doc and also add a test to check this (esp in case of mismatch)
thanks

iTrooz:
Hello ! I added tests in case the IP is mismatched, although this specific logic wasn't touched by my change (I don't even check the IP if check_server_ip = false)

Is this enough, or do you want me to add a test in https://github.com/mozilla/sccache/blob/main/tests/dist.rs to test the feature of this PR all the way from config file to the condition ?

PR #2711, opened 2026-05-15
Strip basedirs from cmdline args and direct-mode hash key

solarispika:
Follow-up to #2521. The basedirs strip ran only on preprocessor output, leaving two leaks:

1. `common_args` is `OsStr::hash`ed verbatim, so path-bearing flags (e.g. `-external:I <path>`) embed absolute paths in the key.
2. On Windows, MSVC's preprocessor emits `__FILE__` strings in C source form (`\\` escapes); `normalize_win_path` then maps each `\\` to `//` and never matches a basedir.
3. The direct-mode key (`preprocessor_cache_entry_hash_key`) hashes args the same un-stripped way.

Fix in 3 commits, one per gap. With `SCCACHE_BASEDIRS` unset the new code is byte-identical to before. No `CACHE_VERSION` bump — existing entries were per-workdir and unreachable anyway.

## Tests

Unit test per commit (args strip preprocessor mode, args strip direct mode, byte-level backslash collapse, Windows full round-trip). `cargo test --lib --release` passes on macOS 15.2 / apple-clang 17, Debian 12 / gcc-12, Windows 11 / MSVC 14.50. End-to-end: 100% cache hit rate across our 10-cell matrix (was 0% before).

## Not in this PR

- Concatenated path args (`-IC:\path`, `/imsvc<path>`) — `strip_basedirs`'s boundary check rejects mid-arg matches. On MSVC these flow through `preprocessor_args` (not `common_args`), so they're stripped via the preprocessor-output path already.
- `PreprocessorCacheEntry` still stores include paths absolute, so direct-mode lookups across workspaces fall back to the preprocessor-mode key (which now hashes correctly). Cross-workspace direct-mode hits are a separate PR.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -1566,7 +1566,11 @@ impl<'a> HashKeyParams<'a> {
         m.update(CACHE_VERSION);
         m.update(self.language.as_str().as_bytes());
         for arg in self.arguments {
-            arg.hash(&mut HashToDigest { digest: &mut m });
+            let arg_bytes = arg.as_encoded_bytes();
+            let stripped = strip_basedirs(arg_bytes, self.basedirs);
+            m.update(&stripped);
+            // Separator so adjacent args don't blur together.
+            m.update(&[0u8]);
         }
         for hash in self.extra_hashes {
             m.update(hash.as_bytes());
@@ -1759,6 +1763,64 @@ mod test {
             b"/home/user1/project".to_vec(), // This should match (longest)
         ];
         assert_eq!(h1, hash_with_basedirs(preprocessed1, &multi_basedirs));
+    }
+
+    #[test]
+    fn test_hash_key_args_basedirs() {
+        // Args carry absolute paths into the cache key for any flag whose
+        // disposition ends up in `common_args` with the path on its own
+        // (Separated disposition). On MSVC, `-external:I <path>` is the
+        // canonical case. Verify that basedirs strip neutralises the
+        // per-workspace difference.
+        let basedirs = [
+            b"/home/user1/project".to_vec(),
+            b"/home/user2/project".to_vec(),
+        ];
+        let preprocessed = b"int main() { return 0; }";
+
+        let args1: Vec<OsString> = ovec![
+            "-c",
+            "-external:I",
+            "/home/user1/project/include",
+            "-o",
+            "/home/user1/project/build/main.o"
+        ];
+        let args2: Vec<OsString> = ovec![
+            "-c",
+            "-external:I",
+            "/home/user2/project/include",
+            "-o",
+            "/home/user2/project/build/main.o"
+        ];
+
+        let h1 = HashKeyParams::new("abcd", Language::C, &args1, preprocessed)
+            .with_basedirs(&basedirs)
+            .compute();
+        let h2 = HashKeyParams::new("abcd", Language::C, &args2, preprocessed)
+            .with_basedirs(&basedirs)
+            .compute();
+        assert_eq!(
+            h1, h2,
+            "Args with workspace-relative absolute paths should hash equal under basedirs"
+        );
+
+        // Without basedirs, same args produce divergent hashes — proves args
+        // (not just preprocessor output) feed into the key.
+        let h1_nb = HashKeyParams::new("abcd", Language::C, &args1, preprocessed).compute();
+        let h2_nb = HashKeyParams::new("abcd", Language::C, &args2, preprocessed).compute();
+        assert_neq!(h1_nb, h2_nb);
+
+        // Sanity: adjacent args must not blur — the inter-arg separator
+        // matters. `["-Ia", "b"]` and `["-I", "ab"]` are distinct cmdlines.
+        let args_a: Vec<OsString> = ovec!["-Ia", "b"];
+        let args_b: Vec<OsString> = ovec!["-I", "ab"];
+        let h_a = HashKeyParams::new("abcd", Language::C, &args_a, preprocessed)
+            .with_basedirs(&basedirs)
+            .compute();
+        let h_b = HashKeyParams::new("abcd", Language::C, &args_b, preprocessed)
+            .with_basedirs(&basedirs)
+            .compute();
+        assert_neq!(h_a, h_b);
     }
 
     #[test]
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -1448,6 +1448,38 @@ impl pkg::ToolchainPackager for CToolchainPackager {
 /// The cache is versioned by the inputs to `HashKeyParams::compute`.
 pub const CACHE_VERSION: &[u8] = b"12";
 
+/// Collapse `\\` byte pairs into a single `\`.
+#[cfg(target_os = "windows")]
+fn collapse_escaped_backslashes(input: &[u8]) -> Vec<u8> {
+    let mut iter = input.iter().peekable();
+    std::iter::from_fn(move || {
+        let &b = iter.next()?;
+        if b == b'\\' {
+            iter.next_if_eq(&&b'\\');
+        }
+        Some(b)
+    })
+    .collect()
+}
+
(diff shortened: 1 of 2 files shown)


sylvestre:
Given if it is clearly written by AI, could you please refactor to only keep the relevant comments ? 

Also a long comment #0 isn't helpful

src/compiler/c.rs
```diff
@@ -1448,6 +1448,28 @@ impl pkg::ToolchainPackager for CToolchainPackager {
 /// The cache is versioned by the inputs to `HashKeyParams::compute`.
 pub const CACHE_VERSION: &[u8] = b"12";
 
+/// Collapse `\\` byte pairs into a single `\`. Used to undo the C/C++
+/// source-form escaping of backslashes before basedirs prefix-stripping
+/// runs against preprocessor output text. See the call site in
+/// [`HashKeyParams::compute`] for the precise reason this is Windows-only
+/// at runtime; we compile it on every platform so its unit test can run
+/// in any CI matrix.
+#[cfg_attr(not(target_os = "windows"), allow(dead_code))]
```
sylvestre: please don't add allow dead code

src/compiler/c.rs
```diff
@@ -1448,6 +1448,28 @@ impl pkg::ToolchainPackager for CToolchainPackager {
...
+fn collapse_escaped_backslashes(input: &[u8]) -> Vec<u8> {
+    let mut out = Vec::with_capacity(input.len());
```
sylvestre: please refactor to avoid the mut if possible

src/compiler/c.rs
```diff
@@ -1565,8 +1587,19 @@ impl<'a> HashKeyParams<'a> {
         m.update(&[self.plusplus as u8]);
         m.update(CACHE_VERSION);
         m.update(self.language.as_str().as_bytes());
+        // Strip basedirs from each cmdline arg before hashing. Path-bearing
```
sylvestre: why adding these comments?

src/compiler/c.rs
```diff
@@ -1580,8 +1613,30 @@ impl<'a> HashKeyParams<'a> {
...
+        //
+        // On Windows, collapse `\\` (escaped backslash in C/C++ source-form
+        // string literals) into `\` first. MSVC's preprocessor expands
+        // `__FILE__` inside macros like `_wassert(L"...", L"<file>", line)`
+        // and keeps the path as a *source-form* string with escaped
+        // backslashes. Without this collapse, `normalize_win_path` inside
+        // `strip_basedirs` turns each `\` into `/` byte-for-byte, so the
+        // source-level `\\path` becomes `//path` and never matches a basedir
+        // stored as `/path`.
+        #[cfg(target_os = "windows")]
```
sylvestre: move that into a function 
you added too many cfg here
it makes the code too hard to read

solarispika:
Revised, please check.

src/compiler/c.rs
```diff
@@ -1448,6 +1448,38 @@ impl pkg::ToolchainPackager for CToolchainPackager {
 /// The cache is versioned by the inputs to `HashKeyParams::compute`.
 pub const CACHE_VERSION: &[u8] = b"12";
 
+/// Collapse `\\` byte pairs into a single `\`.
+#[cfg(any(target_os = "windows", test))]
```
sylvestre: why ", test" ?

solarispika: The function is pure byte manipulation, so I kept the test cross-platform to get cheap coverage from the Linux/macOS CI matrix too. Since the function only runs on Windows in production, gating both the helper and the test under `#[cfg(target_os = "windows")]` is also defensible — the trade-off is the byte logic then only runs in the Windows CI legs.

Happy to either way. Which do you prefer?

sylvestre: just windows is probably enough :)

solarispika: Done, Windows-only now.

src/compiler/c.rs
```diff
@@ -1566,7 +1598,11 @@ impl<'a> HashKeyParams<'a> {
         m.update(CACHE_VERSION);
         m.update(self.language.as_str().as_bytes());
         for arg in self.arguments {
-            arg.hash(&mut HashToDigest { digest: &mut m });
+            let arg_bytes = arg.as_encoded_bytes();
```
06393993, line 1601: I am not a maintainer nor a contributor, but I am a user of sccache.

I am wondering if we should use `OsString::as_encoded_bytes` here with `strip_basedirs` and `self.basedirs`.

`arg_bytes` here seems to be bytes with platform specific encoding, while `self.basedirs` is in UTF-8, and can cause problems on Windows.

This is the doc for [`OsString::as_encoded_bytes`](https://doc.rust-lang.org/std/ffi/struct.OsString.html#method.as_encoded_bytes):

> The byte encoding is an unspecified, platform-specific, self-synchronizing superset of UTF-8. By being a self-synchronizing superset of UTF-8, this encoding is also a superset of 7-bit ASCII.

`self.basedirs` is initialized by `storage.basedirs()`.  Let's follow how `<DiskCache as Storage>::basedirs` is implemented and initialized, we can tell it's ultimately from `config.basedirs`. Checking `Config::from_env_and_file_configs`, we can tell `basedirs` are populated from `basedirs_raw` by converting the path to `Utf8TypedPathBuf` before normalization and converting to bytes. Therefore, `basedirs` are encoded in UTF-8 not in a platform specific way.

I am wondering if the following is better:

```rust
            let arg = arg.to_str()
                .and_then(|arg| {
                    let arg = match strip_basedirs(arg.as_bytes(), self.basedirs) {
                        Cow::Owned(arg) => {
                            let arg = String::from_utf8(arg).ok()?;
                            let arg = OsString::from(arg);
                            Cow::Owned(arg)
                        }
                        Cow::Borrowed(arg) => {
                            let arg = str::from_utf8(arg).ok()?;
                            let arg = OsStr::new(arg);
                            Cow::Borrowed(arg)
                        }
                    };
                    Some(arg)
                })
                .unwrap_or(Cow::Borrowed(arg));
            arg.hash(&mut HashToDigest { digest: &mut m });
```

The same applies to the similar change in `src/compiler/preprocessor_cache.rs`. Thanks.

solarispika: @06393993 Good catch on the encoding mismatch. On Windows `OsString::as_encoded_bytes()` is WTF-8 and `self.basedirs` is plain UTF-8, so formally they don't agree. In practice I don't think it bites, and the rewrite has costs that aren't obvious:

1. WTF-8 only diverges from UTF-8 on unpaired surrogates. Paths coming out of real build tools are valid Unicode, so the bytes are the same there.
2. `strip_basedirs` already normalises the input on Windows (lowercases ASCII, `\` → `/`) and the non-ASCII branch decodes UTF-8 with an explicit fall-through for invalid sequences (`src/util.rs:1090-1095`). An unpaired surrogate just fails to match; it doesn't mis-match.
3. Basedirs themselves can't represent non-Unicode paths. They go through `String` (file) or `to_string_lossy()` (env) into `Utf8TypedPathBuf` (`src/config.rs:1175`), so the needle is guaranteed UTF-8 and cannot match an ill-formed segment of the haystack on either path. Whatever mismatch the rewrite "fixes" isn't fixable at this layer anyway.
4. The proposed shape changes the cache key. Hashing via `OsString`'s `Hash` impl writes a length prefix; the current code writes raw bytes plus a null separator. Not behaviour-preserving, so it'd need `CACHE_VERSION` / `FORMAT_VERSION` bumps to avoid silent cache corruption.

If we want to be defensive about it without paying that price, the narrowest version I can see is:

```rust
arg.to_str()
   .map(str::as_bytes)
   .unwrap_or_else(|| arg.as_encoded_bytes())
```

Same bytes for valid Unicode, current behaviour otherwise. I'm not convinced it earns the extra ceremony, since the basedirs side already constrains what can ever match.

@sylvestre what do you prefer — leave it as-is, or take the `to_str()` fallback?

PR #2713, opened 2026-05-17
feat: add `force_remote_build` config param to disable local job handling

iTrooz:
This PR adds a `force_remote_build` config parameter to the sccache client. This can be useful in 2 cases:
- when debugging if your builds are actually sent to the build servers instead of being processed locally
- when you don't want to ensure no compute is used on your local client

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -3313,6 +3313,82 @@ LLVM version: 6.0",
             assert_eq!(COMPILER_STDERR, res.stderr.as_slice());
         }
     }
+
+    #[cfg(feature = "dist-client")]
+    #[test]
+    fn test_compiler_get_cached_or_compile_dist_force_remote() {
+        drop(env_logger::try_init());
+        let creator = new_creator();
+        let f = TestFixture::new();
+        let gcc = f.mk_bin("gcc").unwrap();
+        let runtime = Runtime::new().unwrap();
+        let pool = runtime.handle().clone();
+        let dist_client =
+            Arc::new(test_dist::ForceRemoteClient(
+                test_dist::ErrorRunJobClient::new(),
+            ));
+        // Write a dummy input file so the preprocessor cache mode can work
+        std::fs::write(f.tempdir.path().join("foo.c"), "whatever").unwrap();
+        let storage = DiskCache::new(
+            f.tempdir.path().join("cache"),
+            u64::MAX,
+            &pool,
+            Default::default(),
+            CacheMode::ReadWrite,
+            vec![],
+        );
+        let storage = Arc::new(storage);
+        // Pretend to be GCC.
+        next_command(
+            &creator,
+            Ok(MockChild::new(exit_status(0), "compiler_id=gcc", "")),
+        );
+        let c = get_compiler_info(
+            creator.clone(),
+            &gcc,
+            f.tempdir.path(),
+            &[],
+            &[],
+            &pool,
+            None,
+        )
+        .wait()
+        .unwrap()
+        .0;
+        // The preprocessor invocation.
+        next_command(
+            &creator,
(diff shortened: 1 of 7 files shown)


src/compiler/compiler.rs
```diff
@@ -864,6 +864,10 @@ where
         Some(ref client) => client.rewrite_includes_only(),
         _ => false,
     };
+    let force_remote_build = match dist_client {
```
sylvestre, line 867: a bit above, when `dist_compile_cmd` is None we just compile locally - shouldn't force_remote_build also refuse there? otherwise "no compute on the client" isn't really guaranteed

src/compiler/compiler.rs
```diff
@@ -3306,6 +3313,84 @@ LLVM version: 6.0",
...
+                arguments,
+                cwd.to_path_buf(),
+                vec![],
+                CacheControl::ForceRecache,
+                pool,
+            )
+            .wait();
+        assert!(
+            result.is_err(),
+            "Expected dist compilation to fail with force_remote, but it succeeded"
```
sylvestre, line 3391: the test passes as soon as anything errors - please assert we don't fall back locally (ie the error is the dist one)

src/compiler/compiler.rs
```diff
@@ -3668,8 +3765,60 @@ mod test_dist {
...
         }
+        fn force_remote_build(&self) -> bool {
+            false
+        }
         fn get_custom_toolchain(&self, _exe: &Path) -> Option<PathBuf> {
             None
         }
     }
+
+    pub struct ForceRemoteClient(pub Arc<dyn dist::Client>);
```
sylvestre, line 3776: 50 lines of pure delegation just to flip one bool - could you please add a flag to the existing mock client instead?

src/config.rs
```diff
@@ -744,6 +744,7 @@ pub struct DistConfig {
     #[serde(deserialize_with = "deserialize_size_from_str")]
     pub toolchain_cache_size: u64,
     pub rewrite_includes_only: bool,
+    pub force_remote_build: bool,
```
sylvestre, line 747: please document the new param in docs/Configuration.md (and docs/Distributed.md)

src/dist/http.rs
```diff
@@ -1127,17 +1124,21 @@ mod client {
                 .pool_max_idle_per_host(0)
                 .build()
                 .context("failed to create an async HTTP client")?;
-            let client_toolchains =
-                cache::ClientToolchains::new(cache_dir, cache_size, toolchain_configs)
-                    .context("failed to initialise client toolchains")?;
+            let client_toolchains = cache::ClientToolchains::new(
+                &config.cache_dir,
```
sylvestre, line 1128: this used to be `config.cache_dir.join("client")`, the subdir is lost here - regression?

src/server.rs
```diff
@@ -148,15 +148,16 @@ pub struct DistClientContainer {
 #[cfg(feature = "dist-client")]
 pub struct DistClientConfig {
     // Reusable items tied to an SccacheServer instance
-    pool: tokio::runtime::Handle,
+    pub pool: tokio::runtime::Handle,
```
sylvestre, line 151: pub(crate) is enough here, these aren't meant for external users

tests/oauth.rs
```diff
@@ -52,11 +52,8 @@ fn config_with_dist_auth(
         cache: Default::default(),
         dist: sccache::config::DistConfig {
             auth: auth_config,
-            scheduler_url: None,
             cache_dir: tmpdir.join("unused-cache"),
-            toolchains: vec![],
-            toolchain_cache_size: 0,
-            rewrite_includes_only: true,
+            ..sccache::config::DistConfig::default()
```
sylvestre, line 56: this silently flips rewrite_includes_only from true to false, intended?

PR #2716, opened 2026-05-25
Minimize thread numbers for the client

badumbatish:
Minimize thread numbers for the client, its querying only after all.

This commit doesn't try to limit the thread size to 1. I was worried that there might be some hidden interacts that might lead to deadlock but would be very glad to be enlightened on this


badumbatish:
i think #2704 might be solving for the same issue. I'd love to have the PR merged

ahartmetz:
#2704 has been merged!

PR #2717, opened 2026-05-25
Add -fno-cxx-modules to clang's known -Xclang flags

badumbatish:
Downstream Bazel's C++ toolchain passes `-Xclang -fno-cxx-modules` in its compiler sanity checks. Since -fno-cxx-modules wasn't in sccache's known flag list, the xclang argument parser in gcc.rs treated it as an UnknownFlag and bailed with CannotCache.


sylvestre:
please add a test

badumbatish:
> please add a test

apologies for the late actions, please see the latest PR

PR #2720, opened 2026-05-27
feat: avoid sccache wrapper when resolving compiler

iTrooz:
This PR allows you to do things like  `/bin/sccache /usr/lib/sccache/gcc main.c`

It can seem weird at first, but the same mechanism already existed for ccache. Personally, I had an issue with a build system that insisted on putting sccache as a wrapper, when I already had gcc as a sccache symlink, which is what prompted that PR

there's a change of behaviour for absolute path (they are now still changed to avoid wrappers) so I would understand if you didn't accept this PR


iTrooz:
Thanks !

sylvestre:
thank you and sorry for the latency!

PR #2728, opened 2026-06-06
dist: use bubblewrap for overlayFS mounts, run `sccache-dist` without root

### LINKED ISSUE #326: Get a better story around the build server running as root
Currently the build server needs to be run as root because it creates an overlay filesystem to mount in the bubblewrap sandbox. Some people are understandably unnerved by running a network daemon as root, so it would be good if we had a better story here. It may not be technically possible to do what we need to do without root, but we could at least have the daemon fork on startup and then have one process drop privileges and run the bulk of the code unprivileged (all the networking etc) with a little bit of IPC to the privileged process to do the things that need privileges.

### LINKED ISSUE #1660: What work needs to be done to allow `sccache-dist server` to run without root?
I've wanted to set up sccache a couple of times, but each time the root/sudo requirement for the build server ends up being a blocker. I can't use docker or podman either, unfortunately.

I admit this is a pretty niche situation, but if there's an easy solution to needing root I'd love to hear it! I believe the root requirement is due to using `bubblewrap`, and I assume we can't get around _that_.

### LINKED ISSUE #1688: `overlay` are not unmounted
After upgrading to v0.4.0, I started to have `sccache-dist`  being sluggish (~5 min to complete 10% of `mozilla-central`'s `configure` when a full build without  cache is ~4min).

After quick investigation, it is on server side. Trying to `CTRL+C` the `sccache-dist server` process results in:
 - long time before processes are killed
 - unable to restart the process after with:
```
sccache-dist: error: Overlay builder failed to start
sccache-dist: caused by: Failed to clean up builder directory
sccache-dist: caused by: failed to remove directory `XXX/sccache/build`
sccache-dist: caused by: Device or resource busy (os error 16)
```

A look at `mount` shows 1052 instances of mounted `overlay` similar to:
```
overlay on XXX/build/builds/746755680e580184e658dbc0c5af280e31b72bf026f2d64adb215757f9099101-769/target type overlay (rw,relatime,lowerdir=XXX/sccache/build/toolchains/746755680e5
80184e658dbc0c5af280e31b72bf026f2d64adb215757f9099101,upperdir=XXX/sccache/build/builds/746755680e580184e658dbc0c5af280e31b72bf026f2d64adb215757f9099101-769/upper,workdir=XXX/sccache/buil
d/builds/746755680e580184e658dbc0c5af280e31b72bf026f2d64adb215757f9099101-769/work)
```

rrbutani:
> [!NOTE]
> I've opened this PR to get feedback on whether the direction here is acceptable. There are still a few items that need to be addressed (see `pending` below) before this is ready for merge.

The approach here is more or less the same as in #1628: use `CLONE_NEWUSER` (unprivileged user namespace) to be able to create a mount namespace + mount the overlayFS mounts without permissions

Unlike #1628, this PR leverages [`bubblewrap`'s overlay mount functionality](https://github.com/containers/bubblewrap/commit/f371022ad6af6309cebb4ccb9635b19ed4196ac0) (added in `v0.11.0`) to create the overlayFS mounts at the `sccache-dist` → `bwrap` compile command process boundary. This has a couple of upsides:
  - sidesteps the [safety issues](https://github.com/mozilla/sccache/blob/9fb6a2542c0d2e82846252be19bb377e5ca8885b/src/bin/sccache-dist/build.rs#L447-L458) in #1628 caused by forking the `sccache-dist` process
  - avoids the [issue](https://github.com/mozilla/sccache/issues/1688) with overlay fs mounts not being unmounted once the compile command finishes that ultimately led to #1628 being [reverted](https://github.com/mozilla/sccache/pull/1689)
    + though, I think this issue was actually caused by #1628 changing the default sandboxing codepath (i.e. `current_ns`) to call `CLONE_NEWNS` [in the `sccache-dist`'s `rouille` request handler thread](https://github.com/mozilla/sccache/blob/9fb6a2542c0d2e82846252be19bb377e5ca8885b/src/bin/sccache-dist/build.rs#L422-L429) instead of [in the thread spawned for a compile](https://github.com/mozilla/sccache/blob/c72eed63d78c8550b93133e9fa461a0c6ef7a30d/src/bin/sccache-dist/build.rs#L272-L276) (and not anything to do with the unprivileged user namespace logic)

other notes:
  - previously deleting the build directory/the per-compile directories did not run into permission issues because `sccache-dist` was running as root

pending/open questions:
  - [ ] fallback support for running on Linux <5.11 and/or with bubblewrap versions older than v0.11.0?
  - [ ] fallback support for running on machines where unprivileged user namespaces are disabled?
  - [ ] manual checks for whiteout files/files with `userxattr.overlay` `xattr`s in the compile inputs tarball?
    + (see [these docs](https://docs.kernel.org/filesystems/overlayfs.html#whiteouts-and-opaque-directories) for context)
    + previously this issue was sidestepped by unpacking directly into the overlay mountpoint (target dir); now that we're having `bubblewrap` handle creating the mountpoint this isn't an option
    + unclear to what degree this is an issue (does not seem like users could do anything nefarious with whiteouts or opaque dirs — worst case a toolchain file is masked?)
    + also: `tar::Archive` [defaults to not restoring `xattrs`](https://docs.rs/tar/latest/tar/struct.Archive.html#method.set_unpack_xattrs) so I think the only consideration is whiteouts?

---

closes #326, closes #1660, closes #1688, supersedes #2706

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,7 @@ version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "cipher",
  "cpufeatures",
 ]
@@ -198,7 +198,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
 dependencies = [
  "addr2line",
- "cfg-if 1.0.0",
+ "cfg-if",
  "libc",
  "miniz_oxide",
  "object",
@@ -271,7 +271,7 @@ dependencies = [
  "arrayref",
  "arrayvec",
  "cc",
- "cfg-if 1.0.0",
+ "cfg-if",
  "constant_time_eq",
 ]
 
@@ -351,12 +351,6 @@ dependencies = [
  "libc",
  "shlex",
 ]
-
-[[package]]
-name = "cfg-if"
-version = "0.1.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4785bdd1c96b2a846b2bd7cc02e86b6b3dbf14e7e53446c4f54c92a361040822"
 
 [[package]]
 name = "cfg-if"
@@ -452,7 +446,7 @@ dependencies = [
  "getrandom 0.2.11",
  "glob",
  "libc",
- "nix 0.30.1",
+ "nix",
  "serde",
  "serde_json",
  "statrs",
@@ -491,7 +485,7 @@ version = "4.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d9de586cc7e9752fc232f08e0733c2016122e16065c4adf0c8a8d9e370749ee"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "clap",
  "codspeed",
  "condtype",
@@ -655,7 +649,7 @@ version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b540bd8bc810d3885c6ea91e2018302f68baba2129ab3e88f32389ee9370880d"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
 ]
 
 [[package]]
@@ -664,7 +658,7 @@ version = "0.8.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3a430a770ebd84726f584a90ee7f020d28db52c6d02138900f22341f866d39c"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


rrbutani:
@sylvestre friendly ping; is there anything I can do to make this PR easier to review?

luhenry:
Getting that PR to be merged would unblock PyTorch's build of sccache on riscv64. See https://github.com/pytorch/pytorch/pull/182278/changes#diff-0a862d15a4e6dff44196719a594b983aece596c7dc38eba416d83e12a95206e6 for details

sylvestre:
having tests that cover this change would be helpful :)

luhenry:
@rrbutani gentle ping on that issue? happy to have a look on adding tests a well.

rrbutani:
Thanks for the ping, apologies for the delay.

I've updated the existing `sccache-dist` tests to run `sccache-dist server` as an unprivileged user.

The first commit in this PR has only these changes; running (for example) `cargo test --test dist --features dist-tests -- test_dist_basic` at that commit will (as we expect) fail with `EPERM`. The next commit (unchanged from earlier) updates `sccache-dist server` to work w/o privileges and gets the `dist-tests` passing again.

> having tests that cover this change would be helpful :)

@sylvestre Is ^ sufficient?

rrbutani:
@sylvestre friendly ping

rrbutani:
@sylvestre could I get another CI run for this PR?

rrbutani:
@sylvestre friendly ping

PR #2733, opened 2026-06-10
fix(gcs): support authorized user ADC credentials

vladpetric:
## Summary

- support Google `authorized_user` ADC JSON files for GCS by exchanging the refresh token for an access token and passing that token to OpenDAL
- keep non-`authorized_user` GCS credential files on the existing `credential_path` path
- report multilevel storage as writable when at least one configured level is writable, so a writable L0 disk cache can still accept writes when a remote L1 is read-only

## Motivation

Local Google ADC files created by `gcloud auth application-default login` use `type: authorized_user`. The existing OpenDAL credential-path path rejects those files and falls back to metadata-server auth, which fails on normal developer machines.

This lets local sccache users authenticate GCS via their existing user ADC file while preserving the existing service-account behavior.

## Testing

- `cargo fmt --manifest-path /home/vlad/build/sccache/Cargo.toml --check`
- `cargo test --manifest-path /home/vlad/build/sccache/Cargo.toml cache::multilevel`
- `cargo build --manifest-path /home/vlad/build/sccache/Cargo.toml --release`
- local smoke test with disk L0 + GCS L1 using an `authorized_user` ADC file:
  - sccache daemon starts with both levels
  - GCS L1 is readable
  - read-only GCS L1 does not prevent writes to disk L0
  - repeated C compile hits disk L0 on the second invocation

### DIFF
--- a/src/cache/gcs.rs
+++ b/src/cache/gcs.rs
@@ -21,7 +21,8 @@ use opendal::{
     services::Gcs,
 };
 use reqwest::Client;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
+use std::fs;
 use url::Url;
 
 use super::http_client::set_user_agent;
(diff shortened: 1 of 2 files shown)


sylvestre:
a bunch of jobs are failing

sylvestre:
are you still interested in this one? if not, i will close it, thanks

PR #2735, opened 2026-06-12
Initial client-side implementation

glandium:


src/commands.rs
```diff
@@ -548,9 +552,43 @@ where
...
+    stdout: &mut dyn Write,
+    stderr: &mut dyn Write,
+) -> Result<i32>
+where
+    T: CommandCreatorSync,
+{
+    match response {
+        CompileResponse::CompileStarted => {
+            return handle_compile_finished(
+                finished.expect("CompileStarted must have a result"),
```
sylvestre: On a mid-compile server disconnect, handle_compile_response now passes (CompileStarted, None) here, so .expect() panics instead of falling back to local compilation as the old code did — handle None by falling through to the local-compile path and add a disconnect regression test.

ahartmetz:
I have a feature wish while you're at it: It would be cool to auto-retry failed remote builds locally and have always-on stderr output and maybe statistics for it so that one can possibly investigate and fix the problem.

Sometimes a file that a remote build requires isn't captured due to a missing sccache feature.

Sometimes a builder is in a borked state for whatever reason. I've observed some intermittent problem with extracting(?) the compiler archive on builders and unfortunately I haven't managed to properly diagnose and fix it. Even if that was fixed, distributed anything can fail for lots of reasons up to and including flaky hardware - I've seen that happen several times when I worked in an office with an Icecream cluster made of the local PCs.

The ideal fixes for these two+ kinds of failures are quite different, but retrying locally is a nice catch-all workaround to keep the builds succeeding and minimize wasted developer time.

glandium:
> I have a feature wish while you're at it: It would be cool to auto-retry failed remote builds locally and have always-on stderr output and maybe statistics for it so that one can possibly investigate and fix the problem.
> 
> Sometimes a file that a remote build requires isn't captured due to a missing sccache feature.
> 
> Sometimes a builder is in a borked state for whatever reason. I've observed some intermittent problem with extracting(?) the compiler archive on builders and unfortunately I haven't managed to properly diagnose and fix it. Even if that was fixed, distributed anything can fail for lots of reasons up to and including flaky hardware - I've seen that happen several times when I worked in an office with an Icecream cluster made of the local PCs.
> 
> The ideal fixes for these two+ kinds of failures are quite different, but retrying locally is a nice catch-all workaround to keep the builds succeeding and minimize wasted developer time.

This is unrelated to what this PR does. You may want to file an issue with steps to reproduce.

ahartmetz:
> This is unrelated to what this PR does. You may want to file an issue with steps to reproduce.

The first and last paragraph are somewhat closely related, the middle two are just motivation and the usefulness is by no means limited to these. Only if distributed sccache was absolutely perfect including with future compiler changes would the feature be not useful. I don't have steps to reproduce the specific bug from the third paragraph or I'd probably have fixed it myself.

I'm also considering implementing local retry myself - it's similar enough to 1ffa8d52c8b029892ded5b86b9c9f30e77a34bf5 which also works by essentially re-invoking the build with different arguments. In that case, it would be nice of you to keep that possible extension in mind to avoid accidental difficulties for it.

sylvestre:
@ahartmetz please open an issue (or prepare a PR) to make sure it isn't forgotten
this PR is already big enough

ahartmetz:
> @ahartmetz please open an issue (or prepare a PR) to make sure it isn't forgotten this PR is already big enough

https://github.com/mozilla/sccache/issues/2745

glandium:
> The first and last paragraph are somewhat closely related

This change doesn't impact distributed compilation.

ahartmetz:
> > The first and last paragraph are somewhat closely related
> 
> This change doesn't impact distributed compilation.

No, but it could. That's why I called it a wish.

PR #2736, opened 2026-06-13
fix: strip SCCACHE_BASEDIRS from escaped-backslash paths on Windows

### LINKED ISSUE #2737: SCCACHE_BASEDIRS has no effect on Windows (escaped backslashes in preprocessor output never match)
## Summary

`SCCACHE_BASEDIRS` has no effect on Windows. The same source compiled from
two different checkout directories produces different cache keys, so cache
entries are never shared across roots — the feature silently does nothing.

## Environment

- Platform: Windows
- Compilers: MSVC (`cl`) and clang both affected
- `SCCACHE_BASEDIRS` set to the absolute checkout roots

## Steps to reproduce

1. Set `SCCACHE_BASEDIRS` to a list of absolute base directories.
2. Build the same source tree from two different absolute directories
   (e.g. `D:\dev\user1\project` and `E:\work\user2\checkout`), as CMake/Ninja
   do when they pass absolute source paths.
3. Compile a file that ends up with an absolute path in the preprocessed
   output (any `#line`/linemarker, `__FILE__`, or `assert()` expansion —
   which build systems routinely produce).

**Expected:** the second build is a cache hit; basedirs are stripped so the
hash is independent of the checkout root.

**Actual:** every build is a miss; basedirs are never stripped.

## Root cause

On Windows, paths inside preprocessor output are emitted as **C string
literals**. MSVC `#line` directives and clang GNU linemarkers quote them
with escaped backslash separators:

```
#line 1 "C:\\Users\\me\\project\\src\\main.c"
```

`normalize_win_path` (`src/util.rs`) converts every backslash to a forward
slash, so each escaped separator `\\` becomes a **double** forward slash:

```
c://users//me//project//src//main.c
```

But the normalized basedir uses single slashes (`c:/users/me/project/`), so
`strip_basedirs` never finds a boundary match and the absolute path survives
into the hash input.

## Fix

PR #2736 makes `strip_basedirs` also search for a doubled-separator variant
of each basedir, so escaped paths are stripped in addition to the plain
form (mixed and UNC cases covered). Non-Windows builds are unchanged
(`cfg(target_os = "windows")`-gated).

rconde01:
# fix: strip `SCCACHE_BASEDIRS` from escaped-backslash paths on Windows

Fixes #2737.

## Problem

`SCCACHE_BASEDIRS` has no effect on Windows. The same source compiled
from two different checkout directories produces different cache keys, so
nothing is shared across roots — the feature silently does nothing.

### Root cause

On Windows, paths inside preprocessor output are emitted as **C string
literals**. MSVC `#line` directives and clang GNU linemarkers quote them
with escaped backslash separators:

```
#line 1 "C:\\Users\\me\\project\\src\\main.c"
```

`normalize_win_path` converts every backslash to a forward slash, so each
escaped separator `\\` becomes a **double** forward slash:

```
c://users//me//project//src//main.c
```

But the normalized basedir uses single slashes (`c:/users/me/project/`),
so `strip_basedirs` never finds a boundary match, and the absolute path is
left in the preprocessed output that feeds the hash.

## Fix

In `strip_basedirs`, also search for a doubled-separator variant of each
basedir, produced by a new `double_path_separators` helper. Escaped paths
are now stripped in addition to the existing plain form; mixed
escaped/plain occurrences and UNC basedirs are handled too.

The extra needle and the helper are gated behind
`cfg(target_os = "windows")`, so non-Windows builds and behavior are
unchanged.

## Tests

New unit tests (Windows-only):

- `util::tests::test_strip_basedir_windows_escaped_backslashes` — MSVC
  `#line`, clang linemarker, and mixed escaped/plain occurrences.
- `util::tests::test_strip_basedir_windows_escaped_unc` — escaped UNC
  basedir.
- `compiler::c::test::test_hash_key_basedirs_windows_escaped_backslashes`
  — the same source under two different basedir roots hashes equally, and
  differs without basedirs.

All three fail before the change and pass after. `cargo test --lib`,
`cargo fmt --check`, and `cargo clippy --lib` are clean on the changed
code (the single pre-existing clippy warning in `msvc.rs` is untouched).

### Manual end-to-end verification

Reproduced with real toolchains on Windows: building the same source from
two distinct absolute directories with `SCCACHE_BASEDIRS` set goes from
**0/2** cache hits (unpatched) to **1/2** (patched) for both clang and
MSVC.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -1761,6 +1761,42 @@ mod test {
         assert_eq!(h1, hash_with_basedirs(preprocessed1, &multi_basedirs));
     }
 
+    #[cfg(target_os = "windows")]
+    #[test]
+    fn test_hash_key_basedirs_windows_escaped_backslashes() {
+        // On Windows, paths in preprocessor output are C string literals with
+        // escaped backslashes (e.g. MSVC: #line 1 "D:\\dir\\file.c", clang:
+        // # 1 "D:\\dir\\file.c"). The same source compiled from two different
+        // directories must produce the same hash when both are basedirs.
+        let args = ovec!["a", "b", "c"];
+        let preprocessed1 =
+            b"#line 1 \"D:\\\\dev\\\\user1\\\\project\\\\src\\\\main.c\"\nint main() { return 0; }";
+        let preprocessed2 =
+            b"#line 1 \"E:\\\\work\\\\user2\\\\checkout\\\\src\\\\main.c\"\nint main() { return 0; }";
+        // As normalized by Config: lowercase, forward slashes, trailing slash
+        let basedirs = [
+            b"d:/dev/user1/project/".to_vec(),
+            b"e:/work/user2/checkout/".to_vec(),
+        ];
+
+        let hash_with_basedirs = |output: &[u8], dirs: &[Vec<u8>]| {
+            HashKeyParams::new("abcd", Language::C, &args, output)
+                .with_basedirs(dirs)
+                .compute()
+        };
+
+        // Same hash with different absolute (escaped backslash) paths
+        let h1 = hash_with_basedirs(preprocessed1, &basedirs);
+        let h2 = hash_with_basedirs(preprocessed2, &basedirs);
+        assert_eq!(h1, h2);
+
+        // Different hashes without basedirs
+        assert_neq!(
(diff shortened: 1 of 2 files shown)


sylvestre:
in the future, please keep only the most important information in comment #0

PR #2740, opened 2026-06-18
Fix preprocessor cache include validation

fallintoplace:
## Summary
- keep validating cached includes after the first match when `ignore_time_macros` is enabled
- add a regression test where the first header is unchanged and a same-size second header changes

## Why
The old control flow returned immediately after hashing the first include in the `ignore_time_macros` path. If a later include changed, the preprocessor cache could still accept the stale result.

## Testing
- `cargo fmt --check`
- `cargo test -q compiler::preprocessor_cache::test::test_preprocessor_cache_result_checks_all_includes_when_ignoring_time_macros --lib`
- `cargo test -q compiler::preprocessor_cache --lib`
- `cargo test -q --lib`

### DIFF
--- a/src/compiler/preprocessor_cache.rs
+++ b/src/compiler/preprocessor_cache.rs
@@ -254,7 +254,11 @@ impl PreprocessorCacheEntry {
 
             if config.ignore_time_macros {
                 match Digest::reader_sync(file) {
-                    Ok(new_digest) => return include.digest == new_digest,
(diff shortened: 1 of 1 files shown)


sylvestre:
@fallintoplace in the future, don't bother adding "Testing" in comment #0, it doesn't add value for us
thanks

sylvestre:
Thanks for your PR

PR #2743, opened 2026-06-19
Micro cleanups

ahartmetz:
I found these in a local "hacks branch", and since small review units are popular here, why not submit them...


ahartmetz:
Needs Rust 1.88.0 for one of the changes to work without unstable features... shelving it until then

PR #2747, opened 2026-06-20
Bundle relocated interpreter's libdir into the dist toolchain package

jetm:
## Problem

`sccache-dist` packages a toolchain's shared libraries by parsing `ldd`
output, which resolves a binary's `NEEDED` libraries against the host's
dynamic loader. Yocto/OpenEmbedded "uninative" cross toolchains ship a
relocated glibc whose loader has a built-in search path pointing at its
own sysroot. For those binaries `ldd` reports host paths (e.g.
`/usr/lib/libm.so.6`), but inside the build sandbox the relocated loader
searches its own sysroot lib dir — where those libraries were never
packaged. The remote compile then dies with `libm.so.6: cannot open
shared object file` and silently falls back to local compilation, so
distribution never actually runs.

## Solution

When a packaged executable's `PT_INTERP` lives outside the standard host
loader directories (`/lib`, `/lib64`, `/usr/lib`, `/usr/lib64`), also
bundle the interpreter's own directory. That directory holds the
`libc`/`libm` the relocated loader resolves against, so they land at the
absolute path the loader searches inside the sandbox. Standard host
toolchains are unaffected: their interpreter is under a standard loader
directory, so the existing `ldd`-only path is preserved.

## Key changes

- Add `read_elf_interpreter()` to read a binary's `PT_INTERP` via the
  already-vendored `object` crate.
- Add `relocated_interpreter_libdir()` to classify an interpreter as
  relocated and return the directory to bundle.
- In `add_executable_and_deps()`, bundle that directory's contents when a
  relocated interpreter is detected.
- Unit test covering standard vs. relocated interpreter classification.

## Reviewer notes

- No behavior change for standard host toolchains — their interpreter is
  under `/lib*`, so nothing extra is bundled.
- Verified against a real OpenEmbedded aarch64 cross toolchain: a full
  Yocto image build distributed 634 compiles to the build server with 0
  dist errors and no `libm.so.6` failures, and the packaged toolchain tar
  now contains the loader's `libc`/`libm` at the uninative sysroot path.

### DIFF
--- a/src/dist/pkg.rs
+++ b/src/dist/pkg.rs
@@ -98,6 +98,20 @@ mod toolchain_imp {
         }
 
         pub fn add_executable_and_deps(&mut self, executable: PathBuf) -> Result<()> {
+            // A relocated program interpreter (e.g. a Yocto/OE uninative
+            // toolchain) resolves its libc/libm against its own sysroot, not the
+            // host paths ldd reports. Bundle the interpreter's directory so those
+            // libraries exist where the loader looks inside the build sandbox.
+            if let Some(libdir) = read_elf_interpreter(&executable)
+                .and_then(|interp| relocated_interpreter_libdir(&interp))
+            {
+                self.add_dir_contents(&libdir).with_context(|| {
+                    format!(
+                        "Failed to bundle relocated interpreter libdir {}",
+                        libdir.display()
+                    )
+                })?;
+            }
             let mut remaining = vec![executable];
             while let Some(obj_path) = remaining.pop() {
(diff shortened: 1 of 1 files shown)


jetm:
Superseded by #2750, which combines this with the rest of the OpenEmbedded/Yocto distributed-compile fixes into a single series. Closing in favor of that PR.

ahartmetz:
Note, you will have more luck with smaller pull requests. sccache has limited reviewer capacity and smaller units of review help to get your changes reviewed.

PR #2748, opened 2026-06-20
Improve toolchain cache reliability

ahartmetz:
These changes do not seem to fix, at least not fully, a problem with broken toolchains that I've observed, but are probably improvements anyway.


sylvestre:
it would be nice to have a test to cover this if this isn't too complex

ahartmetz:
> it would be nice to have a test to cover this if this isn't too complex

It's probably doable, and maybe writing the tests will help me come up with more ideas on how this stuff can fail.

PR #2749, opened 2026-06-23
Fix ToolchainPackager cfg gate to build on ppc64le/s390x

cryptomilk:
On Linux ppc64le/s390x, the blanket `impl<T: Send> ToolchainPackager for T` in pkg.rs is active (non-x86_64/aarch64 path), but the test-only PanicToolchainPackager impl in cache.rs was gated on only `target_os = "linux"`, causing a conflicting-impl error on those architectures.

### DIFF
--- a/src/dist/cache.rs
+++ b/src/dist/cache.rs
@@ -302,7 +302,10 @@ mod client {
                 Box::new(PanicToolchainPackager)
             }
         }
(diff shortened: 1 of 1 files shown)


cryptomilk:
Build succeeded with the patch at: https://koji.fedoraproject.org/koji/taskinfo?taskID=146963651

Issue #2745, opened 2026-06-19
Feature: Retry build locally after remote failure


ahartmetz:
A remote build may fail for several reasons, the most common ones being:

- Missing support for a compiler / build system feature in sccache, for example a required file isn't included in the file set sent to the remote builder
- A flaky remote builder, for hardware or other (usually) spurious reasons

In such situations, it would be very nice to retry jobs locally in order to avoid blocking the development work that sccache is supposed to accelerate. It is mandatory that the user can see that and why (as far as possible) a build was rerun locally. Optionally, the information could be included in the statistics that sccache gathers.

A tricky aspect is that using sccache has the nice side effect of limiting the total number of parallel jobs running on a machine, including the machine initiating a build. It would be nice to retain that somewhat accidental feature, probably by still allocating jobs through the server.

AJIOB:
This issue looks like a possible solution for #2700

Issue #2700, opened 2026-05-07
regression: sccache-dist build fails on assembler including files


lissyx:
I have a setup with `sccache-dist` to build gecko, and recently after upgrading to `sccache v0.15.0` I started to hit build failures:
```
 0:03.39 W sccache: Job failed on server 192.168.1.112:10501:
 0:03.40 W clang: warning: argument unused during compilation: '-D DEBUG=1' [-Wunused-command-line-argument]
 0:03.40 W clang: warning: argument unused during compilation: '-D ICU_DATA_FILE="/home/alex/firefox/gecko-cinnabar/config/external/icu/data/icudt78l.dat"' [-Wunused-command-line-argument]
 0:03.40 W clang: warning: argument unused during compilation: '-D ICU_DATA_SYMBOL=icudt78_dat' [-Wunused-command-line-argument]
 0:03.40 E /home/alex/firefox/gecko-cinnabar/config/external/icu/data/icu_data.S:28:9: error: Could not find incbin file '/home/alex/firefox/gecko-cinnabar/config/external/icu/data/icudt78l.dat'
 0:03.40 E .incbin "/home/alex/firefox/gecko-cinnabar/config/external/icu/data/icudt78l.dat"
 0:03.40 E         ^
 0:03.40 sccache: Compiler killed by signal 1
```

I could narrow it down to #2545, and making a local revert of the feature unblocks the build

lissyx:
@AJIOB I see on #2556 you mentionned #2557, but I think it is on current `v0.15.0` tag, so maybe something else needs to be done there?

AJIOB:
Hi @lissyx ,

As I can see in the log, you try to include file  `/home/alex/firefox/gecko-cinnabar/config/external/icu/data/icudt78l.dat`, which is not presented on your disk. Am I correct?

Does it works without sccache run?

Can you provide the whole command line to that sccache/clang call, please?

Is it reproduced with local sccache only (without distributed compiling)?

lissyx:
> Hi [@lissyx](https://github.com/lissyx) ,
> 
> As I can see in the log, you try to include file `/home/alex/firefox/gecko-cinnabar/config/external/icu/data/icudt78l.dat`, which is not presented on your disk. Am I correct?

I believe the error is coming from the server side, because client side it's there.

> 
> Does it works without sccache run?

It think it works with sccache without dist, it's the setup on CI so I suspect it would have been caught

> 
[...]

PR #2750, opened 2026-06-23
Make distributed compilation work for OpenEmbedded/Yocto builds

jetm:
This series makes `sccache --dist` usable end-to-end for OpenEmbedded/Yocto
(BitBake) builds, where it previously panicked, failed, or silently fell back
to local on several distinct issues. Validated by building a qemuarm64
core-image-minimal entirely through a single-node sccache-dist cluster:
userspace do_compile distributes (e.g. busybox 509/509) and the kernel
distributes 1124/1128 compiles.

It combines and supersedes #2746 and #2747 into one OE/BitBake-support series.

1. Don't panic when a finished compile has neither code nor signal (#2746).
   sccache-dist synthesizes an ExitStatus with neither exit code nor signal for
   some abnormal remote compiles; get_signal panicked on .expect("must have
   signal"). Return Option<i32> and handle the both-absent case.

2. Bundle the relocated interpreter's libdir into the dist toolchain package
   (#2747). OE cross-toolchains use an absolute-path uninative loader whose
   libc/libm live in the uninative sysroot, not the host paths ldd reports;
   bundle the interpreter's libdir so the toolchain resolves in the sandbox.

3. Fix distributed compile of relative input paths. The dist command used the
   raw input path while the inputs packager shipped the preprocessed content at
   the absolute, simplified path, so out-of-tree builds (../sources/foo.c)
   failed "No such file or directory" on the server. Use the same path in both.

4. Log distributed-compile decisions at info level. The distribute-vs-local
   decision was debug-only and only failures warned; log it at info so
   SCCACHE_LOG=info shows the distribute/fallback breakdown. Emitted only when
   SCCACHE_LOG is set.

5. Fall back to local on distributed-compile failure. A remote failure is often
   a distribution artifact (e.g. a kernel object that .incbin's a binary the
   packager cannot ship) rather than a real compiler error; recompile locally
   so a dist failure never breaks a build that compiles fine locally.

Draft: validated single-node so far; marking ready after a two-machine cluster
run confirms it.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -973,7 +973,7 @@ where
                 )
             })?;
 
-        let mut jc = match jres {
+        let jc = match jres {
             dist::RunJobResult::Complete(jc) => jc,
             dist::RunJobResult::JobNotFound => bail!("Job {} not found on server", job_id),
         };
@@ -1039,19 +1039,26 @@ where
         );
 
         if jc.output.code != 0 {
-            // Add server info to help diagnose host-specific failures, e.g. due to flaky hardware.
-            // Failed builds are not cached so this tampering should not cause too much trouble.
-            let server_info = format!("sccache: Job failed on server {}:\n", server_id.addr());
-            jc.output
-                .stderr
-                .splice(0..0, server_info.as_bytes().to_vec());
+            // A non-zero remote result is frequently a distribution artifact
+            // rather than a genuine compiler error: e.g. an object that
+            // .incbin's a binary the inputs packager did not ship (the kernel's
+            // vdso/dtb/embedded-config wrappers), which the build-server cannot
+            // assemble. Fall back to a local recompile via the or_else below - it
+            // either succeeds (confirming a dist-only artifact) or reproduces the
+            // real error locally, so a remote failure never breaks a build that
+            // would compile fine locally. This only affects failing dist
+            // compiles; successful ones are returned unchanged above.
+            bail!(
+                "distributed compile on {} returned exit code {}; recompiling locally",
+                server_id.addr(),
+                jc.output.code
+            );
         }
 
         info!(
-            "[{}]: Distributed compilation finished on {} (exit code {})",
+            "[{}]: Distributed compilation finished on {}",
             out_pretty,
-            server_id.addr(),
-            jc.output.code
+            server_id.addr()
         );
         Ok((DistType::Ok(server_id), jc.output.into()))
     };
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -877,7 +877,10 @@ where
     let dist_client = match dist_compile_cmd.clone().and(dist_client) {
         Some(dc) => dc,
         None => {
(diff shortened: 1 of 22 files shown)


src/compiler/compiler.rs
```diff
@@ -1036,14 +1039,27 @@ where
...
-                .splice(0..0, server_info.as_bytes().to_vec());
+            // A non-zero remote result is frequently a distribution artifact
+            // rather than a genuine compiler error: e.g. an object that
+            // .incbin's a binary the inputs packager did not ship (the kernel's
+            // vdso/dtb/embedded-config wrappers), which the build-server cannot
+            // assemble. Fall back to a local recompile via the or_else below - it
+            // either succeeds (confirming a dist-only artifact) or reproduces the
+            // real error locally, so a remote failure never breaks a build that
+            // would compile fine locally. This only affects failing dist
+            // compiles; successful ones are returned unchanged above.
```
AJIOB, line 1119: This looks like a fix for #2700.

IMO, please, do a dedicated PR with that fix & test it to mitigate possible future regressions

AJIOB: And it implements #2745, isn't it?

src/compiler/compiler.rs
```diff
@@ -877,15 +877,18 @@ where
     let dist_client = match dist_compile_cmd.clone().and(dist_client) {
         Some(dc) => dc,
         None => {
-            debug!("[{}]: Compiling locally", out_pretty);
+            info!(
```
sylvestre, line 908: should be done  in a different pr

src/compiler/compiler.rs
ahartmetz: To implement #2745, it probably(?) makes sense to wait until #2735 is merged, then reuse some of its code.

AJIOB: Maybe yes. IMO, your code part should resolve that problem

ahartmetz: I actually haven't written any code in that area (yet, at least)

sylvestre:
please split this PR into smaller PRs.
To set expectations, it won't be merged otherwise

jetm:
Thanks @sylvestre, @AJIOB, and @ahartmetz for the reviews here.

I'm closing this PR. Since I opened it, the scope has grown well beyond the original change: continued testing on real OpenEmbedded/Yocto builds surfaced a range of further distributed-compile problems and their fixes (toolchain packaging, Rust std and `--target` resolution, scheduler job-accounting under high concurrency, connection reuse, and local fallback on failure). As a single 43-commit PR this is too large to review well.

I'm classifying the work by concern and splitting it into a series of smaller, independently reviewable PRs, landing the self-contained fixes first. I expect to start opening them this week, once a final round of validation on large OE/Yocto builds finishes (those runs take a while to complete). I'll link the new PRs back here as they go up.

Sorry for the noise here, on this PR and the earlier ones I closed. I got ahead of myself after some good early results on a handful of recipes, and opened them before I had tested broadly enough across the rest.

sylvestre:
no worries, i made that mistake so many times in my career ;)

Issue #2510, opened 2025-12-05
"Failed to stat include file ...: No such file or directory" with meson projects


dezgeg:
To reproduce, first pick an example meson project and build it once:
````
git clone https://github.com/MusicPlayerDaemon/mpc.git
cd mpc
meson setup build
ninja -C build
````

Now rebuild one file with debug prints on:
````
SCCACHE_LOG=debug SCCACHE_START_SERVER=1 SCCACHE_NO_DAEMON=1 sccache &
touch src/args.c && ninja -C build
````

Output:
````
ninja: Entering directory `build'
[0/2] Compiling C object mpc.p/src_args.c.o[2025-12-05T15:42:36Z DEBUG sccache::server] handle_client: compile
[2025-12-05T15:42:36Z DEBUG sccache::server] check_compiler: Supported compiler
[2025-12-05T15:42:36Z DEBUG sccache::server] parse_arguments: Ok: ["-Impc.p", "-I.", "-I..", "-I../src", "-I../subprojects/libmpdclient/include", "-Isubprojects/libmpdclient", "-I../subprojects/libmpdclient", "-fdiagnostics-color=always", "-D_FILE_OFFSET_BITS=64", "-Wall", "-Winvalid-pch", "-Wextra", "-Wpedantic", "-std=c99", "-O0", "-g", "-D_GNU_SOURCE", "-fvisibility=hidden", "-Wcast-qual", "-Wdouble-promotion", "-Wmissing-declarations", "-Wmissing-prototypes", "-Wshadow", "-Wstrict-prototypes", "-Wunused", "-Wstrict-prototypes", "-Wvla", "-Wwrite-strings", "-Wno-deprecated-declarations", "-MD", "-MQ", "mpc.p/src_args.c.o", "-MF", "mpc.p/src_args.c.o.d", "-o", "mpc.p/src_args.c.o", "-c", "../src/args.c"]
[...]

PR #2751, opened 2026-06-24
Manually cherry-pick fix for `normalize_path()`

SomeTiramisu:
As discussed in https://github.com/mozilla/sccache/issues/2510#2751
`normalize_path()` was copied from Cargo before commit
https://github.com/rust-lang/cargo/commit/7a6eaf96672360f8e0c52061054d8c708b9dbb62
which had an issue with ..s in processor output path (at least). This prevent
meson, or any exotic build environment to cache preprocessed objects.

This PR manually cherry pick the fix, but lack any associated test (yet)

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -913,7 +913,10 @@ fn process_preprocessor_line(
             None => include_path,
         }
     } else {
-        let path_buf = decode_path(include_path)?;
+        let mut path_buf = decode_path(include_path)?;
+        if path_buf.is_relative() {
+            path_buf = cwd.join(path_buf);
+        }
         let normalized = normalize_path(&path_buf);
         if normalized == path_buf {
             // `None` is a marker that the normalization is the same
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -2301,5 +2301,169 @@ mod test {
             // hash of `b"contents"`
             "a93900c371d997927c5bc568ea538bed59ae5c960021dcfe7b0b369da5267528",
         );
+
+        // `normalize_path()` case `^../`
+        let mut include_files = HashMap::new();
+        let fs_impl = TestFs {
+            metadata_results: Mutex::new(
+                [(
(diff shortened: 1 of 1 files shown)


AJIOB:
Hi @SomeTiramisu ,

Please, add the test to exclude the possible regression for that.

SomeTiramisu:
> Please, add the test to exclude the possible regression for that.

I'm new with the rust toolchain, so far I haven't found what obscure test triggered the `Component::ParentDir =>` path. Since this does not fail with the patch, i assume there is no regression there. The worst I can do is return an even more invalid path.

What I can do is testing the improved accuracy of `normalize_path()`  in case of `..`; adding test cases in `test_process_preprocessor_line_fs_access()` that would fail before the patch. 


Curiously, `msvc.rs` have it's own way of normalizing, `clang.rs` omit it completely

SomeTiramisu:
It's whether I'm on the right track or completely lost. I fear going too deep for what looks like an edge case optimisation (allow preprocessor cache when include path contain a `..`).

No more warning is a thing, but does this really improve performance? I still have to benchmark this (I will use mpc like in https://github.com/mozilla/sccache/issues/2510#issue-3699500735)

SomeTiramisu:
This now looks like a two in one PR. On one hand `normalize_path()` is now more correct with new code path tested. On the other hand `process_preprocessor_line()` now normalize on absolute path. The latter fix the original issue.

Benchmark on over optimistic scenario (full clean rebuild, same code) with sccache, on a meson project (GIMP in my case) shows improvement on build time (because preprocessor cache is now working) at the cost of a bit slower caching.

SomeTiramisu:
Rebase on main, CI is happy, I'm now confident enought to make it a genuine PR

I still find the tests a bit too verbose for my taste

SomeTiramisu:
Rebase on latest release, no merge issue, no code change

AJIOB:
Hi @sylvestre,

Is it OK for you?

AJIOB (approved):

PR #2752, opened 2026-06-26
 doc: document client-side and direct modes in Architecture.md 

sylvestre:


docs/Architecture.md
```diff
@@ -14,8 +15,158 @@ This schema shows at high level how sccache works.
...
+  backend and the accumulated statistics.
+
+The two communicate over a local IPC connection. There are two ways to divide
+the work between them.
+
+### Server-side mode (default)
+
+The CLI forwards the whole compilation to the daemon. The daemon runs the
+direct-mode preprocessor cache lookup, computes the hash, looks up the object
+cache, runs the compiler on a miss, stores the result, and streams the outputs
```
glandium: might be worth clarifying here because there are two types of outputs: output file and output streams (stdout, stderr). The former are written out on disk by the server.

docs/Architecture.md
```diff
@@ -14,8 +15,158 @@ This schema shows at high level how sccache works.
...
+
+`StorageGetPath` lets the CLI read a cached entry straight off disk when the
+backend exposes a local path; for backends that don't (S3, Redis, …) the CLI
+falls back to fetching the raw bytes with `StorageGetRaw`. Direct-mode entries
+are exchanged with `StorageGetPreprocessorEntry` / `StoragePutPreprocessorEntry`.
+Because each CLI process accumulates its own statistics, it flushes them to the
+daemon with `RecordStats` before exiting.
+
+Client-side mode is enabled with the `SCCACHE_CLIENT_SIDE` environment variable
+(or the `client_side_mode` config key). It is the recommended mode and is
```
glandium: It's kind of weird to say it's the recommended mode. If it really were, it would be the default.

glandium (changes requested):

PR #2754, opened 2026-06-27
cmdline: warn instead of recursing when invoked as a non-sccache name that is sccache itself

sylvestre:
When sccache is run under a name other than `sccache`, it acts as a compiler masquerade and prepends `sccache` to argv. If that name resolves back to the sccache binary itself (no real compiler of that name on PATH), the prepend made sccache wrap itself, and the `<compiler> -vV` rustc-detection probe re-entered the same path under the same name -- recursing forever and spawning a server-side process per core (a fork bomb that just hangs). Seen when an sccache binary was installed at a path whose filename was not `sccache`.

Detect that case and warn, running as a normal sccache invocation instead of prepending the name. Add an integration test that runs the binary under a non-sccache, non-PATH name and asserts it warns and behaves as sccache rather than hanging.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -231,7 +231,21 @@ pub fn try_parse() -> Result<Command> {
                             Ok(full_path) => args[0] = full_path.into(),
                             Err(_) => {}
                         }
-                        args.insert(0, env!("CARGO_PKG_NAME").into());
+
+                        // If $name resolves back to sccache itself, masquerading would
+                        // make it wrap itself and recurse forever (`<compiler> -vV`). Warn
+                        // and run as plain sccache instead of prepending the name.
+                        if PathBuf::from(&args[0]).canonicalize().ok()
+                            == env::current_exe().ok().and_then(|p| p.canonicalize().ok())
+                        {
+                            warn!(
+                                "sccache invoked as `{}`, which resolves to sccache itself; running as sccache (put a real `{}` on PATH to wrap it)",
+                                exe.display(),
+                                exe_filename.to_string_lossy()
+                            );
+                        } else {
+                            args.insert(0, env!("CARGO_PKG_NAME").into());
+                        }
                     }
                 }
             }
--- a/tests/sccache_args.rs
+++ b/tests/sccache_args.rs
@@ -57,6 +57,33 @@ fn test_gcp_arg_check() -> Result<()> {
     Ok(())
 }
 
+/// Running the sccache binary under a name that is not `sccache` and is not a
+/// real compiler on PATH must NOT make sccache wrap itself (which used to recurse
+/// forever via the `<compiler> -vV` detection probe). It should warn and behave
+/// as a normal sccache invocation instead. The test timeout guards the old hang.
+#[cfg(unix)]
+#[test]
+fn test_masquerade_as_self_warns_instead_of_recursing() -> Result<()> {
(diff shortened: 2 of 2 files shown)


src/cmdline.rs
```diff
@@ -231,7 +231,21 @@ pub fn try_parse() -> Result<Command> {
...
-                        args.insert(0, env!("CARGO_PKG_NAME").into());
+
+                        // If $name resolves back to sccache itself, masquerading would
+                        // make it wrap itself and recurse forever (`<compiler> -vV`). Warn
+                        // and run as plain sccache instead of prepending the name.
+                        if PathBuf::from(&args[0]).canonicalize().ok()
+                            == env::current_exe().ok().and_then(|p| p.canonicalize().ok())
+                        {
+                            warn!(
+                                "sccache invoked as `{}`, which resolves to sccache itself; running as sccache (put a real `{}` on PATH to wrap it)",
```
glandium, line 242: This should use CARGO_PKG_NAME for the warning message, arguably.

src/cmdline.rs
```diff
@@ -231,7 +231,21 @@ pub fn try_parse() -> Result<Command> {
                             Ok(full_path) => args[0] = full_path.into(),
                             Err(_) => {}
                         }
-                        args.insert(0, env!("CARGO_PKG_NAME").into());
+
+                        // If $name resolves back to sccache itself, masquerading would
+                        // make it wrap itself and recurse forever (`<compiler> -vV`). Warn
+                        // and run as plain sccache instead of prepending the name.
+                        if PathBuf::from(&args[0]).canonicalize().ok()
+                            == env::current_exe().ok().and_then(|p| p.canonicalize().ok())
```
glandium, line 239: Nit: when both sides fail, we get None == None, which makes it true and emit the warning and do the masquerading, which may not represent reality because we really don't know what's up. It might be better to emit a different warning in that case.

src/cmdline.rs
```diff
@@ -231,7 +231,21 @@ pub fn try_parse() -> Result<Command> {
...
+                        {
+                            warn!(
+                                "sccache invoked as `{}`, which resolves to sccache itself; running as sccache (put a real `{}` on PATH to wrap it)",
+                                exe.display(),
+                                exe_filename.to_string_lossy()
```
glandium, line 244: you should be able to use exe_filename.display()

src/cmdline.rs
```diff
@@ -231,7 +231,21 @@ pub fn try_parse() -> Result<Command> {
                             Ok(full_path) => args[0] = full_path.into(),
                             Err(_) => {}
                         }
-                        args.insert(0, env!("CARGO_PKG_NAME").into());
+
+                        // If $name resolves back to sccache itself, masquerading would
+                        // make it wrap itself and recurse forever (`<compiler> -vV`). Warn
+                        // and run as plain sccache instead of prepending the name.
```
glandium, line 237: I'm not sure this is the appropriate behavior. Like the typical case would be calling `cc -o foo.o -c foo.c` where `cc` is `sccache`, and there is no other `cc` in the `PATH`. This would make that command equivalent to `sccache -o foo.o -c foo.c` and fail because -o and -c are not sccache flags, which is a weird error mode. It would be better to say `cc` couldn't be found, probably doing so in the Err case above rather than add a somewhat redundant check.

glandium (changes requested):

PR #2755, opened 2026-06-29
gcc/clang: cache and distribute builds using quoted @response files

avikivity:
Compilations that pass arguments via a `@file` response file were silently
treated as uncacheable whenever the file contained a single or double quote.

Change response-file parsing to handle quoting correctly.

Add unit tests covering the tokenizer edge cases and an end-to-end test
verifying a response file with a quoted argument parses and stays cacheable.

### DIFF
--- a/docs/ResponseFiles.md
+++ b/docs/ResponseFiles.md
@@ -13,8 +13,8 @@ As defined by the [gcc docs](https://gcc.gnu.org/onlinedocs/gcc-4.6.3/gcc/Overal
 5. The response file may itself contain additional @file options; any such options will be processed recursively.
 
 Implementation details:
-- The gcc implementation in sccache supports all of these **except** #3. If a response file contains **any** quotations (`"` or `'`), the @file arg is treated literally and not removed (and its content not processed).
-- Additionally, sccache will not expand concatenated arguments such as `-include@foo` (see [#150](https://github.com/mozilla/sccache/issues/150#issuecomment-318586953) for more on this).
+- The gcc implementation in sccache supports all of these.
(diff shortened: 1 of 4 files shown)


avikivity:
Updates:
 - improved coverage
 - un-XFAILED tests that covered this bug

ahartmetz:
I think it would be worth spending some time to think about the commit messages and how to make them shorter without losing the essence. They read (AI)-sloppy. Producing text may be cheap, but reading it isn't.

avikivity:
While I generally agree that AI is overly verbose, I think it's better to say more than less. It's always possible not to read a commit message, but it's not possible to infer information that was not written in the first place.

I'll defer to @sylvestre here and adjust as he wishes.

sylvestre:
agreed, your comment #0 is too long and should focus only on what matters

ahartmetz:
> While I generally agree that AI is overly verbose, I think it's better to say more than less. It's always possible not to read a commit message, but it's not possible to infer information that was not written in the first place.
> 
I think commit messages should leave out what can be trivially found in the code or otherwise inferred without much thinking (I sometimes write too much, too, reviewing some past commit messages). If I really want to know what a commit does, I use git log -p.

avikivity:
Update: trimmed changelog of excess details.

ahartmetz:
At first I thought "Hey, some of that removed detail was useful!", but actually, that is in the comments, so 👍

CMake has started default-enabling C++ modules support, which currently breaks compilation with sccache, so fixing blockers for modules support decreases friction even for such "accidental" users of modules.

avikivity:
@sylvestre please review

avikivity:
Update: update docs/ResponseFiles.md

avikivity:
@sylvestre a gentle ping

avikivity:
@sylvestre another small ping

sylvestre:
thanks

PR #2760, opened 2026-07-01
[ci] enforce 'zizmor' checks for GitHub Actions configurations

jameslamb:
👋🏻 hi, `sccache` user for a few years, first-time contributor.

This PR proposes introducing `zizmor`, a static analyzer that enforces stricter security rules for GitHub Actions (https://github.com/zizmorcore/zizmor). I don't have any affiliation with that project, just genuinely think its suggestions are very useful.

## Changes

* adds `zizmor` pre-commit hook, pinned to its latest release
* pins most third-party GitHub Actions uses to commit SHAs instead of mutable tags
  - *kept the same versions you were already getting, any upgrades/downgrades in this PR are mistakes*
* passes inputs through `env:` to avoid script injection (https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable)
* adds a "dependency cooldown" for dependabot (a minimum age that releases must reach before dependabot recommends updating to them)
* adds explicit permissions to workflows that were missing it (ref: https://docs.github.com/en/actions/reference/security/secure-use#restricting-permissions-for-tokens)

### How I tested this

Ran `pre-commit` and `actionlint` locally

```shell
pre-commit run --all-files
actionlint .github/workflows/*
```

The only other way I know to test changes like this is to actually run CI.

### Related work

This would help with @sylvestre 's draft PR here: #2636

And would avoid findings like https://github.com/mozilla/sccache/pull/2729#discussion_r3375056305 that GitHub Advanced Security is adding to PRs here.

### Handling this PR

I know this is a sensitive and large-ish changeset and I'm a random person from the internet. I'm happy to break this into smaller PRs if you like, or for maintainers to close this and make these changes themselves.

Thanks for your time and consideration.

### DIFF
--- a/.github/actions/artifact_failure/action.yml
+++ b/.github/actions/artifact_failure/action.yml
@@ -9,6 +9,8 @@ runs:
   steps:
     - name: pack failure artifacts
       shell: bash
+      env:
+        INPUTS_NAME: ${{ inputs.name }}
       run: |
         echo "Current running processes"
         ps uax
@@ -23,8 +25,8 @@ runs:
             --exclude='docs' \
             --exclude='bins' \
             --exclude='.git' \
-            -zcf target/failure-${{ inputs.name }}.tar.gz .
-    - uses: actions/upload-artifact@v4
+            -zcf target/failure-${INPUTS_NAME}.tar.gz .
+    - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       with:
         name: ${{ inputs.name }}
         path: target/failure-${{ inputs.name }}.tar.gz
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,12 @@ updates:
     schedule:
       interval: weekly 
     open-pull-requests-limit: 8
+    cooldown:
+      default-days: 7
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: weekly
     open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: ci
 on: [ push, pull_request ]
 
+permissions:
+  contents: read
+
 jobs:
   lint:
     name: ${{ matrix.component }} ${{ matrix.os }}
@@ -19,7 +22,9 @@ jobs:
             cargo_cmd: clippy --locked --all-targets -- -D warnings -A unknown-lints -A clippy::type_complexity -A clippy::new-without-default
     steps:
       - name: Clone repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        with:
+          persist-credentials: false
 
       - name: Install rust
         uses: ./.github/actions/rust-toolchain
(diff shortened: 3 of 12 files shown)


.github/actions/artifact_failure/action.yml
```diff
@@ -23,8 +25,8 @@ runs:
             --exclude='docs' \
             --exclude='bins' \
             --exclude='.git' \
-            -zcf target/failure-${{ inputs.name }}.tar.gz .
-    - uses: actions/upload-artifact@v4
+            -zcf target/failure-${INPUTS_NAME}.tar.gz .
+    - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```
jameslamb, line 29: dependabot knows how to read these commit SHAs and will update the comments with versions too.

Example: https://github.com/lightgbm-org/LightGBM/pull/7329/changes

.github/dependabot.yml
```diff
@@ -5,8 +5,12 @@ updates:
     schedule:
       interval: weekly 
     open-pull-requests-limit: 8
+    cooldown:
+      default-days: 7
```
jameslamb, line 9: I chose this value arbitrarily.

And value `> 0` will be sufficient to slow the spread of problematic packages in the event of a supply chain attack.

.github/zizmor.yml
```diff
@@ -0,0 +1,4 @@
+rules:
+  unpinned-images:
+    ignore:
+      - ci.yml
```
jameslamb, line 4: Suppresses this finding:

```text
error[unpinned-images]: unpinned image references
   --> .github/workflows/ci.yml:264:5
    |
264 |     container: ${{ fromJson(matrix.container || '{"image":null}') }}
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ container image may be unpinned
    |
    = note: audit confidence → Lo
```

Looking at the configuration around that line, it's pretty clear that you're fine with unpinned images.

https://github.com/mozilla/sccache/blob/68cd3308d5b29747381df8ca3df89934f47bbc35/.github/workflows/ci.yml#L253

I think this one is a false positive and not relevant for this project.

.github/workflows/ci.yml
```diff
@@ -448,9 +463,13 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ build, lint, test ]
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
+    permissions:
+      contents: write
```
jameslamb, line 467: This `release` job needs elevated permissions for this:

https://github.com/mozilla/sccache/blob/68cd3308d5b29747381df8ca3df89934f47bbc35/.github/workflows/ci.yml#L490-L492

This is one thing I've really liked about `zizmor` in other projects... it encourages you to make such things explicit and scope them down narrowly.

All other jobs in this workflow should be able to work with just `contents: read`, this one needs `contents: write`.

.github/workflows/ci.yml
```diff
@@ -130,10 +139,12 @@ jobs:
       CARGO_PROFILE_DEV_DEBUG: '0' # save disk space
     steps:
       - if: ${{ contains(matrix.os, 'windows') }}
-        uses: ilammy/msvc-dev-cmd@v1
+        uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
```
jameslamb, line 142: `ci.yml` didn't run because of this change.

> _The action ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 is not allowed in mozilla/sccache because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: ..._

https://github.com/mozilla/sccache/actions/runs/28532952024

It doesn't show in that output, but I suspect you have a rule like `ilammy/msvc-dev-cmd@v1` or similar in the repo (organization-wide) allowlist for third-party actions.

If you like the direction of this PR, then the choices for that list are:

* use permissive ranges (`ilammy/msvc-dev-cmd@*`)
* use commit SHAs (`ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756`) and update it manually whenever you want to bump versions of actions

PR #2769, opened 2026-07-15
Fix description of Unix socket-based Redis connection

Toorero:
OpenDAL parses the URI with http::Uri which _expects_ a host. If no host is specified parsing is going to fail.


Toorero:
You can see the problem with parsing [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=76cfe1d9b288030bcb9194a70815d5f0)

Issue #2775, opened 2026-07-21
RFC: making sccache-dist work for OpenEmbedded/Yocto host-mode distributed builds


jetm:
I've been getting sccache-dist working for OpenEmbedded/Yocto host-mode
distributed builds, and the work has grown well past a single reviewable change.
I originally opened it as #2750, which I've now closed and am splitting into
small, focused PRs landed in waves. This issue is the discussion anchor for that
series, especially the scheduler work, before those PRs land.

## What the series covers

Bug fixes (independent, landing first):
- A server-side panic when a finished compile reports neither exit code nor
  signal.
- Local fallback when a distributed compile fails or drops an output, instead of
  failing the whole build.
- gcc/rust distributed-compile correctness: force-local for feature probes and
  PCH generation; ship the rust sysroot std, target spec, and rlibs; toolchain
  packaging fixes for relocated and split-sysroot cross toolchains.

Scheduler and server reliability under high concurrency (a base-first stack):
- Prompt removal of stopped servers; topology-aware, deterministic routing;
  idempotent job-state updates and lost/stale-job reaping; liveness-lease
  reaping; restart-safe job accounting.
- Per-toolchain preparation locks so concurrent compiles stop convoying, and
  HTTP connection reuse.

## Two compatibility surfaces I want to flag up front

1. The liveness-lease reaper adds an `active_jobs` field to the heartbeat
   request, which changes the bincode wire format: scheduler and build servers
   must run matching binaries. Old-server to new-scheduler fails loudly;
   new-server to old-scheduler silently drops the field. I'm happy to gate this
   behind a protocol version if you'd prefer.
2. Connection reuse adds two `DistConfig` keys (with `serde(default)`, so old
   configs keep parsing) and a small `arc-swap` dependency.

## Questions for maintainers

- The scheduler routing work introduces some new policy surface (topology-aware
  placement, a dynamically derived colocated de-weight, advertised core counts).
  It is opt-in via config with defaults that preserve current behavior. Is that
  direction welcome, or would you rather shape it before the PRs land?
- There is a set of load-diagnosis instrumentation I used to find these issues.
  I can submit it, fold the useful parts into the PRs that needed them, or drop
  it. Preference?

I'll link the individual PRs here as they open. Happy to adjust the split,
ordering, or scope to match how you'd like to review this.

PR #2776, opened 2026-07-21
server: don't panic when a finished compile has neither code nor signal

jetm:
Part of #2775 (splitting the OpenEmbedded/Yocto sccache-dist work into focused
PRs; replaces the closed #2750).

## Problem

`get_signal` did `status.signal().expect("must have signal")`, assuming the Unix
invariant that an `ExitStatus` with no exit code was terminated by a signal. That
does not always hold: an `ExitStatus` reconstructed for a distributed compile (or
an abnormal wait status such as `WIFSTOPPED`) can report neither a code nor a
signal. When that happened the `expect()` panicked the compile task, which the
server surfaced as a misleading "Failed to bind socket" and, under load,
repeatedly fell back to local compilation. Found under a Yocto/BitBake host-mode
distributed build.

## Approach

Return `Option<i32>` from `get_signal` and assign it straight into `res.signal`,
so a compile that reports neither code nor signal leaves `res.signal` unset
instead of crashing the in-flight task. The Windows arm returns `None` rather
than panicking; `ExitStatus::code()` is always `Some` there, so the signal branch
is never reached anyway.

## Test

Adds a unit test covering a real terminating signal (SIGKILL) and the
neither-code-nor-signal case (`WIFSTOPPED` via `from_raw`), which previously
panicked.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -133,13 +133,20 @@ fn notify_server_startup(name: Option<&OsString>, status: ServerStartup) -> Resu
 }
 
 #[cfg(unix)]
-fn get_signal(status: ExitStatus) -> i32 {
+fn get_signal(status: ExitStatus) -> Option<i32> {
     use std::os::unix::prelude::*;
-    status.signal().expect("must have signal")
+    // None when the process produced neither an exit code nor a terminating
+    // signal - e.g. an ExitStatus synthesized from a distributed-compile result,
+    // or an abnormal wait status. Previously this was `.expect("must have
+    // signal")`, which panicked the compile task (surfacing as a misleading
+    // "Failed to bind socket") and could be tripped repeatedly under load.
+    status.signal()
 }
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


jetm:
The dist-tests failure here was pre-existing on the base I branched from rather than from this change: main at 7cd4f2c3 fails the same two tests with nothing applied (`test_dist_cargo_build` and `test_dist_cargo_makeflags`, both expecting `dist_compiles == 1` and getting 0): https://github.com/mozilla/sccache/actions/runs/29833785987. I have rebased onto current main, which includes the dist test fixes and the MSRV bump. The new runs need an approval before they start.

On the CodSpeed report: it flags the same 46% regression on this PR and on #2777, and it also notes "Different runtime environments detected". This change turns one `expect()` into an `Option` and adds a test, so I read that number as a baseline artifact rather than a real regression. Happy to dig in if you think otherwise.

PR #2777, opened 2026-07-21
compiler: fall back to local on distributed-compile failure

jetm:
Part of #2775 (splitting the OpenEmbedded/Yocto sccache-dist work into focused
PRs; replaces the closed #2750).

## Problem

Any remote failure during a distributed compile - a non-zero result, a dropped
output, or unmatched metadata - failed the whole build instead of just the
distribution, so a single distributable input that misbehaved remotely broke the
build. Separately, sccache's own log output on stderr corrupted autoconf/libtool
probe output (for example a libtool `-fPIC` probe reading wrapper logs off
stderr).

## Approach

- Treat a distributed-compile failure (a non-zero remote result, or a declared
  output that did not materialize) as a trigger to recompile locally.
- Honor optional outputs so a missing optional artifact does not fail the
  compile.
- Route wrapper log output off the compiler's stderr.
- Log the distribute/local decision at info and any remote failure at debug, so
  the fallback path is diagnosable.

## Evidence

Local fallback is exercised repeatedly per cold Yocto build with zero build
failures; the libtool `-fPIC` probe reproduction is fixed.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -973,7 +973,7 @@ where
                 )
             })?;
 
-        let mut jc = match jres {
+        let jc = match jres {
             dist::RunJobResult::Complete(jc) => jc,
             dist::RunJobResult::JobNotFound => bail!("Job {} not found on server", job_id),
         };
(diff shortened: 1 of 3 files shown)


AJIOB:
Hi, thanks for the PR.

Looks like it also resolves #2700 and #2745

AJIOB:
For discussion: maybe we need a feature flag for that

AJIOB:
Hi @sylvestre ,

Is it OK for you?

jetm:
@AJIOB thanks for looking at this.

Yes, it covers #2745 directly and should also resolve #2700. #2745 asks for the retry-locally behaviour plus visibility, and both are here: the fallback is logged at info (which server, and why it fell back) with the remote's stderr at debug, and the failure is already counted as `dist_errors` in `sccache --show-stats`, so a degraded cluster shows up rather than silently slowing the build. For #2700 the `.incbin` compile fails on the remote and previously took the build down with it; with this change it is retried locally and the build completes.

On the feature flag: I defaulted it to on because the alternative is a hard build failure for something sccache can recover from, which is what #2745 is asking to avoid. The counter-argument I see is CI, where you might prefer a loud failure over silently slower local builds masking a broken cluster. I am happy to add an opt-out for that, either a `dist.fallback` config key defaulting to true or an env override. Want me to add it in this PR, and do you have a preference on the name?

On the red dist-tests: that failure was pre-existing on the base I branched from, not from this change. main at 7cd4f2c3 fails the same two tests with no patches applied (`test_dist_cargo_build` and `test_dist_cargo_makeflags`, both expecting `dist_compiles == 1` and getting 0): https://github.com/mozilla/sccache/actions/runs/29833785987. I have rebased onto current main, which picks up the dist test fixes and the MSRV bump. The new workflow runs are sitting in `action_required`, so they need an approval to start.

PR #2781, opened 2026-07-26
Add support for /d20bforceinline

MatthewJM96:
## Summary

- Add support for an MSVC optimisation flag, `/d20bforceinline`, which causes `__forceinline` to be respected in debug builds.

## Scope

One change in `src/compiler/msvc.rs` to add a pass through for the flag, same as `/d1nodatetime`.

## Validation

Updated pass through test to include the new flag.


src/compiler/msvc.rs
```diff
@@ -2325,6 +2327,7 @@ mod test {
                 "-Qpar-",
                 "-Gw",
                 "/d1nodatetime",
+                "/d20bforceline",
```
AJIOB, line 2330: Typo in flag name, so the CI was failed

PR #2782, opened 2026-07-27
Add support for arg files in Rust

ranger-ross:
This PR adds support for arg files (aka response files) in Rust. 

Cargo is planning to make some changes in https://github.com/rust-lang/cargo/issues/15010 which will likely increase the total number of args passed to rustc, making sccache's lack of support for arg files more painful to users. (especially to windows users)

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -50,7 +50,7 @@ use std::future::Future;
 use std::hash::Hash;
 #[cfg(feature = "dist-client")]
 use std::io;
-use std::io::Read;
+use std::io::{BufReader, Read};
 use std::iter;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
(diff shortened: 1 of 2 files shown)


src/compiler/rust.rs
```diff
@@ -1065,6 +1065,60 @@ counted_array!(static ARGS: [ArgInfo<ArgData>; _] = [
...
+    fn next(&mut self) -> Option<OsString> {
+        loop {
+            let arg = self.stack.pop()?;
+            let file = match arg.split_prefix("@") {
+                Some(arg) => self.cwd.join(arg),
+                None => return Some(arg),
+            };
+
+            let mut contents = String::new();
+            let res = fs_err::File::open(&file).and_then(|mut f| f.read_to_string(&mut contents));
```
Kobzol: This should probably be wrapped in a `BufReader`, to avoid doing many syscalls for each read? Not sure if `read_to_string` already handles that.

ranger-ross: good point! The arg files can be decently large so seems like a good idea to do that.

There are some other reads in this file that as use `read_to_string()` but I think those are reading much smaller files so the they probably get everything in one syscall.

Updated in the latest push

PR #2715, opened 2026-05-20
Bump opendal to 0.58.1 and fix fallout (fixes local GCS cache usage)

flip1995:
## Motivation

- https://github.com/apache/opendal/issues/6287

The gcs cache can currently not be used with `authorized_user` application-default credentials, as this was not supported in `reqsign` version 0.16.5, which `opendal` version `0.55.0` depended on. This was fixed in a later version of reqsign and since has been updated in `opendal`

- https://github.com/apache/opendal/issues/6553

With this, the gcs cache should Just Work :tm: after a `gcloud auth application-default login`. 

## Changes

I split this PR into the following commits for easier commit-by-commit review:

### Bump MSRV to 1.91.0

opendal 0.58.1 has an MSRV of 1.91.0, so sccache needs to match this in order to bump the opendal dependency.

### Bump opendal to 0.58.1

- Required bump of reqwest to 0.13.0
    - Rename and remove features deprecated in that release
    - https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0
    - https://github.com/seanmonstar/reqwest/pull/2891
- Remove direct `reqsign` dependency that was not used in the code
- Add new (optional) opendal-* dependencies that were split out of opendal

### Run cargo update

I first did just a partial Cargo.lock update for only the dependencies that were bumped in this PR. But the churn was quite big when just doing that. So I figured why not do a full, clean `cargo update` of the `Cargo.lock` file.

### Fix fallout in code

- `HttpClientLayer` was removed and now has to be built from the `OperationContext`
- `LoggingLayer` was moved to its own crate
- `allow_anonymous` was deprecated in favor of `skip_signature`
- Use `sha256` instead of `sha1`, as that is quasi-deprecated. I'm not sure if this is actually necessary. I think it fixed some issue in the tests? In any case, I think this is a good change
- Check if `native_certs` are available when building a reqwest client, and if not set the TLS certs to an empty vector. With reqwest 0.13, the client would fail otherwise.

PR #2783, opened 2026-07-28
Release 0.17.0

sylvestre:


weihanglo:
Do we want to wait for <https://github.com/mozilla/sccache/pull/2715> or is it fine we have a follow-up patch version release?

AJIOB:
IMO, #2781 fix looks nice too: just need to fix the test case in changes

sylvestre:
Thanks for the feedback but both aren't ready

PR #2786, opened 2026-07-29
chore(deps): upgrade opendal to 0.57.0

cryptomilk:
Bumps reqwest to 0.13 to match opendal-core 0.57.0's requirement, avoiding two reqwest versions in the dependency graph.

Enables the now-optional layers-logging feature for LoggingLayer, and switches from the deprecated allow_anonymous() to skip_signature().


cryptomilk:
reqwests rustls feature offers only aws-lc-rs.

We have two options:

1. Switch reqwest to native-tls everywhere
2. Install bindgen+clang in CI containers

Xuanwo:
covered by https://github.com/mozilla/sccache/pull/2715

PR #2793, opened 2026-08-05
Bump MSRV to 1.91.0

flip1995:
In preparation of the opendal bump to version 0.58.1. opendal 0.58.1 has an MSRV of 1.91.0, so sccache needs to match this in order to bump the opendal dependency.

cc #2715


Xuanwo (approved):
Thank you! Let's go!

Issue #2763, opened 2026-07-02
Feature request: Patch obj files on windows to point to the correct source file


rconde01:
I'm using multiple worktrees at once using the base dirs functionality. This works fine for getting cache hits. However, if i'm debugging and I hit an exception, Visual Studio will load the source file from location where the obj was originally built instead of the worktree i'm working in. Then i make changes to that file and scratch my head for a second while I wonder why they don't seem to work. This makes sense, but it seems theoretically possible to fix. If we can have the cached obj contain a very long path, you should be able to patch it to the actual path on copy. Or maybe there's an entirely different approach that could work, but I wanted to capture the issue at least.

PR #2794, opened 2026-08-05
rust: support SCCACHE_BASEDIRS across checkout roots 

### LINKED ISSUE #2652: Wire SCCACHE_BASEDIRS into Rust hash key

### LINKED ISSUE #2765: Should exclude --remap-path-prefix from cache key
Say I have a Rust project and I add a worktree. I want debuginfo to be valid across trees, so we normalize using `--remap-path-prefix`. `sccache` normalizes the source directory when basedirs is set, so we get a nice shared cache hit. Now, I add a new worktree. I have to go to all other worktrees and add the new worktree's base to the `--remap-path-prefix`  set, which changes the cache key, blowing away cache that could otherwise be reused.

Shouldn't `sccache` normalize away `--remap-path-prefix` when it's congruent with one of the sccache base directories? Or better yet, add its own `--remap-path-prefix` so shared debuginfo Just Works?

pamarcos:
First, I want to point that I've used GPT-5.6 Sol to assist me with the work. However, I've reviewed the code and tested it.

Rust compilation currently includes checkout-specific paths in its cache key, so the existing `SCCACHE_BASEDIRS` support does not provide cross-checkout cache hits for Rust crates.

This adds Rust support while preserving location sensitivity for path inputs that are not safely covered by compiler remapping. In particular, normalization requires an effective `--remap-path-prefix` covering the working directory with the default or `all` remap scope. Without that compiler-side remapping, the working directory and other path-sensitive inputs remain exact.

This work is based on https://github.com/mozilla/sccache/pull/2678 (Matt remains the author of the imported baseline commit). I also found that there's a similar PR at https://github.com/johnkferguson/sccache/pull/1. Not sure why a PR upstream was never created.

So, I decided to create this PR because the other ones seem not to be followed up, but I'm happy to close this as a duplicate if the others are planned to be merged.

<details>
<summary>Implementation details</summary>

- normalizes only the source side of parsed `--remap-path-prefix` values, preserving the destination,
  mapping precedence, component matching, and final remap scope;
- hashes the effective remapped identity of every source file and preserves original identity when
  multiple sources map to the same virtual path;
- keeps `env!`, `option_env!`, profiles, plugins, and unremapped paths location-sensitive;
- normalizes a small allowlist of path-valued `CARGO_*` variables, including
  `CARGO_INSTALL_ROOT`, only when the working directory is remapped, there are no explicit externs,
  and crate search paths contain no dynamic libraries;
- frames normalized values and advances the Rust cache-key version from `6` to `7`;
- falls back to local compilation when a distributed path transformer cannot safely preserve a user
  remap; and
- declines `SCCACHE_BASEDIRS` prefix stripping for non-ASCII Windows values.

This deliberately avoids scanning arbitrary argument, linker, environment, or replacement-path bytes. Those values may affect compiler output without being rewritten by `rustc`, so stripping them from the key could return an artifact containing paths or semantics from another checkout.

</details>

<details>
<summary>Full provenance and related discussions</summary>

### Direct lineage

| Reference | Relevance |
|---|---|
| https://github.com/mozilla/sccache/issues/35 | Original request for a `CCACHE_BASEDIR` equivalent. |
| https://github.com/mozilla/sccache/pull/104 | Early unmerged basedir implementation. |
| https://github.com/mozilla/sccache/pull/2521 | Mikhail Shiryaev's merged `SCCACHE_BASEDIRS` foundation for C/C++. |
| https://github.com/mozilla/sccache/pull/2270 | Made Rust `--remap-path-prefix` compilations cacheable. |
| https://github.com/mozilla/sccache/pull/2651 | Stopped separately hashing `CARGO_ENCODED_RUSTFLAGS`, which often carries the remap. |
| https://github.com/mozilla/sccache/issues/2652 | Weihang Lo's direct request for Rust support and warning that key-only normalization is unsafe. |
| https://github.com/vercel/sccache/pull/10 | Matt Mastracci's original Rust prototype. |
| https://github.com/mozilla/sccache/pull/2678 | Matt Mastracci's direct upstream implementation and the baseline for this PR. |

### Independent implementations and experiments

| Reference | Relevance |
|---|---|
| https://github.com/johnkferguson/sccache/pull/1 | Glob/worktree experiment, per-token argument diagnosis, `CARGO_INSTALL_ROOT`, and reported 98.6% hit rate. |
| https://github.com/manoelcalixto/sccache/pull/1 | Independent opt-in linked-worktree implementation with extensive safety and CI coverage. |
| https://github.com/zackees/zccache/issues/229 | Related request for safe Rust cache sharing across source roots. |
| https://github.com/zackees/zccache/pull/237 | Root-covering remap gate and path-sensitive Rust integration tests in `zccache`. |
| https://github.com/moriyoshi/sccache-wrapper | Standalone normalized-artifact cache wrapper for cross-worktree reuse. |
| https://github.com/NVIDIA/OpenShell/pull/2379 | Operational evidence of zero Rust cross-worktree hits before normalization. |
| https://github.com/NVIDIA/OpenShell/pull/2475 | Follow-up experiment using `SCCACHE_BASEDIRS`. |

### Cache-key and artifact-safety context

| Reference | Relevance |
|---|---|
| https://github.com/mozilla/sccache/issues/196 | Historical report of shared Rust dependencies missing across projects. |
| https://github.com/mozilla/sccache/pull/206 | Earlier path-independent Rust key attempt, left unmerged because artifacts differed by path. |
| https://github.com/mozilla/sccache/pull/345 | Added Rust working-directory hashing because the path is embedded in rlibs. |
| https://github.com/mozilla/sccache/issues/986 | Discussion of whether and why the working directory belongs in cache keys. |
| https://github.com/mozilla/sccache/issues/2494 | Discussion of which `CARGO_*` variables should affect Rust keys. |
| https://github.com/mozilla/sccache/pull/2495 | Alternative proposal to stop hashing redundant or irrelevant `CARGO_*` variables. |
| https://github.com/mozilla/sccache/pull/2673 | Analogous parsed source-side normalization for C/C++ prefix-map arguments. |
| https://github.com/mozilla/sccache/pull/2711 | Related C/C++ argument normalization and per-argument key framing. |
| https://github.com/mozilla/sccache/issues/2763 | Demonstrates the debugger/source-path consequence of serving an artifact from another checkout. |
| https://github.com/mozilla/sccache/issues/2765 | Daniel Colascione's direct request to normalize congruent Rust `--remap-path-prefix` values. |
| https://github.com/mozilla/sccache/issues/2595 | Worktree workflow and dynamic basedir discovery requirements. |
| https://github.com/mozilla/sccache/issues/2737 | Windows basedir matching bug report. |
| https://github.com/mozilla/sccache/pull/2736 | Windows escaped-separator fix in shared basedir handling. |
| https://github.com/rust-lang/cargo/issues/10915 | Upstream evidence that moving `CARGO_HOME` changes path-sensitive artifacts/fingerprints. |
| https://github.com/rust-lang/cargo/issues/12137 | Cargo `trim-paths` and remap-scope tracking issue. |

</details>

## Benchmarks building ClickHouse

Running on a AMD Ryzen 9 9950X with 96G

| Scenario | Average build time | Sample SD | C/C++ hits | C/C++ misses | Rust hits | Rust misses |
|---|---:|---:|---:|---:|---:|---:|
| ccache | 142.73s | 2.03s | 16,614 | 2 | Not wrapped | Not wrapped |
| sccache with Rust | **54.91s** | 0.63s | 16,730 | 2 | 822 | 0 |
| sccache with Rust, client-side | 154.72s | 0.27s | 16,730 | 2 | 808 | 0 |
| sccache C/C++ only | 144.03s | 0.78s | 16,614 | 2 | Not wrapped | Not wrapped |

The interesting part here is that using the new client-side architecture, it slows down a lot. I'll probably create a separate issue to point that.

Closes: https://github.com/mozilla/sccache/issues/2652
Closes: https://github.com/mozilla/sccache/issues/2765
Related: https://github.com/mozilla/sccache/pull/2678
Related: https://github.com/mozilla/sccache/issues/2595

### DIFF
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -4571,22 +4571,24 @@ proc_macro false
                 Some(format!("{prefix}=/workspace").into()),
             )]
         };
+        let joined =
+            |prefix: &str, suffix: &str| Some(Path::new(prefix).join(suffix).into_os_string());
         assert!(super::remap_path(Path::new("/home/user/project"), &remap("/home/user")).is_some());
         assert!(super::remap_path(Path::new("/home/user"), &remap("/home/user=part")).is_none());
         assert_eq!(
             super::remap_path(Path::new("/home/user"), &remap("/")),
-            Some("/workspace/home/user".into())
+            joined("/workspace", "home/user")
         );
         assert_eq!(
             super::remap_path(Path::new("/home/user/project"), &remap("/home/user/")),
-            Some("/workspace/project".into())
+            joined("/workspace", "project")
         );
         assert!(
             super::remap_path(Path::new("/home/username/project"), &remap("/home/user")).is_none()
         );
         assert_eq!(
             super::remap_path(Path::new("/home/a=b/project"), &remap("/home/a=b")),
-            Some("/workspace/project".into())
+            joined("/workspace", "project")
         );
 
         let overlapping = vec![
@@ -4598,7 +4600,7 @@ proc_macro false
         ];
         assert_eq!(
             super::remap_path(Path::new("/home/user/project"), &overlapping),
-            Some("/last/project".into())
+            joined("/last", "project")
         );
 
         let mut diagnostics = remap("/home/user");
--- a/src/compiler/rust.rs
+++ b/src/compiler/rust.rs
@@ -175,20 +175,31 @@ fn is_path_cargo_env(var: &OsString) -> bool {
     ) || var.as_encoded_bytes().starts_with(b"CARGO_BIN_EXE_")
 }
 
+fn omits_dep_info_target(flag: &OsString, value: Option<&OsString>) -> bool {
+    flag == "-Z"
+        && value.and_then(|value| value.to_str()).is_some_and(|value| {
+            value.split_once('=').map_or(value, |(name, _)| name) == "dep-info-omit-d-target"
+        })
+}
+
+#[derive(Clone, Debug)]
+struct DepInfoTemplate {
+    target_dependencies: String,
+    tail: String,
+}
+
 fn rewrite_dep_info_targets(
     dep_info: &Path,
     outputs: &HashMap<String, ArtifactDescriptor>,
     cwd: &Path,
+    template: &DepInfoTemplate,
 ) -> Result<()> {
     let dep_info = cwd.join(dep_info);
     let deps = fs::read_to_string(&dep_info).context("Failed to read cached Rust dep-info")?;
-    let mut rewritten = String::with_capacity(deps.len());
-    let mut matched_output = false;
-    let mut scanning_output_targets = true;
+    let mut output_targets = Vec::new();
 
     for line in deps.split_inclusive('\n') {
-        if !scanning_output_targets || line.trim().is_empty() {
-            rewritten.push_str(line);
+        if line.trim().is_empty() {
             continue;
         }
 
@@ -232,22 +243,28 @@ fn rewrite_dep_info_targets(
                 }
             }
         }
-        let Some((separator, _, output)) = matched else {
-            if !matched_output {
+        let Some((_, _, output)) = matched else {
+            if output_targets.is_empty() {
                 bail!(
                     "No output targets matched cached Rust dep-info {}",
                     dep_info.display()
                 );
             }
-            scanning_output_targets = false;
-            rewritten.push_str(line);
-            continue;
+            break;
         };
 
-        matched_output = true;
-        rewritten.push_str(&output.path.to_string_lossy());
-        rewritten.push_str(&line[separator..]);
+        output_targets.push(output);
     }
+
+    let mut rewritten = String::with_capacity(deps.len());
+    for (index, output) in output_targets.iter().enumerate() {
+        rewritten.push_str(&output.path.to_string_lossy());
+        rewritten.push_str(&template.target_dependencies);
+        if index + 1 < output_targets.len() {
+            rewritten.push('\n');
+        }
+    }
+    rewritten.push_str(&template.tail);
 
     if rewritten != deps {
         let parent = dep_info
@@ -433,6 +450,8 @@ pub struct RustCompilation {
     crate_types: CrateTypes,
     /// If dependency info is being emitted, the name of the dep info file.
     dep_info: Option<PathBuf>,
+    /// Dependency records generated locally for the current invocation.
+    dep_info_template: Option<DepInfoTemplate>,
     /// The current working directory
     cwd: PathBuf,
     /// The environment variables
@@ -462,7 +481,7 @@ async fn get_source_files_and_env_deps<T>(
     cwd: &Path,
     env_vars: &[(OsString, OsString)],
     pool: &tokio::runtime::Handle,
-) -> Result<(Vec<PathBuf>, Vec<(OsString, OsString)>)>
+) -> Result<(Vec<PathBuf>, Vec<(OsString, OsString)>, DepInfoTemplate)>
 where
     T: CommandCreatorSync,
 {
@@ -494,7 +513,7 @@ where
         })
         .await?;
 
-    parsed.map(move |(files, env_deps)| {
+    parsed.map(move |(files, env_deps, template)| {
         trace!(
             "[{}]: got {} source files and {} env-deps from dep-info in {}",
             crate_name,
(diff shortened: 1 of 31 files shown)


sylvestre:
Long comment #0 written by a llm isn't useful... Please make it shorter

pamarcos:
> Long comment #0 written by a llm isn't useful... Please make it shorter

Roger. I cut most of it. Left the implementation details available in case someone's interested, though.

src/compiler/rust.rs
```diff
@@ -62,6 +62,120 @@ use std::time;
...
+    matches!(
+        var.to_str(),
+        Some(
+            "CARGO_HOME"
+                | "CARGO_INSTALL_ROOT"
+                | "CARGO_MANIFEST_DIR"
+                | "CARGO_MANIFEST_PATH"
+                | "CARGO_TARGET_DIR"
+                | "CARGO_TARGET_TMPDIR"
+                | "CARGO_WORKSPACE_DIR"
```
weihanglo: This is LLM hallucination. At least Cargo doesn't set any `CARGO_WORKSPACE_DIR` env var

src/compiler/rust.rs
```diff
@@ -1572,6 +1737,26 @@ where
             .cloned()
             .collect();
         env_vars.sort();
+        // Procedural macros can read inherited environment variables without reporting them in
+        // rustc dep-info. They can be explicit externs or resolved from a crate search path.
+        let normalize_cargo_paths = !basedirs.is_empty()
+            && remap_path(&cwd, &os_string_arguments).is_some()
+            && env_vars.iter().any(|(var, _)| is_path_cargo_env(var))
```
weihanglo: Env sanitization is better to be be split out to its own PR

src/compiler/rust.rs
pamarcos: I removed it from the built-in allowlist and added a negative assertion in 25b2f7f. Thanks for catching this.

src/compiler/rust.rs
pamarcos: I can still split it in a separate PR if you insist, but I didn't do it at first because is a hard requirement for this feature to work with cargo. Otherwise, AFAIU without this we could only make the rustc compiler support cross-worktree cache hits.

weihanglo: Yeah, the rationale is well understood. Just feel like a fix of <https://github.com/mozilla/sccache/issues/2494> could be shipped incrementally without too much burden on review side, and has value on its own, and is also easy to revert/bisect if something bug pops up.

I not a maintainer btw. You could keep your original comment I don't mind.

pamarcos: Alright, I'll keep it as it is for now waiting for a maintainer to give their opinion on it. I'll do whatever makes everyone's life easier

src/compiler/rust.rs
weihanglo: If it is never be set, it makes little sense to have a negative assertion.

pamarcos: That's right, removed at https://github.com/mozilla/sccache/pull/2794/commits/8fca89f22946ea4c8fe0053397a29ba33d4475ef

packplusplus:
We're running into these cache path issue with our shared ci runners writing back to shared s3 (multiple runners on a single host with randomized build dirs). I'm running a POC on if this works for us. I feel like cleaning up the failed ci jobs would be a positive step in getting it merged while you wait for a maintainer. 

Edit: The POC work is interesting. The conservative decision `no explicit externs, and crate search paths contain no dynamic libraries` made the cache efficacy for several of our builds negligible. An ai slopped version that allows ordinary `.rlib` externs and `-L dependency` normalization gave better caching (closer to 25%), but I'm not sure the safety of it (certainly not enough to propose to upstream). 

Edit 2: the slopped version is too risky, it 100% had a physical path from a cached library artifact. there may still be a path, but i'm wary of diving too deep on this.

PR #2795, opened 2026-08-05
fix(c): invalidate cache for .include and .incbin directives

bleggett:
This fixes a really nasty bug with sccache for C code where `.incbin` and `.include` directives weren't fully checked for cache consistency before being returned in the build cache.

This led to a case where a Linux kernel build that used `sccache` ended up with kmods signed with the *wrong signing key*, because the kernel's `certs/system_certificates.S` embeds the cert via assembler .incbin, and `sccache` would mistakenly return a stale signing key from a _previous_ build :upside_down_face: 

Ex:
certs/system_certificates.S embeds the cert via assembler .incbin:

```
  asm
  __module_cert_start:
        .incbin "certs/signing_key.x509"
  __module_cert_end:
        .incbin "certs/x509_certificate_list"
```

The sizes come from label arithmetic, not from the source text. So the preprocessed text of that file is byte-identical on every build regardless of which key is in signing_key.x509. `sccache` hashes the preprocessed output, gets a hit, and hands back an object containing the _old_ certificate.

`ccache` refuses to cache any translation unit containing .incbin precisely because of this (https://github.com/ccache/ccache/issues/136, documented in the ccache manual https://ccache.dev/manual/4.13.6.html). 

`sccache` has the detector but only wires it into preprocessor cache mode, which isn't enough.

Tested with a kernel build using this SHA, mismatch issue goes away. Also added some more unit tests.

I also did a pass to see if there are other similarly-shaped problems where the cache hashing/keying might be broken and return the wrong thing and found a few more minor ones, mostly around plugins/specs/struct layout randomization, those are fixed in the second commit.

### DIFF
--- a/src/compiler/c.rs
+++ b/src/compiler/c.rs
@@ -32,7 +32,7 @@ use crate::util::{
 use async_trait::async_trait;
 use fs_err as fs;
 use std::borrow::Cow;
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::ffi::{OsStr, OsString};
 use std::fmt;
 use std::hash::Hash;
@@ -503,6 +503,9 @@ where
                         );
                         return Ok(HashResult {
                             key,
+                            // Preprocessor cache mode is explicitly disabled for any
+                            // translation unit containing an `.incbin`.
+                            cacheable: Cacheable::Yes,
                             compilation: Box::new(CCompilation {
                                 parsed_args: self.parsed_args.clone(),
                                 is_locally_preprocessed: false,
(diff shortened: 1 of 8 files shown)


AJIOB:
This also should fix #2700

AJIOB:
Hi @bleggett ,

Looks like you have conflicts. Please, rebase your changes

bleggett:
> Hi @bleggett ,
> 
> Looks like you have conflicts. Please, rebase your changes

Rebased, thanks!

AJIOB:
Thanks for the fixes, @bleggett 

Is it ok for you, @sylvestre ?

src/compiler/c.rs
```diff
@@ -694,6 +793,174 @@ const PRAGMA_GCC_PCH_PREPROCESS: &[u8] = b"pragma GCC pch_preprocess";
...
+/// output alone therefore misses the dependency entirely: the named file can
+/// change while the cache key stays the same, and a stale object gets served,
+/// which we do not want.
+///
+/// `.include` names assembly source, so its contents have to be scanned in turn.
+/// That is the caller's job, since it is the one reading the files.
+fn find_asm_dependencies(bytes: &[u8]) -> Vec<AsmDependency> {
+    let mut dependencies = Vec::new();
+    let mut offset = 0;
+    while offset < bytes.len() {
```
sylvestre: .include/.incbin matching seems to be  too loose (c.rs:870)  no line-start requirement, no ?

src/compiler/c.rs
```diff
@@ -694,6 +793,174 @@ const PRAGMA_GCC_PCH_PREPROCESS: &[u8] = b"pragma GCC pch_preprocess";
...
+
+    let (rest, skipped_whitespace) = skip_asm_whitespace(after_directive);
+    let (quoting, body) = if let Some(body) = rest.strip_prefix(b"\\\"") {
+        (OperandQuoting::Escaped, body)
+    } else if let Some(body) = rest.strip_prefix(b"\"") {
+        (OperandQuoting::Plain, body)
+    } else if skipped_whitespace {
+        // gas requires a quoted file name. Not finding one means the operand was
+        // produced some other way, so we cannot tell what this depends on.
+        return Some((FileOperand::Unparsable, 0));
```
sylvestre, line 873: Since the caller matches .include/.incbin anywhere in the buffer with no line-start check, this bails out to Unparsable on ordinary C like opts.include = 1; or a ".include %s\n" string  literal, making the whole TU non-cacheable 
 the match needs a statement/line-boundary check before reaching here.

src/compiler/c.rs
bleggett: Yep, thanks. Fixed/added some tests.

src/compiler/c.rs
bleggett: Yep, thanks. Tighened up, and added some tests.

sylvestre:
> This branch has conflicts that must be resolved

sorry!

bleggett:
> > This branch has conflicts that must be resolved
> 
> sorry!

Rebased, thanks

bleggett:
Rebased again.

PR #2800, opened 2026-08-08
add a monitor action

sylvestre:


src/cmdline.rs
```diff
@@ -132,6 +147,18 @@ fn get_clap_command() -> clap::Command {
             flag_infer_long("show-adv-stats")
                 .help("show advanced cache statistics")
                 .action(ArgAction::SetTrue),
+            flag_infer_long("monitor")
+                .help("watch cache statistics live in a terminal dashboard")
+                .action(ArgAction::SetTrue),
+            flag_infer_long("monitor-interval")
+                .help("polling interval of `--monitor`, in seconds (0.2 to 60)")
```
AJIOB:
```suggestion
                .help(format!("polling interval of `--monitor`, in seconds ({} to {})", MIN_MONITOR_INTERVAL, MAX_MONITOR_INTERVAL))
```

nit: single responsibility

docs/Monitoring.md
```diff
@@ -0,0 +1,135 @@
...
+```
+sccache --monitor                          # poll once a second
+sccache --monitor --monitor-interval 0.25  # poll four times a second
+```
+
+`--monitor-interval` accepts 0.2 to 60 seconds; anything outside that range is
+rejected rather than silently clamped.
+
+The monitor connects to the same address as every other sccache client, so
+`SCCACHE_SERVER_PORT` and `SCCACHE_SERVER_UDS` are honoured:
```
glandium, line 30: honored (honoured` is British English).

docs/Monitoring.md
```diff
@@ -0,0 +1,135 @@
...
+
+```
+env SCCACHE_SERVER_UDS=$HOME/sccache.sock sccache --monitor
+```
+
+If no server is running, the monitor does *not* start one: it shows
+`disconnected` and attaches as soon as a server appears. It also survives the
+server being stopped and restarted underneath it.
```
glandium, line 38: I feel like `sccache --monitor` starting a server wouldn't be a bad thing, but this doesn't have to block merging this.

docs/Monitoring.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Panes
```
glandium, line 40: The UI should be self-explanatory, and we shouldn't need to document how it works in this file. Moreover, this is a maintenance nightmare.

docs/Monitoring.md
```diff
@@ -0,0 +1,135 @@
...
+
+The file is followed the way `tail -f` does: the monitor reads the last 64 KiB
+at startup and appends what arrives after that, keeping the most recent 10,000
+lines to scroll back through. It does not need the file to exist yet — it will
+pick it up when it appears — and if the file is truncated or replaced it starts
+over rather than going quiet. Lines are coloured by level, and a line with no
+level of its own, such as the middle of a panic backtrace, keeps the colour of
+the line above it. Any colour codes already in the file are stripped: logging is
+set up before the daemon redirects its stderr, so a server started from a
+terminal writes the escapes env_logger chose for a tty into the log.
```
glandium, line 98: Implementation details are not really that interesting to put in this doc.

docs/Monitoring.md
```diff
@@ -0,0 +1,135 @@
...
+
+The server has no logging RPC: it logs by having its stderr redirected to the
+file named by `SCCACHE_ERROR_LOG`, with `SCCACHE_LOG` setting the verbosity (see
+[Debugging](../README.md#debugging)). So start the server with a log and point
+the monitor at it:
+
+```
+SCCACHE_ERROR_LOG=/tmp/sccache.log SCCACHE_LOG=debug sccache --start-server
+sccache --monitor --monitor-log /tmp/sccache.log
+```
```
glandium, line 84: While --monitor-log defaulting SCCACHE_ERROR_LOG makes it a little more convenient, the fact that server needs to be started with this set makes the log monitoring not all that interesting. You might as well... tail -f the file yourself.

docs/Monitoring.md
```diff
@@ -0,0 +1,135 @@
...
+level of its own, such as the middle of a panic backtrace, keeps the colour of
+the line above it. Any colour codes already in the file are stripped: logging is
+set up before the daemon redirects its stderr, so a server started from a
+terminal writes the escapes env_logger chose for a tty into the log.
+
+Note that the monitor silences *its own* logging while the dashboard is up.
+Otherwise a `SCCACHE_LOG` exported for the whole shell would have this process
+writing log lines onto the terminal it is drawing the dashboard on.
+
+## Sharing a terminal
```
glandium, line 104: Starting from here, there's a lot of implementation details that are not that interesting to be documented here. These kind of things are better in comments close to the relevant code.

README.md
```diff
@@ -252,6 +258,15 @@ The list of features can be found in the `Cargo.toml` file, `[features]` section
 
 By default, `sccache` builds with support for all storage backends, but individual backends may be disabled by resetting the list of features and enabling all the other backends. Refer the [Cargo Documentation](http://doc.crates.io/manifest.html#the-features-section) for details on how to select features with Cargo.
 
+The `monitor` feature is *not* part of the default build, and neither the
+default nor the `all` feature set pulls it in — enable it explicitly if you want
```
glandium, line 262: why not put it in `all`? I mean, if `all` doesn't enable all features, then what does?
(and yes, that does mean default would need to not be all)

src/cmdline.rs
```diff
@@ -151,7 +151,7 @@ fn get_clap_command() -> clap::Command {
                 .help("watch cache statistics live in a terminal dashboard")
                 .action(ArgAction::SetTrue),
             flag_infer_long("monitor-interval")
-                .help("polling interval of `--monitor`, in seconds (0.2 to 60)")
+                .help(format!("polling interval of `--monitor`, in seconds ({} to {})", MIN_MONITOR_INTERVAL, MAX_MONITOR_INTERVAL))
```
glandium, line 154: rustfmt is not happy with this

.github/workflows/ci.yml
glandium, line 19: clippy is not running on the monitor code (btw, it's also not running on the dist-server code)

src/monitor.rs
```diff
@@ -0,0 +1,1942 @@
...
+use crate::protocol::{Request, Response};
+use crate::server::{DistInfo, PerLanguageCount, ServerInfo, ServerStats};
+use crate::util::fmt_duration_as_secs;
+use tail::LogEvent;
+
+/// How many samples of history to keep for the sparklines.
+const HISTORY: usize = 512;
+/// Bounds on the poll interval, adjustable with `+`/`-`.
+const MIN_INTERVAL: Duration = Duration::from_millis(200);
+const MAX_INTERVAL: Duration = Duration::from_secs(60);
```
glandium, line 59: This duplicates MIN/MAX_MONITOR_INTERVAL in cmdline.rs.

src/monitor.rs
```diff
@@ -0,0 +1,1942 @@
...
+        }
+    }
+
+    fn reset_history(&mut self) {
+        self.prev = None;
+        self.rates = Rates::default();
+        self.hist_requests.clear();
+        self.hist_hits.clear();
+        self.hist_misses.clear();
+    }
```
glandium, line 606: how about hist_size, hist_growth, trims, trimmed?
History should probably be in a separate struct with its own `impl Default` (`#[derive]` would probably be enough), making resetting history easier to maintain.

src/cmdline.rs
```diff
@@ -132,6 +147,18 @@ fn get_clap_command() -> clap::Command {
...
+                .action(ArgAction::SetTrue),
+            flag_infer_long("monitor-interval")
+                .help(format!("polling interval of `--monitor`, in seconds ({} to {})", MIN_MONITOR_INTERVAL, MAX_MONITOR_INTERVAL))
+                .value_name("SECS")
+                .value_parser(clap::value_parser!(f64))
+                .default_value("1"),
+            flag_infer_long("monitor-log")
+                .help("log file for `--monitor` to follow [default: $SCCACHE_ERROR_LOG]")
+                .value_name("FILE")
+                .value_parser(clap::value_parser!(PathBuf)),
```
glandium, line 161: both these flags should require --monitor

scripts/try-monitor.sh
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env bash
+#
+# Try out `sccache --monitor` against a throwaway server.
```
glandium, line 3: This is not a test, and apparently nothing is running the script. Is that intended?

src/monitor.rs
```diff
@@ -0,0 +1,1942 @@
...
+                self.updates += 1;
+                self.last_update = Some(now);
+                self.error = None;
+            }
+            Sample::Dist(info) => self.dist = Some(*info),
+            Sample::Zeroed => {
+                self.reset_history();
+                // Drop the pre-zero snapshot too, otherwise the counters keep
+                // showing the old totals until the next poll lands. The poller
+                // re-polls immediately after zeroing, so this is one frame.
```
glandium, line 588: This doesn't happen while paused. Whether that's a problem or not is another question.

glandium (changes requested):
Note: I haven't dug deep in the monitor implementation itself yet, but there's already enough substance in this review for a first round.

PR #2802, opened 2026-08-10
feat: add Microsoft Entra ID (passwordless) auth for the Azure Blob backend

### LINKED ISSUE #1118: Azure Active Directory authentication support
The Azure storage support currently only supports authenticating with shared keys. Sccache should also support Azure Active Directory authentication, in order to support storage accounts that disable shared key authentication.

babrekel:
Closes #1118.

## Summary

sccache's Azure Blob backend currently supports only connection-string
(shared-key) authentication, which fails against storage accounts that have
shared-key access disabled — an increasingly common security baseline (and one
some Azure Policy configurations mandate).

This adds a **Microsoft Entra ID (passwordless)** path: when no connection
string is configured, the Azblob operator is built without an account key or
SAS token, so OpenDAL (via its `reqsign` dependency) resolves an Entra
credential from the ambient environment and signs each request with an OAuth
bearer token.

## Behavior

- **Shared key (unchanged):** `SCCACHE_AZURE_CONNECTION_STRING` keeps its
  existing behavior, byte-for-byte.
- **Entra ID (new):** with no connection string, select the account via either
  - `SCCACHE_AZURE_STORAGE_ACCOUNT` — endpoint synthesized as
    `https://{account}.blob.core.windows.net`, or
  - `SCCACHE_AZURE_ENDPOINT` — a full endpoint for sovereign clouds / custom
    DNS, validated as an `https` URI (plain `http` accepted only for a loopback
    host).

  Credentials then resolve from the environment: a service principal
  (`AZURE_TENANT_ID` / `AZURE_CLIENT_ID` / `AZURE_CLIENT_SECRET`), workload
  identity (`AZURE_FEDERATED_TOKEN_FILE`), or a managed identity via IMDS.
- The two modes are **mutually exclusive**, enforced on both the env and
  file-config surfaces; blank values from a file config are treated as absent.
- A container configured with no auth source: an ambient env var only warns and
  disables the backend (so a stray variable can't take down an unrelated cache),
  whereas an explicit `[cache.azure]` block errors clearly at startup.

No new dependencies — `opendal/services-azblob` and `reqsign` are already pulled
in by the existing `azure` feature.

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`,
  `cargo check --no-default-features` (with and without `--features azure`), and
  `cargo test --features azure` all pass.
- Unit tests cover endpoint synthesis / precedence / trailing-slash handling,
  `https` + host-injection validation, empty-string normalization, auth-mode
  selection, mutual exclusivity (both operands), the no-auth error, and config
  parsing across the env and TOML surfaces.
- The passwordless *request* path (IMDS / token exchange) cannot run in CI
  without a live account; a key-less `Azblob` build is confirmed to engage
  reqsign's OAuth loader chain rather than going anonymous. In practice this
  path has been running as the compilation-cache backend for a CI fleet,
  authenticating to Azure Blob Storage via workload identity
  (`AZURE_FEDERATED_TOKEN_FILE`).

## Docs

`docs/Azure.md` and `docs/Configuration.md` are updated with the new variables,
precedence, mutual exclusivity, the `https`/loopback rule, and an Azurite note.

### DIFF
--- a/docs/Azure.md
+++ b/docs/Azure.md
@@ -1,11 +1,79 @@
 # Azure
 
-To use Azure Blob Storage, you'll need your Azure connection string and an _existing_ Blob Storage container name.  Set the `SCCACHE_AZURE_CONNECTION_STRING`
-environment variable to your connection string, and `SCCACHE_AZURE_BLOB_CONTAINER` to the name of the container to use.  Note that sccache will not create
-the container for you - you'll need to do that yourself.
+sccache can store its cache in an Azure Blob Storage container using either a
+shared-key connection string or Microsoft Entra ID (passwordless) authentication.
+In both cases set `SCCACHE_AZURE_BLOB_CONTAINER` to the name of an _existing_
+container — sccache will not create it for you.
 
-You can also define a prefix that will be prepended to the keys of all cache objects created and read within the container, effectively creating a scope. To do that use the `SCCACHE_AZURE_KEY_PREFIX` environment variable. This can be useful when sharing a bucket with another application.
+## Shared key (connection string)
 
-The `SCCACHE_AZURE_RW_MODE` environment variable can be set to `READ_ONLY` to make sccache use this backend in read-only mode. The default is `READ_WRITE`.
+Set the `SCCACHE_AZURE_CONNECTION_STRING` environment variable to your storage
+account connection string.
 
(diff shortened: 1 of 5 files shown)


sylvestre:
thanks!

PR #2806, opened 2026-08-12
fix(dist): close piped stdin before waiting on the child

### LINKED ISSUE #2805: sccache-dist: Docker builder hangs indefinitely — `docker cp -` is never sent EOF
A build server configured with `[builder] type = "docker"` never completes a job. The first `docker cp - <container>:/` invocation blocks forever, so the build server holds the job until the client times out and falls back to local compilation.

`CommandExt::check_piped` (`src/bin/sccache-dist/build.rs`) moves the child's stdin out of the `Child` and keeps it alive across `wait_with_output()`. The write end of the pipe therefore stays open while the parent waits for the child to exit, and `docker cp -` does not exit until it sees EOF on stdin.

## Environment

- sccache v0.17.0 (`sccache-dist` built with `--features dist-server`)
- Code path unchanged at `67683cc` (current `main` at time of writing)
- Build servers running Docker 27.5.1 (where the hang was observed)
- The `docker cp` behaviour below re-confirmed on Docker 29.7.2, the current release — it is not specific to an old client
- Affects `BuilderType::Docker` only; the `overlay` and `pot` builders do not use `check_piped`

## Steps to reproduce

1. Run a scheduler and a build server, with the server configured as:

```toml
[builder]
type = "docker"
```

2. From a client with `[dist] scheduler_url` pointing at the scheduler, compile any single C file:

```sh
sccache gcc -c hello.c -o hello.o
```

## Expected

The job is built in the container and the object returned to the client.

## Actual

The build server makes no progress. The target container remains in `Created`, a `docker cp` process remains alive indefinitely, and no job ever completes. The client eventually reports a distributed failure and compiles locally:

```text
Could not perform distributed compile, falling back to local
```

## Analysis

```rust
// src/bin/sccache-dist/build.rs:48
fn check_piped(&mut self, pipe: &mut dyn FnMut(&mut ChildStdin) -> Result<()>) -> Result<()> {
    let mut process = self
        .stdin(Stdio::piped())
        .spawn()
        .context("Failed to start command")?;
    let mut stdin = process
        .stdin
        .take()
        .expect("Requested piped stdin but not present");
    pipe(&mut stdin).context("Failed to pipe input to process")?;
    let output = process
        .wait_with_output()
        .context("Failed to wait for process to return")?;
    check_output(&output)
}
```

`stdin` is an owned `ChildStdin` produced by `.take()`. It is dropped at the end of the function's scope, which is *after* `wait_with_output()` returns.

`Child::wait_with_output` closes only the stdin still held by the `Child`:

```rust
pub fn wait_with_output(mut self) -> io::Result<Output> {
    drop(self.stdin.take());
    ...
}
```

Because `check_piped` already took it, `self.stdin` is `None` and that `drop` is a no-op. The pipe's write end remains open for the duration of the wait.

The closure cannot close it either: its parameter is `&mut ChildStdin`, so the callee has no way to drop the value. Only `check_piped` can.

### `docker cp -` requires EOF, not just a complete archive

This can be confirmed without sccache. A complete, valid tar is written, then the write end is held open:

```sh
docker run --rm alpine sh -c 'mkdir -p /s && echo hi > /s/f && tar -cf - -C /s f' > /tmp/t.tar
docker run --rm -i alpine tar -tf - < /tmp/t.tar     # archive is valid and complete

docker create --name cptest alpine true

# control — writer closes immediately
time ( cat /tmp/t.tar ) | docker cp - cptest:/          # returns in ~0s

# writer stays open for 20s after the complete archive
time ( cat /tmp/t.tar; sleep 20 ) | docker cp - cptest:/  # returns after 20s
```

Observed on Docker 29.7.2, timing `docker cp` itself:

```text
control:    docker cp - cptest:/   0.019 total
held open:  docker cp - cptest:/  20.013 total
```

`docker cp -` waits for the writer to close even when it already holds a complete archive. The same result was obtained on an earlier client, so this is long-standing behavior rather than a regression in a particular release.

## Affected call sites

Both are in `DockerBuilder`:

- `build.rs:720` — `make_image`, copying the toolchain tar
  (`Failed to copy toolchain tar into container`)
- `build.rs:771` — `perform_build`, copying the inputs tar
  (`Failed to copy inputs tar into container`)

Both call sites already `drop()` the corresponding *reader* immediately after the call (`drop(toolchain_rdr)` at `:725`, `drop(inputs_rdr)` at `:776`); the writer is the one left open.

## Suggested fix

Drop `stdin` before waiting:

```rust
pipe(&mut stdin).context("Failed to pipe input to process")?;
drop(stdin);
let output = process
    .wait_with_output()
    .context("Failed to wait for process to return")?;
```

An alternative consistent with the existing `// Should really take a FnOnce/FnBox` note is to pass `ChildStdin` by value into the closure so the callee owns and drops it.

I have this change running and can open a PR if it is wanted.

oguzhanmeteozturk:
Fixes #2805

## What

`CommandExt::check_piped` drops the `ChildStdin` it took from the `Child` before
calling `wait_with_output()`, so the child sees EOF on stdin.

```diff
     pipe(&mut stdin).context("Failed to pipe input to process")?;
+    // `process.stdin` was moved out by the `.take()` above, so
+    // `wait_with_output()` has no stdin left to close. Without this the write
+    // end stays open for the rest of the function and a child that reads to EOF
+    // never returns.
+    drop(stdin);
     let output = process
         .wait_with_output()
         .context("Failed to wait for process to return")?;
```

## Why

`check_piped` takes stdin out of the `Child`:

```rust
let mut stdin = process.stdin.take().expect("Requested piped stdin but not present");
```

`stdin` is owned and lives until the end of the function, so it is still open
while `wait_with_output()` blocks. `Child::wait_with_output` closes only the
stdin still held by the `Child` (`drop(self.stdin.take())`), which is `None`
here, making that a no-op.

The closure cannot close it either — its parameter is `&mut ChildStdin`, so the
callee cannot drop the value. Only `check_piped` can.

The two callers are `DockerBuilder::make_image` (`build.rs:720`) and
`DockerBuilder::perform_build` (`build.rs:771`), both of which pipe a tar into
`docker cp - <container>:/`. `docker cp -` reads until EOF, so with the write end
held open it never exits, and every job on a `type = "docker"` build server
blocks indefinitely with its container left in `Created`.

Both call sites already `drop()` the corresponding reader right after the call
(`drop(toolchain_rdr)`, `drop(inputs_rdr)`); this makes the writer symmetric.

## Reproducing the underlying behaviour

That `docker cp -` requires EOF rather than merely a complete archive can be
shown without sccache:

```sh
docker run --rm alpine sh -c 'mkdir -p /s && echo hi > /s/f && tar -cf - -C /s f' > /tmp/t.tar
docker create --name cptest alpine true

time ( cat /tmp/t.tar ) | docker cp - cptest:/            # docker cp:  0.019s
time ( cat /tmp/t.tar; sleep 20 ) | docker cp - cptest:/  # docker cp: 20.013s
```

The archive is complete and valid in both cases; only the writer's behaviour
differs. Measured on Docker 29.7.2, and reproduced on an earlier client, so this
is not specific to a particular release.

## Testing

- Before: a single `sccache gcc -c hello.c` against a `type = "docker"` build
  server made no progress; the container stayed in `Created` and the client fell
  back to local compilation.
- After: the same compile completes and is reported as a distributed compile.
- Larger workload: a Linux 6.12 `defconfig` build distributed 2,867 compilations
  across 11 build servers with no scheduler allocation failures.

## Scope

One statement, in a helper used only by the Docker builder. The `overlay` and
`pot` builders do not call `check_piped`. No public API or configuration change.

### DIFF
--- a/src/bin/sccache-dist/build.rs
+++ b/src/bin/sccache-dist/build.rs
@@ -55,6 +55,11 @@ impl CommandExt for Command {
             .take()
             .expect("Requested piped stdin but not present");
         pipe(&mut stdin).context("Failed to pipe input to process")?;
+        // `process.stdin` was moved out by the `.take()` above, so
+        // `wait_with_output()` has no stdin left to close. Without this the write
+        // end stays open for the rest of the function and a child that reads to
+        // EOF never returns.
+        drop(stdin);
         let output = process
             .wait_with_output()
             .context("Failed to wait for process to return")?;


sylvestre:
it will need a test to make sure it won't regress

please make comment #0 a bit shorter in the future

oguzhanmeteozturk:
Writing the test exposed a gap: all nine test_dist_* tests go through DistSystem::add_server, which hardcodes BuilderType::Overlay. Nothing ever sets type = "docker", so the Docker builder had no coverage. sccache_server_cfg now takes a BuilderType, there’s a new add_docker_server() that mounts the daemon socket, however the harness image (Dockerfile.sccache-dist) will need the Docker client. Happy to push it

PR #2722, opened 2026-05-28
Fix `nvcc` dryrun parsing for CUDA 13.3

mbrobbel:
`sccache` assumed `nvcc`-generated `cicc` and `ptxas` inputs were always at a fixed offset near the end of the command line. With CUDA 13.3.33, `cicc` dryrun output can place `--simt-only` after the input file, which makes the old positional parser treat `--simt-only` as the input instead of `kernel.cpp1.ii`.

That can prevent `sccache` from grouping the `cicc` and `ptxas` device compilation steps correctly. Later, `fatbinary` fails because the expected `.cubin` files were never produced.

## Changes

- Find `cicc` inputs by `.cpp1.ii` extension instead of fixed position.
- Find `ptxas` inputs by `.ptx` extension instead of fixed position.
- Keep the old positional fallback for unexpected command shapes.
- Add regression coverage for `cicc` input before `--simt-only`.
- Add regression coverage for `ptxas` input detection by `.ptx`.
- Add `nvcc` dryrun grouping coverage for CUDA 13.3-style `--simt-only`.

PR #2809, opened 2026-08-14
Don't cache ptxas PTX-verification passes

tinglvv:
Follow up from https://github.com/mozilla/sccache/pull/2722, we discovered issue while building torchao._C_mxfp8 extension: https://github.com/pytorch/pytorch/actions/runs/30797328141/job/91700085985

"nvcc": nvcc_args
+ [
    "-gencode=arch=compute_100,code=sm_100",
    "-gencode=arch=compute_120,code=compute_120",
],

For a virtual-only gencode such as -gencode arch=compute_120,code=compute_120 there is no SASS to emit, so output is not cacheable. nvcc runs ptxas purely to syntax-check the PTX and passes no -o at all:

  ptxas -arch=compute_120 -m64 kernel.compute_120.ptx

group_nvcc_subcommands_by_compilation_stage marks every ptxas sub-command Cacheable::Yes, and the caching path requires an "obj" output that only -o populates, so the whole compile dies in cicc::generate_compile_commands with `Missing "cubin" file output`. Any target carrying a +PTX architecture can hit this.

This fix adds a Cacheable::No clause to route the sub-command to the path nvcc.rs already uses for non-cacheable steps: build the command and run it directly. 

Test Plan:
Carried as a source patch in PyTorch CI, where it fixes the inductor build that compiles torchao against CUDA 13.2:

```
python .ci/pytorch/smoke_test/smoke_test.py --package=torchonly
```

### DIFF
--- a/src/compiler/nvcc.rs
+++ b/src/compiler/nvcc.rs
@@ -792,7 +792,15 @@ where
                     }
                     false
                 });
-                (env_vars.clone(), Cacheable::Yes, group)
+                // For a virtual-only gencode (e.g. -gencode arch=compute_120,
+                // code=compute_120) nvcc runs ptxas with no -o just to verify the
+                // PTX. There is no artifact to cache, so run it directly.
+                let cacheable = if args.iter().any(|arg| arg == "-o") {
+                    Cacheable::Yes
+                } else {
+                    Cacheable::No
+                };
+                (env_vars.clone(), cacheable, group)
             }
             // cudafe++ _must be_ cached, because the `.module_id` file is unique to each invocation (new in CTK 12.8)
             Some("cudafe++") => {


tinglvv:
cc @mbrobbel @sylvestre please review

mike-wendt:
@tinglvv thanks for this! I was just making a PR to fix this gap as well

I've verified this fix works (full `compiler::` suite green, 315/315) and reproduced the bug as you reported in #2722 with local mocks

Thanks again

sylvestre:
Could you please add a test to make sure we don't regress? thanks

mbrobbel (approved):

sylvestre:
@mbrobbel i am confused why you approved this change? it doesn't have tests :/

mbrobbel:
> @mbrobbel i am confused why you approved this change? it doesn't have tests :/

Oh you are right, I'm sorry about that. Can I suggest the following patch for the regression test?

```diff
diff --git a/src/compiler/nvcc.rs b/src/compiler/nvcc.rs
index 8f618efd..d455ce94 100644
--- a/src/compiler/nvcc.rs
+++ b/src/compiler/nvcc.rs
@@ -1592,6 +1592,9 @@ mod test {
 #$ cl -D__CUDA_ARCH__=800 -D__CUDACC__ -E -x c++ "kernel.cu" > "kernel.cpp1.ii"
 #$ cicc --device-c -arch compute_80 --module_id_file_name "kernel.module_id" --gen_c_file_name "kernel.cudafe1.c" --stub_file_name "kernel.cudafe1.stub.c" --gen_device_file_name "kernel.cudafe1.gpu" "kernel.cpp1.ii" --simt-only -o "kernel.ptx"
 #$ ptxas -arch=sm_80 --compile-only "kernel.ptx" -o "kernel.cubin"
+#$ cl -D__CUDA_ARCH__=1200 -D__CUDACC__ -E -x c++ "kernel.cu" > "kernel.compute_120.cpp1.ii"
+#$ cicc --device-c -arch compute_120 "kernel.compute_120.cpp1.ii" -o "kernel.compute_120.ptx"
+#$ ptxas -arch=compute_120 -m64 "kernel.compute_120.ptx"
 #$ fatbinary --create="kernel.fatbin" "--image3=kind=elf,sm=80,file=kernel.cubin" --embedded-fatbin="kernel.fatbin.c" --device-c
 #$ cl -D__CUDA_ARCH__=800 -D__CUDACC__ -c -x c++ "kernel.cudafe1.cpp" -Fo"kernel.o"
 "#
@@ -1604,6 +1607,9 @@ mod test {
 #$ gcc -D__CUDA_ARCH__=800 -D__CUDACC__ -E -x c++ "kernel.cu" -o "kernel.cpp1.ii"
 #$ cicc --device-c -arch compute_80 --module_id_file_name "kernel.module_id" --gen_c_file_name "kernel.cudafe1.c" --stub_file_name "kernel.cudafe1.stub.c" --gen_device_file_name "kernel.cudafe1.gpu" "kernel.cpp1.ii" --simt-only -o "kernel.ptx"
 #$ ptxas -arch=sm_80 --compile-only "kernel.ptx" -o "kernel.cubin"
+#$ gcc -D__CUDA_ARCH__=1200 -D__CUDACC__ -E -x c++ "kernel.cu" -o "kernel.compute_120.cpp1.ii"
+#$ cicc --device-c -arch compute_120 "kernel.compute_120.cpp1.ii" -o "kernel.compute_120.ptx"
+#$ ptxas -arch=compute_120 -m64 "kernel.compute_120.ptx"
 #$ fatbinary --create="kernel.fatbin" "--image3=kind=elf,sm=80,file=kernel.cubin" --embedded-fatbin="kernel.fatbin.c" --device-c
 #$ gcc -D__CUDA_ARCH__=800 -D__CUDACC__ -c -x c++ "kernel.cudafe1.cpp" -o "kernel.o"
 "#
@@ -1647,7 +1653,7 @@ mod test {
             .find(|group| {
                 group
                     .iter()
-                    .any(|cmd| cmd.exe.file_stem().and_then(|stem| stem.to_str()) == Some("cicc"))
+                    .any(|cmd| cmd.args.iter().any(|arg| arg == "kernel.cubin"))
             })
             .expect("missing device compile group");

@@ -1671,6 +1677,17 @@ mod test {
             .find(|cmd| cmd.exe.file_stem().and_then(|stem| stem.to_str()) == Some("ptxas"))
             .unwrap();
         assert!(ptxas.args.iter().any(cicc::is_ptxas_input));
+        assert_eq!(Cacheable::Yes, ptxas.cacheable);
+
+        let virtual_ptxas = groups
+            .iter()
+            .flatten()
+            .find(|cmd| {
+                cmd.exe.file_stem().and_then(|stem| stem.to_str()) == Some("ptxas")
+                    && cmd.args.iter().any(|arg| arg == "kernel.compute_120.ptx")
+            })
+            .expect("missing virtual-only ptxas command");
+        assert_eq!(Cacheable::No, virtual_ptxas.cacheable);
     }

     #[test]
```

src/compiler/nvcc.rs
```diff
@@ -792,7 +792,15 @@ where
                     }
                     false
                 });
-                (env_vars.clone(), Cacheable::Yes, group)
+                // For a virtual-only gencode (e.g. -gencode arch=compute_120,
+                // code=compute_120) nvcc runs ptxas with no -o just to verify the
+                // PTX. There is no artifact to cache, so run it directly.
+                let cacheable = if args.iter().any(|arg| arg == "-o") {
```
sylvestre, line 798: there's already test_group_nvcc_subcommands_with_simt_only_cicc_input just below, please add one with a virtual-only gencode ptxas line and assert Cacheable::No

sylvestre, line 798: only bare `-o`? ptxas also takes `--output-file` and `-o=file`, would those slip through as Yes?

src/compiler/nvcc.rs
```diff
@@ -792,7 +792,15 @@ where
                     }
                     false
                 });
-                (env_vars.clone(), Cacheable::Yes, group)
+                // For a virtual-only gencode (e.g. -gencode arch=compute_120,
```
sylvestre: 3 lines of comment for a 5 lines change, could you shorten it please :)

PR #2810, opened 2026-08-17
Add SCCACHE_ALLOW_INCREMENTAL to keep caching the non-incremental parts of a build

PHPCraftdream:
Adds an opt-in `SCCACHE_ALLOW_INCREMENTAL=1`. Default behaviour is unchanged.

sccache refuses to run when `CARGO_INCREMENTAL` is set (#1767). That refusal is all-or-nothing, but cargo passes `-C incremental=` only to workspace members and path dependencies, so the registry dependencies of the same build are still cacheable. With the flag set, the incremental invocations take the `CannotCache` path `compiler/rust.rs` has had since 2018, and the rest of the build is cached as before.

`CARGO_INCREMENTAL`/`CARGO_BUILD_INCREMENTAL` are no longer hashed into the Rust cache key: they reach rustc only as `-C incremental=`, already part of the hashed argument list. `CACHE_VERSION` is not bumped, since this only merges keys whose entries are identical. Entries cached with `CARGO_INCREMENTAL` explicitly set will miss once.

Tests: the default refusal still fires; with the flag, the registry dep hits an entry written by a non-incremental build while the workspace crate bypasses the cache and gets a real incremental session dir.

Docs: README, `docs/Rust.md`, `docs/Configuration.md`.

The second commit is an unrelated `clippy::for_kv_map` fix I needed locally to get a clean `-D warnings` run; happy to drop it.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -891,19 +891,28 @@ pub fn run_command(cmd: Command) -> Result<i32> {
         } => {
             trace!("Command::Compile {{ {:?}, {:?}, {:?} }}", exe, cmdline, cwd);
 
-            let incr_env_strs = ["CARGO_BUILD_INCREMENTAL", "CARGO_INCREMENTAL"];
-            incr_env_strs
-                .iter()
-                .for_each(|incr_str| match env::var(incr_str) {
-                    Ok(incr_val) if incr_val == "1" => {
-                        println!(
-                            "sccache: incremental compilation is prohibited: Unset {} to continue.",
-                            incr_str
-                        );
-                        std::process::exit(1);
-                    }
-                    _ => (),
-                });
+            // Incremental crates cannot be cached, hence the refusal below.
+            // Lifting it leaves only the crates cargo hands `-C incremental=`
+            // uncached (compiler/rust.rs already treats those as CannotCache).
+            // Compared against "1" so a global opt-in can be overridden per build.
+            let allow_incremental = env::var("SCCACHE_ALLOW_INCREMENTAL").as_deref() == Ok("1");
+            if !allow_incremental {
+                let incr_env_strs = ["CARGO_BUILD_INCREMENTAL", "CARGO_INCREMENTAL"];
+                incr_env_strs
+                    .iter()
+                    .for_each(|incr_str| match env::var(incr_str) {
+                        Ok(incr_val) if incr_val == "1" => {
+                            println!(
+                                "sccache: incremental compilation is prohibited: Unset {} to \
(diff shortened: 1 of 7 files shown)


src/compiler/rust.rs
```diff
@@ -1579,10 +1579,21 @@ where
...
+            // CARGO_INCREMENTAL/CARGO_BUILD_INCREMENTAL only affect rustc output through
+            // the `-C incremental=` argument cargo derives from them, which is already
+            // hashed as part of the argument list (and marks the invocation non-cacheable
+            // anyway). Hashing the env var too would needlessly fork the cache between
+            // incremental-enabled and incremental-disabled builds of identical deps.
+            // No CACHE_VERSION bump: this only merges previously-distinct keys whose
+            // entries are output-identical (the env var itself does not change rustc
+            // output, and `env!("CARGO_INCREMENTAL")` readers are still tracked through
+            // dep-info env-deps), so existing entries stay valid under the merged key.
             if var == "CARGO_MAKEFLAGS"
```
sylvestre: do we need such a long comment?

src/commands.rs
```diff
@@ -891,19 +891,50 @@ pub fn run_command(cmd: Command) -> Result<i32> {
...
+            // and still reported as such by `--show-stats`. So the
+            // incremental crates compile incrementally and uncached, while
+            // everything else is cached exactly as before.
+            //
+            // Compared against "1" rather than merely tested for presence (as
+            // SCCACHE_IGNORE_SERVER_IO_ERROR above already does), so that an
+            // opt-in configured globally -- a user environment variable, or
+            // cargo's `[env]` table, which is injected into every wrapper
+            // process cargo spawns -- can still be turned back off for a single
+            // build with SCCACHE_ALLOW_INCREMENTAL=0.
```
sylvestre: can you please update your agent to not write long comments? thanks

src/compiler/rust.rs
PHPCraftdream: Trimmed to two lines, matching the one-line-per-variable style of the list above it.

src/commands.rs
PHPCraftdream: Point taken. Cut to four lines here and trimmed the rest of the PR the same way: 70 lines of added comment removed, 25 kept, code unchanged. The reasoning now lives in the PR description instead.

sylvestre:
similar feedback about comment 0
this is way too long
i am not interested to read a full llm output :(

PHPCraftdream:
Fair. Cut it down to the essentials.

Issue #2470, opened 2025-10-24
CUDA + MSVC + Ninja Multi-Config error: Could not parse shell line


nikolaseu:
Using the latest sccache v0.12.0 installed with scoop I get an error using Ninja Multi-Config generator in CMake v4.1.2. It also happened with sccache 0.11.

Here's a small project demonstrating the issue (based on [vectorAdd official CUDA sample](https://github.com/NVIDIA/cuda-samples/tree/master/Samples/0_Introduction/vectorAdd)):

[example project.zip](https://github.com/user-attachments/files/23121489/example.project.zip)

Configure log when `msvc-x64-multi-config` preset is used:

```
proc] Executing command: C:\Users\nikol\scoop\shims\cmake.EXE -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe "-DCMAKE_CONFIGURATION_TYPES=Debug;RelWithDebInfo;Release" -DCMAKE_CROSS_CONFIGS=all -DCMAKE_DEFAULT_BUILD_TYPE=Release -DCMAKE_DEFAULT_CONFIGS=Release -S C:/Users/nikol/Projects/tmp/sccache_cuda_multi_config -B C:/Users/nikol/Projects/tmp/sccache_cuda_multi_config/build/msvc-x64-multi-config -G "Ninja Multi-Config"
[cmake] -- Found sccache: C:/Users/nikol/scoop/shims/sccache.exe, using SCCACHE_DIRECT=true, SCCACHE_DIR=
[cmake] -- Using:
[cmake]     CMAKE_C_COMPILER_LAUNCHER=C:/Users/nikol/scoop/shims/sccache.exe
[cmake]     CMAKE_CXX_COMPILER_LAUNCHER=C:/Users/nikol/scoop/shims/sccache.exe
[cmake]     CMAKE_CUDA_COMPILER_LAUNCHER=C:/Users/nikol/scoop/shims/sccache.exe
[cmake] 
[cmake] -- The C compiler identification is MSVC 19.44.35217.0
[cmake] -- The CXX compiler identification is MSVC 19.44.35217.0
[cmake] -- The CUDA compiler identification is NVIDIA 13.0.88 with host compiler MSVC 19.44.35217.0
[cmake] -- Detecting C compiler ABI info
[cmake] -- Detecting C compiler ABI info - done
[...]

Issue #2808, opened 2026-08-14
nvcc: server deterministically stops responding (deadlock/crash) while compiling ONNX Runtime's generated MoE GEMM launchers (Windows, CUDA 13.3, MSVC host)


Kataglyphis:
Building ONNX Runtime v1.28.0's CUDA provider on Windows with sccache as `CMAKE_CUDA_COMPILER_LAUNCHER`, the sccache **server deterministically stops responding** partway through the compile — every concurrent client then fails with:

```
sccache: error: failed to execute compile
sccache: caused by: error reading compile response from server
sccache: caused by: Failed to read response header
sccache: caused by: An existing connection was forcibly closed by the remote host. (os error 10054)
```

Two independent full builds died at **4909.2 s and 4911.5 s** (±2 s!) on the same TU family: `onnxruntime/contrib_ops/cuda/llm/moe_gemm/launchers/fused_moe_gemm_sm80_{f16,bf16}.generated.cu` (cutlass-generated launchers in the `onnxruntime_providers_cuda_llm` OBJECT library). The determinism across runs rules out a random race trigger.

We also observed the same failure *shape* earlier in the build (different TU region, no determinism verified) as a **pure hang**: sccache server + 9 clients all at 0 % CPU, zero backend connections while the WebDAV endpoint answered HTTP 200 externally — a full-fleet flatline for 40+ minutes until the processes were killed manually. In the two deterministic cases above, a watchdog killed the server ~3 minutes after CPU flatlined, which is what surfaced the 10054s — so the underlying event may be a **server-side deadlock in the nvcc decomposition under concurrency** rather than a crash; we cannot distinguish from client-side evidence alone.

## Environment

| Component | Version |
|---|---|
| sccache | built from source at `e9b15a35f7240a7edd1b9644583edb388c6cb5f9` (includes the #2722 CUDA 13.3 dryrun fix; no nvcc-relevant commits on `main` since) |
| Host compiler | MSVC 14.51 (VS 18 BuildTools), `nvcc -ccbin=cl.exe` |
| CUDA | 13.3.33 |
| OS | Windows Server Core ltsc2025 container (process isolation), Win 11 25H2 host |
| Backend | WebDAV (dufs) with `SCCACHE_MULTILEVEL_CHAIN=disk,webdav`, `SCCACHE_DIR=C:\sccache`, `SCCACHE_CACHE_SIZE=15G` |
| Parallelism | ninja `-j9`, ~1900 CUDA TUs total |

[...]

PR #2811, opened 2026-08-18
nvcc (Windows): protect escaped quotes in dryrun lines before flattening backslashes

Kataglyphis:
## Problem

On Windows, `nvcc --dryrun` prints string-valued defines with escaped quotes:

```
-D "FILE_NAME=\"onnxruntime_providers_cuda.dll\""
```

`fold_env_vars_or_split_into_exe_and_args` flattens `\` to `/` *before*
tokenizing the line (to normalize paths), which turns every `\"` into `/"`
and collapses the line's quote structure at the first such define.
`shlex::split` then packs everything after it into one giant token — on a
real ONNX Runtime v1.28.0 compile we measured a **493-character token
containing ~30 `-D` pairs** (`USE_CUDA=1`, `USE_FLASH_ATTENTION=1`,
`USE_MEMORY_EFFICIENT_ATTENTION=1`, `USE_FP8_KV_CACHE=1`, ...).

The host preprocess that produces `cpp4.ii` therefore runs **without** those
defines, `cudafe++` generates no stubs for the `#ifdef`-guarded explicit
instantiations, and the loss propagates silently through ptx/cubin into the
final object. It surfaces hours later as `lld-link: undefined symbol`
(`BiasSoftmaxImpl<double>`, `QkvToContext<*, __nv_fp8_e4m3>`,
`run_memory_efficient_attention`) — a silent miscompile, not a build error.
onnxruntime-genai's CUDA TUs fail hard (nvcc exit -2) through the same path.

Likely the mechanism behind long-standing Windows reports such as #1077, and
the remaining actionable half of #2808 (whose deadlock we could no longer
reproduce once our storage backend was healthy — see the issue addendum).

## Reproduction (single command)

ONNX Runtime v1.28.0, `onnxruntime/contrib_ops/cuda/math/bias_softmax_impl.cu`,
compiled with the exact command from ORT's build.ninja (extract via
`ninja -t commands`), CUDA 13.3.1, four `-gencode` archs, cl.exe host:

* bare nvcc: 3,189 defined symbols
* `sccache nvcc` (cold cache): 2,628 — every `double` instantiation's
  `__device_stub__` missing, float/half intact
* intermediates: `cudafe1.stub.c` ddd-marker counts 88 (bare) vs 0 (wrapped);
  the wrapped `cpp4.ii` lacks exactly the `#ifdef USE_CUDA` instantiation

## Fix

Protect the `\"` escapes with a sentinel (`\u{1}`) around the backslash
flatten and restore them afterwards, so `shlex` sees the original quoting.
One hunk in `fold_env_vars_or_split_into_exe_and_args`, plus a regression
test (`test_group_nvcc_subcommands_preserves_escaped_quotes_in_defines`)
that feeds an escaped-quote dryrun line through the grouping and asserts
every define survives as its own token (the giant-token fingerprint —
`" -D "` inside a single argument — is asserted absent).

With the patch, the reproducer's object matches bare nvcc symbol-for-symbol
(3,189 == 3,189; only 1:1-substituted `??_C@` string literals differ, the
expected module-id naming divergence).

---
\U0001F916 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/compiler/nvcc.rs
+++ b/src/compiler/nvcc.rs
@@ -1096,14 +1096,6 @@ fn fold_env_vars_or_split_into_exe_and_args(
     // The rest of the lines are subcommands, so parse into a vec of [cmd, args..]
 
     let mut line = if cfg!(target_os = "windows") {
-        // Protect escaped quotes (\") before flattening path separators:
-        // nvcc --dryrun prints string-valued defines as -D "FILE_NAME=\"x.dll\"",
-        // and replacing '\' with '/' first turns \" into /", collapsing the
-        // line's quote structure. shlex then mis-groups everything after the
-        // first such define into one giant token, so the host preprocess runs
-        // WITHOUT those -D's and #ifdef-guarded instantiations silently
-        // disappear from the final object (dropped-instantiation miscompile;
-        // observed as lld-link undefined symbols on ONNX Runtime).
         let line = line
             .replace("\\\"", "\u{1}")
             .replace("\"\"", "\"")
@@ -1575,13 +1567,6 @@ mod test {
     #[test]
     #[cfg(windows)]
     fn test_group_nvcc_subcommands_preserves_escaped_quotes_in_defines() {
-        // nvcc --dryrun on Windows prints string-valued defines with escaped
-        // quotes (-D "FILE_NAME=\"x.dll\""). The line mangling used to flatten
-        // '\' to '/' before tokenization, collapsing the quote structure:
-        // shlex then packed every argument after the first such define into
-        // one giant token, the preprocess ran without those -D's, and
-        // #ifdef-guarded instantiations silently vanished from the final
-        // object (observed as lld-link undefined symbols on ONNX Runtime).
         let bin_dir = tempfile::tempdir().unwrap();
 
         for exe in ["nvcc", "cl", "cudafe++", "cicc", "ptxas", "fatbinary"] {
--- a/src/compiler/nvcc.rs
+++ b/src/compiler/nvcc.rs
@@ -1096,11 +1096,21 @@ fn fold_env_vars_or_split_into_exe_and_args(
     // The rest of the lines are subcommands, so parse into a vec of [cmd, args..]
 
     let mut line = if cfg!(target_os = "windows") {
+        // Protect escaped quotes (\") before flattening path separators:
+        // nvcc --dryrun prints string-valued defines as -D "FILE_NAME=\"x.dll\"",
+        // and replacing '\' with '/' first turns \" into /", collapsing the
+        // line's quote structure. shlex then mis-groups everything after the
+        // first such define into one giant token, so the host preprocess runs
+        // WITHOUT those -D's and #ifdef-guarded instantiations silently
+        // disappear from the final object (dropped-instantiation miscompile;
+        // observed as lld-link undefined symbols on ONNX Runtime).
         let line = line
+            .replace("\\\"", "\u{1}")
             .replace("\"\"", "\"")
             .replace(r"\\?\", "")
             .replace('\\', "/")
-            .replace(r"//?/", "");
+            .replace(r"//?/", "")
+            .replace('\u{1}', "\\\"");
         match host_compiler {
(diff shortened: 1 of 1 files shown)


Kataglyphis:
Additional data point: the reproduction is ONNX-Runtime-version-independent — the same single-TU replay against **ORT v1.29.0** (tag `2e2543f`) shows the identical loss: bare nvcc 3,189 defined symbols vs wrapped 2,628. The trigger is ORT's long-standing `-DFILE_NAME=\"...dll\"" / `-DVER_STRING=\"...\"` quoted-string defines, which any version emits.

Kataglyphis:
Design notes for reviewers - considered alternatives and one known edge:

**Why sentinel-protect instead of tokenize-first:** the deeper issue is that the Windows normalization does whole-line string replaces *before* tokenization. A tokenize-then-normalize-per-token refactor would remove this class entirely, but it touches all the platform assumptions in `fold_env_vars_or_split_into_exe_and_args`; this PR stays minimal inside the existing design. Happy to attempt the refactor in a follow-up if that direction is preferred - the regression test added here survives it either way.

**Known edge:** an argument whose literal path ends in a backslash immediately before a closing quote (`-I"C:\dir\"`) would now parse as an escaped quote. We have not observed nvcc (CTK 12/13) dryrun output emitting that shape - paths print with forward slashes or no trailing separator, and `\"` appears only in string-valued defines - and the previous code mangled such a line differently-but-also-incorrectly (`\` -> `/` inside the quoted path). Flagging it for completeness.

**Sentinel choice:** `\u{1}` cannot occur in dryrun output; can switch to a longer improbable marker if preferred.

src/compiler/nvcc.rs
```diff
@@ -1096,11 +1096,21 @@ fn fold_env_vars_or_split_into_exe_and_args(
     // The rest of the lines are subcommands, so parse into a vec of [cmd, args..]
 
     let mut line = if cfg!(target_os = "windows") {
+        // Protect escaped quotes (\") before flattening path separators:
```
sylvestre: do we need really such a long comment?

src/compiler/nvcc.rs
```diff
@@ -1562,6 +1572,105 @@ mod test {
...
+    #[test]
+    #[cfg(windows)]
+    fn test_group_nvcc_subcommands_preserves_escaped_quotes_in_defines() {
+        // nvcc --dryrun on Windows prints string-valued defines with escaped
+        // quotes (-D "FILE_NAME=\"x.dll\""). The line mangling used to flatten
+        // '\' to '/' before tokenization, collapsing the quote structure:
+        // shlex then packed every argument after the first such define into
+        // one giant token, the preprocess ran without those -D's, and
+        // #ifdef-guarded instantiations silently vanished from the final
+        // object (observed as lld-link undefined symbols on ONNX Runtime).
```
sylvestre: same comment

src/compiler/nvcc.rs
Kataglyphis: You are right. I will remove those long comments. Only helped me in the first place
to understand whats happening :)

Kataglyphis:
Pushed a follow-up addressing both threads:

- **Comment length:** the explanation now lives as a short doc comment on an extracted pure helper, `normalize_windows_dryrun_line` — the inline blocks are gone (building on the comment-trim commit).
- **Codecov patch coverage:** the 4 uncovered lines were the `cfg!(windows)` normalization branch, which the Linux coverage job can never execute. With the logic extracted into a platform-independent helper, the new `test_normalize_windows_dryrun_line_preserves_escaped_quotes` exercises those lines on every platform; the Windows-only end-to-end grouping test stays as the integration check.

Kataglyphis:
Now the changes are minimal without lenghty comments. 
Test will now only be executed on windows CI which should help CodeCov to be happy.

nikolaseu:
I assume this will also fix #2470 ?

Kataglyphis:
> I assume this will also fix #2470 ?

Yes. Verified locally on my setup.
Thanks for linking. Wasn't aware about it.

Issue #2595, opened 2026-02-06
SCCACHE_BASEDIRS support for Worktree workflow


howardjohn:
SCCACHE_BASEDIRS added in https://github.com/mozilla/sccache/commit/7e02a2f4f067e2fb5a438be6b4dc2d7534e3ddd9 is a great step to help enable cache hits across copies of repos, but doesn't work well for common git worktree workflows, since we need to statically set each base dir. This breaks a few workflows.

For example, consider a project at `/home/projects/hello-world`. Some common workflows I have seen are...
* Worktrees in `/home/projects/hello-world/.worktrees/<branchname>/...`
* `$HOME/.local/share/opencode/worktree/<some UUID>/<some name>/...`
etc

It would be nice to be able to have these all use the shared cache. This is especially important with AI coding use cases (which is a common usage for worktrees), where all new work is done in a new branch. Usage of sccache (if this can be supported) would cut multiple minute build times at the start of each piece of work

howardjohn:
Actually looking a bit more closely it seems like this only works for C anyways not rust so it wouldn't be relevant for my use

HuaHuaY:
I have tried `SCCACHE_BASEDIRS` and it seems to work. For examples, my project is at `/foo` and one of my worktree is at `/foo/.worktree/bar`. `SCCACHE_BASEDIRS` should be `/foo:/foo/.worktree/bar`.

johnkferguson:
Was able to get cross-worktree Rust dedup working in a fork and wanted to share what was needed to actually get it working end-to-end,. Fork is on top of the existing [Vercel `rust-basedirs` work](https://github.com/vercel/sccache).

1. **Glob patterns** in `SCCACHE_BASEDIRS` (so `$HOME/repos/*/worktrees/*/*/*` works without enumerating random worktree names).
2. **Per-token argv strip before concatenation** — cargo emits `--remap-path-prefix /abs=virtual` as two tokens, and the post-concat blob scan can't see the `/` boundary. Stripping each token in isolation catches it.
3. **Strip every `CARGO_*` env value**, not a hand-coded allowlist. `devenv` injects `CARGO_INSTALL_ROOT` per-project; the existing allowlist missed it and that one var broke the hash key for every workspace crate.

[...]

PR #2813, opened 2026-08-18
rust: normalize cache keys with SCCACHE_BASEDIRS

tcashel:
Rust builds of the same source in different worktrees currently get different cache keys. This makes Rust hashing honor `SCCACHE_BASEDIRS`, using the longest matching root for source paths, remap sources, relevant environment values, and the working directory. The Rust cache-key version advances to 7.

This is opt-in and changes cache keys, not embedded artifact paths. A cached crate using `env!("CARGO_MANIFEST_DIR")` can retain the original checkout path. Tests cover cross-root reuse, source-change misses, argument framing, and Windows path forms.

Rebased onto current upstream. Formatting, clippy, and the locked unit/integration suite pass on Apple silicon. CUDA cases skip locally; native Windows and CUDA validation are left to CI.

Request-scoped configuration follows in #2814. Related to #2595. Developed with OpenAI Codex.


tcashel:
@sylvestre This is ready for review. #2814 and #2838 are stacked follow-ups. The new CI runs are also waiting for maintainer approval. Thanks! Please let me know if i'm doing something wrong, in code or in process!

PR #2814, opened 2026-08-18
server: scope SCCACHE_BASEDIRS to compiler requests

### LINKED ISSUE #2595: SCCACHE_BASEDIRS support for Worktree workflow

tcashel:
One sccache daemon currently captures `SCCACHE_BASEDIRS` at startup, requiring a restart when another worktree needs a different root. This lets each compile request override the file-configured fallback. An absent value uses the fallback; an empty value disables it; an invalid relative path fails that request. Concurrent requests remain isolated in daemon and client-side modes.

Depends on #2813; the [request-only diff](https://github.com/tcashel/sccache/compare/codex/rust-basedirs...codex/request-scoped-basedirs) excludes that prerequisite. The storage wrapper preserves the current upstream raw-cache read API. Tests cover overrides, fallback, statistics, invalid paths, and concurrent requests in both modes.

Formatting, clippy, and the locked unit/integration suite pass on Apple silicon. CUDA cases skip locally; native Windows and CUDA validation are left to CI.

Fixes #2595. Developed with OpenAI Codex.


benches/sccache_bench.rs
```diff
@@ -867,6 +867,23 @@ fn strip_basedirs_multiple(bencher: Bencher) {
     bencher.bench(|| black_box(strip_basedirs(black_box(&output), black_box(&basedirs))));
 }
 
+#[divan::bench(args = [0, 1, 8, 32])]
```
sylvestre: please add the benchmark in a different pr

tcashel: Done, thank you for the comment.  PR here for ref: https://github.com/mozilla/sccache/pull/2838

ghoulspa:
Built a528aca on Windows (x86_64-pc-windows-gnu, rustc 1.98) and have been using it with a Bevy workspace across git worktrees. With `SCCACHE_BASEDIRS` set per checkout through cargo's `[env] SCCACHE_BASEDIRS = { value = "", relative = true }` and no `basedirs` in the server config, a fresh worktree gets 365/365 Rust hits; on 0.18.0 the same build missed `serde_core`/`serde` (`env-dep:OUT_DIR`) and every path dependency (`CARGO_MANIFEST_DIR`), about 80 crates. Details and repro in #2870.

src/cache/basedirs.rs
```diff
@@ -0,0 +1,185 @@
...
+use async_trait::async_trait;
+use bytes::Bytes;
+
+use crate::cache::multilevel::MultiLevelStats;
+use crate::cache::{Cache, CacheMode, CacheWrite, GetPathResult, Storage};
+use crate::compiler::PreprocessorCacheEntry;
+use crate::config::PreprocessorCacheModeConfig;
+use crate::errors::*;
+
+struct BasedirsStorage {
```
sylvestre, line 27: 150 lines of forwarding just to carry a Vec<Vec<u8>>, and the day someone adds a method to Storage it will silently bypass this wrapper. could we pass the basedirs next to the storage into generate_hash_key instead?

src/commands.rs
```diff
@@ -719,7 +719,15 @@ where
 pub fn run_command(cmd: Command) -> Result<i32> {
     // Config isn't required for all commands, but if it's broken then we should flag
     // it early and loudly.
-    let config = &Config::load()?;
+    let config = if matches!(
+        &cmd,
+        Command::StartServer | Command::InternalStartServer | Command::ShowStats(_, _)
+    ) {
+        Config::load_with_file_basedirs()?
```
sylvestre, line 726: hardcoding the command list here is fragile. if the daemon only uses the file value and client-side gets it from the handshake, is the env branch in config_from_env still used at all?

src/compiler/rust.rs
```diff
@@ -235,7 +238,90 @@ static ALLOWED_EMIT: LazyLock<HashSet<&'static str>> =
...
+        return;
+    };
+    let source = strip_path_basedirs(&encoded[..separator], basedirs);
+    let replacement = &encoded[separator..];
+    digest.update(&((source.len() + replacement.len()) as u64).to_le_bytes());
+    digest.update(source);
+    digest.update(replacement);
+}
+
+fn encode_rust_value(value: &OsStr) -> &[u8] {
```
sylvestre, line 290: do we need this wrapper? as_encoded_bytes() reads fine on its own.

src/server.rs
```diff
@@ -1169,12 +1179,29 @@ where
...
+    ) -> Result<Arc<dyn Storage>> {
+        let Some((_, value)) = env_vars
+            .iter()
+            .rev()
+            .find(|(name, _)| is_basedirs_env(name))
+        else {
+            return Ok(self.storage.clone());
+        };
+
+        let basedirs = config::parse_basedirs_env(value)?;
```
sylvestre, line 1203: a relative value here makes handle_compile return Err, so the whole compile dies instead of falling back to the real compiler. what does the user actually see? and please add a test for it, it's in the PR description but not covered.

src/util.rs
```diff
@@ -1369,6 +1369,153 @@ fn pathname_positions(arg: &[u8]) -> impl Iterator<Item = usize> + '_ {
...
+    } else if windows_path_starts_with(value, b"//?/unc/") {
+        Some(WindowsVerbatimPrefix::Unc)
+    } else {
+        None
+    }
+}
+
+/// Strips the longest configured base directory from a complete path.
+/// Configured base directories must be normalized and end with a slash.
+#[doc(hidden)]
```
sylvestre, line 1452: why `#[doc(hidden)] pub` and not pub(crate)? it's only used inside the crate.

sylvestre:
this patch is a bit too large to my taste, sorry

PR #2825, opened 2026-08-27
Forward target options during detection

Joldiges:
Compiler identification runs a preprocessing probe without preserving target selection. Clang-compatible cross-compilers that require an explicit target can therefore fail detection before a compilation is considered for caching.

Forward GNU/Clang `--target` and `-target` options into the detection probe, including values supplied through response files. Normalize single-dash `-target` to the separated form accepted by standard Clang and add coverage for all joined and separated forms.

This change was primarily made to fix an assumed regression in v0.17 where informing SCcache that `ARM-Clang` was `Clang-like` was now failing

GitHub copilot was used for the debugging and resolution of this issue.


sylvestre:
please fix the conflict

Joldiges:
> PR fixes C/C++ compiler vendor detection for Clang-compatible cross-compilers by forwarding target-selection flags into the preprocessing probe used during detection (including when those flags come from `@response` files). This prevents detection failures when a compiler requires an explicit target to

Done, thank you

Accidentally done in 2 pushes.  Sorry about that and the extra CI time/cost consumed

PR #2827, opened 2026-08-28
Optimize child process spawns by skipping job server configs

rnk:
Configuring the job server so that all child processes can access it causes Rust's subprocess library to switch from `posix_spawn` to `fork`, which supports running arbitrary code in a pre-exec context. Fork is, unfortunately, very expensive. You really want to use vfork or posix_spawn, which IIRC uses that under the hood, in order to get fast process launching.

This really doesn't matter in the grand scheme of things because sccache direct mode (the defatult) doesn't launch a ton of processes when it gets a cache hit, but if you turn it off, it will launch may pre-processing jobs, and the overhead is observable in a profiler, which is how I (well, Claude) found this:

> Measured on a 2456-file LLVM rebuild (X86 only, Release, clang, `-j16`,
> 16 cores) where every compile is a cache hit, so the preprocessor spawn is
> essentially all the server does. Four interleaved rounds per arm:
> 
>     | arm    | wall             | server CPU       | of which system |
>     |--------|------------------|------------------|-----------------|
>     | before | 40.7 s (+-0.7)   | 28.2 s (+-0.4)   | 22.8 s          |
>     | after  | 33.5 s (+-0.2)   | 10.6 s (+-0.1)   |  5.8 s          |

It seemed like a reasonably small contribution that might make a good first PR.

The code is AI generated, and I'll take another pass to try to simplify it. I think it has the wrong job server default.

### DIFF
--- a/src/compiler/cicc.rs
+++ b/src/compiler/cicc.rs
@@ -319,6 +319,7 @@ pub fn generate_compile_commands(
         arguments,
         env_vars: env_vars.to_owned(),
         cwd: cwd.to_owned(),
+        share_jobserver: false,
     };
 
     #[cfg(not(feature = "dist-client"))]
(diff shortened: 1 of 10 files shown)


sylvestre:
Nice to see you here :)

PR #2832, opened 2026-08-29
msvc: support lots of flags

AJIOB:
This PR is partially done with AI.

AI search to flags we're not listed and I select what should we do with them.

The full list on flags is gone from https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=msvc-180

For now we support some rarely used MSVC cl.exe flags and provide better details why we skipped caching something (not just "multiple files") for many of flags.

I skipped the flags support from #2831 to minimize conflicts while backporting commits.

P.S. The list of MSVC flags may still not fully complete, but we will have better flags coverage for it


AJIOB:
Hi @sylvestre ,

Graceful ping

PR #2837, opened 2026-09-01
cache: backfill raw multilevel reads

pentlander:
Client-side mode reads cache entries through `MultiLevelStorage::get_raw()`. That path bypassed per-level statistics and remote-to-local backfill, leaving local caches cold.

Record raw-read hits and misses and share the existing asynchronous backfill implementation with regular reads.


pentlander:
My apologies, this was from a rogue agent.

PR #2838, opened 2026-09-02
bench: measure basedir stripping from compiler arguments

tcashel:
Adds a benchmark for the existing `strip_basedirs_from_arg` helper. Each iteration strips 100 path arguments with 0, 1, 8, or 32 configured roots; 90% of paths are outside those roots. Fixture assertions verify matching and unchanged outputs before timing.

This PR changes only `benches/sccache_bench.rs` and applies directly to current upstream. It no longer includes or depends on #2813. It measures the compiler-argument helper already in main, rather than the Rust-specific matcher proposed there.

Formatting, clippy, the locked unit/integration suite, and all four benchmark cases pass on Apple silicon. CUDA cases skip locally. Developed with OpenAI Codex.


docs/Configuration.md
```diff
@@ -179,7 +179,7 @@ Note that some env variables may need sccache server restart to take effect.
...
+* `SCCACHE_BASEDIRS` base directory (or directories) to strip from paths for cache key computation. This is similar to ccache's `CCACHE_BASEDIR` and enables cache hits across different absolute paths when compiling the same source code. Multiple directories can be separated by `;` on Windows hosts and by `:` on any other operating system. When multiple directories are specified, the longest matching prefix is used. Path matching is **case-insensitive** on Windows and **case-sensitive** on other operating systems. For Rust, sccache normalizes matching absolute source arguments, the source side of `--remap-path-prefix`, Cargo path variables, tracked environment dependency values that are absolute paths, and the current working directory. Environment variable takes precedence over file configuration. Only absolute paths are supported; relative paths will cause an error and prevent the server from starting. This setting changes cache keys but does not rewrite paths embedded in artifacts; a Rust artifact can retain an absolute path from the compilation that populated the cache.
```
sylvestre, line 182: sorry but i just would like the benchmark change in this PR

docs/Rust.md
```diff
@@ -9,5 +9,6 @@ sccache includes support for caching Rust compilation. This includes many caveat
 * Procedural macros that read files from the filesystem may not be cached properly.
 * `rustc`'s incremental compilation needs to be disabled. See [The Cargo Book](https://doc.rust-lang.org/cargo/reference/profiles.html#incremental)
 * Crates that invoke the system linker cannot be cached. Examples are `bin`, `dylib`, `cdylib`, and `proc-macro` crates.
+* `SCCACHE_BASEDIRS` normalizes matching paths in cache-key inputs, but it does not rewrite paths embedded in artifacts. For example, a crate that uses `env!("CARGO_MANIFEST_DIR")` can retain the path from the compilation that populated a shared cache entry.
```
sylvestre, line 12: same

benches/sccache_bench.rs
```diff
@@ -867,6 +867,43 @@ fn strip_basedirs_multiple(bencher: Bencher) {
     bencher.bench(|| black_box(strip_basedirs(black_box(&output), black_box(&basedirs))));
 }
 
+/// Most dependency paths fall outside the configured checkout roots.
```
sylvestre, line 870: the 0 case returns immediately in the helper, so it only measures the loop —-worth keeping?

benches/sccache_bench.rs
```diff
@@ -867,6 +867,43 @@ fn strip_basedirs_multiple(bencher: Bencher) {
...
+#[divan::bench(args = [0, 1, 8, 32])]
+fn strip_basedirs_from_arg_mostly_unmatched(bencher: Bencher, root_count: usize) {
```
sylvestre, line 872: all the inputs here are bare paths, but this helper is about compiler arguments — could you please bench `-I/path`, `-ffile-prefix-map=/path=/x` too? that's where pathname_positions does the work.

benches/sccache_bench.rs
```diff
@@ -867,6 +867,43 @@ fn strip_basedirs_multiple(bencher: Bencher) {
...
+    let basedirs: Vec<Vec<u8>> = (0..root_count)
+        .map(|i| format!("/workspaces/project/checkouts/worktree-{i:02}/").into_bytes())
+        .collect();
+    let paths: Vec<Vec<u8>> = (0..100)
+        .map(|i| {
```
sylvestre, line 877: why 7? please document or just use i / 10

PR #2845, opened 2026-09-11
gcc: don't require .gcno for assembly compiled with coverage flags

### LINKED ISSUE #2275: clang coverage test compile error with precompile headers
Hi,

Similar to #1652, I got a build error when a combination of clang + test coverage + precompile headers:

```
sccache: error: failed to zip up compiler outputs
sccache: caused by: failed to zip up compiler outputs
sccache: caused by: failed to open file `"/home/yestyle/code/sccache-clang-pch/build_clang/CMakeFiles/main.dir/cmake_pch.hxx.gcno"`
sccache: caused by: failed to open file `/home/yestyle/code/sccache-clang-pch/build_clang/CMakeFiles/main.dir/cmake_pch.hxx.gcno`
sccache: caused by: No such file or directory (os error 2)
```

sccache version is 0.8.2 and no issues when building with g++.

Minimum reproducible example can be found in https://github.com/yestyle/sccache-clang-pch (build steps in `build.sh`).

Also similar to the fix for #1652, I tested a local build of sccache by changing `optional` to `true` [here](https://github.com/mozilla/sccache/blob/877746cd784e29536cdee1a18e18b8d38044b730/src/compiler/gcc.rs#L612) and it works. Not sure if there are any better solutions. Thanks.

cstrahan:
## Problem

Caching an assembly compile that carries the gcov/coverage flags aborts the build:

```console
$ sccache cc --coverage -c foo.S -o foo.o
sccache: error: failed to zip up compiler outputs
sccache: caused by: failed to open file `"foo.gcno"`
sccache: caused by: failed to open file `foo.gcno`: No such file or directory (os error 2)
```

`--coverage` / `-ftest-coverage` set `outputs_gcno = true`, and in `gcc::parse_arguments`
the `.gcno` note file is then added to the expected outputs as **required**
(`optional: false`). For C/C++/ObjC the compiler does emit a `.gcno`, but an assembly
input (`.s` / `.S`) accepts the coverage flags and produces **no** `.gcno`. When sccache
packages the compiler outputs for caching it can't find the file and fails the whole
compile.

This only surfaces because assembly is cacheable: the "Assembly language support" change
(d9d2eb6) added `Language::Assembler` / `Language::AssemblerToPreprocess`, so assembly
compiles now go through output packaging. Before that they were `CannotCache` and ran the
compiler directly, so the missing `.gcno` never mattered. It's the same class of bug as
the missing-`.gcno` case in #2275.

Real-world hit: any project that builds a `.S` under `-Db_coverage`/`--coverage`
(e.g. breakpad's `breakpad_getcontext.S`) can no longer be built through sccache with a
coverage build.

## Fix

Mark the `.gcno` output `optional` for assembly languages. C/C++/ObjC are unchanged — the
`.gcno` stays required there, so a genuinely-missing note file for those languages is still
an error.

```rust
let optional = matches!(
    language,
    Language::Assembler | Language::AssemblerToPreprocess
);
outputs.insert("gcno", ArtifactDescriptor { path: gcno, optional });
```

## Test

Added `test_parse_arguments_coverage_assembly_gcno_optional`, asserting that
`--coverage -c foo.S` yields a `gcno` output with `optional: true` (the existing
`test_parse_arguments_coverage_outputs_gcno` continues to assert `optional: false` for
`foo.cpp`).

Verified end to end with a from-source build:
- `sccache cc --coverage -c foo.S` now succeeds and is cached;
- `sccache cc --coverage -c foo.c` still emits and caches `foo.gcno` (restored on a cache
  hit), i.e. no regression for C/C++ coverage.

Fixes #2275.


AJIOB:
Hi @sylvestre ,

Is it ok for you?

sylvestre:
what about clang ?
I would like to have also a test in tests/integration

PR #2849, opened 2026-09-11
Fix dist overlay: persistent toolchain dirs and sysroot packaging

justdoGIT:
## Summary

Two independent fixes for sccache-dist that affect Void Linux (and potentially other distros with non-standard sysroot layouts).

### Fix 1: OverlayBuilder crash on persistent toolchain dirs

When sccache-dist restarts, the in-memory `toolchain_dir_map` is wiped, but the on-disk toolchain directories survive. On the next build, `prepare_overlay_dirs` hits the `else` branch (toolchain not in the map) and calls `fs::create_dir` on the existing directory, which returns an `AlreadyExists` error and fails the build.

Guard `create_dir` with an `exists()` check. The `archive_id` is a content hash, so a pre-existing directory for the same id already holds the correct unpacked toolchain and is safe to reuse.

### Fix 2: Package rustlib/ instead of whole sysroot lib dir

On distros where `rustc --print sysroot` reports `/usr` (e.g. Void Linux), the `ToolchainPackager` joins `LIBS_DIR` ("lib") to the sysroot and packages the entire `/usr/lib` tree. This drags the full system library set (several GiB) into every distributed toolchain archive, blowing past the toolchain cache size and making toolchain packaging prohibitively slow.

Package the `rustlib/` subtree instead, which is the actual Rust standard library. Fall back to the old behavior (whole lib dir) if `rustlib/` does not exist, preserving compatibility with non-standard sysroot layouts.

## Testing

- Built and deployed sccache-dist with both fixes on a 4-node LAN cluster (Void Linux)
- Verified toolchain packaging completes in seconds instead of timing out
- Verified sccache-dist restarts no longer fail with AlreadyExists errors
- All existing unit tests pass


justdoGIT:
Superseded by #2850 which consolidates all fixes onto a single branch.

PR #2861, opened 2026-09-17
Read the server startup timeout from SCCACHE_SERVER_STARTUP_TIMEOUT_MS

gronke:
The client waits `server_startup_timeout_ms` for the server, which checks its cache storage before it reports ready; a busy remote backend can need more than the ten-second default.
The value is only settable through the configuration file, so a CI job that configures its backend through the environment has to write a file for this one number.

## Changes

- `SCCACHE_SERVER_STARTUP_TIMEOUT_MS` is read in `config_from_env` through `number_from_env_var`, so a malformed value fails like the other numeric variables.
- It takes precedence over `server_startup_timeout_ms` in the file, as `SCCACHE_BASEDIRS` and `SCCACHE_CLIENT_SIDE` do.
- `docs/Configuration.md` lists the variable and notes the precedence at the file key.
- A test covers environment over file, file alone, and neither.
- Reject zero startup timeout.


src/config.rs
```diff
@@ -1343,22 +1348,29 @@ impl Config {
...
             client_side_mode: env_client_side_mode,
+            server_startup_timeout_ms: env_server_startup_timeout_ms,
         } = env_conf;
         conf_caches.merge(cache);
 
+        // Environment variable takes precedence over file config if it is set
+        let server_startup_timeout_ms =
+            env_server_startup_timeout_ms.or(file_server_startup_timeout_ms);
+        if let Some(0) = server_startup_timeout_ms {
+            bail!("server_startup_timeout_ms must be greater than 0");
```
sylvestre, line 1369: this now also rejects 0 in an existing config file, which used to work

intended? if yes please say it in the docs, else only check the env value

sylvestre, line 1369: when the value comes from the env var, the error names the file key, a bit confusing

src/config.rs
```diff
@@ -1343,22 +1348,29 @@ impl Config {
...
+        }
+        let server_startup_timeout =
+            server_startup_timeout_ms.map(std::time::Duration::from_millis);
+
         // Environment variable takes precedence over file config if it is set
```
sylvestre, line 1374: same comment twice in a row, one is enough

src/config.rs
```diff
@@ -1858,6 +1871,97 @@ fn config_overrides() {
     );
 }
 
+#[test]
+fn config_server_startup_timeout_overrides() {
```
sylvestre, line 1875: the test never sets SCCACHE_SERVER_STARTUP_TIMEOUT_MS, it only tests the merge

could you please add one that goes through config_from_env() like the SCCACHE_BASEDIRS tests, including a malformed value?

src/config.rs
```diff
@@ -1858,6 +1871,97 @@ fn config_overrides() {
...
+        cache: Default::default(),
+        dist: Default::default(),
+        server_startup_timeout_ms: None,
+        basedirs: vec![],
+        client_side_mode: false,
+    };
+    let config = Config::from_env_and_file_configs(env_conf, file_conf).unwrap();
+    assert_eq!(config.server_startup_timeout, None);
+
+    // Test that zero is rejected, from env and from file
```
sylvestre, line 1933: the same 4 blocks with only one field changing, a small table would be shorter, no?

